PR #22, opened 2026-03-12
Fix browse setup when Playwright Chromium is missing

AkbarDevop:
## Summary
- make `./setup` verify that Playwright Chromium can actually launch
- install Chromium automatically if the browser runtime is missing
- update the install docs to match the new setup behavior

## Why
On a clean install, `./setup` can succeed after compiling `browse`, but the first `browse` command can still fail because the Playwright browser binaries were never downloaded.

I reproduced it by running `./setup` with an empty `PLAYWRIGHT_BROWSERS_PATH`, then calling `browse/dist/browse goto https://example.com`. `setup` exited successfully, but `browse` failed with `Executable doesn't exist ... chrome-headless-shell`.

This patch makes `./setup` act like a reliable one-shot installer for `/browse`, which is what the README already suggests.

## Verification
- `bun test`
- clean temp clone with empty `PLAYWRIGHT_BROWSERS_PATH`:
  - `./setup`
  - `browse/dist/browse goto https://example.com`
  - `browse/dist/browse text`

### DIFF
--- a/BROWSER.md
+++ b/BROWSER.md
@@ -152,7 +152,7 @@ The browser automation layer is built on [Playwright](https://playwright.dev/) b
 ### Prerequisites
 
 - [Bun](https://bun.sh/) v1.0+
-- Playwright's Chromium (installed automatically by `bun install`)
+- Playwright's Chromium (installed by `./setup`, or manually via `bunx playwright install chromium`)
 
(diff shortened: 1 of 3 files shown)


garrytan:
Overlaps #40, both stale. Closing.

PR #25, opened 2026-03-12
fix: add Windows compatibility for browse server

DanMcInerney:
## Summary

- **Root cause:** Bun's subprocess management hangs when Playwright calls `chromium.launch()` on Windows. Node.js works fine with the same Playwright code.
- Replace `Bun.serve()` with `node:http` `createServer()` — works in both Bun and Node
- Replace `Bun.serve()` port scanning with `node:net` `createServer()`
- On Windows, spawn the server via `node:child_process` with `detached: true` running a pre-bundled `dist/server.js` under Node. On Unix, behavior is unchanged (Bun runs `.ts` source directly).
- Fix `resolveServerScript()` dev-mode path check that excluded Windows paths

## How it works

| Platform | CLI binary | Server runtime | Server source |
|----------|-----------|---------------|---------------|
| Linux/Mac | `browse` (Bun compiled) | `bun run src/server.ts` | TypeScript source |
| Windows | `browse.exe` (Bun compiled) | `node dist/server.js` | Pre-bundled JS |

The build step now also produces `dist/server.js` via `bun build --target node --packages external` for the Windows/Node path.

## Test plan

- [x] Verified `chromium.launch()` hangs under Bun on Windows, works under Node
- [x] Rebuilt CLI + server bundle successfully
- [x] `browse goto <url>` starts server and navigates
- [x] Server persists between CLI calls (no "Starting server..." on subsequent commands)
- [x] `browse status`, `browse text`, `browse screenshot`, `browse snapshot -i` all work
- [x] Unix path (Bun) is unchanged — only Windows path diverges

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -11,6 +11,7 @@
 
 import * as fs from 'fs';
 import * as path from 'path';
+import { spawn as nodeSpawn } from 'node:child_process';
 
 const PORT_OFFSET = 45600;
 const BROWSE_PORT = process.env.CONDUCTOR_PORT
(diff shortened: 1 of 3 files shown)


garrytan:
Closing — we can't support Windows at this time. The Bun.serve → node:http migration adds significant complexity and a dual-runtime server path. Appreciate the investigation into the Chromium launch hang though.

PR #33, opened 2026-03-13
fix(cookie-picker): pass port explicitly to jsonResponse and errorResponse

### LINKED ISSUE #32: fix(cookie-picker): jsonResponse references undefined `url` — all API routes return HTML error
## Bug

The cookie picker UI fails with:
```
Unexpected token '<', "<!doctype "... is not valid JSON
```

## Root Cause

In `browse/src/cookie-picker-routes.ts`, the `jsonResponse()` helper (line 34) references `url` to build the CORS header, but `url` is not in scope — it only exists as a parameter inside `handleCookiePickerRoute()`.

```typescript
// line 29-37 — `url` is undefined here
function jsonResponse(data: any, status = 200): Response {
  return new Response(JSON.stringify(data), {
    status,
    headers: {
      'Content-Type': 'application/json',
      'Access-Control-Allow-Origin': `http://127.0.0.1:${parseInt(url.port, 10) || 9400}`,
      //                                                        ^^^ ReferenceError
    },
  });
}
```

Every API call (`/cookie-picker/browsers`, `/domains`, etc.) throws `ReferenceError: url is not defined`. Bun catches this and returns its default HTML error page (`<!doctype html>...`). The frontend's `api()` function calls `res.json()` on that HTML → parse error.

`errorResponse()` has the same issue since it delegates to `jsonResponse()`, so error handling also fails.

## Suggested Fix

Extract `port` once at the top of `handleCookiePickerRoute()` and pass it explicitly:

```typescript
function jsonResponse(data: any, port: number, status = 200): Response {
  return new Response(JSON.stringify(data), {
    status,
    headers: {
      'Content-Type': 'application/json',
      'Access-Control-Allow-Origin': `http://127.0.0.1:${port}`,
    },
  });
}

function errorResponse(message: string, code: string, port: number, status = 400, action?: string): Response {
  return jsonResponse({ error: message, code, ...(action ? { action } : {}) }, port, status);
}

export async function handleCookiePickerRoute(url, req, bm) {
  const port = parseInt(url.port, 10) || 9400; // extracted once
  // ... pass `port` to all jsonResponse() and errorResponse() calls
}
```

Full diff: https://github.com/garrytan/gstack/compare/main...fix/cookie-picker-json-response-scope

## Environment

- macOS, Chrome browser
- Introduced in commit `f7b9532` (Phase 3.5)

Jah-yee:
## Bug

The cookie picker API routes fail with `ReferenceError: url is not defined` because the `jsonResponse()` and `errorResponse()` helper functions reference `url` which is not in their scope - it only exists as a parameter in `handleCookiePickerRoute()`.

Every API call (`/cookie-picker/browsers`, `/domains`, `/import`, `/remove`, `/imported`) throws `ReferenceError`.

## Fix

Extract `port` at the top of `handleCookiePickerRoute()` and pass it explicitly to all `jsonResponse()` and `errorResponse()` calls.

Fixes #32

### DIFF
--- a/browse/src/cookie-picker-routes.ts
+++ b/browse/src/cookie-picker-routes.ts
@@ -26,18 +26,18 @@ const importedCounts = new Map<string, number>();
 
 // ─── JSON Helpers ───────────────────────────────────────────────
 
-function jsonResponse(data: any, status = 200): Response {
+function jsonResponse(data: any, port: number, status = 200): Response {
   return new Response(JSON.stringify(data), {
     status,
     headers: {
       'Content-Type': 'application/json',
-      'Access-Control-Allow-Origin': `http://127.0.0.1:${parseInt(url.port, 10) || 9400}`,
+      'Access-Control-Allow-Origin': `http://127.0.0.1:${port}`,
     },
   });
 }
 
-function errorResponse(message: string, code: string, status = 400, action?: string): Response {
-  return jsonResponse({ error: message, code, ...(action ? { action } : {}) }, status);
+function errorResponse(message: string, code: string, port: number, status = 400, action?: string): Response {
+  return jsonResponse({ error: message, code, ...(action ? { action } : {}) }, port, status);
 }
 
 // ─── Route Handler ──────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


garrytan:
Likely stale against current main. Closing.

Issue #18, opened 2026-03-12
[Security] Sensitive values leaked into stdout and model transcripts


fredluz:
## Summary

Several commands reflect secrets, credentials, and sensitive field values directly into command output. Since Browse is designed for LLM agents, this output is typically fed back into model context and persisted in chat logs, terminal history, and orchestration traces.

## Affected Code

- `browse/src/write-commands.ts:97` — `type` command echoes all typed text:
  ```ts
  return `Typed "${text}"`;
  ```
- `browse/src/write-commands.ts:156` — `cookie` command echoes cookie values:
  ```ts
  return `Cookie set: ${name}=${value}`;
  ```
- `browse/src/write-commands.ts:166` — `header` command echoes header values:
  ```ts
  return `Header set: ${name}: ${value}`;
  ```
- `browse/src/read-commands.ts:68` — `forms` command dumps password field values:
  ```ts
  value: input.value || undefined,
  ```
- `browse/src/read-commands.ts:187` — `storage set` echoes stored values:
  ```ts
  return `Set localStorage["${key}"] = "${value}"`;
  ```

## Impact

Two concrete examples:
- `header "Authorization:Bearer sk-live-..."` leaks the bearer token into stdout/model context
- `forms` on a page with a filled password field dumps the password as plaintext (including `"type": "password"`, `"value": "supersecret"`)

In an agentic workflow, these values end up in LLM context windows, log files, and potentially shared traces.

## Suggested Fix

- Redact echoed values by default for `type`, `cookie`, `header`, and `storage set` (e.g., `Cookie set: session=****`)
- Never emit `value` for `type=password` fields in `forms` output
- Consider a `--show-secrets` flag gated behind explicit consent if raw output is ever needed

## Severity

**Medium** — credential and secret leakage via stdout in an LLM-agent context

garrytan:
Fixed. All sensitive values are now redacted in browse output:
- `type` returns character count instead of echoing text
- `cookie` redacts value with `****`
- `header` redacts Authorization, Cookie, X-API-Key, X-Auth-Token values
- `forms` redacts password field values with `[redacted]`
- `storage set` no longer echoes stored values

PR #168, opened 2026-03-18
fix: redact sensitive values in storage read command

xr843:
## Summary

Addresses #18 (remaining unfixed part).

Most of issue #18 was already fixed in prior commits:
- ✅ `cookie` — redacts value as `****`
- ✅ `header` — redacts sensitive headers
- ✅ `type` — returns only character count
- ✅ `forms` — redacts password field values
- ✅ `storage set` — doesn't output the value

**Remaining:** The `storage` read command dumps all `localStorage`/`sessionStorage` values without redaction via `JSON.stringify(storage, null, 2)`, potentially exposing tokens, API keys, and session data into stdout and model transcripts.

## Changes

| File | Change |
|------|--------|
| `browse/src/read-commands.ts` | Redact values for keys matching sensitive patterns; add `--show-secrets` flag |

**Redacted patterns** (case-insensitive): `token`, `secret`, `key`, `password`, `auth`, `credential`, `session`

Key names are preserved so agents can see what's stored without seeing the secret values. Example output:

```json
{
  "localStorage": {
    "theme": "dark",
    "authToken": "[redacted]",
    "api_key": "[redacted]"
  }
}
```

A `--show-secrets` flag bypasses redaction when explicitly needed.

## Test plan

- [x] All 173 tests in `browse/test/commands.test.ts` pass
- [x] Pre-existing `gstack-config` failure confirmed on `main` (not related)
- [ ] Verify `storage` command redacts keys containing "token", "password", etc.
- [ ] Verify `storage --show-secrets` shows full values

### DIFF
--- a/browse/src/read-commands.ts
+++ b/browse/src/read-commands.ts
@@ -289,6 +289,17 @@ export async function handleReadCommand(
         localStorage: { ...localStorage },
         sessionStorage: { ...sessionStorage },
       }));
+      const showSecrets = args.includes('--show-secrets');
+      if (!showSecrets) {
+        const sensitivePattern = /token|secret|key|password|auth|credential|session/i;
+        for (const store of [storage.localStorage, storage.sessionStorage]) {
+          for (const k of Object.keys(store)) {
+            if (sensitivePattern.test(k)) {
+              store[k] = '[redacted]';
+            }
+          }
+        }
+      }
       return JSON.stringify(storage, null, 2);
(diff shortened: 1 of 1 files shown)


xr843:
Hi @garrytan, gentle ping on this one as well. It redacts sensitive values (tokens, passwords, etc.) in the storage read command output, with a --show-secrets flag to bypass when needed.

PR #213, opened 2026-03-19
Translate README from Linkedin-speak to plain English

tornikeo:
Some people debate whether Garry Tan's gstack is useful. I looked at the README and noticed that it contained LinkedIn-speak. Used Kagi to translate from LinkedIn Speak to plain English and submitted a PR. I'm sure this will get merged quickly!


Frizlab:
@kkevindev Though I agree with the spirit of the PR in principle, I honestly think at some point the joke should stop.

PR #221, opened 2026-03-19
fix: use three-dot diff for scope drift detection in /review

roblambell:
## Summary
- `/review` Step 1.5 (Scope Drift Detection) used two-dot `git diff origin/<base> --stat`, which shows the full tree difference between HEAD and the base ref
- On rebased feature branches, this includes commits already on the base branch, producing false-positive "scope drift" findings for changes the author did not introduce
- Switched to three-dot `git diff origin/<base>...HEAD --stat` (merge-base diff), which shows only changes introduced on the feature branch
- This matches what `/ship` already uses for its line-count stat

## How it was found
Contributor mode filed a field report after `/review` flagged unrelated version bumps and date changes as scope drift on a rebased feature branch. The changes came from main via rebase, not from the feature work.

## Test plan
- [x] `bun test` passes
- [x] Validated live: ran `/review` on a rebased feature branch — scope drift step no longer flags rebased-in changes from main


16francej:

PR #240, opened 2026-03-20
fix: repair workflow YAML parsing and lint CI

itstimwhite:
## Summary
- rewrite the two diff-check workflow steps as block scalars so GitHub can parse the workflow YAML
- preserve the existing skill docs freshness behavior for Claude and Codex generation
- add a lightweight actionlint workflow to catch future workflow syntax regressions

## Verification
- ruby YAML parse for .github/workflows/skill-docs.yml
- ruby YAML parse for .github/workflows/actionlint.yml

## Notes
- left unrelated local changes untouched (`bin/gstack-config`, `pnpm-lock.yaml`)

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,8 @@
+name: Workflow Lint
+on: [push, pull_request]
+jobs:
+  actionlint:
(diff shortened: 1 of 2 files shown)


16francej:

PR #275, opened 2026-03-21
fix: support Linux Chromium cookie import

### LINKED ISSUE #265: cookie-import-browser fails on Linux — macOS-only path detection
My claude code on wsl (debian) claims the cookie's can't be imported for this reason.  I'm not sure if it is lying to me, but thought I'd pass this along.
----
 Title: cookie-import-browser fails on Linux — macOS-only path detection

  Environment:
  - OS: Ubuntu on WSL2 (Linux 6.6.87.2-microsoft-standard-WSL2)
  - gstack version: 0.9.4.1
  - Browser: google-chrome-stable installed at /usr/bin/google-chrome-stable
  - Chrome profile: ~/.config/google-chrome/Default/
  - Cookie DB exists: ~/.config/google-chrome/Default/Cookies (36KB, valid SQLite)

  What happens:
  $ browse cookie-import-browser
  [browse] Starting server...
  No Chromium browsers found. Supported: Comet, Chrome, Arc, Brave, Edge

  Root cause:

  findInstalledBrowsers() in browse/src/cookie-import-browser.ts (line ~106) is hardcoded to macOS paths:

  export function findInstalledBrowsers(): BrowserInfo[] {
    const appSupport = path.join(os.homedir(), 'Library', 'Application Support');
    return BROWSER_REGISTRY.filter(b => {
      const dbPath = path.join(appSupport, b.dataDir, 'Default', 'Cookies');
      try { return fs.existsSync(dbPath); } catch { return false; }
    });
  }

  It only checks ~/Library/Application Support/<browser>/Default/Cookies (macOS). On Linux, Chrome stores cookies at
  ~/.config/google-chrome/Default/Cookies, which is never checked.

  Additionally, the cookie decryption uses macOS Keychain (keychainService field in BROWSER_REGISTRY). On Linux, Chrome encrypts cookies
  using PBKDF2 with either a GNOME Keyring-stored key or the hardcoded password "peanuts" with salt "saltysalt" (1 iteration, 16-byte
  AES-128-CBC key).

  What needs to change:
  1. Platform-aware path detection — check ~/.config/google-chrome/ on Linux, ~/Library/Application Support/Google/Chrome/ on macOS
  2. Platform-aware cookie decryption — use PBKDF2 "peanuts" key (or GNOME Keyring/libsecret) on Linux instead of macOS Keychain
  3. Update BROWSER_REGISTRY with per-platform dataDir values

  Linux Chrome data directories:
  - Chrome: ~/.config/google-chrome/
  - Brave: ~/.config/BraveSoftware/Brave-Browser/
  - Edge: ~/.config/microsoft-edge/
  - Chromium: ~/.config/chromium/

AliFozooni:
## Summary
- detect Chromium cookie databases from Linux profile directories in addition to macOS paths
- support Linux cookie decryption for Chromium `v10` cookies and best-effort `v11` key lookup via `secret-tool`
- add regression coverage for Linux profile discovery/import and update the related browse docs/help text

## Testing
- `/tmp/bun-install/bin/bun test browse/test/cookie-import-browser.test.ts`
- `PATH="/tmp/bun-install/bin:$PATH" /tmp/bun-install/bin/bun test browse/test/cookie-picker-routes.test.ts`

Closes #265.

### DIFF
--- a/BROWSER.md
+++ b/BROWSER.md
@@ -247,7 +247,7 @@ Tests spin up a local HTTP server (`browse/test/test-server.ts`) serving HTML fi
 | `browse/src/read-commands.ts` | Non-mutating commands: `text`, `html`, `links`, `js`, `css`, `is`, `dialog`, `forms`, etc. Exports `getCleanText()`. |
 | `browse/src/write-commands.ts` | Mutating commands: `goto`, `click`, `fill`, `upload`, `dialog-accept`, `useragent` (with context recreation), etc. |
 | `browse/src/meta-commands.ts` | Server management, chain routing, diff (DRY via `getCleanText`), snapshot delegation. |
-| `browse/src/cookie-import-browser.ts` | Decrypt Chromium cookies via macOS Keychain + PBKDF2/AES-128-CBC. Auto-detects installed browsers. |
+| `browse/src/cookie-import-browser.ts` | Decrypt Chromium cookies from macOS and Linux browser profiles using platform-specific safe-storage key lookup. Auto-detects installed browsers. |
 | `browse/src/cookie-picker-routes.ts` | HTTP routes for `/cookie-picker/*` — browser list, domain search, import, remove. |
 | `browse/src/cookie-picker-ui.ts` | Self-contained HTML generator for the interactive cookie picker (dark theme, no frameworks). |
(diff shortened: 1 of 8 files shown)


16francej:

PR #330, opened 2026-03-22
fix: install a Codex-safe root skill

tsubasakong:
## Summary
- cap generated Codex frontmatter descriptions at the host's 1024-character limit
- point Codex setup's `~/.codex/skills/gstack` root symlink at `.agents/skills/gstack` so Codex loads the generated Codex root skill instead of the Claude-format repo root
- add regression coverage for both the description limit and the Codex root install path, and regenerate the affected generated skill docs

## Test Plan
- [x] `bun test test/gen-skill-docs.test.ts`
- [x] `bun run gen:skill-docs --dry-run`
- [x] `bun run gen:skill-docs --host codex --dry-run`
- [x] `bun run skill:check`
- [x] `HOME=$(mktemp -d) bash ./setup --host codex` and verified `~/.codex/skills/gstack` points to `.agents/skills/gstack/`

## Notes
- `bun test` is still red on current `main` outside this diff. In my run the unrelated failures were in `test/skill-e2e-browse.test.ts` and multiple assertions/timeouts in `browse/test/commands.test.ts`.
- `autoplan/SKILL.md` and `.agents/skills/gstack-autoplan/SKILL.md` were already stale relative to the current generator output, so they were refreshed here to keep the generated-file checks green.

### DIFF
--- a/.agents/skills/gstack-autoplan/SKILL.md
+++ b/.agents/skills/gstack-autoplan/SKILL.md
@@ -27,6 +27,9 @@ _PROACTIVE=$(~/.codex/skills/gstack/bin/gstack-config get proactive 2>/dev/null 
 _BRANCH=$(git branch --show-current 2>/dev/null || echo "unknown")
 echo "BRANCH: $_BRANCH"
 echo "PROACTIVE: $_PROACTIVE"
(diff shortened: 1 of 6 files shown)


16francej:

Issue #333, opened 2026-03-22
Codex CLI doesn't follow symlinks in .agents/skills/ — skills not discovered


cpdcpd:
## Problem

`create_agents_sidecar()` in `setup` creates symlinks in `.agents/skills/` for each gstack skill:

```
.agents/skills/gstack-qa -> /path/to/.claude/skills/gstack/.agents/skills/gstack-qa/
```

Codex CLI (v0.116.0) does not follow symlinks when scanning `.agents/skills/` for skill directories. Only real directories (like `supabase-postgres-best-practices`) are discovered. This means gstack skills are invisible to Codex's automatic skill discovery.

## Evidence

When asked to list available skills, Codex reports:
> "The repo-local .agents/skills only contains the Supabase skill"

It then manually explores `.claude/skills/gstack/.agents/skills/` to find them — proving the SKILL.md files are correct, just not discovered via the symlink path.

## Root cause

In `setup`, the `create_agents_sidecar()` function uses `ln -snf` for skill directories:
```bash
ln -snf "$skill_dir" "$root_skill_dir/$skill_name"
```

Codex likely uses a directory listing method that filters out symlinks (e.g., checking `d_type` or using `stat` instead of following links).

## Workaround

Replace symlinks with real directory copies:
```bash
for link in .agents/skills/gstack-*; do
  if [ -L "$link" ]; then
    target=$(readlink "$link")
    rm "$link"
    cp -r "$target" "$link"
  fi
done
```

## Suggested fix

In `create_agents_sidecar()`, use `cp -r` instead of `ln -snf` when `--host codex` is specified. This duplicates ~30KB of SKILL.md files per skill but ensures automatic discovery works. The root `gstack` sidecar (which has runtime assets like `bin/` and `browse/`) should remain symlinked to avoid duplicating large binaries — or alternatively, copy SKILL.md and symlink only the heavy directories.

## Environment

- Codex CLI: v0.116.0
- gstack: v0.10.2.0
- macOS Darwin 25.3.0

## Additional issue: global ~/.codex/skills/gstack/ missing runtime assets

[...]

PR #334, opened 2026-03-22
fix: keep root gstack skill loadable in codex

malikrohail:
## Summary

- keep the root `gstack` skill description under Codex CLI's current 1024-character frontmatter limit
- add a regression test covering the two root skill surfaces Codex loads today: `SKILL.md` and `.agents/skills/gstack/SKILL.md`
- preserve the existing root-skill behavior while trimming only the overlong frontmatter text

## Why this PR

On March 22, 2026, I tested current `main` with `codex-cli 0.116.0` and reproduced a live Codex compatibility error:

```text
failed to load skill /Users/test/.codex/skills/gstack/SKILL.md: invalid description: exceeds maximum length of 1024 characters
```

I also saw the same root-skill failure for:

```text
/Users/test/.codex/skills/gstack/.agents/skills/gstack/SKILL.md
```

I could not reproduce the symlink-specific claim from issue #333 on current `main`, but this root-skill frontmatter failure is a confirmed Codex loading bug on the current codepath.

PR #326 does not address this area. It does not touch the root `gstack` skill frontmatter or Codex root-skill loading.

Related to #333.

## Verification

- `bun test`
- `./setup --host codex`
- `codex exec --ephemeral -C /Users/test/.codex/skills/gstack -s read-only -o /tmp/gstack-codex-skills-after.txt "List any installed skills you have available. Return only names, comma-separated."`

## Result

Before this change, Codex emitted the root `gstack` `invalid description` error.

After this change:

- `gstack` appears in the installed skill list
- the root `gstack` `invalid description: exceeds maximum length of 1024 characters` error is gone
- the new static regression test keeps both root skill descriptions under the current Codex limit

### DIFF
--- a/.agents/skills/gstack/SKILL.md
+++ b/.agents/skills/gstack/SKILL.md
@@ -7,38 +7,10 @@ description: |
   ~100ms per command. Use when you need to test a feature, verify a deployment, dogfood a
   user flow, or file a bug with evidence.
   
-  gstack also includes development workflow skills. When you notice the user is at
-  these stages, suggest the appropriate skill:
-  - Brainstorming a new idea → suggest /office-hours
-  - Reviewing a plan (strategy) → suggest /plan-ceo-review
-  - Reviewing a plan (architecture) → suggest /plan-eng-review
-  - Reviewing a plan (design) → suggest /plan-design-review
-  - Auto-reviewing a plan (all reviews at once) → suggest /autoplan
-  - Creating a design system → suggest /design-consultation
-  - Debugging errors → suggest /investigate
-  - Testing the app → suggest /qa
-  - Code review before merge → suggest /review
-  - Visual design audit → suggest /design-review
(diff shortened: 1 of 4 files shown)


malikrohail:
@garrytan tested on **March 22, 2026** with `codex-cli 0.116.0`.

This is intentionally the smallest fix I could verify end-to-end:

- no setup redesign
- no new Codex metadata generation
- no broad skill rewrites
- just bring the root `gstack` skill description under Codex's current 1024-character frontmatter limit
- lock it with a regression test for both `SKILL.md` and `.agents/skills/gstack/SKILL.md`

Before this patch, Codex emitted the root-skill load error during a real `./setup --host codex` + `codex exec` flow. After this patch, `gstack` loads successfully and that specific `invalid description` error is gone.

I also checked PR #326 while narrowing this down. It does not touch the root `gstack` frontmatter or Codex root-skill loading path, so it does not solve this bug.

PR #336, opened 2026-03-22
feat: add optional prefix for claude skill installs

### LINKED ISSUE #284: feat: add skills prefix
Since it's many skills and often installing to `.claude` user dir, an option to enable skill prefix when running `setup` will be helpful, e.g. `gstack-office-hour`.

malikrohail:
## Summary

- add `./setup --prefix <value>` for Claude skill installs
- apply the prefix only to Claude-side per-skill symlink names
- avoid double-prefixing skills that already start with the chosen prefix
- add setup coverage and document the new flag in the README

## Problem

Issue #284 asks for an option to prefix skill names during setup so a user-level Claude install does not drop many bare names like `review`, `ship`, and `qa` directly into `~/.claude/skills/`.

Current `main` still links Claude skills without any optional prefix.

## Fix

- parse `--prefix` / `--prefix=<value>` in `setup`
- pass the value only into `link_claude_skill_dirs`
- keep Codex behavior unchanged because Codex skills already use `gstack-*`
- preserve already-prefixed skill names such as `gstack-upgrade`
- document the flag in `README.md`

Fixes #284.

## Verification

- `bun test`
- temp install repro:
  - create a temp `.claude/skills/gstack` symlink to this checkout
  - run `./setup --host claude --prefix gstack-`
  - verify `gstack-review` exists, bare `review` does not, and `gstack-upgrade` is not double-prefixed


16francej:

PR #351, opened 2026-03-23
fix: support $CLAUDE_CONFIG_DIR for custom config paths (#349)

### LINKED ISSUE #349: Support $CLAUDE_CONFIG_DIR for custom config paths
## Problem

Claude Code supports `$CLAUDE_CONFIG_DIR` to override the default `~/.claude` directory. However, gstack hardcodes `~/.claude/skills/gstack` throughout the codebase (480 occurrences across 50 files). When a user runs Claude Code with a custom `$CLAUDE_CONFIG_DIR`, gstack silently references the default `~/.claude` directory instead — reading the wrong memories, settings, and project config, and writing state to the wrong location.

### Source-of-truth files (7 — must change)

| File | Role |
|------|------|
| `scripts/gen-skill-docs.ts` | Defines `skillRoot`, `binDir`, `browseDir` on lines 42-45; embeds paths into all generated preamble/postamble templates |
| `setup` | Install script — creates symlinks in `~/.claude/skills/`, rewrites paths for kiro/codex |
| `scripts/skill-check.ts` | Validates generated output contains correct host paths |
| `browse/src/find-browse.ts` | Binary discovery — searches `~/.claude/skills/gstack/browse/dist/browse` |
| `bin/gstack-global-discover.ts` | Session discovery across `~/.claude/projects` |
| `bin/dev-setup` | Creates `.claude/skills/gstack` dev symlink |
| `bin/dev-teardown` | Cleans up dev symlinks |

### Template files with their own hardcoded paths (~17 `.tmpl` files)

Several templates (notably `gstack-upgrade/SKILL.md.tmpl` with 19 refs, `codex/SKILL.md.tmpl` with 5, `autoplan/SKILL.md.tmpl` with 7) contain literal `~/.claude/skills/gstack` in bash blocks beyond what the generator injects. These need fixing too.

### Auto-fixed (no direct changes needed)

28 generated `SKILL.md` files — regenerated from templates when sources are fixed.

### Also need updates

- 6 test files (verify path correctness in generated output)
- 8 documentation files (CLAUDE.md, ARCHITECTURE.md, etc.)

## Proposed fix

Replace hardcoded `~/.claude` with `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` at each layer:

1. **`gen-skill-docs.ts`** — change `skillRoot`, `localSkillRoot`, `binDir`, `browseDir` to use the dynamic pattern
2. **`setup`** — resolve install target via `${CLAUDE_CONFIG_DIR:-$HOME/.claude}`
3. **`find-browse.ts`** — check `$CLAUDE_CONFIG_DIR` when searching for the browse binary
4. **`bin/dev-setup`, `bin/dev-teardown`** — same pattern
5. **`.tmpl` files** — replace literal `~/.claude/skills/gstack` with the dynamic pattern in bash blocks
6. Regenerate all SKILL.md files

## Relation to #289

Issue #289 (host integration contract RFC) proposes a broader abstraction for multi-host support. `$CLAUDE_CONFIG_DIR` support is a concrete, narrower fix that can land independently — it fixes a real bug for Claude Code users today without waiting for the full host adapter architecture. The fix naturally aligns with #289's direction since it makes path resolution dynamic rather than hardcoded.

## Impact

Users running Claude Code with a custom `$CLAUDE_CONFIG_DIR` will have gstack silently reference the default `~/.claude` directory. This means gstack reads the wrong memories, settings, and project config — and writes state (sessions, analytics, skill data) to the wrong location. No error or warning is shown.

zehua:
## Summary

Fixes #349.

Claude Code supports `$CLAUDE_CONFIG_DIR` to override the default `~/.claude` directory. This PR makes gstack correctly resolve paths when running with a custom config directory.

**Approach:** Rather than baking `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` into every SKILL.md file (which would produce a massive diff across dozens of generated files), the path is resolved **once at install time** in `setup` — the same approach already used for Kiro. `link_claude_skill_dirs` now copies each SKILL.md with `sed` substitution and symlinks all other assets. Source SKILL.md files stay clean with plain `~/.claude`.

**Changes:**

- **`setup`** — `link_claude_skill_dirs` copies SKILL.md with sed substitution (handles both `~/` and `$HOME/` forms); symlinks all other skill assets; also rewrites root gstack SKILL.md when `CLAUDE_CONFIG_DIR` differs from default
- **`browse/src/find-browse.ts`** — compute `claudeDir` from `CLAUDE_CONFIG_DIR` once, use as pre-resolved entry in `globalDirs` (local workspace check keeps relative `.claude` marker unchanged)
- **`bin/gstack-global-discover.ts`** — resolve `projectsDir` from `CLAUDE_CONFIG_DIR`
- **`test/gen-skill-docs.test.ts`** — updated setup install behaviour assertions

No SKILL.md or `.tmpl` files changed. Future path changes only need to touch `setup`.

## Design options explored

Four approaches were considered before landing on the final solution:

**Option 1: Bash variable in every SKILL.md** *(rejected)*
Replace `~/.claude` with `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` everywhere in SKILL.md files. Rejected because SKILL.md files are prose/markdown read by LLMs, not bash scripts — bash syntax in prose is fragile and non-portable. Also produces a massive diff across all SKILL.md and `.tmpl` files, with a runtime dependency on bash variable expansion at every skill invocation.

**Option 2: Bash variable injected via `gen-skill-docs.ts`** *(rejected)*
Have the generator substitute `HOST_PATHS.skillRoot` with `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` so generated SKILL.md files contain the bash variable. Same problem as Option 1 — bash syntax in prose. Also broke CI: committed SKILL.md files had `~/.claude` but `gen:skill-docs` generated `${CLAUDE_CONFIG_DIR:-...}`, causing the freshness check to fail.

**Option 3: Setup-time substitution — always substitute** *(intermediate)*
`link_claude_skill_dirs` always copies SKILL.md with `sed -e "s|~/.claude|${CLAUDE_CONFIG_DIR:-$HOME/.claude}|g"`, baking in the resolved literal path at install time. Better — no runtime bash dependency, no SKILL.md diffs. Problem: default users got `$HOME/.claude` (expanded at setup time) baked in instead of `~/.claude` — a slightly different string and less portable.

**Option 4: Setup-time substitution — conditional only** *(final)*
`link_claude_skill_dirs` only runs sed when `CLAUDE_CONFIG_DIR` is actually set; otherwise does a plain `cp` preserving the original `~/.claude`:

```bash
if [ -n "${CLAUDE_CONFIG_DIR:-}" ]; then
  sed -e "s|~/.claude|$claude_dir|g" ... > "$target/SKILL.md"
else
  cp "$skill_dir/SKILL.md" "$target/SKILL.md"
fi
```

- Default users: SKILL.md unchanged, `~/.claude` stays as-is (portable)
- Custom-path users: literal resolved path baked in at install time (no bash dependency, no runtime env requirement)
- Source `.tmpl` and SKILL.md files stay clean — zero diff on any skill docs
- Mirrors how Kiro and Codex installs already work in `setup`

## `~/.gstack` and multi-profile isolation

`~/.gstack` is intentionally **shared across all Claude Code profiles** — it is user-level state, not profile-level state:

- `config.yaml` (telemetry, auto_upgrade) — same preferences for all profiles
- Update check cache / snooze — no need to re-prompt per profile
- `.completeness-intro-seen`, `.telemetry-prompted` — one-time per user
- `projects/{slug}/...` (design docs, test plans, reviews) — already repo-scoped by slug, naturally separated
- `sessions/`, `analytics/` — user-level telemetry

The one edge case where sharing causes a real problem: **two different gstack versions installed for two different profiles**. The upgrade markers (`just-upgraded-from`, `last-update-check`) could get confused.

Two pre-existing issues worth noting (out of scope for this PR):

1. **Inconsistent override env vars** — bin scripts use `GSTACK_STATE_DIR`; `gstack-review-log`/`gstack-review-read` use `GSTACK_HOME`. Should be unified.
2. **Hardcoded `~/.gstack` in preamble** — the generated preamble bakes `~/.gstack/sessions/`, `~/.gstack/analytics/`, and marker files as hardcoded strings, so `GSTACK_STATE_DIR` is partially ignored even when set.

## Test plan

- [x] `bun test` passes (all skill validation + gen-skill-docs tests pass)
- [x] `bun run gen:skill-docs` -- SKILL.md files unchanged from upstream (zero diff)
- [x] `bun run gen:skill-docs --host codex` -- Codex SKILL.md files still use `$GSTACK_ROOT`
- [ ] Manual: set `CLAUDE_CONFIG_DIR=/tmp/custom-claude` and run `setup` -- verify installed SKILL.md files contain `/tmp/custom-claude` instead of `~/.claude`

### DIFF
--- a/setup
+++ b/setup
@@ -189,8 +189,9 @@ fi
 mkdir -p "$HOME/.gstack/projects"
 
 # ─── Helper: install Claude skill subdirectories into a skills parent directory ──
-# Copies SKILL.md with ~/.claude path substitution (supports CLAUDE_CONFIG_DIR);
-# symlinks all other files so changes to assets are immediately visible.
+# Copies SKILL.md with ~/.claude path substitution when CLAUDE_CONFIG_DIR is set;
+# otherwise copies as-is (keeps ~/.claude, portable for default users).
+# Symlinks all other files so changes to assets are immediately visible.
 link_claude_skill_dirs() {
   local gstack_dir="$1"
   local skills_dir="$2"
@@ -205,10 +206,16 @@ link_claude_skill_dirs() {
       # Remove old whole-dir symlink from previous installs
       [ -L "$target" ] && rm -f "$target"
       mkdir -p "$target"
-      # Copy SKILL.md with path substitution (handles both ~/ and $HOME/ forms)
-      sed -e "s|~/.claude|$claude_dir|g" \
-          -e "s|\$HOME/\.claude|$claude_dir|g" \
-          "$skill_dir/SKILL.md" > "$target/SKILL.md"
+      # Copy SKILL.md, substituting ~/.claude only when CLAUDE_CONFIG_DIR is set.
+      # Default users keep ~/.claude as-is (portable); custom-path users get the
+      # literal resolved path baked in (no runtime bash dependency).
+      if [ -n "${CLAUDE_CONFIG_DIR:-}" ]; then
+        sed -e "s|~/.claude|$claude_dir|g" \
+            -e "s|\$HOME/\.claude|$claude_dir|g" \
+            "$skill_dir/SKILL.md" > "$target/SKILL.md"
+      else
+        cp "$skill_dir/SKILL.md" "$target/SKILL.md"
+      fi
       # Symlink all other entries in the skill directory
       for item in "$skill_dir"*; do
         item_name="$(basename "$item")"
--- a/test/gen-skill-docs.test.ts
+++ b/test/gen-skill-docs.test.ts
@@ -1239,6 +1239,38 @@ describe('setup script validation', () => {
     expect(fnBody).toContain('ln -snf');
   });
 
+  test('link_claude_skill_dirs copies SKILL.md as-is when CLAUDE_CONFIG_DIR is unset', () => {
+    // When CLAUDE_CONFIG_DIR is not set, ~/.claude should be preserved in the installed SKILL.md
+    const tmp = fs.mkdtempSync(path.join(require('os').tmpdir(), 'gstack-test-'));
+    try {
+      const fakeSkill = path.join(tmp, 'gstack', 'fake-skill');
+      const fakeSkillsDir = path.join(tmp, 'skills');
+      fs.mkdirSync(fakeSkill, { recursive: true });
+      fs.mkdirSync(fakeSkillsDir, { recursive: true });
+      fs.writeFileSync(path.join(fakeSkill, 'SKILL.md'),
+        'run ~/.claude/skills/gstack/bin/gstack-config\n');
+
+      const fnStart = setupContent.indexOf('link_claude_skill_dirs()');
+      const fnEnd = setupContent.indexOf('\n}', setupContent.indexOf('linked[@]}', fnStart)) + 2;
+      const fnBody = setupContent.slice(fnStart, fnEnd);
+
+      const env = { ...process.env };
+      delete env.CLAUDE_CONFIG_DIR;
+      const script = `${fnBody}\nlink_claude_skill_dirs "${path.join(tmp, 'gstack')}" "${fakeSkillsDir}"`;
+      Bun.spawnSync(['bash', '-c', script], {
+        env: { ...env, HOME: process.env.HOME ?? '/tmp' },
+        stdout: 'pipe', stderr: 'pipe',
+      });
+
+      const installed = path.join(fakeSkillsDir, 'fake-skill', 'SKILL.md');
+      expect(fs.existsSync(installed)).toBe(true);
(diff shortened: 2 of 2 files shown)


time-attack:
Thanks for this. The setup approach mutates a tracked source file and interpolates CLAUDE_CONFIG_DIR into sed without safely handling metacharacters or multiple profiles. Rebase onto the current HostConfig design, avoid source mutation, and test default, custom, and secondary profiles with difficult paths.

Issue #289, opened 2026-03-21
RFC: define a standard host integration contract for gstack


loocor:
# RFC: define a standard host integration contract for gstack

## Summary

This is a design proposal / RFC, not just a request for one-off OpenCode compatibility.

gstack already has the beginnings of a multi-host architecture, but today it still behaves more like:

- Claude Code as the primary path
- Codex as a special-case adaptation

The proposal is to make multi-host support explicit and first-class by introducing a host/runtime adapter layer, then using OpenCode as the first new host to validate the design.

The goal is to avoid a future where Claude/Codex/OpenCode/Cursor compatibility grows through one-off branches, hardcoded path rewrites, and incremental string replacement that eventually conflict.

I’m less interested in a one-off OpenCode patch than in creating the right abstraction point for all adjacent runtimes.

## Status

This issue is intended as an RFC / enhancement discussion before implementation.

The immediate goal is to align on the abstraction boundary first, so future OpenCode support and similar runtime work can build on the same model instead of landing as parallel special cases.

## Why now

There is already community demand for OpenCode support:

- issue #16 requests OpenCode / OpenRouter documentation
- PR #86 adds an OpenCode compatibility guide
- there are community repos experimenting with `gstack-on-opencode`

That suggests the demand is real, but the current direction is mostly documentation-level compatibility.
This feels like the right moment to insert a proper abstraction point before multiple partial adaptations diverge.

## What already exists

There is already a strong foundation for this:

- `scripts/gen-skill-docs.ts` has host-aware generation for `claude` and `codex`
- `setup` already supports `--host claude|codex|auto`
- `.agents/skills/` support already exists for non-Claude runtimes
- `browse` is already a standalone CLI, so it should adapt well once the runtime abstraction is clean

## Core idea

Introduce a first-class host integration contract, implemented by host adapters, instead of adding more `if host === ...` branches.

The contract should define both install/discovery behavior and runtime capability translation.
Each host adapter would define things like:

- `skillRoot`
- `localSkillRoot`
- `binDir`
- `skillDiscoveryRoots`
- `frontmatterProfile`
- `interactionPrimitive`
- `skillInvocationPrimitive`
- `toolCapabilityProfile`
- `pathRewriteRules`
- `installStrategy`
- `runtimeAssetStrategy`
- `sessionRunner`

This would let gstack treat Claude, Codex, OpenCode, Cursor, Droid, and similar runtimes as peers rather than "primary host + exceptions".

[...]

PR #364, opened 2026-03-23
RFC #289: land phase-1 host integration contract

loocor:
## Context
- builds the phase-1 implementation checkpoint for RFC #289 (`RFC: define a standard host integration contract for gstack`)
- focuses on the install/runtime boundary first so future OpenCode or adjacent host work can build on a shared contract instead of more one-off branches
- keeps the roadmap intentionally incremental: land the contract foundation now, normalize more host semantics next, then use a second host to validate the model

## What This PR Lands
- introduce `scripts/host-registry.ts` as the single source of truth for host layouts, discovery roots, runtime roots, path rewrites, and generated output locations
- make `~/.gstack` the canonical shared runtime home, with host-specific roots becoming thin discoverability layers and repo-local `.gstack` acting as the workspace sidecar/fallback
- rewire `setup`, `scripts/gen-skill-docs.ts`, `scripts/skill-check.ts`, packaging, and regression tests to the shared contract
- separate generated discoverability artifacts from runtime assets and regenerate the affected Claude/Codex/agents skill docs, including the new `CODEX_PLAN_REVIEW` placeholder support merged from upstream
- fold in the install/runtime hardening that fell out of dogfooding this contract: workspace discovery cleanup, shared-vs-workspace resolution checks, upgrade-path tightening, and a small browse loopback fix for chunked responses

## Why The Diff Is Large
- most of the volume is generated `SKILL.md` output and contract-driven rewrites flowing from the registry / shared-runtime change
- the behavioral core is concentrated in `scripts/host-registry.ts`, `scripts/gen-skill-docs.ts`, `setup`, `scripts/skill-check.ts`, `browse/src/cli.ts`, and the related tests

## Roadmap Progress
- **Phase 1 — install/runtime contract:** landed in this PR
- **Phase 2 — host-specific skill semantics:** still next; interactive prompting, invocation primitives, and metadata/frontmatter policy should keep moving onto the same host contract instead of remaining template-level special cases
- **Phase 3 — second-host validation:** intentionally deferred until the phase-2 surface is clearer; OpenCode remains the most useful candidate to prove the contract rather than arriving as a one-off exception

## Intentionally Not In Scope
- no full OpenCode adapter yet
- no attempt to solve every host/runtime capability difference in one pass
- no broad host expansion beyond the install/runtime contract needed to make future host work coherent

## Test Plan
- [x] `bun test test/gen-skill-docs.test.ts`
- [x] `bun run skill:check`
- [x] `git diff --check`

Refs #289


garrytan:
Thanks for working on this! Closing the PR — the implementation is premature, but the RFC (#289) has good ideas worth keeping in the discussion. Let's continue the conversation there rather than landing code before the design is settled. Appreciate the contribution!

PR #366, opened 2026-03-23
feat(browse): connect to existing Chrome via CDP (BROWSE_CDP_URL)

HMAKT99:
## Summary

- `BROWSE_CDP_URL` env var connects browse server to an existing Chrome instance via CDP instead of launching headless Chromium
- Uses Playwright's `chromium.connectOverCDP()` — no new dependencies
- When connected: reuses existing browser context (cookies, extensions, login state) and attaches to open tabs
- When not set: zero behavior change

## How it works

```
# Without BROWSE_CDP_URL (unchanged):
browse server → chromium.launch() → headless Chromium → new context

# With BROWSE_CDP_URL:
browse server → chromium.connectOverCDP(ws://...) → existing Chrome → existing context + tabs
```

```bash
chrome --remote-debugging-port=9222
BROWSE_CDP_URL=ws://localhost:9222 browse server
```

- Existing browser contexts reused (falls back to new context if none found)
- Existing tabs attached with page handlers registered
- `isHeaded=true` set automatically
- `nextTabId` initialized from existing page count

## 1 file, 46 lines

`browse/src/browser-manager.ts` — `launch()` method extended with CDP branch.

## Test plan
- [x] All 548 existing tests pass
- [x] Without `BROWSE_CDP_URL`: identical behavior
- [x] CDP path: connectOverCDP → reuse context → attach tabs → register handlers
- [x] Fallback: no existing contexts → create new one

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -62,7 +62,19 @@ export class BrowserManager {
   private consecutiveFailures: number = 0;
 
   async launch() {
-    this.browser = await chromium.launch({ headless: true });
+    const cdpUrl = process.env.BROWSE_CDP_URL;
+
+    if (cdpUrl) {
+      // Connect to an existing Chrome instance via CDP WebSocket URL.
+      // Use case: Chrome sidebar extension, remote debugging, real browser sessions.
+      // Launch Chrome with: chrome --remote-debugging-port=9222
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej we are sticking to /browse skill for now

PR #394, opened 2026-03-23
fix: create browse state dir before taking startup lock

pengwk:
## Summary
- ensure the browse state directory exists before acquiring the startup lock
- add a regression test for first run in a fresh repo with no `.gstack/` directory

## Repro
1. Install the latest repo-local gstack into a fresh repo
2. Run `browse goto https://example.com`
3. `browse` prints `Another instance is starting the server, waiting...` and then times out

The root cause is that `ensureServer()` tried to create `.gstack/browse.json.lock` before `.gstack/` existed, then treated the resulting `ENOENT` like a held lock.

## Related context
- Issue: #358 `browse: acquireServerLock fails silently when .gstack/ dir does not exist` by @nicezic
- Similar PR: #374 `fix(browse): ensure .gstack/ dir exists before acquiring lock` by @zuoc1993
- This PR takes the same fix direction and also adds an automated regression test for the fresh-repo first-start case

## Testing
- `bun test browse/test/commands.test.ts --filter "CLI lifecycle"`

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -262,6 +262,11 @@ async function ensureServer(): Promise<ServerState> {
     }
   }
 
+  // The lockfile lives under config.stateDir. Ensure the directory exists
+  // before taking the lock so first run in a fresh repo doesn't misread
+  // ENOENT as "another instance is starting".
+  ensureStateDir(config);
+
   // Acquire lock to prevent concurrent restart races (TOCTOU)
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej Please close this PR. Current main already creates the browser state directory before using its startup lock, resolving the problem tracked in #358. This old conflicting branch is no longer useful.

PR #397, opened 2026-03-23
feat: add hate-driven-development skill

build000r:
## Summary
- add a new `/hate-driven-development` skill that scans recent X mentions of gstack and ranks the worst posts deterministically
- persist `newest_id` state so the default run only looks since the last successful breakfast
- add deterministic scoring tests and wire the skill into README, AGENTS, docs, routing, and validation surfaces

## Verification
- `npx -y bun run gen:skill-docs`
- `npx -y bun run gen:skill-docs --host codex`
- `npx -y bun test`
- `npx -y bun run skill:check`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,3 +3,6 @@
 
 # Required for LLM-as-judge evals (bun run test:eval)
 ANTHROPIC_API_KEY=sk-ant-your-key-here
+
+# Optional for /hate-driven-development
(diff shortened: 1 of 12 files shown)


garrytan:
Thanks for the creative idea! Closing — "hate-driven development" is a fun concept but not something we'd ship as a real skill. It also has merge conflicts. Appreciate the humor and the effort!

PR #409, opened 2026-03-23
feat: ./setup --enforcement for plan enforcement

Yann-BILIEN:
Sharing how I use gstack in case it's useful to others.

I run multiple Claude Code tabs in parallel, typically 5 or more, each working on a different part of a feature. The plan review skills (/plan-eng-review, /plan-ceo-review, /plan-design-review) are what make that possible: they force real decisions before the agent starts touching code.

The issue I kept running into is that those decisions don't carry across tabs or sessions. Tab 2 doesn't know what tab 1 decided. The agent drifts, contradicts a constraint that was already agreed, or re-opens a question that was already closed. I'd end up checking in constantly, which defeats the point of running in parallel.

What fixed it for me was adding Rippletide (open source) on top. It hooks into Claude Code and injects prior plan decisions before every prompt, and validates writes against them before files are touched. Once that's in place, I can actually leave the tabs running. The enforcement happens at the hook level so I don't have to be the one enforcing it.

This PR just adds `--enforcement` to `./setup` so that running `npx rippletide-code@latest connect` is one step instead of something you have to look up separately:

```sh
./setup --enforcement
```

Completely optional, without it gstack works exactly as before. If you try it, check `CLAUDE.md` afterward since Rippletide generates its own and the `## gstack` section needs to still be there.


Yann-BILIEN (commented):
Adding plan enforcement

Issue #407, opened 2026-03-23
bun build --compile fails on virtiofs mounts (devcontainers)


richardrigby:
## Summary

`bun build --compile` fails inside devcontainers that use **virtiofs**-mounted filesystems. This breaks the `./setup` script — the browse binary at `browse/dist/browse` is never rebuilt, leaving a stale binary that can't locate `server.ts`. The result is every `$B` command hanging with:

```
[browse] Another instance is starting the server, waiting...
[browse] Timed out waiting for another instance to start the server
```

## Root Cause

Bun's `--compile` flag writes output to a temp file (e.g. `/run/host_virtiofs/.../.bun-build`) and then does an atomic `rename()` to the target path. On virtiofs mounts, this rename fails with `ENOENT`:

```
[9ms]  bundle  2 modules
failed to rename /run/host_virtiofs/.../.189f99fd7dcbfffb-00000000.bun-build to browse/dist/browse: ENOENT
```

The `setup` script has `|| true` at the end of the build pipeline, so this failure is silently swallowed and setup reports success:

```
gstack ready (claude).
  browse: /path/to/browse/dist/browse
```

The stale compiled binary has wrong embedded `import.meta.dir` and `process.execPath` values, so `resolveServerScript()` in `cli.ts` can't find `server.ts`. The server process never starts, and `acquireServerLock()` in `ensureServer()` times out waiting for a state file that will never appear.

## Reproduction

1. Run VS Code / Cursor with a devcontainer on macOS (Docker Desktop with virtiofs)
2. Install gstack and run `./setup`
3. Observe `bun build --compile` failure in output (easy to miss due to `|| true`)
4. Run any `$B` command → hangs and times out

## Workaround

Replace the compiled binary with a shell wrapper that delegates to `bun run`:

```sh
#!/bin/sh
SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
exec bun run "$SCRIPT_DIR/../src/cli.ts" "$@"
```

This works because `bun` itself runs fine on virtiofs — only `bun build --compile` is broken (it's a known Bun issue with cross-mount atomic renames).

## Suggested Fix Options

1. **Shell wrapper fallback in `setup`**: After `bun build --compile` fails, detect the failure and write the shell wrapper automatically. This is the simplest fix and works everywhere.

[...]

PR #412, opened 2026-03-24
fix: fall back to wrappers when bun compile fails

malikrohail:
## Summary

Fixes the failure mode behind #407 where `bun build --compile` can fail on virtiofs/devcontainer mounts but `./setup` still reports success.

### What changed
- Move build orchestration into `scripts/build-artifacts.sh`
- Replace the blanket `|| true` build tail with explicit `build_or_wrap` handling
- When `bun build --compile` fails, write executable `bun run` wrappers for:
  - `browse/dist/browse`
  - `browse/dist/find-browse`
  - `bin/gstack-global-discover`
- Keep the normal compiled-binary path unchanged when compile succeeds
- Add static coverage for the new build helper in `test/gen-skill-docs.test.ts`

## Why this fixes the bug

On virtiofs mounts, Bun can fail the final rename step for compiled binaries. The old `build` command hid that failure behind `|| true`, so `setup` could continue with stale artifacts.

This change makes the failure explicit and recovers by installing wrapper entrypoints instead of leaving stale binaries in place.

## Verification

- `bun run build`
- `bun test test/gen-skill-docs.test.ts -t "build script validation"`
- `bun test`

Refs #407

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -11,11 +11,26 @@
 
 import * as fs from 'fs';
 import * as path from 'path';
+import * as os from 'node:os';
+import { request as httpRequest } from 'node:http';
+import { spawn as nodeSpawn } from 'node:child_process';
 import { resolveConfig, ensureStateDir, readVersionHash } from './config';
 
 const config = resolveConfig();
+
+export function isWsl(
+  env: Record<string, string | undefined> = process.env,
+  platform: string = process.platform,
+  release: string = os.release()
+): boolean {
+  if (platform !== 'linux') return false;
+  const lowerRelease = release.toLowerCase();
+  return Boolean(env.WSL_DISTRO_NAME || env.WSL_INTEROP || lowerRelease.includes('microsoft'));
+}
+
 const IS_WINDOWS = process.platform === 'win32';
-const MAX_START_WAIT = IS_WINDOWS ? 15000 : 8000; // Node+Chromium takes longer on Windows
+const IS_WSL = isWsl();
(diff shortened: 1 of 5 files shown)


malikrohail:
> Great work on this PR! We hit a related but distinct issue on WSL2 (filed as #494).
> 
> **Your PR fixes:** compile failure → wrapper fallback. ✅
> 
> **Our additional finding:** Even when compile succeeds, `Bun.spawn` inside the compiled binary hangs on WSL because stdio pipes don't connect properly. The process tree spawns fine (bun + chrome-headless all visible via pstree), but the parent CLI never receives data from the pipe.
> 
> **Second issue we found:** If the system has `HTTP_PROXY` set (common on WSL with clash/v2ray), the health check `curl http://127.0.0.1:$port/health` gets routed through the proxy → 502. Adding `--noproxy '*'` or `no_proxy=localhost,127.0.0.1` to internal HTTP calls fixes this.
> 
> **Our workaround:** A bash wrapper (~80 lines) that uses `nohup bun run server.ts` instead of `Bun.spawn`, and adds `--noproxy '*'` to all curl calls. This could be integrated into the `build_or_wrap` approach as an additional fallback layer.
> 
> Would be happy to contribute this as a follow-up PR if useful.

I will patch a commit with the fixes you just mentioned momentarily.

time-attack:
Thanks for this. The branch combines the virtiofs fallback with separate WSL behavior and would omit parts of the current five-artifact build flow. Split the WSL work, rebase the fallback onto the current build script, preserve all artifacts, and verify virtiofs and WSL paths separately.

Issue #280, opened 2026-03-21
Skill chain state detection broken: cross-branch JSONL isolation + /review not logging


chancheuklap:
## Problem

Skills in the gstack workflow chain cannot detect outputs from previous skills. This breaks the intended workflow state machine where each skill feeds into the next.

**Three reported failures:**

| Upstream Skill | Downstream Skill | Expected | Actual |
|---|---|---|---|
| /office-hours | /plan-ceo-review | Finds design doc | ❌ Branch-scoped filename mismatch |
| /plan-ceo-review | /plan-eng-review | Finds CEO plan + review log | ❌ No discovery code for `ceo-plans/` + JSONL is branch-isolated |
| /review | /ship | Review Readiness Dashboard shows CLEAR | ❌ /review writes no JSONL entry; dashboard doesn't recognize it |

### 1. JSONL review log is branch-scoped but workflow is cross-branch

`gstack-review-log` writes to `$BRANCH-reviews.jsonl`. When skills run on different branches (e.g., `/office-hours` on `main`, then `/plan-eng-review` on `feat/my-feature`), all prior review signals are invisible because `gstack-review-read` only reads the current branch's file.

### 2. /review skill never writes to the review log

The `/review` skill (pre-landing code review) does NOT write a `"skill":"review"` or `"skill":"plan-eng-review"` entry to the JSONL. It only conditionally writes `design-review-lite` and `codex-review`.

The `/ship` Review Readiness Dashboard only looks for: `plan-ceo-review`, `plan-eng-review`, `plan-design-review`, `design-review-lite`, `codex-review`. Running `/review` leaves zero trace that `/ship` can parse → dashboard shows "Eng Review: 0 runs / NOT CLEARED" even after a full review.

### 3. plan-eng-review has no CEO plan discovery code

`plan-eng-review` only globs for `*-design-*.md` (office-hours output). It has zero code to read `~/.gstack/projects/$SLUG/ceo-plans/` where `/plan-ceo-review` writes its output.

### 4. File naming uses branch name, breaking cross-branch discovery

`office-hours` writes files named `{user}-{branch}-design-{datetime}.md`. Downstream skills glob with `*-$CURRENT_BRANCH-design-*.md`. When branches differ, the glob fails. The fallback (`*-design-*.md` sorted by mtime) is non-deterministic.

## Impact

[...]

PR #414, opened 2026-03-24
fix: preserve workflow handoffs across branches

malikrohail:
## Summary

Fixes the still-open root causes from #280 that make multi-step workflows lose context when skills run on different branches.

### What changed
- Add `bin/gstack-find-artifact` to resolve the best design doc or CEO plan for the current repo/branch
- Make `gstack-review-log` write to a project-scoped `reviews.jsonl` in addition to the legacy branch file
- Make `gstack-review-read` prefer the project-scoped log and fall back to all legacy per-branch logs when needed
- Update `/plan-eng-review` to load both the current design doc and the active CEO plan via the helper instead of branch-fragile globs
- Update `/plan-ceo-review` and `/autoplan` to use the shared design-doc lookup helper
- Add behavioral tests covering cross-branch review visibility, legacy log fallback, branch-aware design doc lookup, and CEO plan lookup

## Why this fixes the bug

Current `main` already has the narrow `/review` logging fix, but workflow continuity still breaks because:
- review discovery is effectively branch-scoped
- design doc lookup still depends on branch-specific filenames
- `/plan-eng-review` has no direct CEO-plan discovery path

This PR moves those handoffs to project scope and gives the planning skills a shared artifact lookup path instead of ad hoc globs.

## Verification

- `bun run gen:skill-docs`
- `bun test test/workflow-handoff.test.ts`
- `bun test`

Refs #280

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -474,7 +474,7 @@ Then prepend a one-line HTML comment to the plan file:
 ### Step 2: Read context
 
 - Read CLAUDE.md, TODOS.md, git log -30, git diff against the base branch --stat
-- Discover design docs: `ls -t ~/.gstack/projects/$SLUG/*-design-*.md 2>/dev/null | head -1`
+- Discover design docs: `~/.claude/skills/gstack/bin/gstack-find-artifact design-doc 2>/dev/null || true`
 - Detect UI scope: grep the plan for view/rendering terms (component, screen, form,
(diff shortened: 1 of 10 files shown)


time-attack:
@malikrohail Rewrite this to resolve current conflicts, restore validation and enqueue behavior, and prevent reviews from unrelated branches from clearing ship readiness.

PR #459, opened 2026-03-24
fix: gracefully degrade AskUserQuestion in remote control mode

manoj30075:
## Summary

- **Problem:** When Claude Code is used via the Anthropic app's remote control feature, `AskUserQuestion` renders an interactive terminal UI (arrow keys + Enter to select) that the user **cannot interact with**.
- **Fix:** The preamble instructions now tell Claude to detect remote control from the conversation context and fall back to plain text questions. Zero config — nothing for users to set up.

## How it works

The preamble instructions (shared by all 23+ skills) now include:

> Check if remote control is active by looking at BOTH:
> 1. The `REMOTE_CONTROL` value from the preamble bash output
> 2. The conversation history — if you see `Remote Control connecting…` without a subsequent `Remote Control disconnected.`, remote control is active

When detected, Claude presents questions as plain text with numbered options instead of calling `AskUserQuestion`. This applies to all skill questions including telemetry, lake intro, and skill-specific prompts.

## Why this approach

| Approach | Result |
|----------|--------|
| TTY detection (`! tty`) | Always triggers — Claude Code's Bash tool never has a real TTY |
| JWT token (`"application": "ccr"`) | Persists after disconnect, not real-time |
| Env var from Anthropic | Doesn't exist yet ([anthropics/claude-code#38428](https://github.com/anthropics/claude-code/issues/38428)) |
| CLAUDE.md instruction | Works but clutters install prompt and needs upgrade patching |
| **Conversation context detection** | **Zero config, self-contained in preamble, works immediately** |

## Changes

| File | What |
|------|------|
| `scripts/resolvers/preamble.ts` | Remote control detection in bash + conversation-based detection in instructions + plain text fallback |
| `*/SKILL.md` (23 files) | Auto-regenerated from preamble change |

## Test plan

- [x] `bun run gen:skill-docs` succeeds (both default and codex host)
- [x] `bun test` passes (329 pass, 11 pre-existing browse/playwright failures)
- [x] Manual test: Connected via remote control → Claude detected from conversation context → questions rendered as plain text
- [x] Manual test: Disconnected → Claude detected disconnect → back to normal AskUserQuestion
- [x] Verified no changes to README install prompt, CLAUDE.md, or upgrade flow

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -50,6 +50,15 @@ _TEL_START=$(date +%s)
 _SESSION_ID="$$-$(date +%s)"
 echo "TELEMETRY: ${_TEL:-off}"
 echo "TEL_PROMPTED: $_TEL_PROMPTED"
+# Detect remote control mode (Anthropic app remote control)
+# TTY detection is unreliable (Claude Code's Bash tool never has a real TTY).
+# Use explicit env var or check if gstack-config has it enabled.
+_REMOTE_CONTROL="${GSTACK_REMOTE_CONTROL:-}"
+if [ -z "$_REMOTE_CONTROL" ]; then
+  _REMOTE_CONTROL=$(~/.claude/skills/gstack/bin/gstack-config get remote_control 2>/dev/null || echo "")
+fi
+[ -z "$_REMOTE_CONTROL" ] && _REMOTE_CONTROL=0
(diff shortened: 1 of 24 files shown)


time-attack:
@manoj30075 Rewrite this against the current modular preamble and generated files, resolve conflicts, and add focused tests for remote-control fallback behavior.

PR #396, opened 2026-03-23
feat: add GitHub Copilot CLI support

ridermw:
## Summary

Add first-class GitHub Copilot CLI support to gstack. All 28 skills now work with the standalone Copilot CLI (`copilot` binary, GA February 2026) alongside the existing Claude Code and Codex support.

This is a purely additive change — no existing Claude Code or Codex functionality is altered.

Related: #289 (host integration contract RFC). This PR follows the existing Codex pattern for pragmatic incremental support. If the adapter abstraction from #289 lands, Copilot support can be migrated onto it — the changes here are scoped and self-contained.

## Motivation

The GitHub Copilot CLI went GA in February 2026 as a standalone binary, separate from the old `gh copilot` extension. It has its own config directory (`~/.copilot/`), session storage (`~/.copilot/session-state/{id}/`), and skill discovery. gstack should support it as a first-class host so users on Copilot CLI get the same experience as Claude Code and Codex users.

## Architecture decisions

- **Copilot shares `.agents/skills/` with Codex.** Both are non-Claude hosts using the same skill discovery path. Only `--host codex` is run during generation; Copilot-specific paths (`~/.copilot/skills/gstack`) are rewritten at `setup --host copilot` time via sed.
- **Build only runs `--host codex`, not `--host copilot`.** Running both would overwrite the same `.agents/skills/` directory. This avoids the conflict entirely.
- **Session scanning walks subdirectories.** Copilot CLI stores sessions as `~/.copilot/session-state/{session-id}/` directories containing `events.jsonl` and `workspace.yaml`, not flat files. `scanCopilot()` walks these subdirs.
- **E2E runner forwards auth.** Spawning `copilot -p` with a temp HOME loses login state. The runner copies `config.json` and `hosts.json` from the real `~/.copilot/` to the temp HOME, and `GH_TOKEN` passes through via env.

## Copilot CLI facts

| Aspect | Value |
|--------|-------|
| Binary | `copilot` (standalone, not `gh copilot`) |
| Install | `npm install -g @github/copilot` (also: brew, winget, install script) |
| Config directory | `~/.copilot/` (overridable via `COPILOT_HOME`) |
| Session data | `~/.copilot/session-state/{session-id}/` |
| Interactive mode | `copilot` |
| One-shot mode | `copilot -p "prompt"` |
| Requirements | Node.js 22+, npm 10+, GitHub Copilot subscription |
| Docs | https://docs.github.com/copilot/how-tos/set-up/install-copilot-cli |

[...]

PR #487, opened 2026-03-25
Add GitHub Copilot CLI support via Claude-compatible skills

ridermw:
## Summary
- add explicit `./setup --host copilot` support using Copilot CLI's documented `.claude/skills` compatibility
- document the lowest-friction Copilot install path and troubleshooting flow
- lock the new setup contract into the existing static setup tests

## Why this shape
GitHub Copilot CLI already supports skills from `.claude/skills` and `~/.claude/skills`, so this keeps the change set small and upstream-friendly. It avoids introducing a new Copilot-only generator mode while still making Copilot a documented, first-class supported host.

## Testing
- `bun test`
- automated Copilot smoke test with a temp `~/.claude/skills/gstack` install via `./setup --host copilot`
- verified Copilot skill discovery (`What skills do you have available?`)
- verified named skill invocation (`Use the /browse skill ...`)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -232,6 +232,11 @@ To add a browse command, add it to `browse/src/commands.ts`. To add a snapshot f
 
 gstack generates SKILL.md files for two hosts: **Claude** (`.claude/skills/`) and **Codex** (`.agents/skills/`). Every template change needs to be generated for both.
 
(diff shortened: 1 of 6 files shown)


garrytan:
Closing as dup of #396 — same author, same goal. Tracking there.

PR #493, opened 2026-03-25
fix(browse): resolve EADDRINUSE race condition on Windows via hardened polyfill and retry logic

### LINKED ISSUE #486: Windows: browse server fails to start — findPort() race condition in Node.js polyfill
## Summary

On Windows, `browse goto <url>` fails 100% of the time with "Server failed to start within 15s". The root cause is a race condition in `findPort()` when running under the Node.js polyfill (`bun-polyfill.cjs`).

## Environment

- Windows 11 Pro
- gstack v0.11.18.2
- Node.js v20.11.1
- Bun 1.3.11

## Root cause

`findPort()` in `server.ts` tests port availability by calling `Bun.serve({ port })` then immediately `testServer.stop()`. In real Bun, `Bun.serve()` is synchronous — the port binds and releases instantly.

In the Node.js polyfill, `Bun.serve()` wraps `http.createServer()` + `server.listen()`, which is **async**. `stop()` wraps `server.close()`, also async. Neither has completed by the time `findPort()` returns the port.

The timeline:
1. `findPort()` calls polyfill `Bun.serve({ port: X })` → queues async `listen()`
2. `testServer.stop()` → queues async `close()`
3. `findPort()` returns port X (neither listen nor close has executed yet)
4. `await browserManager.launch()` — event loop runs, test server's `listen()` **now binds port X**
5. Real `Bun.serve({ port: X })` → **EADDRINUSE** because the test server's `close()` hasn't completed

## Reproduction

Minimal script proving the race condition:

```javascript
// Save as test-race.js, run with: node test-race.js
const path = require('path');
const polyfillPath = path.join(
  process.env.USERPROFILE, '.claude', 'skills', 'gstack', 'browse', 'dist', 'bun-polyfill.cjs'
);
require(polyfillPath);

async function test() {
  const port = 10000 + Math.floor(Math.random() * 50000);
  // Simulate findPort():
  const testServer = Bun.serve({ port, fetch: () => new Response('ok') });
  testServer.stop();
  // Simulate browserManager.launch() delay:
  await new Promise(r => setTimeout(r, 500));
  // Simulate real server bind:
  const http = require('http');
  const s = http.createServer((req, res) => res.end('ok'));
  s.on('error', (e) => {
    console.log('FAILED:', e.message); // Always hits EADDRINUSE
    process.exit(1);
  });
  s.listen(port, '127.0.0.1', () => {
    console.log('SUCCESS');
    s.close();
  });
}
test();
```

This fails 100% of the time on Windows.

## Fix

Replace `Bun.serve()` port testing in `findPort()` with `net.createServer()` using proper async bind/close:

```javascript
async function findPort() {
  const net = require('net');
  function checkPort(port) {
      const srv = net.createServer();
      srv.once('error', (err) => reject(err));
      srv.listen(port, '127.0.0.1', () => {
        srv.close(() => resolve(port));
      });
    });
  }
  // ... same retry logic, using await checkPort(port) instead of Bun.serve()
}
```

This properly waits for both bind AND close to complete before returning the port.

Verified locally — 5/5 cold starts succeed, browse commands work normally.

## Related issues

- #448, #454, #458 — prior Windows browse fixes (stdio format, process detection)
- #77, #254 — earlier Bun/Windows pipe transport issues

This is distinct from the stdio array fix in v0.11.18.2 — that fixed the spawn format, this fixes the port selection race condition.

BemnetMussa:
Fixes #486

## Summary
This PR resolves #486 by addressing a **TOCTOU (Time-of-Check to Time-of-Use)** race condition in the `browse` server startup, specifically affecting Windows users running under the Node.js polyfill.

It introduces a "Mechanical Resilience" pattern that allows the server to gracefully recover from port collisions instead of crashing the daemon.

---

## The Problem
On Windows/Node.js, `server.listen()` is an asynchronous operation.

The current implementation uses `findPort()` to check for a free port, but due to the delay between the check and the actual `listen()` call, the port can be reclaimed by the OS or another process.

Because the Node.js polyfill handled this via an asynchronous `'error'` event rather than a synchronous exception, the error was going unhandled, leading to a process crash.

---

### 1. Polyfill Hardening (`browse/src/bun-polyfill.cjs`)
- Updated `Bun.serve` to return a `ready` Promise.
- Attached one-time `listening` and `error` listeners **before** the `listen()` call to ensure no events are missed.
- Implemented a "zero-leak" cleanup strategy that removes startup listeners once the server is bound or fails.

### 2. Retry Orchestration (`browse/src/server.ts`)
- Introduced `startServerWithRetry`.
- Instead of relying on pre-check validation (`findPort()`), the server now attempts to bind directly.
- On `EADDRINUSE`, the logic:
  - Catches the error
  - Increments the port
  - Retries (up to 5 attempts)

This ensures the `browse` daemon remains stable even under high concurrency or port contention.

---

## Technical Rigor & Evidence

I verified the fix by:
- Injecting a 10-second artificial delay in `server.ts`
- Manually occupying the selected port using a background script to force a collision

### Before Fix
```
node:events:496
throw er; // Unhandled 'error' event
```

### After Fix
```
[browse] Node HTTP server runtime error: listen EADDRINUSE: address already in use 127.0.0.1:45789
[browse] Port 45789 in use, retrying... (1/5)
[browse] Server running on http://127.0.0.1:45790 (PID: 46840)
```

---

## Impact

- **Zero New Dependencies**  
  Uses only native Node.js and Bun APIs.

- **Backward Compatible**  
  Native Bun runtime behavior remains unchanged.

- **Improved UX**  
  Eliminates a major cause of "Browse server won't start" issues on Windows.

### DIFF
--- a/browse/src/bun-polyfill.cjs
+++ b/browse/src/bun-polyfill.cjs
@@ -16,10 +16,11 @@ const { spawnSync, spawn } = require('child_process');
 globalThis.Bun = {
   serve(options) {
     const { port, hostname = '127.0.0.1', fetch } = options;
+    let currentPort = port;
 
     const server = http.createServer(async (nodeReq, nodeRes) => {
       try {
-        const url = `http://${hostname}:${port}${nodeReq.url}`;
+        const url = `http://${hostname}:${currentPort}${nodeReq.url}`;
         const headers = new Headers();
         for (const [key, val] of Object.entries(nodeReq.headers)) {
           if (val) headers.set(key, Array.isArray(val) ? val[0] : val);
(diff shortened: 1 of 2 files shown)


16francej:

PR #500, opened 2026-03-26
fix: generate Claude runtime under the install root

yasuhirofujii-medley:
## Summary
- generate Claude skill docs into the install root instead of reading directly from the source checkout
- link top-level Claude slash commands from the generated runtime root
- add setup regression coverage for runtime-root linking and repo-root generation

## Why
This keeps the Claude install self-contained under the chosen install root and makes the top-level skill links point at the runtime that setup actually generated.

## Testing
- bun test ./test/gen-skill-docs.test.ts --test-name-pattern 'setup script validation'


16francej:

PR #501, opened 2026-03-26
fix: infer Claude install root from setup location

yasuhirofujii-medley:
## Summary
- infer Claude global vs project-local installs from where setup lives instead of from the caller's pwd
- keep explicit --local and --global overrides
- add setup regression coverage for inferred install roots and override handling

## Why
Vendored project-local installs currently depend on the caller's working directory. Running setup from inside .claude/skills/gstack can create a nested .claude/skills/gstack/.claude/skills tree instead of installing into the project root.

This PR switches the default to path-based inference so:
- ~/.claude/skills/gstack/setup resolves to global Claude install
- PROJECT/.claude/skills/gstack/setup resolves to project-local Claude install

## Notes
- This branch currently includes the runtime-root commit from PR1 because both branches target upstream main.
- Review this after PR1, then it can be rebased down to the install-inference-only delta.

## Testing
- bun run gen:skill-docs --host codex
- bun test ./test/gen-skill-docs.test.ts --test-name-pattern 'setup script validation'

### DIFF
--- a/scripts/gen-skill-docs.ts
+++ b/scripts/gen-skill-docs.ts
@@ -22,6 +22,9 @@ import { generatePlanCompletionAuditShip, generatePlanCompletionAuditReview, gen
 
 const ROOT = path.resolve(import.meta.dir, '..');
 const DRY_RUN = process.argv.includes('--dry-run');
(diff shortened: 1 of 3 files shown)


16francej:

Issue #399, opened 2026-03-23
bug: package.json version out of sync with VERSION file on main


simonemacario:
PR #360 bumped `VERSION` to `0.11.10.0` but missed `package.json` (still `0.11.9.0`).

The existing test catches it:

```
gen-skill-docs > package.json version matches VERSION file
Expected: "0.11.10.0"
Received: "0.11.9.0"
```

One-line fix: #400

PR #523, opened 2026-03-26
fix: auto-upgrade marker no longer masks newer remote versions

### LINKED ISSUE #515: Auto-upgrade misses updates when just-upgraded-from marker is stale
## Bug

When a `just-upgraded-from` marker file persists across sessions, `gstack-update-check` short-circuits and never fetches the remote VERSION — even if the remote has moved far ahead.

## Reproduction

1. gstack auto-upgrades from 0.9.3.0 → 0.9.4.0 (writes `~/.gstack/just-upgraded-from` with `0.9.3.0`)
2. That session ends before the marker is consumed (or the marker persists for another reason)
3. 42 commits land on `origin/main`, pushing VERSION to 0.11.21.0
4. New session starts → `gstack-update-check` runs
5. **Step 2 (line 112-122)** finds the marker, outputs `JUST_UPGRADED 0.9.3.0 0.9.4.0`, writes `UP_TO_DATE 0.9.4.0` to cache, and **exits immediately**
6. Step 4 (remote fetch) never runs
7. User stays on 0.9.4.0, missing 42 commits

## Root Cause

`gstack-update-check` lines 112-122:

```bash
if [ -f "$MARKER_FILE" ]; then
  OLD="$(cat "$MARKER_FILE" 2>/dev/null | tr -d '[:space:]')"
  rm -f "$MARKER_FILE"
  rm -f "$SNOOZE_FILE"
  mkdir -p "$STATE_DIR"
  echo "UP_TO_DATE $LOCAL" > "$CACHE_FILE"
  if [ -n "$OLD" ]; then
    echo "JUST_UPGRADED $OLD $LOCAL"
  fi
  exit 0  # ← Never checks remote
fi
```

The script assumes "just upgraded = up to date" which isn't true if the remote advanced between the upgrade and the next session.

## Suggested Fix

After consuming the marker, still do the remote VERSION fetch (or at minimum, don't write `UP_TO_DATE` to cache — let it fall through to Step 4):

```bash
if [ -f "$MARKER_FILE" ]; then
  OLD="$(cat "$MARKER_FILE" 2>/dev/null | tr -d '[:space:]')"
  rm -f "$MARKER_FILE"
  rm -f "$SNOOZE_FILE"
  if [ -n "$OLD" ]; then
    echo "JUST_UPGRADED $OLD $LOCAL"
  fi
  # Don't exit — fall through to remote check in case
  # more updates landed since the upgrade
fi
```

## Environment

- gstack v0.9.4.0 → v0.11.21.0 (42 commits missed)
- macOS Darwin 24.6.0
- `auto_upgrade: true` in `~/.gstack/config.yaml`
- Global git install at `~/.claude/skills/gstack/`

## Workaround

Run `/gstack-upgrade` manually or `gstack-update-check --force` to bust the cache.

gregario:
## Problem

When a `just-upgraded-from` marker file persists across sessions, `gstack-update-check` short-circuits at Step 2 (line 112–122) — it writes `UP_TO_DATE` to cache and exits immediately, **never fetching the remote VERSION**. Users silently stay on old versions, potentially missing dozens of commits.

Reported in #515: a user upgraded from 0.9.3.0 → 0.9.4.0, then missed 42 commits because the marker caused Step 4 (remote fetch) to be skipped entirely.

## Root cause

```bash
# Step 2 — before fix
if [ -f "$MARKER_FILE" ]; then
  ...
  echo "UP_TO_DATE $LOCAL" > "$CACHE_FILE"  # ← premature
  ...
  exit 0  # ← never reaches remote check
fi
```

The script assumes "just upgraded = up to date", which isn't true if the remote advanced between the upgrade and the next session.

## Fix

Remove the `exit 0` and the premature `UP_TO_DATE` cache write. The script now:

1. Consumes and deletes the marker (same as before)
2. Clears the snooze file (same as before)
3. Outputs `JUST_UPGRADED` for the preamble (same as before)
4. **Falls through** to Step 3 (cache check) → Step 4 (remote fetch)

If no further updates exist, Step 4 writes `UP_TO_DATE` to cache as normal. If a newer version landed, the user sees both `JUST_UPGRADED` and `UPGRADE_AVAILABLE` in the same session.

## Changes

- `bin/gstack-update-check` — remove early exit and premature cache write from Step 2 (+2/−3 lines)
- `browse/test/gstack-update-check.test.ts` — add two regression tests:
  - **Path C2**: marker present + remote has newer version → outputs both notices, cache reflects `UPGRADE_AVAILABLE`
  - **Path C3**: marker present + remote matches local → outputs `JUST_UPGRADED`, cache reflects `UP_TO_DATE`

## Test plan

- [x] All 35 existing + 2 new tests pass (`bun test browse/test/gstack-update-check.test.ts`)
- [x] Full `bun test` passes (only pre-existing #399 version mismatch fails — unrelated)
- [x] Existing "just-upgraded clears snooze file" test still passes
- [x] Existing "outputs JUST_UPGRADED and deletes marker" test still passes (falls through to remote fetch, which writes `UP_TO_DATE` on unreachable remote)

Fixes #515

### DIFF
--- a/bin/gstack-update-check
+++ b/bin/gstack-update-check
@@ -113,12 +113,11 @@ if [ -f "$MARKER_FILE" ]; then
   OLD="$(cat "$MARKER_FILE" 2>/dev/null | tr -d '[:space:]')"
   rm -f "$MARKER_FILE"
   rm -f "$SNOOZE_FILE"
-  mkdir -p "$STATE_DIR"
-  echo "UP_TO_DATE $LOCAL" > "$CACHE_FILE"
   if [ -n "$OLD" ]; then
     echo "JUST_UPGRADED $OLD $LOCAL"
   fi
-  exit 0
+  # Don't exit — fall through to remote check in case
+  # more updates landed since the upgrade
 fi
 
 # ─── Step 3: Check cache freshness ──────────────────────────
--- a/browse/test/gstack-update-check.test.ts
+++ b/browse/test/gstack-update-check.test.ts
@@ -92,6 +92,35 @@ describe('gstack-update-check', () => {
     expect(cache).toContain('UP_TO_DATE');
   });
 
+  // ─── Path C2: Just-upgraded marker + newer remote ──────────
+  test('just-upgraded marker does not mask newer remote version', () => {
+    writeFileSync(join(gstackDir, 'VERSION'), '0.4.0\n');
+    writeFileSync(join(stateDir, 'just-upgraded-from'), '0.3.3\n');
+    writeFileSync(join(gstackDir, 'REMOTE_VERSION'), '0.5.0\n');
+
+    const { exitCode, stdout } = run();
+    expect(exitCode).toBe(0);
+    // Should output both the just-upgraded notice AND the new upgrade
+    expect(stdout).toContain('JUST_UPGRADED 0.3.3 0.4.0');
+    expect(stdout).toContain('UPGRADE_AVAILABLE 0.4.0 0.5.0');
+    // Cache should reflect the upgrade available, not UP_TO_DATE
+    const cache = readFileSync(join(stateDir, 'last-update-check'), 'utf-8');
+    expect(cache).toContain('UPGRADE_AVAILABLE 0.4.0 0.5.0');
+  });
+
+  // ─── Path C3: Just-upgraded marker + remote matches local ──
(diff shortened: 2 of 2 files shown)


gregario:
Merged via #552. Closing.

PR #526, opened 2026-03-26
feat: add Claude Code plugin marketplace support

rhiever:
## Summary

- **Plugin marketplace distribution.** gstack can now be installed via the Claude Code plugin system. Two commands: `/plugin marketplace add garrytan/gstack` then `/plugin install gstack@garrytan-gstack`. No git clone, no setup script, no bun dependency.
- **28 skills discovered correctly.** A new `skills/` directory contains symlinks to every existing skill directory. The plugin system requires `skills/<name>/SKILL.md` for discovery. Symlinks are followed during plugin caching, so skill content is self-contained in the installed copy. Verified with `claude plugin validate .` and live install testing.
- **Zero changes to existing files or structure.** All skill directories stay where they are. The `skills/` symlinks point back to them. Existing git clone + setup install path is completely unaffected.
- **README updated.** New "Claude Code Plugin Marketplace" section between the Codex install and "See it work." Documents the install/upgrade commands, the `gstack:` skill namespace, and the tradeoffs vs. the full git clone install.

## What works via plugin install

All 28 skill prompts load and are invocable as `/gstack:<name>`. Skills that are pure instructions work fully: `/office-hours`, `/plan-ceo-review`, `/plan-eng-review`, `/plan-design-review`, `/design-consultation`, `/review`, `/ship`, `/investigate`, `/cso`, `/retro`, `/document-release`, `/autoplan`, `/codex`, `/land-and-deploy`, `/setup-deploy`, `/connect-chrome`, `/gstack-upgrade`, `/unfreeze`.

## What doesn't work (and why)

The plugin system copies files but doesn't run build scripts. Three things break:

| Issue | Affected skills | Root cause |
|-------|----------------|------------|
| No browse binary | `/browse`, `/qa`, `/qa-only`, `/benchmark`, `/canary`, `/design-review`, `/setup-browser-cookies` | `browse/dist/browse` is compiled by `bun run build` during `./setup`. The plugin system has no post-install build hook. `node_modules` (Playwright, puppeteer-core) are also missing. |
| Hardcoded paths in preambles | All skills (non-fatal) | Every SKILL.md preamble references `~/.claude/skills/gstack/bin/*` for update checks, session tracking, and telemetry. Plugin cache uses a different path (`~/.claude/plugins/cache/...`). Fails silently due to `2>/dev/null \|\| true`, so skills still load fine. |
| Hook scripts unreachable | `/careful`, `/freeze`, `/guard` | These register `PreToolUse` hooks pointing to shell scripts at `~/.claude/skills/gstack/*/bin/*.sh`. The path doesn't exist in the plugin cache, so the safety hooks won't fire. |

## What would fix the remaining gaps

These are out of scope for this PR but would close the gap for full-parity plugin installs:

1. **Resolve paths via `${CLAUDE_PLUGIN_ROOT}`** instead of hardcoded `~/.claude/skills/gstack/`. This variable is set by Claude Code at runtime for plugin content. Would require updating the preamble resolver in `scripts/resolvers/preamble.ts` and the hook paths in skill frontmatter.
2. **Ship a pre-built binary or add a `SessionStart` hook** that runs `bun install && bun run build` on first use. Pre-built binaries are platform-specific (arm64 vs x86, macOS vs Linux), so a build-on-first-use hook may be more practical.

## Files added

| File | Purpose |
|------|---------|
| `.claude-plugin/plugin.json` | Plugin manifest: name, version (0.12.2.0), description, metadata |
| `.claude-plugin/marketplace.json` | Marketplace catalog: owner, plugin source (`./`), category, tags |
| `skills/*` (28 symlinks) | Plugin-compatible discovery paths pointing to existing skill directories |

## Test plan

- [x] `claude plugin validate .` passes
- [x] `claude --plugin-dir .` discovers all 28 skills as `/gstack:<name>`
- [x] Local path marketplace add + install discovers all 28 skills
- [x] Remote GitHub marketplace add + install discovers all 28 skills
- [x] Uninstall + marketplace remove cleans up completely
- [x] Existing `./setup` install path still works (no files moved or renamed)


apalevich:
It will make the skills much more convenient to install and update. Could you approve and merge please? @garrytan

tomclues:
Worth expecting ：）

PR #532, opened 2026-03-26
feat: add yc_alumni config to skip YC apply pitch

### LINKED ISSUE #502: Skip YC plug
Hi Garry 👋 ,

Found about this through my co-founder today.

We are already part of YC, so if possible I would like to skip the "apply to YC" plugs.

Is there an option somewhere, or the best course of action is forking the repo?

So far I like it, I only used once to ship code to prod but working on a new feature next.

Thank you for sharing this with us. 🙇

keshav55:
## Summary

Closes #502 — YC founders asked to skip the "apply to YC" plea in /office-hours.

### Change

Adds a config check before Beat 3 (Garry's Personal Plea) in /office-hours:

```bash
gstack-config set yc_alumni true
```

When set, the plea is skipped and the skill goes straight to next-skill recommendations.

### Why

From #502: "We are already part of YC, so if possible I would like to skip the 'apply to YC' plugs."

### Test plan

- [x] 365/365 skill validation tests pass
- [x] Config check uses existing `gstack-config` infrastructure

### DIFF
--- a/office-hours/SKILL.md
+++ b/office-hours/SKILL.md
@@ -1128,6 +1128,15 @@ One more thing.
 
 ### Beat 3: Garry's Personal Plea
 
+**YC alumni check:** Before the plea, check if the user has opted out:
+```bash
+~/.claude/skills/gstack/bin/gstack-config get yc_alumni 2>/dev/null || echo "false"
+```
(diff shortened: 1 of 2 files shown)


time-attack:
garry already rejected this @16francej  can you close

PR #533, opened 2026-03-26
fix: auto-upgrade respects auto_upgrade config

abhishekray07:
## Summary
- `gstack-update-check` now checks `auto_upgrade` config before outputting `UPGRADE_AVAILABLE`
- When `auto_upgrade: true`, attempts best-effort upgrade on **both** cache-hit and slow-fetch paths
- Skips if install dir has local modifications (dirty-tree guard)
- Failure falls through gracefully to manual prompt

## Changes
- **`bin/gstack-update-check`** — added `try_auto_upgrade()` with dirty-tree check, called from both output paths
- **7 new tests** covering both paths, failure modes, snooze priority, dirty tree

## Related
- #515 (stale marker — avoids worsening, doesn't fix)

## Test plan
- [x] All 40 update-check tests pass (33 existing + 7 new)

### DIFF
--- a/bin/gstack-update-check
+++ b/bin/gstack-update-check
@@ -99,6 +99,56 @@ check_snooze() {
   return 1  # snooze expired
 }
 
+# ─── Auto-upgrade helper ──────────────────────────────────
+# try_auto_upgrade <local_version> <remote_version>
+#   Attempts auto-upgrade if auto_upgrade config is true.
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej — recommend closing this PR in favor of #1942. It puts fetch, hard reset, and setup work into a passive update-check path, which can discard local changes. Current main already handles auto-upgrade elsewhere, and #1942 is the safer consolidation point.

PR #549, opened 2026-03-27
fix: resolve codex exec -C repo root eagerly to prevent wrong-project reviews (v0.12.6.0)

garrytan:
## Summary

When gstack runs in Conductor with multiple workspaces, `codex exec -C "$(git rev-parse --show-toplevel)"` evaluates in whatever cwd the background shell inherits, which may be a different project. Codex silently reviews the wrong codebase.

**Fix:** All 12 `codex exec` commands now resolve `_REPO_ROOT` eagerly at the top of each bash block and reference the stored value in `-C`.

### Bug Fix
- `codex/SKILL.md.tmpl` (3 occurrences): challenge, consult new session, consult resumed session
- `autoplan/SKILL.md.tmpl` (3 occurrences): CEO voice, design voice, eng voice
- `scripts/resolvers/review.ts` (3 occurrences): codex second opinion, adversarial step, plan review
- `scripts/resolvers/design.ts` (3 occurrences): design review lite, design sketch, design outside voices

### Dead Code Removal
- Removed 6 dead function definitions from `scripts/gen-skill-docs.ts` (695 lines). These were moved to `scripts/resolvers/` months ago but never deleted. They had diverged from the live versions and contained the old vulnerable pattern.

### Regression Test
- New test in `gen-skill-docs.test.ts` scans all `.tmpl` and resolver `.ts` source files for `codex exec` commands using inline `$(git rev-parse --show-toplevel)` in the `-C` flag. Prevents reintroduction.

## Pre-Landing Review
No issues found.

## Plan Completion
All items DONE: 12 occurrences fixed, dead code removed, regression test added.

## Test plan
- [x] `bun test` passes (skill validation + gen-skill-docs quality + regression test)
- [x] `bun run gen:skill-docs` regenerates all SKILL.md files successfully
- [x] `grep -rn 'codex exec' **/*.tmpl scripts/resolvers/*.ts | grep 'git rev-parse --show-toplevel'` returns zero results


samarri-sys:
The eager resolution pattern makes sense — race conditions in multi-workspace setups are exactly the kind of bug that's hard to reproduce and easy to dismiss. The regression test scanning for inline `$(git rev-parse)` in `-C` flags is a nice guardrail. Prevents the pattern from creeping back in during future contributions.

PR #555, opened 2026-03-27
fix: auth-gate /refs and /activity/* endpoints

ryankc33:
## Summary
- Require bearer auth on `/refs`, `/activity/stream`, and `/activity/history` endpoints — previously any local process or website could read browse state without authentication
- Remove `Access-Control-Allow-Origin: *` headers from those endpoints so web pages can no longer read responses via cross-origin fetch
- Add query-parameter token fallback to `validateAuth` for `EventSource` compatibility (it cannot set custom headers)
- Update Chrome extension callers (`background.js`, `sidepanel.js`) to pass auth tokens on all three endpoints

## Test plan
- [x] New `browse/test/endpoint-auth.test.ts` — 12 tests covering all three endpoints
- [x] Unauthenticated requests return 401
- [x] Wrong token returns 401
- [x] Valid bearer header returns 200
- [x] Query-parameter token returns 200 (EventSource compat)
- [x] `Access-Control-Allow-Origin` header is absent from authenticated responses
- [x] Existing `sidebar-integration.test.ts` still passes (13/13)
- [x] Existing `activity.test.ts` still passes (18/18)

### DIFF
--- a/browse/src/server.ts
+++ b/browse/src/server.ts
@@ -41,7 +41,14 @@ const IDLE_TIMEOUT_MS = parseInt(process.env.BROWSE_IDLE_TIMEOUT || '1800000', 1
 
 function validateAuth(req: Request): boolean {
   const header = req.headers.get('authorization');
-  return header === `Bearer ${AUTH_TOKEN}`;
+  if (header === `Bearer ${AUTH_TOKEN}`) return true;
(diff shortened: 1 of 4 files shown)


time-attack:
@16francej Please close this PR. The affected browser endpoints were secured more safely in #595 and later hardening. This old version can put auth tokens in URLs, where they may leak into logs, so it must not be merged.

PR #558, opened 2026-03-27
fix: use $HOME instead of ~ in preamble for Windows compatibility

### LINKED ISSUE #475: GStack’s persistence of user reading “Boil the Lake” essay is not cross-platform compatible
Every time I try to run a GStack skill on Windows, I am prompted to read the “Boil the Lake” essay again. Look, I get it… the first few times I did read it again, because I thought the GStack was trying to reinforce the message. 

After the fifth read through, though, I realized that I was being reminded each time because the GStack itself failed to persist the file that indicates I have in fact read the essay. 

I think the error lies here: 
```
User answered Claude's questions:
L • Want to open the Boil the Lake essay in your browser? → No thanks
en • Bash(touch ~/.gstack/.completeness-intro-seen)
L Running...
Bash command
touch ~/ .gstack/.completeness-intro-seen
Mark completeness intro as seen
```

Note that these forward slashes break in a win32 environment, so the file is never persisted. I actually don’t mind right now, because I still like to read the essay before embarking on a GStack marathon, but I’m sure in a few weeks I will be okay to skip it.

HMAKT99:
## Summary

- replace bare `~/.gstack/` with `"$HOME/.gstack/"` in all preamble bash blocks
- affects marker file checks, touch, mkdir, find, and analytics appends
- update zsh-safe glob test to match new path format
- prose/documentation references left as `~/.gstack/`

## Problem

Every gstack skill invocation on Windows re-prompts the user to read the "Boil the Lake" essay, re-asks about telemetry consent, and re-asks about proactive mode — even after the user already answered.

The preamble checks marker files with bare `~`:
```bash
_LAKE_SEEN=$([ -f ~/.gstack/.completeness-intro-seen ] && echo "yes" || echo "no")
```

Tilde expansion is not guaranteed in all Windows shell contexts (Git Bash subshells, MSYS2, CI runners). When `~` doesn't expand, the `[ -f` check always returns false and the marker is never found.

## Reproduction

1. Windows 10/11 with Git Bash
2. Run any gstack skill (e.g. `/office-hours`)
3. Answer the Boil the Lake, telemetry, and proactive prompts
4. Run another gstack skill
5. All three prompts appear again

## Root Cause

`touch ~/.gstack/.completeness-intro-seen` creates the file at a literal `~/.gstack/` path (or fails silently). The next invocation checks `[ -f ~/.gstack/.completeness-intro-seen ]` which also fails to expand `~`. The marker is never written or never found.

## Fix

Replace `~` with `"$HOME"` in all 15 executable bash paths in `scripts/resolvers/preamble.ts`. `$HOME` is set by the OS on every platform and expands reliably in all bash contexts.

Updated paths:
- `mkdir -p`, `touch`, `find` for sessions directory
- `[ -f ... ]` checks for `.completeness-intro-seen`, `.telemetry-prompted`, `.proactive-prompted`
- `>> ... skill-usage.jsonl` and `eureka.jsonl` appends
- `rm -f ... .pending-*` cleanup

## Verification

```
bun run gen:skill-docs --dry-run
  29 FRESH, 0 STALE

  563 pass, 1 fail (pre-existing: package.json version mismatch on main)
```

Fixes #475.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -21,13 +21,13 @@ allowed-tools:
 ```bash
 _UPD=$(~/.claude/skills/gstack/bin/gstack-update-check 2>/dev/null || .claude/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
 [ -n "$_UPD" ] && echo "$_UPD" || true
-mkdir -p ~/.gstack/sessions
-touch ~/.gstack/sessions/"$PPID"
-_SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
-find ~/.gstack/sessions -mmin +120 -type f -delete 2>/dev/null || true
+mkdir -p "$HOME/.gstack/sessions"
+touch "$HOME/.gstack/sessions/$PPID"
+_SESSIONS=$(find "$HOME/.gstack/sessions" -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
+find "$HOME/.gstack/sessions" -mmin +120 -type f -delete 2>/dev/null || true
 _CONTRIB=$(~/.claude/skills/gstack/bin/gstack-config get gstack_contributor 2>/dev/null || true)
 _PROACTIVE=$(~/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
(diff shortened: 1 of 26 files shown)


time-attack:
@16francej close pls

PR #564, opened 2026-03-27
fix: fall back to dirname/unknown when git context is absent

collinstraka-clov:
## Problem

`gstack-review-log` uses `set -euo pipefail`, which treats empty strings as unbound variables. When run outside a git repo (or in a repo with no remote configured), `gstack-slug` produces empty strings for `SLUG` and `BRANCH`, causing:

```
gstack-review-log: line 8: SLUG: unbound variable
gstack-review-log: line 9: BRANCH: unbound variable
```

This breaks any gstack review skill (e.g. `/autoplan`, `/plan-ceo-review`) when the working directory isn't a git repo with a remote.

## Fix

Two changes in `gstack-slug`:
1. Add `|| true` to the git commands so `set -e` doesn't abort when git context is missing
2. Fall back to `basename "$PWD"` for `SLUG` and `"unknown"` for `BRANCH` when the values are empty

## Repro

Run any gstack review skill in a directory with no git remote configured.

### DIFF
--- a/bin/gstack-slug
+++ b/bin/gstack-slug
@@ -6,10 +6,13 @@
 # Security: output is sanitized to [a-zA-Z0-9._-] only, preventing
 # shell injection when consumed via source or eval.
 set -euo pipefail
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej Please close this PR. Current main already handles missing Git context and includes stronger safe slug caching. This branch targets the old helper, so its useful idea has already been incorporated.

PR #586, opened 2026-03-27
feat: custom prefix + SKILL_NAME templating — /g-ship, /g-qa in one command (v0.12.12.0)

lucaslim:
## Summary

- **Custom prefix word.** Setup now offers 3 choices: short names (\`/qa\`), namespaced (\`/gstack-qa\`), or a custom word you type (\`g\`, \`gs\`, \`mytools\`). The dash separator is automatic.
- **\`{{SKILL_NAME}}\` template variable.** SKILL.md templates now use \`name: {{SKILL_NAME}}\` in frontmatter. The generator resolves this to the prefixed name at build time (e.g. \`name: g-ship\` when prefix is \`g\`). Claude Code uses the \`name:\` field as the slash command trigger — this is what makes \`/g-ship\` actually work, not just a renamed directory.
- **Live reload.** \`gstack-config set skill_prefix g\` regenerates all installed SKILL.md files immediately — no setup re-run needed.
- **\`--prefix <word>\` flag.** Takes a word argument now (\`--prefix g\`) instead of being a toggle. \`--no-prefix\` unchanged.
- **Universal cleanup.** Any prefix → any prefix cleans up old generated dirs first, using \`AUTO-GENERATED from SKILL.md.tmpl\` as the detection signal.

## What changed

| File | Change |
|------|--------|
| \`bin/gstack-config\` | Validation, empty string handling, backward compat (\`true\`→\`gstack\`), live regen trigger, \`readlink -f\` path resolution, \`|\` sed delimiter |
| \`setup\` | \`SKILL_PREFIX\` is now a string word; 3-option interactive prompt; \`validate_prefix()\`; \`cleanup_generated_skill_dirs()\` replaces 2 old cleanup functions; \`GSTACK_SKIP_REGEN=1\` prevents double regen |
| \`scripts/gen-skill-docs.ts\` | \`{{SKILL_NAME}}\` resolution from dir name + prefix; \`--output-root <dir>\` for installing to target dir without modifying source |
| \`scripts/resolvers/preamble.ts\` | Updated prefix instruction — handles arbitrary word, not just legacy \`"true"\` |
| 29 SKILL.md + templates | \`name: {{SKILL_NAME}}\` in all templates; regenerated with updated preamble text |
| \`test/gen-skill-docs.test.ts\` | 8 updated tests for string-based prefix, universal cleanup, 3-option prompt |

## How the name templating works

Every SKILL.md template has \`name: {{SKILL_NAME}}\` in its frontmatter. When \`gen-skill-docs.ts\` runs:

1. It derives the base name from the directory (\`ship\`, \`qa\`, etc.)
2. If a prefix is set (e.g. \`g-\`), it prepends it: \`g-ship\`
3. The generated \`SKILL.md\` gets \`name: g-ship\`
4. Claude Code reads that \`name:\` field and registers \`/g-ship\` as the trigger

Without this, renaming the directory to \`g-ship/\` would create the file path but Claude Code would still look for a different name. The \`name:\` field is the actual hook.

## Test plan

- [x] \`bun test\` — 591 pass, 0 fail
- [x] \`{{SKILL_NAME}}\` resolves correctly with and without prefix (gen-skill-docs tests)
- [x] \`gstack-config get skill_prefix\` returns \`gstack\` when stored value is \`true\` (backward compat)
- [x] \`gstack-config set skill_prefix g\` validates format and triggers regen
- [x] Setup 3-option prompt tested (choice 1/2/3, custom word input, validation)
- [x] Universal cleanup detects by \`AUTO-GENERATED\` header, not prefix string
- [x] \`GSTACK_SKIP_REGEN=1\` suppresses double regen when called from setup
- [x] PATH-invoked \`gstack-config\` resolves gstack dir via \`readlink -f\`

## Design doc

APPROVED by autoplan (CEO + Eng reviews): \`~/.gstack/projects/lucaslim-gstack/lucaslim-lucaslim-skill-prefix-frontmatter-sync-design-20260327-131400.md\`

9 MUST-FIX items from eng review — all addressed before coding began.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Type `gstack-config set skill_prefix g` and every gstack skill immediately switc
 ### Added
 
 - **Custom prefix word.** Setup now offers three choices: short names (`/qa`, `/ship`), namespaced (`/gstack-qa`, `/gstack-ship`), or a custom word you type yourself (`g`, `gs`, `mytools` — anything lowercase, 1-10 chars). The system appends the `-` separator.
+- **`{{SKILL_NAME}}` template variable.** SKILL.md templates now use `name: {{SKILL_NAME}}` in their frontmatter. The generator resolves this to the actual prefixed name at build time (e.g. `name: g-ship` when prefix is `g`). Claude Code uses the `name:` field to register the slash command trigger, so this is what makes `/g-ship` actually work — not just a renamed directory.
 - **Live prefix reload.** `gstack-config set skill_prefix <word>` regenerates all installed SKILL.md files immediately. Switching from `gstack` to `g` takes one command, not a full re-run of setup.
 - **`--prefix <word>` on setup.** The `--prefix` flag now takes a word argument (`--prefix g`) instead of being a toggle. `--no-prefix` still works as before.
 - **Universal skill cleanup.** Switching between any two prefix configurations (including custom words) now cleans up all previously generated dirs before regenerating. Detection uses the `AUTO-GENERATED from SKILL.md.tmpl` header in each file.


garrytan:
Won't merge, but thanks for using gstack\!

PR #587, opened 2026-03-27
fix(security): commit bun.lock to pin dependency versions

### LINKED ISSUE #566: Security: bun.lock is gitignored - unpinned dependencies resolved on every install/upgrade (similar to LiteLLM malware propagation)
### Summary
bun.lock is listed in .gitignore, which means every ./setup and every auto-upgrade resolves dependencies fresh from the npm registry using floating semver ranges. This is the same class of supply chain vulnerability that was exploited in the LiteLLM attack on March 24, 2026.

### The problem
package.json declares dependencies with ^ ranges:
"diff": "^7.0.0",
"playwright": "^1.58.2",
"puppeteer-core": "^24.40.0"

Since bun.lock is gitignored, bun install (called by ./setup) resolves the latest compatible version from npm on every run. A compromised version of any dependency (or transitive dependency) within the ^ range would be silently pulled and executed.
This is compounded by the auto-upgrade flow. When auto_upgrade: true, the upgrade runs git reset --hard origin/main && ./setup, which calls bun install && bun run build — re-resolving the full dependency tree mid-session with no user confirmation.

### LiteLLM parallel
Three days ago, TeamPCP compromised LiteLLM by stealing a CI/CD publish token and uploading malicious versions to PyPI. The attack harvested SSH keys, cloud credentials, Kubernetes tokens, and CI/CD secrets. LiteLLM's pipeline pulled an unpinned dependency (Trivy) which was the entry point.
The gstack install/upgrade path has the same structural vulnerability: unpinned deps in a build pipeline that executes during auto-update.

References:
https://snyk.io/articles/poisoned-security-scanner-backdooring-litellm/

boinger:
## Summary

- Remove `bun.lock` from `.gitignore` and commit the lockfile
- Pins all dependency versions deterministically — `bun install` no longer resolves from npm on every `./setup` or auto-upgrade

## Problem

`bun.lock` was gitignored. Every `./setup` and every auto-upgrade (`git reset --hard origin/main && ./setup`) resolved dependencies fresh from npm using floating `^` semver ranges. A compromised package within range would be silently pulled and executed.

This is the same class of supply chain vulnerability exploited in the [LiteLLM attack](https://snyk.io/articles/poisoned-security-scanner-backdooring-litellm/) (2026-03-24).

## Fix

Commit `bun.lock`. Now `bun install` reads the lockfile and installs exact pinned versions. Registry resolution only happens when dependencies are explicitly added or updated.

Closes #566

## Test plan

- [x] `bun install` produces deterministic output from the committed lockfile
- [x] `bun test` passes (610/611 — 1 pre-existing failure: VERSION/package.json mismatch on main)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ bin/gstack-global-discover
 .gstack-worktrees/
 /tmp/
 *.log
-bun.lock
 *.bun-build
 .env
 .env.local
--- a/bun.lock
+++ b/bun.lock
@@ -1,0 +1,196 @@
+{
+  "lockfileVersion": 1,
+  "configVersion": 1,
+  "workspaces": {
+    "": {
+      "name": "gstack",
+      "dependencies": {
+        "diff": "^7.0.0",
+        "playwright": "^1.58.2",
+        "puppeteer-core": "^24.40.0",
+      },
+      "devDependencies": {
+        "@anthropic-ai/sdk": "^0.78.0",
+      },
+    },
+  },
+  "packages": {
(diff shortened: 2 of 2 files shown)


garrytan:
Thanks for flagging this supply chain risk! `bun.lock` is now tracked in the repo (commit `9c4ae901`). All dependencies are pinned to exact versions. Appreciate the security awareness!

PR #612, opened 2026-03-29
fix: replace find -delete with find -exec rm to avoid Safety Net false positive

### LINKED ISSUE #474: Safety Net plugin blocks preamble cleanup commands
## Problem

Users running [claude-code-safety-net](https://github.com/kenryu42/claude-code-safety-net) alongside gstack hit false-positive blocks on two preamble commands:

1. `find ~/.gstack/sessions -mmin +120 -type f` with the delete flag — blocked by Safety Net's rule against find with delete
2. `rm -rf` operations outside cwd (e.g., `/tmp/` cleanup) — blocked when cwd is `~`

### Impact

Every skill invocation that runs the standard preamble will fail on the first blocked command, requiring manual workarounds. The browse daemon itself works fine once the preamble is bypassed.

### Suggested fix

Replace the blocked patterns with Safety Net-compatible alternatives:

```bash
# Instead of find with delete flag:
find ~/.gstack/sessions -mmin +120 -type f -exec rm {} +

# Or even simpler:
find ~/.gstack/sessions -mmin +120 -type f | xargs rm -f 2>/dev/null
```

These accomplish the same cleanup without triggering Safety Net's rules.

### Environment
- gstack (latest from cc-marketplace)
- Safety Net v0.8.1
- macOS (Darwin 25.4.0)

boinger:
## Summary

- Replace `find -delete` with `find -exec rm -f -- {} +` in preamble session cleanup
- Fixes false positive blocks for users running [claude-code-safety-net](https://github.com/kenryu42/claude-code-safety-net)
- 25 of 30 skills affected (5 have no preamble bash block)

## Details

The preamble runs on every skill invocation to clean up stale session files older than 2 hours. The `-delete` flag triggers Safety Net's rule against destructive find commands, blocking the entire skill before it starts.

`-exec rm -f -- {} +` is functionally identical: batches deletions, handles filenames with spaces, and the `--` separator prevents dash-prefixed filenames from being misinterpreted as flags.

Verified against Safety Net's [find.ts source](https://github.com/kenryu42/claude-code-safety-net/blob/main/src/core/analyze/find.ts): it blocks `-delete` and `-exec rm -rf` but not `-exec rm -f`. Ran the actual `analyzeFind()` function against both commands to confirm.

## Test plan

- [x] `bun test` passes (exit 0)
- [x] `bun run gen:skill-docs --dry-run` passes
- [x] No `find.*-delete` in any generated SKILL.md
- [x] `find -exec rm -f --` present in all 25 preamble-enabled skills
- [x] Safety Net detection test: old command blocked, new command passes, rm -rf control blocked

Fixes #474

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -24,7 +24,7 @@ _UPD=$(~/.claude/skills/gstack/bin/gstack-update-check 2>/dev/null || .claude/sk
 mkdir -p ~/.gstack/sessions
 touch ~/.gstack/sessions/"$PPID"
 _SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
-find ~/.gstack/sessions -mmin +120 -type f -delete 2>/dev/null || true
+find ~/.gstack/sessions -mmin +120 -type f -exec rm -f -- {} + 2>/dev/null || true
 _CONTRIB=$(~/.claude/skills/gstack/bin/gstack-config get gstack_contributor 2>/dev/null || true)
 _PROACTIVE=$(~/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
 _PROACTIVE_PROMPTED=$([ -f ~/.gstack/.proactive-prompted ] && echo "yes" || echo "no")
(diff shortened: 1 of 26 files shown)


boinger:
Closing — upstream fixed this in #640 (v0.13.4.0) by changing `find -delete` to `find -exec rm {} +`. Our PR added `-f` and `--` for defensive hardening, but the core Safety Net issue is resolved.

PR #615, opened 2026-03-29
security: track cookie-imported domains and scope cookie imports

halbert04:
## Summary

- Adds cookie origin tracking to `BrowserManager` — every cookie import path now records which domains were imported
- The `cookie-import-browser` direct CLI mode now requires either `--domain <domain>` (scoped) or `--all` (explicit opt-in) for non-interactive use
- The `cookie-import` (JSON file) path also tracks imported domains
- The `--all` flag works but emits a warning recommending `--domain` for tighter scoping

## Why this matters

When cookies are imported without domain scoping, the headless browser gets access to every authenticated session in the user's real browser — Gmail, GitHub, banking, corporate SSO. A prompt injection via web page content could instruct the agent to navigate to a sensitive domain and exfiltrate data.

By tracking which domains had cookies imported, downstream commands can restrict operations to those origins. This PR is the **foundation** for origin-pinned JS execution (separate PR) — the `BrowserManager` now knows which domains have imported cookies.

## Changes

| File | What changed |
|------|-------------|
| `browse/src/browser-manager.ts` | Added `cookieImportedDomains` set + `trackCookieImportDomains()`, `getCookieImportedDomains()`, `hasCookieImports()` methods |
| `browse/src/write-commands.ts` | `cookie-import-browser`: added `--all` flag with warning, domain tracking on import. `cookie-import`: added domain tracking |

## Backward compatibility

- `cookie-import-browser chrome --domain .example.com` — works as before, now also tracked
- `cookie-import-browser chrome` (no flags) — opens picker UI as before (unchanged)
- `cookie-import-browser chrome --all` — new explicit opt-in for all cookies
- `cookie-import cookies.json` — works as before, now also tracked

No breaking changes. The `--all` flag is additive.

## Test plan

- [ ] `cookie-import-browser chrome --domain .github.com` imports only github cookies and tracks the domain
- [ ] `cookie-import-browser chrome --all` imports all cookies with a warning
- [ ] `cookie-import-browser chrome` (no flags) opens the picker UI as before
- [ ] `cookie-import cookies.json` tracks domains from the JSON file
- [ ] `bm.hasCookieImports()` returns true after any import, false before

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -57,6 +57,9 @@ export class BrowserManager {
   private dialogAutoAccept: boolean = true;
   private dialogPromptText: string | null = null;
 
+  // ─── Cookie Origin Tracking ────────────────────────────────
+  private cookieImportedDomains: Set<string> = new Set();
+
   // ─── Handoff State ─────────────────────────────────────────
   private isHeaded: boolean = false;
   private consecutiveFailures: number = 0;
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks for building the cookie domain tracking foundation! Landed with co-author credit as part of the security architecture. Ships in the next release.

PR #617, opened 2026-03-29
security: add persistent command audit log to browse server

halbert04:
## Summary

- Every browse server command is now logged to `.gstack/browse-audit.jsonl` as append-only JSONL
- Each entry records: timestamp, command, args, page origin, duration, status, error, whether cookies are imported, and connection mode
- The audit log persists across server restarts (unlike the in-memory ring buffers)
- All writes are best-effort — audit failures never block command execution

## Why this matters

The browse server executes commands on behalf of AI agents — navigating pages, clicking elements, running JavaScript, importing cookies. When something goes wrong (unexpected navigation, data exfiltration, prompt injection), there's currently no persistent forensic trail. The in-memory ring buffers (console, network, dialog) are capped at 50K entries and lost on server restart.

The audit log creates a persistent record of every command the server processed. Example use cases:

```bash
# What commands ran in the last session?
cat .gstack/browse-audit.jsonl | jq .

# Which commands ran with imported cookies? (elevated risk)
cat .gstack/browse-audit.jsonl | jq 'select(.hasCookies == true)'

# What JS was executed and on which origins?
cat .gstack/browse-audit.jsonl | jq 'select(.cmd == "js" or .cmd == "eval")'

# What failed?
cat .gstack/browse-audit.jsonl | jq 'select(.status == "error")'
```

## Changes

| File | What changed |
|------|-------------|
| `browse/src/audit.ts` | **New file.** `initAuditLog()` and `writeAuditEntry()` — append-only JSONL writer with truncation |
| `browse/src/config.ts` | Added `auditLog` path to `BrowseConfig` (`.gstack/browse-audit.jsonl`) |
| `browse/src/server.ts` | Initialize audit log on startup, write entries on command success and error |
| `browse/src/browser-manager.ts` | Added `markCookiesImported()` and `hasCookieImports()` for the `hasCookies` audit field |
| `browse/src/write-commands.ts` | Call `markCookiesImported()` on `cookie-import` and `cookie-import-browser` |

## Design decisions

- **Append-only, never truncated by server.** The console/network/dialog logs are cleared on each server start. The audit log is not — it accumulates across sessions. Users can rotate or truncate it manually if it grows large.
- **Args truncated to 200 chars, errors to 300 chars.** Prevents the log from growing unboundedly if large payloads are passed (e.g., `js` with a long expression).
- **`hasCookies` boolean flag.** Makes it trivial to filter for elevated-risk commands — any command that ran while browser cookies were imported from the user's real browser.
- **Independent of other security PRs.** The `hasCookieImports()` tracking added here is minimal (just a boolean). It's compatible with but doesn't depend on #615 (domain-level tracking).

## Test plan

- [ ] Start browse server, run a few commands (`goto`, `snapshot`, `js`), verify `.gstack/browse-audit.jsonl` contains entries
- [ ] Import cookies, run commands, verify `hasCookies: true` in subsequent entries
- [ ] Trigger an error (e.g., `js` with invalid expression), verify error entry is logged
- [ ] Restart server, verify audit log is preserved (not cleared)
- [ ] Verify `jq` queries above work against the log format

### DIFF
--- a/browse/src/audit.ts
+++ b/browse/src/audit.ts
@@ -1,0 +1,65 @@
+/**
+ * Persistent command audit log — forensic trail for all browse server commands.
+ *
+ * Writes append-only JSONL to .gstack/browse-audit.jsonl. Unlike the in-memory
+ * ring buffers (console, network, dialog), the audit log persists across server
+ * restarts and is never truncated by the server. Each entry records:
+ *
+ *   - timestamp, command, args (truncated), page origin
+ *   - duration, status (ok/error), error message if any
+ *   - whether cookies were imported (elevated security context)
+ *   - connection mode (headless/headed)
+ *
+ * All writes are best-effort — audit failures never cause command failures.
+ */
+
+import * as fs from 'fs';
+
+export interface AuditEntry {
+  ts: string;
+  cmd: string;
+  args: string;
+  origin: string;
+  durationMs: number;
+  status: 'ok' | 'error';
+  error?: string;
+  hasCookies: boolean;
+  mode: 'launched' | 'headed';
+}
+
+const MAX_ARGS_LENGTH = 200;
+const MAX_ERROR_LENGTH = 300;
+
+let auditPath: string | null = null;
+
+export function initAuditLog(logPath: string): void {
+  auditPath = logPath;
+}
+
+export function writeAuditEntry(entry: AuditEntry): void {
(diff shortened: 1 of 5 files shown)


garrytan:
Thanks for the audit log! Persistent forensic trail for all commands. Landed with co-author credit. Ships in the next release.

PR #631, opened 2026-03-29
fix: add Write tool to sidebar agent + surface empty responses

sternryan:
## Summary

- Adds `Write` to `--allowedTools` in both `server.ts` (line 433) and `sidebar-agent.ts` (line 165), enabling the sidebar agent to create files (CSVs, exports, etc.) when users ask
- Surfaces a "Done (no text output)" notice in the sidebar UI when Claude finishes without producing text output, instead of leaving an empty response bubble
- Adds `.agent-empty` CSS class for the notice styling

Addresses the P1 TODO: "Sidebar agent needs Write tool + better error visibility"

## Changes

| File | Change |
|------|--------|
| `browse/src/server.ts` | Add `Write` to `--allowedTools` |
| `browse/src/sidebar-agent.ts` | Add `Write` to fallback `--allowedTools` |
| `extension/sidepanel.js` | Show "Done (no text output)" when `agent_done` fires with no `agentTextEl` |
| `extension/sidepanel.css` | Add `.agent-empty` style (muted, italic, 12px) |

## Test plan

- [x] `bun test test/skill-validation.test.ts test/gen-skill-docs.test.ts` -- 601/601 pass
- [ ] Manual: ask sidebar to "write a CSV file" and verify it creates the file
- [ ] Manual: send a message that produces only tool output (no text) and verify the "Done (no text output)" notice appears

### DIFF
--- a/browse/src/server.ts
+++ b/browse/src/server.ts
@@ -430,7 +430,7 @@ function spawnClaude(userMessage: string, extensionUrl?: string | null): void {
 
   const prompt = `${systemPrompt}\n\n<user-message>\n${escapedMessage}\n</user-message>`;
   const args = ['-p', prompt, '--model', 'opus', '--output-format', 'stream-json', '--verbose',
-    '--allowedTools', 'Bash,Read,Glob,Grep'];
+    '--allowedTools', 'Bash,Read,Write,Glob,Grep'];
   if (sidebarSession?.claudeSessionId) {
     args.push('--resume', sidebarSession.claudeSessionId);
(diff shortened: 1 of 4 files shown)


sternryan:
Closing as diverged. This targets browse/src/sidebar-agent.ts, which no longer exists after the v1.57 browse rewrite — the sidebar/write path is now handled by browser-skill-write.ts and a restructured extension. Would need a fresh implementation against current main rather than a rebase. Thanks!

PR #635, opened 2026-03-29
fix: bundle server into browse binary, eliminate fallback chain

sternryan:
## Summary

- The compiled `browse` binary now bundles `server.ts` via `bun build --compile`, eliminating the fragile `resolveServerScript()` fallback chain that caused bugs in v0.3.2
- When invoked with `--server`, the binary runs as the persistent Chromium daemon instead of the CLI -- the CLI spawns *itself* with `--server` rather than searching for `server.ts` on disk
- Dev mode (`bun run browse/src/cli.ts`) is unchanged and still uses `resolveServerScript()` via lazy resolution

## Technical approach

**The problem:** `resolveServerScript()` tries multiple filesystem paths to find `server.ts` at runtime (adjacent to `cli.ts`, relative to `execPath`, `BROWSE_SERVER_SCRIPT` env). In compiled binaries where `import.meta.dir` is inside `$bunfs`, this chain falls through to the `execPath`-relative check, which fails when the binary isn't installed in the expected `browse/dist/browse` directory structure.

**The fix:**

1. **Server mode gate** (top of `cli.ts`): Check `process.argv` for `--server`. When present, `import('./server')` triggers server.ts's top-level `start()` call. The CLI's `main()` is gated on `!IS_SERVER_MODE`.

2. **Compiled binary detection**: `IS_COMPILED = import.meta.dir.includes('$bunfs')` -- Bun's compiled binaries use a virtual filesystem prefix.

3. **startServer() priority chain**:
   - Windows: Node.js launcher (unchanged)
   - Compiled binary: `Bun.spawn([process.execPath, '--server'], ...)`
   - Dev mode: `Bun.spawn(['bun', 'run', getServerScript()], ...)`

4. **Build change**: Added `--external electron` to the browse binary build command. Since `server.ts` is now bundled, Bun follows the import chain through `playwright-core` which has an unused Electron loader dependency.

5. **Lazy server script resolution**: `resolveServerScript()` is now only called when actually needed (dev mode), not eagerly at module load time.

## Test plan

- [x] `bun run build` compiles successfully (377 modules bundled into browse binary)
- [x] `bun test test/skill-validation.test.ts test/gen-skill-docs.test.ts` -- 601 tests pass
- [x] `bun test browse/test/config.test.ts` -- 31 tests pass (includes `resolveServerScript` unit tests)
- [ ] Manual: verify `browse/dist/browse status` auto-starts server via `--server` flag
- [ ] Manual: verify `bun run dev status` still works via dev mode path
- [ ] Manual: verify `BROWSE_SERVER_SCRIPT=/path/to/server.ts browse status` still works

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -2,19 +2,38 @@
  * gstack CLI — thin wrapper that talks to the persistent server
  *
  * Flow:
- *   1. Read .gstack/browse.json for port + token
- *   2. If missing or stale PID → start server in background
- *   3. Health check + version mismatch detection
- *   4. Send command via HTTP POST
- *   5. Print response to stdout (or stderr for errors)
+ *   1. If invoked with --server, run in server mode (bundled server.ts)
+ *   2. Read .gstack/browse.json for port + token
+ *   3. If missing or stale PID → start server in background
+ *   4. Health check + version mismatch detection
+ *   5. Send command via HTTP POST
+ *   6. Print response to stdout (or stderr for errors)
  */
 
 import * as fs from 'fs';
 import * as path from 'path';
 import { resolveConfig, ensureStateDir, readVersionHash } from './config';
 
+// ─── Server Mode Gate ─────────────────────────────────────────
+// When the compiled binary is invoked with --server, it runs as the
+// persistent Chromium daemon instead of the CLI. This eliminates the
+// fragile resolveServerScript() fallback chain for finding server.ts
+// on disk — server code is bundled into the same binary via
+// bun build --compile.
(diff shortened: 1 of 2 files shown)


sternryan:
Closing as diverged — this is based on v0.3.2 and browse/src/cli.ts was rewritten in the v1.57 line, so the PR conflicts and won't rebase mechanically. Worth noting the underlying issue is still real: current cli.ts retains the resolveServerScript() fallback chain this PR set out to eliminate. If useful I can re-file a fresh, smaller PR against current main that bundles server.ts into the compiled binary. Thanks!

PR #646, opened 2026-03-30
fix: codex review broken — [PROMPT] and --base are mutually exclusive

ziyangliu-666:
## Summary

`/codex review` has been completely broken since codex CLI started enforcing mutual exclusivity between `[PROMPT]` and `--base`. Every invocation fails with:

```
```

**Root cause:** The skill template passes both a prompt string (filesystem boundary instructions) and `--base <branch>` to `codex review`, but the codex CLI (v0.117.0) treats these as mutually exclusive arguments.

**Fix:**
- **Default reviews** (no custom instructions): use `--base <base>` only — codex's built-in review handles the diff correctly without a prompt
- **Custom instructions** (e.g., `/codex review focus on security`): use `[PROMPT]` only, with inline diff instructions ("Review changes against `<base>`. Run `git diff origin/<base>`...") so codex knows what to diff

Also removes `--enable web_search_cached` from all 3 modes (review, challenge, consult) — this feature flag is deprecated in codex v0.117.0 (`codex features list` shows `deprecated`).

## Tested

Both paths verified on codex v0.117.0:

| Scenario | Before | After |
|----------|--------|-------|
| `codex review --base main` (default) | `error: [PROMPT] cannot be used with --base` | exit 0, clean review |
| `codex review "custom prompt"` (with instructions) | `error: [PROMPT] cannot be used with --base` | exit 0, full review with 4×P1 + 1×P2 findings |

## Test plan
- [x] `codex review --base main` — exits 0, reports "no changes" (expected on main)
- [x] `codex review "custom prompt"` — exits 0, produces full review output
- [x] `bun test` — 507 pass, 16 fail (pre-existing, identical on clean main)
- [x] `bun run gen:skill-docs` — regenerates cleanly

### DIFF
--- a/codex/SKILL.md
+++ b/codex/SKILL.md
@@ -468,26 +468,30 @@ Run Codex code review against the current branch diff.
 TMPERR=$(mktemp /tmp/codex-err-XXXXXX.txt)
 ```
 
-2. Run the review (5-minute timeout). **Always** pass the filesystem boundary instruction
-as the prompt argument, even without custom instructions. If the user provided custom
-instructions, append them after the boundary separated by a newline:
+2. Run the review (5-minute timeout). Use `timeout: 300000` on the Bash call.
+
+**IMPORTANT:** `codex review` does NOT allow `[PROMPT]` and `--base` together — they are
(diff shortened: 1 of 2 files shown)


16francej:

PR #654, opened 2026-03-30
feat: --no-browser setup flag and version v0.13.11.0

jimmckeeth:
Merge latest upstream and add --no-browser setup flag with config persistence. Skip building the browse binary and installing Playwright Chromium with ./setup --no-browser. The choice persists to ~/.gstack/config.yaml so subsequent runs auto-detect the preference.


time-attack:
Thanks for this. The branch mixes the browser flag with unrelated Copilot changes, edits generated files rather than their templates, and does not prove reversible persisted state. Isolate the browser state change on the current host registry and templates, then add isolated HOME tests for disable, persistence, re-enable, and ownership-safe removal.

PR #657, opened 2026-03-30
feat: Windows setup.ps1 with pre-requisites and delegates to ./setup via Git Bash

jimmckeeth:
Windows installer PowerShell script that encapsulates `./setup` via Git Bash. 

On Windows, you can use the PowerShell wrapper instead of `./setup`. It auto-installs `Node.js`, `Bun`, and `Git` via `winget` if they're missing, then delegates to `./setup` inside Git Bash. Git Bash has the same home path as Windows (unlike WSL) so the sills install when run outside Git Bash too.

```powershell
.\setup.ps1 --keep-open --auto
#              ^          ^ Passed to setup
#              | Handled by setup.ps1
```

The window stays open for 20 seconds after setup completes so you can read the output. Control this with flags:

| Flag            | Behavior                                                 |
| ----------------| -------------------------------------------------------- |
| *(default)*     | Prompts "Autoclose [Y/n]" — auto-closes after 20s        |
| `--keep-open`   | Drops to an interactive bash shell after setup completes |
| `--auto-close`  | Closes the window immediately on completion              |

All other flags (`--no-browser`, `--prefix`, `--no-prefix`, `--host`, etc.) are forwarded to `./setup`.


time-attack:
@jimmckeeth Rewrite the wrapper to pass arguments without interpolating them into shell source, and remove unrelated stale changes.

PR #660, opened 2026-03-30
fix(setup): make gstack skills discoverable and usable in Factory Droid

### LINKED ISSUE #661: Factory Droid skill discovery is broken for gstack installs
## Problem

Skills installed via `cd ~/gstack && ./setup --host factory` are not reliably discoverable by Factory Droid. The generated skill files exist on disk, but Droid cannot discover or load them. As a result, gstack skills do not appear in the `/` menu.

## Reproduction

```bash
cd ~/gstack && ./setup --host factory
# Then restart Droid and type /
# Expected: gstack skills (qa, review, ship, etc.) appear
# Actual: no gstack skills visible
```

## Root Cause

The installation produces three incompatibilities with Droid's discovery model:

### 1. Wrong symlink type

Droid follows only **relative** symlinks in `~/.factory/skills/`. The current setup creates **absolute** symlinks:

```bash
ln -snf /Users/will/gstack/.factory/skills/gstack-qa ~/.factory/skills/gstack-qa
```

Absolute symlinks are ignored by Droid.

### 2. Skills not in Droid's read directory

Droid reads skill files from `~/.agents/skills/`. The current setup installs skills to `~/gstack/.factory/skills/` (a project subdirectory) and creates symlinks from `~/.factory/skills/` pointing back into the repo. Even if the symlinks were correct, Droid would need skills to exist as real directories in `~/.agents/skills/`.

### 3. Wrong `sourceType` in lockfile

The lockfile registration uses `sourceType: "local"`:

```json
{
  "sourceType": "local",
  "sourceUrl": "file:///Users/will/gstack/.factory/skills/gstack-qa"
}
```

Droid's lockfile reader **completely ignores** entries with `sourceType: "local"`. Every working skill (57 total in the registry) uses `sourceType: "github"`.

## How Droid Discovers Skills

Droid uses a two-tier discovery system:

- `~/.factory/skills/` — Droid scans here, follows relative symlinks
- `~/.agents/skills/` — Droid reads skill files from here
- `~/.agents/.skill-lock.json` — Droid's skill registry (only sourceType=github loaded)

A skill appears in `/` when **both**:
1. A relative symlink exists in `~/.factory/skills/` pointing to `~/.agents/skills/{skill}`
2. The lockfile entry has `sourceType: "github"`

## Proposed Fix

The `link_factory_skill_dirs()` function needs to:

1. **Copy** skills as real directories into `~/.agents/skills/` (not symlink from repo)
2. **Create relative symlinks** in `~/.factory/skills/` pointing to `../../.agents/skills/{skill}`
3. **Register** in `~/.agents/.skill-lock.json` with `sourceType: "github"`

Note: On a fresh profile (no existing `~/.agents/.skill-lock.json`), the lockfile registration would be skipped unless the existing lockfile check is made conditional or the lockfile is created.

## Verification

After a fix is applied, these commands should show the correct state:

```bash
# Relative symlinks exist
ls -l ~/.factory/skills/ | grep gstack

# Real directories exist
ls -l ~/.agents/skills/ | grep gstack

# Lockfile entries use sourceType=github
python3 -c "
import json
with open('$HOME/.agents/.skill-lock.json') as f:
    lock = json.load(f)
for name, entry in lock['skills'].items():
    if name.startswith('gstack'):
        print(f'{name}: sourceType={entry[\"sourceType\"]}')"
```

## Skill Name

After fixing, skills are invoked by their `name:` field in `SKILL.md`, not the directory name:

| Directory | name field | Invoked as |
|-----------|-----------|------------|
| gstack-qa | qa | /qa |
| gstack-review | review | /review |

## References

- Droid skill discovery: `~/.factory/skills/` (relative symlinks only)
- Droid skill files: `~/.agents/skills/` (real directories)
- Droid registry: `~/.agents/.skill-lock.json` (sourceType=github required)

morluto:
## Problem

`./setup --host factory` generated Factory-specific skills correctly, but installed the per-skill entries in `~/.factory/skills` as directory symlinks. Droid does not reliably discover those symlinked directories, so gstack skills may be missing from the `/` menu.

## Rewrite

This branch is rebuilt on current `main`, resolving the PR's merge conflicts and retaining the newer runtime-root and cross-platform setup behavior.

Factory skills are now installed as real directories directly in Factory's documented user skill root:

```
~/.factory/skills/gstack-*
```

Each setup run refreshes those installer-owned directories from the generated `.factory/skills/gstack-*` output. The root `gstack` runtime directory remains managed by `create_factory_runtime_root()`.

The rewrite deliberately does not:

- create or write `~/.agents/skills`
- create or modify `~/.agents/.skill-lock.json`
- register local files as fake GitHub-sourced skills
- share Factory-generated files with another host's namespace

## Validation

- `bash -n setup`
- `bun test test/gen-skill-docs.test.ts test/setup-windows-fallback.test.ts` — 411 pass
- isolated install smoke test verifies:
  - an old directory symlink is replaced by a real directory
  - rerunning setup refreshes the copied skill
  - no `~/.agents` directory is created

Closes #661.

### DIFF
--- a/setup
+++ b/setup
@@ -912,7 +912,7 @@ link_factory_skill_dirs() {
   local gstack_dir="$1"
   local skills_dir="$2"
   local factory_dir="$gstack_dir/.factory/skills"
-  local linked=()
+  local installed=()
 
   if [ ! -d "$factory_dir" ]; then
     echo "  Generating .factory/ skill docs..."
@@ -929,14 +929,17 @@ link_factory_skill_dirs() {
       skill_name="$(basename "$skill_dir")"
       [ "$skill_name" = "gstack" ] && continue
       target="$skills_dir/$skill_name"
-      if [ -L "$target" ] || [ ! -e "$target" ]; then
-        _link_or_copy "$skill_dir" "$target"
-        linked+=("$skill_name")
-      fi
+      # Factory documents ~/.factory/skills as its user-level skill root. Keep
+      # the install self-contained there and use a real directory so Droid does
+      # not have to follow a directory symlink during discovery. These gstack-*
+      # paths are installer-owned and are refreshed on every setup run.
+      rm -rf "$target"
+      cp -R "$skill_dir" "$target"
+      installed+=("$skill_name")
     fi
   done
-  if [ ${#linked[@]} -gt 0 ]; then
-    echo "  linked skills: ${linked[*]}"
+  if [ ${#installed[@]} -gt 0 ]; then
+    echo "  installed skills: ${installed[*]}"
   fi
 }
 
--- a/test/gen-skill-docs.test.ts
+++ b/test/gen-skill-docs.test.ts
@@ -2450,6 +2450,19 @@ describe('setup script validation', () => {
     expect(fnBody).not.toContain('_link_or_copy "$gstack_dir" "$codex_gstack"');
   });
 
+  test('Factory installs real skill directories only in its documented user root', () => {
+    const fnStart = setupContent.indexOf('link_factory_skill_dirs()');
+    const fnEnd = setupContent.indexOf('\nlink_opencode_skill_dirs()', fnStart);
+    const fnBody = setupContent.slice(fnStart, fnEnd);
+
+    expect(fnBody).toContain('target="$skills_dir/$skill_name"');
+    expect(fnBody).toContain('rm -rf "$target"');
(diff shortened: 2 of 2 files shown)


time-attack:
@morluto Rewrite this to resolve current conflicts and avoid writing skills into an undocumented user-level agents directory.

PR #678, opened 2026-03-30
feat: add native OpenCode host support

gitfudge0:
Generate and install OpenCode-native skills under .opencode and ~/.config/opencode so gstack works without Claude-specific paths. Update setup, uninstall, platform detection, docs, worktree copying, and tests to keep the host-specific runtime layout consistent.


garrytan:
Thanks @gitfudge0 — OpenCode host shipped in `hosts/opencode.ts`. Closing as superseded.

PR #679, opened 2026-03-30
Match skill responses to the user's language

shrektan:
This adds a small but explicit language rule to the shared skill voice preamble:

- match the user's language by default
- if the user is writing mostly in Chinese, respond in Chinese
- only switch languages when the user asks, or when the source material is clearer in its original language

Why:
Many skills have long English instructions, which can unintentionally override the user's language context. In practice, that can make the agent reply in English even when the user is clearly interacting in Chinese.

Implementation:
- update the shared voice generator in `scripts/resolvers/preamble.ts`
- regenerate affected `SKILL.md` files
- add a validation test so the rule does not regress

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -202,6 +202,8 @@ This only happens once per project. If `HAS_ROUTING` is `yes` or `ROUTING_DECLIN
 
 ## Voice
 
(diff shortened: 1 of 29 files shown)


time-attack:
Thanks for this. The patch targets the former voice resolver, and its presence-only coverage does not address mixed-language prompts, explicit overrides, or preservation of technical text. Implement the rule in the current voice resolver, regenerate all hosts, and test those cases while keeping commands, paths, code, and quotations unchanged.

PR #684, opened 2026-03-30
fix: make /ship reruns resume safely

### LINKED ISSUE #649: fix: /ship Steps 7-8 not idempotent — re-run after partial failure corrupts VERSION and CHANGELOG
## Problem

`/ship` executes a multi-step sequence: version bump → changelog → commit → push → PR create. If Step 7 (`git push`) succeeds but Step 8 (`gh pr create`) fails (GitHub API outage, auth token expiry, rate limiting), re-running `/ship` corrupts state:

- **VERSION gets double-bumped** (e.g., 0.13.5 → 0.13.6 on first run, then 0.13.6 → 0.13.7 on re-run)
- **CHANGELOG gets duplicate entries** (same changes listed under two versions)
- No way to safely resume from where it left off

The dangerous window is narrow (one step) but the blast radius is real — manual cleanup is required.

## Why this matters

`/land-and-deploy` already handles partial failure well (Step 6 offers revert, Step 8 has full revert procedure with branch protection awareness). `/ship` doesn't apply the same rigor despite being the more frequently used skill.

## Suggested fix

Make the critical section idempotent rather than building a rollback framework:

1. **Before push:** check if branch is already pushed (`git rev-parse @{u}` or `git ls-remote`)
2. **Before PR create:** check if a PR already exists for the branch (`gh pr list --head <branch>`)
3. **Before version bump:** check if VERSION was already bumped in a prior run (compare against base branch)

This is ~10 lines of shell, not a new subsystem.

## Reproduction

```bash
# 1. Run /ship on a feature branch
# 2. Simulate gh failure: revoke token or disconnect network after push
# 3. Re-run /ship
# 4. Observe: VERSION bumped again, CHANGELOG has duplicate section
```

## Context

- `/land-and-deploy` already handles this pattern (Steps 6, 7, 8 all have failure recovery)
- Issue #341 covers cross-session resume but not within-session partial failure
- The agent IS the resume mechanism — but it can't safely resume if the steps aren't idempotent

AliFozooni:
## Summary
- add a rerun-safety checkpoint before the `/ship` version bump so a failed PR-creation step resumes instead of double-bumping `VERSION`
- teach the push and PR/MR steps to detect already-pushed branches and existing PRs/MRs before recreating them
- add regression coverage for the generated ship skill text so the resume-safe behavior stays locked in

## Problem
A partial failure between push and PR creation can currently leave `/ship` in a state where rerunning the workflow bumps `VERSION` again and duplicates the top `CHANGELOG` entry.

## Testing
- `git diff --check`
- `node` verification of the new resume-safety strings in `ship/SKILL.md`, `ship/SKILL.md.tmpl`, and `test/gen-skill-docs.test.ts`

Fixes #649

### DIFF
--- a/ship/SKILL.md
+++ b/ship/SKILL.md
@@ -1753,9 +1753,49 @@ staleness detection: if those files are later deleted, the learning can be flagg
 **Only log genuine discoveries.** Don't log obvious things. Don't log things the user
 already knows. A good test: would this insight save time in a future session? If yes, log it.
 
+## Step 3.9: Resume / rerun safety
+
+`/ship` must be restartable after partial failure. If Step 7 succeeds and Step 8 fails, a re-run must resume from the first incomplete step. Never double-bump `VERSION`, never duplicate the top `CHANGELOG` entry, and never open duplicate PRs/MRs.
+
+Capture the current release state:
+
+```bash
+BRANCH=$(git branch --show-current)
+SHIP_METADATA_COMMIT=$(git log <base>..HEAD --grep='^chore: bump version and changelog' --max-count=1 --format='%H' 2>/dev/null || true)
+REMOTE_HEAD=$(git ls-remote --heads origin "$BRANCH" | awk '{print $1}')
+```
+
+**If GitHub:**
+```bash
+HEAD_OWNER=$(gh repo view --json nameWithOwner -q .nameWithOwner 2>/dev/null | cut -d/ -f1)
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej Please close this PR. Safe resume behavior for repeated ship runs already shipped in #693 and was hardened further on main. This older branch relies on fragile commit-message guesses and should not be merged.

PR #686, opened 2026-03-30
fix: patch SKILL.md name: field when --prefix is used

### LINKED ISSUE #620: /setup --prefix bug

garagon:
## Summary

`./setup --prefix` creates symlinks (`gstack-qa → gstack/qa`) but doesn't update the `name:` frontmatter field inside each SKILL.md. Claude Code uses `name:` for skill registration — so the skill registers as `/qa` instead of `/gstack-qa`, making the prefix mode non-functional.

After creating each symlink, `setup` now patches the `name:` field to match the linked name. Same fix applied to `bin/gstack-relink` for prefix changes after initial setup.

## Root cause

`link_claude_skill_dirs()` reads `name:` from SKILL.md (line 279) and computes the correct `link_name` with prefix (line 282-288), but only uses it for the symlink target — the SKILL.md `name:` field stays unchanged. Claude Code ignores the symlink name and reads `name:` from frontmatter.

## What changed

- **`setup`**: after `ln -snf`, patches `name:` in SKILL.md when `link_name != skill_name`
- **`bin/gstack-relink`**: same patch logic for prefix changes after setup

Uses portable sed (tmpfile + mv) — works on both macOS and Linux. No-op for unprefixed mode and already-prefixed skills (`gstack-upgrade`).

## Verified

```
# Before (qa/SKILL.md):
name: qa          ← Claude registers as /qa

# After setup --prefix:
name: gstack-qa   ← Claude registers as /gstack-qa
```

## Test plan

- [x] `sed` patch correctly rewrites `name: qa` → `name: gstack-qa`
- [x] Already-prefixed skills (`gstack-upgrade`) are unaffected (link_name == skill_name)
- [x] Unprefixed mode (`--no-prefix`) does not trigger the patch
- [x] Switching prefix modes via `gstack-relink` patches in both directions

Fixes #620

### DIFF
--- a/bin/gstack-relink
+++ b/bin/gstack-relink
@@ -45,6 +45,10 @@ for skill_dir in "$INSTALL_DIR"/*/; do
   case "$skill" in bin|browse|design|docs|extension|lib|node_modules|scripts|test|.git|.github) continue ;; esac
   [ -f "$skill_dir/SKILL.md" ] || continue
 
+  # Read current name: from frontmatter
+  current_name=$(grep -m1 '^name:' "$skill_dir/SKILL.md" 2>/dev/null | sed 's/^name:[[:space:]]*//' | tr -d '[:space:]')
+  [ -z "$current_name" ] && current_name="$skill"
+
   if [ "$PREFIX" = "true" ]; then
     # Don't double-prefix directories already named gstack-*
     case "$skill" in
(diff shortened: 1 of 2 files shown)


garagon:
Closing — superseded by `bin/gstack-patch-names` (called from `bin/gstack-relink:84` and `setup:783`). The `name:` field rewrite this PR open-coded inside `gstack-relink` is now its own helper script invoked at both setup time and relink time, so the prefix change stays consistent across re-runs. Closing as obsolete.

PR #703, opened 2026-03-31
[codex] fix plan reviews to prefer repo-local design docs

### LINKED ISSUE #682: improvements when single branch uses multiple projects
see if gstack can be improved when a single branch host multiple products as we do. 

LOGS::::::

when i started plan-eng-review skill i got message "That design doc is for Entropy Chess, not Memory Log. Not relevant here. Let me investigate the two Android issues directly."

why that happened? why it did not find design doc for memorylog?

Find DESIGN.md references
Find DESIGN.md references

Good question. The agent found the root cause.

Short answer: The /plan-eng-review skill doesn't know which product you're in. It only knows the monorepo name and the git branch. So when it looked for a cached design doc, it grabbed the most recent one for production branch across the entire monorepo, which happened to be an Entropy Chess design doc from a previous session.

The mechanism:

The skill locates design docs like this:

SLUG=moblizeit-code-repo    # monorepo-level, not product-level
BRANCH=production
ls -t ~/.gstack/projects/$SLUG/*-$BRANCH-design-*.md | head -1

It found ~/.gstack/projects/moblizeit-code-repo/moblizeit-production-design-20260328-085239.md, which was an Entropy Chess admin dashboard design doc from a previous session. Since all your products share the same repo and the same production branch, the skill can't tell them apart.

What it should have done: Read the local products/memorylog/DESIGN.md that's right there in the product directory. But the skill only checks ~/.gstack/projects/ for design docs generated by /office-hours, not the product's own DESIGN.md file.

Why it didn't break anything: The skill correctly detected the mismatch ("That design doc is for Entropy Chess, not Memory Log") and ignored it. So it recovered gracefully, but the initial confusion is real.

The fix would be in gstack itself... the slug system needs to be product-aware in monorepos. Either index design docs by {slug}-{product} instead of just {slug}, or fall back to reading the local DESIGN.md from the current working directory. That's a gstack bug, not a project bug.

malikrohail:
## Summary
- prefer repo-local `DESIGN.md` / `design-system.md` before `~/.gstack/projects/...` history in plan review design-doc discovery
- apply the same fallback after `/office-hours` prerequisite runs so the resumed review stays scoped to the current subproject
- add a generated-skill regression test for the repo-local fallback

## Root Cause
Plan review design-doc discovery only looked in `~/.gstack/projects/$SLUG/...`, which is repo-scoped rather than subproject-scoped. In monorepos or multi-product branches, that could pick up the wrong historical design doc even when the current product already had a local `DESIGN.md`.

## Validation

Fixes #682.

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -517,8 +517,14 @@ After /office-hours completes, re-run the design doc check:
 setopt +o nomatch 2>/dev/null || true  # zsh compat
 SLUG=$(~/.claude/skills/gstack/browse/bin/remote-slug 2>/dev/null || basename "$(git rev-parse --show-toplevel 2>/dev/null || pwd)")
 BRANCH=$(git rev-parse --abbrev-ref HEAD 2>/dev/null | tr '/' '-' || echo 'no-branch')
-DESIGN=$(ls -t ~/.gstack/projects/$SLUG/*-$BRANCH-design-*.md 2>/dev/null | head -1)
-[ -z "$DESIGN" ] && DESIGN=$(ls -t ~/.gstack/projects/$SLUG/*-design-*.md 2>/dev/null | head -1)
+if [ -f DESIGN.md ]; then
+  DESIGN="$PWD/DESIGN.md"
+elif [ -f design-system.md ]; then
+  DESIGN="$PWD/design-system.md"
+else
+  DESIGN=$(ls -t ~/.gstack/projects/$SLUG/*-$BRANCH-design-*.md 2>/dev/null | head -1)
+  [ -z "$DESIGN" ] && DESIGN=$(ls -t ~/.gstack/projects/$SLUG/*-design-*.md 2>/dev/null | head -1)
+fi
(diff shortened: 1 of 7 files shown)


time-attack:
@16francej — recommend closing this PR as superseded by #1125, which owns the unified project-artifact identity design. This patch changes reader preference but does not address writer identity, nested discovery, or multi-repository grouping. Please preserve contributor credit and the useful repo-local document preference as input to #1125.

garrytan:
Fixed in fork-port wave 2: plan reviews now prefer repo-local design docs, and design docs land in the repo as decision records (instead of only `~/.gstack`). The discovery block is a single shared resolver across the plan-review skills, so they can't drift.

Commits: `4a34bd7d` feat(office-hours): design docs land in the repo, written as decision records (#703, #2000); `808344ec` (one source for the design-doc discovery block).
Pinned by: the template carve-guards (`test/helpers/carve-guards.ts` via `test/carve-guard-completeness.test.ts`).

Ships in #2577 (v1.65.0.0).

PR #714, opened 2026-03-31
fix: use content-length header to avoid buffering responses (#711)

### LINKED ISSUE #711: Performance: res.body() buffers entire response for size calculation
Noticed while profiling browse performance on a heavy page.

**File:** `browse/src/browser-manager.ts:947`

\`await res.body()\` buffers the entire response into memory just to calculate byte length. On pages with large assets (videos, big JS bundles), this causes memory spikes for no reason.

The content-length header is right there:

```typescript
const headers = res.headers();
const size = parseInt(headers['content-length'] || '0', 10);
```

Not every response will have the header, but for the ones that do (most static assets), you skip the full buffer.

stedfn:
## Summary
- `requestfinished` handler now reads the `content-length` header first
- Only falls back to `res.body()` when the header is absent
- Avoids buffering entire response bodies (images, videos, large payloads) just to record byte length

Fixes #711

## Test plan
- [ ] `bun test` passes
- [ ] `bun run build` succeeds
- [ ] Network buffer still records accurate sizes for responses with content-length header
- [ ] Responses without content-length header still get their size recorded via body fallback

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -1022,8 +1022,8 @@ export class BrowserManager {
         const res = await req.response();
         if (res) {
           const url = req.url();
-          const body = await res.body().catch(() => null);
-          const size = body ? body.length : 0;
+          const cl = await res.headerValue('content-length');
+          const size = cl != null ? parseInt(cl, 10) : (await res.body().catch(() => null))?.length ?? 0;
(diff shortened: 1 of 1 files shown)


garrytan:
Closing — overlaps with #746 which addresses the same issue. Thank you for the contribution!

PR #741, opened 2026-04-01
fix: sync package.json version with VERSION file

### LINKED ISSUE #399: bug: package.json version out of sync with VERSION file on main

Gonzih:
## Problem

`package.json` shows version `0.15.0.0` but `VERSION` file contains `0.15.1.0`. These are out of sync.

This means `npm show gstack version` (if published), `package.json` version checks, and any tooling that reads `package.json` see the wrong version. Closes #399.

## Fix

Updated `package.json` `"version"` field to match the canonical `VERSION` file: `0.15.1.0`.

## Notes

The `VERSION` file is the source of truth (used by `gstack-update-check`, CHANGELOG headers, etc). `package.json` should always mirror it. The `/ship` skill's idempotency check already guards against double-bumping `VERSION` — but it should also update `package.json` to keep these in sync after each bump.

---
*sent from mStack*

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gstack",
-  "version": "0.15.0.0",
+  "version": "0.15.1.0",
   "description": "Garry's Stack — Claude Code skills + fast headless browser. One repo, one install, entire AI engineering workflow.",
   "license": "MIT",
   "type": "module",


garrytan:
Thanks @Gonzih — closing as stale. The package.json/VERSION sync has been reworked through several /ship Step 12 fixes since this was opened. If drift still occurs on current main, a fresh repro would help.

PR #742, opened 2026-04-01
fix: check AAAA records in DNS rebinding protection (IPv6 bypass)

### LINKED ISSUE #668: [Security] DNS rebinding protection only checks IPv4 — IPv6 AAAA records bypass
## Summary

`resolvesToBlockedIp()` in `browse/src/url-validation.ts:51` only calls `dns.resolve4()` (A records). It never checks `dns.resolve6()` (AAAA records). The `BLOCKED_METADATA_HOSTS` set includes `fd00::` but that is only matched against the raw hostname string, not resolved IPv6 addresses. A DNS rebinding attack via an AAAA record pointing to a blocked IPv6 metadata address bypasses protection.

## Reproduction

```
# Attacker sets up DNS:
# evil.com  AAAA  fd00::
$B goto http://evil.com   # passes validation, reaches metadata endpoint
```

## Fix

PR #664 adds `resolve6` check in parallel with `resolve4`, using `Promise.all` with individual `.catch(() => [])` for resilience.

Found via sqry AST-based semantic code graph analysis.

Gonzih:
## Problem

`resolvesToBlockedIp()` in `browse/src/url-validation.ts` only calls `resolve4()` — it only checks IPv4 A records. An attacker can bypass the DNS rebinding protection by setting up a hostname that has **only** an AAAA (IPv6) record pointing to a cloud metadata endpoint (or an IPv6 equivalent). Closes #668.

Example bypass: a domain with only `AAAA ::ffff:169.254.169.254` (IPv4-mapped IPv6) would sail right through the existing check.

## Fix

- Added `resolve6()` alongside `resolve4()` — both run concurrently via `Promise.allSettled`
- Individual failures (e.g. host has no AAAA records) are silently ignored, same as before
- Added `::ffff:169.254.169.254` and `fe80::1` to `BLOCKED_METADATA_HOSTS` to catch IPv4-mapped and link-local IPv6 metadata patterns
- Used `Promise.allSettled` so a failure on one record type doesn't prevent checking the other

## Why `Promise.allSettled` not `Promise.all`

Many hosts legitimately have no AAAA records. `resolve6()` throws `ENOTFOUND` in those cases. Using `allSettled` collects both results and lets us check all resolved addresses without a try/catch per record type.

---
*sent from mStack*

### DIFF
--- a/browse/src/url-validation.ts
+++ b/browse/src/url-validation.ts
@@ -6,6 +6,8 @@
 const BLOCKED_METADATA_HOSTS = new Set([
   '169.254.169.254',  // AWS/GCP/Azure instance metadata
   'fd00::',           // IPv6 unique local (metadata in some cloud setups)
+  'fe80::1',          // IPv6 link-local (metadata in some setups)
+  '::ffff:169.254.169.254', // IPv4-mapped IPv6 form of link-local metadata
   'metadata.google.internal', // GCP metadata
   'metadata.azure.internal',  // Azure IMDS
 ]);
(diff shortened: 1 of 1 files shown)


garrytan:
Closed — this fix shipped in v0.15.7.0 (PR #810). Cherry-picked with attribution. Thank you @Gonzih for the thorough security audit!

Issue #667, opened 2026-03-30
[Security] cookie-import relative path bypasses validation via symlinks


mr-k-man:
## Summary

The `cookie-import` command in `browse/src/write-commands.ts:286` validates relative paths by checking for `..` via `path.normalize()` but never resolves symlinks. A relative symlink like `./cookies.json` pointing to `/etc/shadow` passes all checks.

## Reproduction

```bash
ln -s /etc/passwd ./cookies.json
$B cookie-import ./cookies.json   # reads /etc/passwd into process memory
```

## Root cause

Custom path validation instead of reusing `validateReadPath` pattern. The `..` check catches traversal but not symlink-based escapes.

## Fix

PR #664 replaces the custom validation with `realpathSync`-based resolution and reads from the resolved path (eliminating TOCTOU gap).

Found via sqry AST-based semantic code graph analysis.

garrytan:
Fixed in commit `47a82775`. Cookie-import path validation now uses `realpathSync` for symlink resolution.

Issue #665, opened 2026-03-30
[Security] validateOutputPath symlink bypass — screenshots/PDFs can write outside safe dirs


mr-k-man:
## Summary

`validateOutputPath()` in `browse/src/meta-commands.ts:20` uses `path.resolve()` but never `fs.realpathSync()` to resolve symlinks. Unlike `validateReadPath()` (fixed in #595), a symlink placed under `/tmp` or `cwd` can redirect screenshot/PDF/responsive writes to arbitrary files.

## Reproduction

```bash
ln -s /etc/crontab /tmp/evil.png
$B screenshot /tmp/evil.png   # overwrites /etc/crontab
```

## Root cause

`validateReadPath` was hardened with `realpathSync` in #595 (security audit remediation), but `validateOutputPath` was missed. The two functions have identical intent but divergent implementations.

## Fix

PR #664 adds `realpathSync` to `validateOutputPath`, matching the `validateReadPath` pattern including symlink resolution on safe directories themselves (macOS `/tmp` → `/private/tmp`).

Found via sqry AST-based semantic code graph analysis.

garrytan:
Fixed in commits `1351cf7d` + `73f5d0b7` (path-security.ts refactor). `validateOutputPath` now uses `realpathSync` on the parent directory. Additional file-level symlink check coming in the next security wave.

PR #745, opened 2026-04-01
fix: validateOutputPath symlink bypass — resolve real path before safe-dir check

Gonzih:
## The Bug

\`validateOutputPath()\` calls \`path.resolve(filePath)\` and then checks if the result is within the safe directories. \`path.resolve\` normalizes \`..\` components lexically but **does not follow symlinks**.

Attack:
\`\`\`bash
ln -s /etc /tmp/evil
# then: screenshot /tmp/evil/cron.d/payload
\`\`\`
\`path.resolve('/tmp/evil/cron.d/payload')\` returns \`/tmp/evil/cron.d/payload\` which passes \`isPathWithin(_, '/tmp')\`, but the actual \`fs.writeFile\` follows the symlink and writes to \`/etc/cron.d/payload\`.

Issue #665. Also related to #667.

## Fix

After the lexical check, resolve the real path of the nearest existing ancestor directory using \`fs.realpathSync\` and re-validate against safe directories. Handles the case where the output file doesn't exist yet (walks up the tree to find the nearest existing parent).

\`\`\`
/tmp/evil/          → realpathSync → /etc/
/tmp/evil/cron.d/   → /etc/cron.d/ (not in SAFE_DIRECTORIES) → BLOCKED
\`\`\`

---

### DIFF
--- a/browse/src/write-commands.ts
+++ b/browse/src/write-commands.ts
@@ -18,9 +18,38 @@ const SAFE_DIRECTORIES = [TEMP_DIR, process.cwd()];
 
 function validateOutputPath(filePath: string): void {
   const resolved = path.resolve(filePath);
+
+  // Basic containment check using lexical resolution only.
+  // This catches obvious traversal (../../../etc/passwd) but NOT symlinks.
   const isSafe = SAFE_DIRECTORIES.some(dir => isPathWithin(resolved, dir));
   if (!isSafe) {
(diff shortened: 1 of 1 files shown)


garrytan:
Closed — this fix shipped in v0.15.7.0 (PR #810). Cherry-picked with attribution. Thank you @Gonzih for the thorough security audit!

Issue #709, opened 2026-03-31
Docker: chmod -R 1777 /tmp sets sticky bit on files


robertoecf:
Came up in a Gemini code review on our repo.

**File:** `.github/docker/Dockerfile.ci:63`

\`chmod -R 1777\` applies the sticky bit recursively to files inside /tmp — sticky bit only makes sense on directories. Line 61 already handles the directory itself correctly, so this line is doing more harm than good (could hit sockets or pipes if they exist).

Probably just remove the recursive chmod since the directory permission on line 61 is already correct.

PR #747, opened 2026-04-01
fix: chmod 1777 /tmp not chmod -R 1777 /tmp in Dockerfile.ci

Gonzih:
## The Bug

\`Dockerfile.ci\` has:

\`\`\`dockerfile
&& chmod 1777 /tmp \      ← correct
&& chmod -R 1777 /tmp     ← wrong + duplicate
\`\`\`

\`chmod -R 1777 /tmp\` recursively sets the sticky bit on every **file** inside \`/tmp\`, not just the directory. The sticky bit on a directory (1777) is well-defined: it prevents users from deleting each other's files. The sticky bit on a file has no defined behavior in modern Linux/POSIX and can confuse tools that inspect permissions (e.g., package managers, installers, security scanners that report "unexpected sticky bit on file").

It also runs twice — the first \`chmod 1777 /tmp\` on the line before is correct. The second with \`-R\` undoes that correctness.

Issue #709.

## Fix

Remove the recursive \`-R\` flag and deduplicate. One \`chmod 1777 /tmp\` is correct.

---

### DIFF
--- a/.github/docker/Dockerfile.ci
+++ b/.github/docker/Dockerfile.ci
@@ -58,6 +58,5 @@ RUN mv /workspace/node_modules /opt/node_modules_cache \
 RUN useradd -m -s /bin/bash runner \
     && chmod -R a+rX /opt/node_modules_cache \
     && mkdir -p /home/runner/.gstack && chown -R runner:runner /home/runner/.gstack \
-    && chmod 1777 /tmp \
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks! chmod fix landed with co-author credit. Ships in the next release.

PR #748, opened 2026-04-01
fix: use os.tmpdir() instead of hardcoded /tmp in cookie-import-browser

### LINKED ISSUE #708: Cross-platform: hardcoded /tmp in cookie-import-browser.ts
Spotted this while reviewing our vendored gstack install.

**File:** `browse/src/cookie-import-browser.ts:390`

The path `/tmp` is hardcoded here. Breaks on Windows. You already have `TEMP_DIR` in `platform.ts` for exactly this — just needs to be used here too:

```typescript
const tmpPath = path.join(TEMP_DIR, "browse-cookies-" + browserName.toLowerCase() + "-" + crypto.randomUUID() + ".db");
```

Small fix, happy to send a PR.

Gonzih:
One-liner. \`/tmp\` doesn't exist on Windows. \`os\` was already imported in the file, just wasn't used here.

Before: \`\`\`\`/tmp/browse-cookies-${browser}-${uuid}.db\`\`\`\`
After: \`os.tmpdir()\` resolves to \`%TEMP%\` on Windows, \`/tmp\` on Unix.

Closes #708.

---

### DIFF
--- a/browse/src/cookie-import-browser.ts
+++ b/browse/src/cookie-import-browser.ts
@@ -386,7 +386,7 @@ function openDb(dbPath: string, browserName: string): Database {
 }
 
 function openDbFromCopy(dbPath: string, browserName: string): Database {
-  const tmpPath = `/tmp/browse-cookies-${browserName.toLowerCase()}-${crypto.randomUUID()}.db`;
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks @Gonzih — closing as stale. cookie-import-browser has been reworked since this was opened. If hardcoded /tmp is still present on current main, a fresh PR against current code would land easier.

Issue #710, opened 2026-03-31
Fragile JSON manipulation via sed in gstack-telemetry-sync


robertoecf:
Found during review.

**File:** `bin/gstack-telemetry-sync:85`

Using \`sed\` with regex to strip fields from JSON is fragile — breaks if values contain commas, escaped quotes, or if field order changes. Since the script already runs in an environment with \`jq\` (or could use bun), this would be more reliable:

```bash
CLEAN=\$(echo "\$LINE" | jq -c 'del(._repo_slug, ._branch, .repo)')
```

Not urgent but avoids silent data corruption in the telemetry pipeline.

garrytan:
Addressed by #749

PR #749, opened 2026-04-01
fix: use jq instead of sed for JSON field stripping in gstack-telemetry-sync

Gonzih:
## The Bug

\`gstack-telemetry-sync\` strips local-only fields before sending events to Supabase using sed:

\`\`\`bash
CLEAN="$(echo "$LINE" | sed \
  -e 's/,"_repo_slug":"[^"]*"//g' \
  -e 's/,"_branch":"[^"]*"//g' \
  -e 's/,"repo":"[^"]*"//g')"
\`\`\`

The \`[^"]*\` pattern breaks on JSON string values containing escaped quotes (\`\\\"\`) or regex metacharacters. This produces corrupt JSON. The Supabase edge function returns a non-2xx response (or 0 inserted), the cursor doesn't advance, and the same events get retried on next sync.

Issue #710.

## Fix

Use \`jq del()\` for proper JSON-aware field removal. The sed code is kept as a fallback for older installs without \`jq\`. \`jq\` is already a required dependency (installed in \`Dockerfile.ci\` and used elsewhere in the codebase).

---

### DIFF
--- a/bin/gstack-telemetry-sync
+++ b/bin/gstack-telemetry-sync
@@ -69,6 +69,37 @@ UNSENT="$(tail -n "+$SKIP" "$JSONL_FILE" 2>/dev/null || true)"
 # ─── Strip local-only fields and build batch ─────────────────
 # Edge function expects raw JSONL field names (v, ts, sessions) —
 # no column renaming needed (the function maps them internally).
+#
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks @Gonzih — closing as stale. gstack-telemetry-sync has been reworked. If the sed/jq concern still applies, a fresh fix against current code would land easier.

Issue #672, opened 2026-03-30
[Security] design serve /api/reload path traversal — arbitrary file read


mr-k-man:
## Summary

`design/src/serve.ts:177-186` — The `/api/reload` endpoint reads `body.html` from the POST request body and passes it directly to `fs.readFileSync()` with only an `existsSync()` check. Any file readable by the process can be served as HTML through the `/` endpoint.

## Reproduction

```bash
# Start the design serve command
$D serve /path/to/board.html

# From another terminal, reload with an arbitrary path
curl -X POST http://127.0.0.1:<port>/api/reload \
  -H 'Content-Type: application/json' \
  -d '{"html": "/etc/passwd"}'

# Then fetch the served page — contains /etc/passwd content
curl http://127.0.0.1:<port>/
```

## Severity

High — arbitrary file read via localhost HTTP server.

## Fix

PR #664 anchors all file reads to the initial HTML file's parent directory using `realpathSync`, blocking symlink and traversal attacks. 3 integration tests added.

Found via sqry AST-based semantic code graph analysis.

garrytan:
Fixed in commit `c151fabf` (PR #664, v0.15.13.0). `/api/reload` now validates paths with `realpathSync` and containment check against the initial HTML file's parent directory.

PR #752, opened 2026-04-01
fix: path traversal in design serve /api/reload — arbitrary file read

Gonzih:
## The Bug

\`POST /api/reload\` in \`design/src/serve.ts\` accepts a JSON body with an arbitrary \`html\` path and reads it with no validation:

\`\`\`typescript
const newHtmlPath = body.html;
if (!newHtmlPath || !fs.existsSync(newHtmlPath)) { ... }  // existence check only
htmlContent = fs.readFileSync(newHtmlPath, "utf-8");       // reads anything
\`\`\`

The design server binds to localhost, but the Chrome extension (or any local process) can POST to it. Attack:

\`\`\`
POST /api/reload {"html": "/etc/passwd"}
GET /  → returns contents of /etc/passwd
\`\`\`

Issue #672.

## Fix

Validate that the reload path resolves within the same directory as the original HTML file. Paths outside are rejected with 403. The agent only ever reloads variant files in the same output directory — legitimate use is unaffected.

---

### DIFF
--- a/design/src/serve.ts
+++ b/design/src/serve.ts
@@ -182,8 +182,20 @@ export async function serve(options: ServeOptions): Promise<void> {
       );
     }
 
+    // Path traversal guard: the reload path must stay within the same directory
+    // as the original HTML file. Arbitrary paths from the POST body would let
+    // any caller read any file on the filesystem via the GET / endpoint.
+    const safeDir = path.resolve(path.dirname(html));
+    const resolvedNew = path.resolve(newHtmlPath);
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks for catching this! The path traversal in `/api/reload` was fixed in commit `c151fabf` (v0.15.13.0). The endpoint now uses `realpathSync` and validates the reload path is within the initial HTML file's parent directory. Appreciate you reporting it!

PR #776, opened 2026-04-03
fix: user-friendly error when OpenAI org is not verified

### LINKED ISSUE #718: Design binary: raw OpenAI 403 error leaks stale URL for org verification
## Bug

When a user's OpenAI organization is not verified, the design binary (`$D generate`, `$D variants`, etc.) displays the raw OpenAI API error response, which contains a stale/broken URL:

```
API error (403): {
  "error": {
    "message": "Your organization must be verified to use the model `gpt-4o`. Please go to: https://platform.openai.com/settings/organization/general and click on Verify Organization. ..."
  }
}
```

The URL `https://platform.openai.com/settings/organization/general` returns a 404. OpenAI appears to have changed their settings page structure.

## Root Cause

`design/src/generate.ts:61-63` throws the raw API response text without parsing:

```typescript
if (!response.ok) {
  const error = await response.text();
  throw new Error(`API error (${response.status}): ${error}`);
}
```

## Suggested Fix

Detect 403 responses containing "organization must be verified" and display a user-friendly message with the correct URL. Something like:

```typescript
if (!response.ok) {
  const error = await response.text();
  if (response.status === 403 && error.includes("organization must be verified")) {
    throw new Error(
      "OpenAI organization verification required.\n" +
      "Go to https://platform.openai.com/settings/organization to verify.\n" +
      "After verification, wait up to 15 minutes for access to propagate."
    );
  }
  throw new Error(`API error (${response.status}): ${error}`);
}
```

This also prevents leaking raw JSON error responses to the user, which is not a great UX.

## Environment

- gstack version: latest (installed via global)
- macOS Darwin 24.6.0
- Triggered by: `$D variants --brief "..." --count 3 --output-dir /path/`

mvanhorn:
When a user's OpenAI org isn't verified, `$D generate` dumps the raw JSON error response — which includes a broken URL (`/settings/organization/general` returns 404). This detects the 403 + "organization must be verified" pattern and shows a clean message with the correct verification URL instead.

Also truncates the generic error fallback to 200 chars, matching `design-to-code.ts`, `iterate.ts`, and `evolve.ts` which already truncate.

### Before / After

demo

### Changes

- `design/src/generate.ts:61-64`: detect 403 org verification error, throw actionable message
- Truncate raw error to `.slice(0, 200)` on the generic fallback path

### Testing

`bun test` passes. No unit test added — triggering this error path requires an unverified OpenAI org. The E2E design tests run against real API and don't cover error responses.

This contribution was developed with AI assistance (Codex).

Fixes #718

### DIFF
--- a/design/src/generate.ts
+++ b/design/src/generate.ts
@@ -60,7 +60,14 @@ async function callImageGeneration(
 
     if (!response.ok) {
       const error = await response.text();
-      throw new Error(`API error (${response.status}): ${error}`);
+      if (response.status === 403 && error.includes("organization must be verified")) {
+        throw new Error(
+          "OpenAI organization verification required.\n"
+          + "Go to https://platform.openai.com/settings/organization to verify.\n"
+          + "After verification, wait up to 15 minutes for access to propagate.",
+        );
+      }
+      throw new Error(`API error (${response.status}): ${error.slice(0, 200)}`);
     }
 
     const data = await response.json() as any;


mvanhorn:
Appreciate the merge. The org verification error was a frustrating one to debug - glad the fix landed.

PR #777, opened 2026-04-03
Add MiniMax image generation fallback

vshen009:
## Summary
- add provider-aware image size normalization so MiniMax and OpenAI can share the same generate flow
- probe OpenAI first and fall back cleanly to MiniMax when available
- add a focused regression test for provider size conversion defaults and mappings

## Testing
- bun run build
- bun test
- bun test test/design-generate.test.ts


time-attack:
@16francej — recommend closing this PR as superseded by #990, which owns provider, privacy, capability, and spending policy. This branch may trigger an unexpected paid request and silently change image providers after failure. Please credit the contributor and preserve MiniMax size mapping, but not paid probing or silent switching.

PR #780, opened 2026-04-03
fix: migration version comparison handles semantic versioning pre-releases

km-git007:
Adds helper functions _base_version() and _version_lte() to the setup migration logic to reliably compare semantic versions that may include pre-release tags (e.g., 0.15.2.0-rc1). By extracting base versions before comparison, we ensure that dev builds in the same release cycle as a migration's target version are handled correctly.

Simplifies the inline version comparison logic (replacing complex sort-based comparisons with clear, testable helpers) and adds documentation explaining the semantic versioning behavior.

Tested with realistic migration scenarios:
- Fresh install: 0.0.0.0 -> 0.15.2.1 (runs v0.15.2.0 migration) ✓
- Already migrated: stays at 0.15.2.0 (skips v0.15.2.0) ✓
- Version bump: 0.15.2.0 -> 0.15.2.1 (runs v0.15.2.1) ✓
- Dev version: 0.15.1.0 -> 0.15.2.0-rc1 (runs v0.15.2.0) ✓
- Future migration: skips v0.15.3.0 when current is 0.15.2.1 ✓

### DIFF
--- a/setup
+++ b/setup
@@ -706,6 +706,27 @@ fi
 # 8. Run pending version migrations
 # Migrations handle state fixes that ./setup alone can't cover (stale config,
 # orphaned files, directory structure changes). Each migration is idempotent.
+
+# Helper: extract base version (strip pre-release suffix like -rc1, -beta, etc.)
+# for reliable semantic version comparison. By comparing base versions, we ensure
+# that development builds (0.15.2.0-rc1) are treated as being in the same release
+# cycle as their final release (0.15.2.0), so migrations run appropriately whether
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for this. Stripping every hyphen suffix does not implement semantic-version prerelease ordering, and the supported suffix-bearing states are not defined. Define the accepted version semantics, revise the comparator accordingly, and add regression cases for upgrades, equal-cycle prereleases, finals, downgrades, invalid inputs, and fresh installs.

PR #803, opened 2026-04-04
fix(design): bind server to localhost and validate reload paths

garagon:
The design comparison board server binds to all interfaces with no authentication. The /api/reload endpoint reads arbitrary files from disk and serves them via GET /.

Tested: started the server, POSTed {"html":"/etc/hosts"} to /api/reload, fetched / and got the file contents. After the fix, the reload returns 403 for paths outside cwd/tmp and the server only listens on 127.0.0.1.

Changes:
- Added hostname to ServeOptions with default '127.0.0.1' (matching browse server pattern at server.ts:1042)
- Added path validation to /api/reload (cwd and tmpdir only)

### DIFF
--- a/design/src/serve.ts
+++ b/design/src/serve.ts
@@ -33,19 +33,21 @@
  */
 
 import fs from "fs";
+import os from "os";
 import path from "path";
 import { spawn } from "child_process";
 
 export interface ServeOptions {
   html: string;
   port?: number;
+  hostname?: string; // default '127.0.0.1' — localhost only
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks for this fix! Both issues are already resolved on main: `design/src/serve.ts` binds to `127.0.0.1` and validates `/api/reload` paths with `realpathSync` + directory containment. Landed in commit `c151fabf` (v0.15.13.0). Appreciate the thorough report!

PR #821, opened 2026-04-05
fix(browse): add path validation to upload command

garagon:
The upload command reads arbitrary files and sends them to page file inputs with no path restriction. Adjacent cookie-import command validates paths against safe directories (TEMP_DIR and cwd) and blocks traversal sequences. Added the same validation pattern to upload.

Change: absolute paths checked against safe directories, relative paths blocked if they contain .. sequences.

### DIFF
--- a/browse/src/write-commands.ts
+++ b/browse/src/write-commands.ts
@@ -399,9 +399,19 @@ export async function handleWriteCommand(
       const [selector, ...filePaths] = args;
       if (!selector || filePaths.length === 0) throw new Error('Usage: browse upload <selector> <file1> [file2...]');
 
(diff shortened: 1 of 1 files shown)


garagon:
CI note: the eval failures require ANTHROPIC_API_KEY which is not available on fork PRs. Free tests (bun test) pass locally. The pre-existing test-2 failure also reproduces on main.

garrytan:
Big thanks for this contribution, @garagon! Your fix has been landed on the garrytan/security-wave-5 branch as part of our community security wave (commit 49d7841a). We credited you as co-author. Ships in v0.15.12.0. Really appreciate you taking the time to find and fix this.

PR #838, opened 2026-04-05
fix(security): add symlink resolution to meta-commands validateOutputPath

### LINKED ISSUE #707: Security: path validation bypass via relative paths in write-commands.ts

```typescript
}
```

Happy to PR this if you want.

Ziadstr:
## Summary

`meta-commands.ts:validateOutputPath` only performed lexical `path.resolve` checking, while `write-commands.ts:validateOutputPath` had a full symlink resolution step added in PR #640 (security audit round 2). A symlink inside `/tmp` pointing outside the safe zone would pass `meta-commands` validation but be blocked by `write-commands`.

This brings `meta-commands.ts:validateOutputPath` to parity with `write-commands.ts:validateOutputPath`.

## The vulnerability

```bash
# Create a symlink inside /tmp pointing to /etc
ln -s /etc /tmp/escape-hatch

# meta-commands validateOutputPath passes: /tmp/escape-hatch/passwd resolves
# lexically to /tmp/escape-hatch/passwd which IS within /tmp
# But the real path is /etc/passwd which is NOT within /tmp
```

Commands using `meta-commands.validateOutputPath` (screenshot, pdf, responsive) would allow writing to paths outside the safe zone via symlinks.

## Fix

Added symlink resolution (identical to the existing `write-commands.ts` implementation):
1. Resolve the real path of the nearest existing ancestor directory via `fs.realpathSync`
2. Re-validate the real path against safe directories
3. Fail closed if resolution fails

## Testing

Added regression test: creates a symlink inside `/tmp` pointing to `/etc`, verifies `validateOutputPath` blocks `join(linkPath, 'passwd')`.

Follows the exact same test pattern as the existing `validateReadPath` symlink test (line 67-75 of `path-validation.test.ts`).

Fixes #707

### DIFF
--- a/browse/src/meta-commands.ts
+++ b/browse/src/meta-commands.ts
@@ -19,9 +19,32 @@ const SAFE_DIRECTORIES = [TEMP_DIR, process.cwd()];
 
 export function validateOutputPath(filePath: string): void {
   const resolved = path.resolve(filePath);
+
+  // Lexical containment check: catches obvious traversal (../../../etc/passwd)
   const isSafe = SAFE_DIRECTORIES.some(dir => isPathWithin(resolved, dir));
   if (!isSafe) {
     throw new Error(`Path must be within: ${SAFE_DIRECTORIES.join(', ')}`);
+  }
+
+  // Symlink check: resolve the real path of the nearest existing ancestor
+  // directory and re-validate. Closes the symlink bypass where a symlink
+  // inside /tmp or cwd points outside the safe zone.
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks for this fix! Symlink resolution in meta-commands validateOutputPath was already landed as commit `fe6880bf` (PR #820, security wave 2, v0.15.13.0). The shared `path-security.ts` module now handles this. Appreciate the contribution!

PR #852, opened 2026-04-06
Fix cookie-import path validation bypass + hardcoded /tmp

### LINKED ISSUE #707: Security: path validation bypass via relative paths in write-commands.ts

### LINKED ISSUE #708: Cross-platform: hardcoded /tmp in cookie-import-browser.ts

urbantech:
## Summary

Two security/portability fixes in the browse cookie-import command.

### #707 — Path validation bypass via relative paths

**Before:** Relative paths like `../../etc/shadow` bypassed the safe-directory check because `path.isAbsolute()` gated the entire validation block. Only the `..` pattern check ran for relative paths, but a relative path without `..` (e.g., `sensitive-file.json` in an unexpected cwd) passed through completely.

**After:** Always `path.resolve()` to absolute, then resolve symlinks (matching `validateOutputPath()`), then check against `SAFE_DIRECTORIES`. No separate code paths for relative vs absolute.

### #708 — Hardcoded `/tmp` in cookie-import-browser.ts

**Before:** `openDbFromCopy()` hardcoded `/tmp/browse-cookies-...`. On Windows, `/tmp` may not exist.

**After:** Uses `path.join(os.tmpdir(), ...)`. The `os` module was already imported.

## Test Plan

- [x] All 224 browse tests pass (`bun test browse/test/`)
- [x] Path validation tests: 29 pass
- [x] Cookie import tests: 22 pass
- [x] Traversal test updated to accept both error messages

Fixes #707, fixes #708

### DIFF
--- a/browse/src/cookie-import-browser.ts
+++ b/browse/src/cookie-import-browser.ts
@@ -386,7 +386,8 @@ function openDb(dbPath: string, browserName: string): Database {
 }
 
 function openDbFromCopy(dbPath: string, browserName: string): Database {
-  const tmpPath = `/tmp/browse-cookies-${browserName.toLowerCase()}-${crypto.randomUUID()}.db`;
+  // Use os.tmpdir() instead of hardcoded /tmp for cross-platform support (#708)
+  const tmpPath = path.join(os.tmpdir(), `browse-cookies-${browserName.toLowerCase()}-${crypto.randomUUID()}.db`);
   try {
     fs.copyFileSync(dbPath, tmpPath);
     // Also copy WAL and SHM if they exist (for consistent reads)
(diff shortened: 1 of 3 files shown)


garrytan:
Thanks for this excellent fix! Your approach was the most complete of the three overlapping PRs (#852, #920, #921). The fix has been landed on the security-wave-triage branch with co-author credit. Ships in the next release. Really appreciate both the fix and the regression test update.

PR #853, opened 2026-04-06
Fix CDPATH breaking all bin/ scripts + fragile JSON stripping via sed

### LINKED ISSUE #710: Fragile JSON manipulation via sed in gstack-telemetry-sync

urbantech:
### #824 — CDPATH breaks all bin/ scripts

Every shell script uses `SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"` to find its location. When `CDPATH` is set (common in power-user dotfiles), `cd` prints the target directory to stdout, prepending garbage to the captured path.

**Fix:** `CDPATH= cd` — unsets CDPATH for the subshell. Standard POSIX fix.

**23 files** across `bin/`, `scripts/`, `browse/scripts/`, `setup`, `supabase/`.

### #710 — Fragile JSON manipulation via sed in telemetry sync

`gstack-telemetry-sync` stripped JSON fields using sed patterns like `s/,"_repo_slug":"[^"]*"//g`. This breaks on:
- Values containing escaped quotes (`\"`)
- Fields in first position (no leading comma)
- Nested objects as values

**Fix:** Replace with `bun -e` JSON.parse/delete/stringify. Falls back to the original line on parse failure. Bun is guaranteed available.

## Test Plan

- [x] CDPATH fix verified: `CDPATH="/usr:/var" SCRIPT_DIR="$(CDPATH= cd ... && pwd)"` returns correct path
- [x] JSON stripping verified: input with `_repo_slug`, `_branch`, `repo` fields correctly stripped
- [x] Fallback on invalid JSON verified: original line preserved

Fixes #824, fixes #710

### DIFF
--- a/bin/dev-setup
+++ b/bin/dev-setup
@@ -11,7 +11,7 @@
 #        bin/dev-teardown    # clean up
 set -e
 
-REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+REPO_ROOT="$(CDPATH= cd "$(dirname "$0")/.." && pwd)"
 
 # 1. Copy .env from main worktree (if we're a worktree and don't have one)
 if [ ! -f "$REPO_ROOT/.env" ]; then
--- a/bin/dev-teardown
+++ b/bin/dev-teardown
@@ -2,7 +2,7 @@
 # Remove local dev skill symlinks. Restores global gstack as the active install.
 set -e
 
-REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+REPO_ROOT="$(CDPATH= cd "$(dirname "$0")/.." && pwd)"
 
 removed=()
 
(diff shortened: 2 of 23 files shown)


garrytan:
Closing — duplicate of several PRs fixing the same CDPATH issue in bin/ scripts. Thank you for the contribution!

PR #856, opened 2026-04-06
Add uninstall link to install section + NO FINDINGS sanity check

### LINKED ISSUE #801: add deinstall steps into the readme
i want to wipe it from anything it touched, memory, settings, claude.md etc.

### LINKED ISSUE #839: Proposal: post-review sanity check when all specialists return NO FINDINGS
## Problem

When the review specialist army dispatches 3-5 agents on a 500+ line diff, there is no validation if all specialists return `NO FINDINGS`. The system trusts their output unconditionally.

In practice, agents can:
- Rationalize away real issues ("this pattern looks intentional")
- Hit context limits and give up silently
- Miss issues due to prompt phrasing or model variance

A 500-line diff with zero findings from 5 specialists is suspicious. Not impossible, but worth a second look.

## Proposed solution: lightweight post-review sanity check

After the specialist merge step, before Fix-First:

1. If diff > 200 lines AND all specialists return `NO FINDINGS`: flag as suspicious
2. Re-dispatch one specialist (testing or security, most likely to find something) with an explicit prompt: "A previous review of this diff found zero issues. Review again with extra scrutiny. Focus on what might have been missed."
3. If the re-run also returns `NO FINDINGS`: accept it (genuine clean diff)
4. If it finds issues: include them in findings with a note "found on second pass"

This adds one extra agent call in the rare case where a large diff gets a clean sweep, zero cost in the common case.

## Why this matters

From experience building a similar review system: we added an "anti-rationalization" hook that fires PostToolUse and catches when agents produce suspiciously clean results on objectively complex code. It's caught real misses where the agent skipped auth checks, missed N+1 queries, or rationalized away a race condition.

The cost is minimal (one conditional re-dispatch), the signal is high, and it prevents false confidence in the review output.

## Alternative: confidence floor

Instead of re-dispatching, require at least one finding on diffs > 200 lines. If all specialists return `NO FINDINGS`, output a warning in the review report: "Large diff reviewed with zero findings. Manual review recommended." This is cheaper but less actionable.

Not submitting a PR for this since it touches the review pipeline architecture and should be a design decision. Happy to implement either approach if there's interest.

urbantech:
### #801 — Add uninstall steps to README

The install section now includes a one-line uninstall reference right after the install command:

> **To uninstall:** `~/.claude/skills/gstack/bin/gstack-uninstall` — or see [Uninstall](#uninstall) for manual steps if you no longer have the repo locally.

The existing Uninstall section (with full manual steps) is already comprehensive — this just makes it discoverable from the install section where new users land.

### #839 — Post-review sanity check when all specialists return NO FINDINGS

Added to both `review/SKILL.md` and `ship/SKILL.md`:

When ALL specialist subagents return "NO FINDINGS", the review now:
1. Verifies the diff is non-empty (`git diff --stat`)
2. Confirms specialists received the diff (checks output for file paths)
3. Warns the user if diff is non-trivial (>50 lines) — may indicate wrong branch, empty diff, or skills not reading files

This catches silent misconfiguration without blocking legitimately clean reviews.

## Test Plan

- [x] README uninstall link renders correctly (anchor to existing section)
- [x] Sanity check logic is consistent between review/SKILL.md and ship/SKILL.md
- [x] Check only triggers when ALL specialists return NO FINDINGS (not when some do)

Fixes #801, fixes #839

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ Open Claude Code and paste this. Claude does the rest.
 
 > Install gstack: run **`git clone --single-branch --depth 1 https://github.com/garrytan/gstack.git ~/.claude/skills/gstack && cd ~/.claude/skills/gstack && ./setup`** then add a "gstack" section to CLAUDE.md that says to use the /browse skill from gstack for all web browsing, never use mcp\_\_claude-in-chrome\_\_\* tools, and lists the available skills: /office-hours, /plan-ceo-review, /plan-eng-review, /plan-design-review, /design-consultation, /design-shotgun, /design-html, /review, /ship, /land-and-deploy, /canary, /benchmark, /browse, /connect-chrome, /qa, /qa-only, /design-review, /setup-browser-cookies, /setup-deploy, /retro, /investigate, /document-release, /codex, /cso, /autoplan, /plan-devex-review, /devex-review, /careful, /freeze, /guard, /unfreeze, /gstack-upgrade, /learn. Then ask the user if they also want to add gstack to the current project so teammates get it.
 
+**To uninstall:** `~/.claude/skills/gstack/bin/gstack-uninstall` — or see [Uninstall](#uninstall) for manual steps if you no longer have the repo locally.
+
 ### Step 2: Team mode — auto-update for shared repos (recommended)
 
 Every developer installs globally, updates happen automatically:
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej Please close this PR. The uninstall documentation was already absorbed with credit, and the clean-review sanity check shipped more strongly in #692. Both parts of this mixed PR are already covered on main.

PR #858, opened 2026-04-06
fix(bin): pass settings path and hook command via env vars

garagon:
## What

Replace direct shell variable interpolation inside `bun -e` scripts with
environment variables read via `process.env` in `gstack-team-init` and
`gstack-settings-hook`.

## Why

Same pattern fixed in #819 (gstack-learnings-search): shell variables
interpolated into JavaScript string literals allow paths containing quotes
to break the string context.

A settings path like `/tmp/it's-a-test/settings.json` produces:
```js
const settingsPath = '/tmp/it's-a-test/settings.json';
//                              ^ JS parse error here
```

Environment variables are read as-is by `process.env` without shell
interpretation.

## Files

- `bin/gstack-team-init`: `$SETTINGS` → `process.env.GSTACK_SETTINGS_PATH`
- `bin/gstack-settings-hook`: `$SETTINGS_FILE` → `process.env.GSTACK_SETTINGS_PATH`, `$HOOK_CMD` → `process.env.GSTACK_HOOK_CMD`

### DIFF
--- a/bin/gstack-settings-hook
+++ b/bin/gstack-settings-hook
@@ -26,10 +26,10 @@ fi
 
 case "$ACTION" in
   add)
-    bun -e "
+    GSTACK_SETTINGS_PATH="$SETTINGS_FILE" GSTACK_HOOK_CMD="$HOOK_CMD" bun -e "
       const fs = require('fs');
-      const settingsPath = '$SETTINGS_FILE';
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks for this fix! The env var approach for shell injection prevention has been landed on the security-wave-triage branch, also applied to gstack-team-init which had the same pattern. Co-author credited. Ships in the next release.

PR #860, opened 2026-04-06
fix(browse): redact form fields with sensitive names, not just type=password

garagon:
## What

Add name and id pattern matching to the forms command redaction logic in `browse/src/read-commands.ts`.

## Why

The `$B forms` command only redacted values for `type="password"` inputs. Hidden inputs and text fields with sensitive names were exposed unredacted in the LLM context:

```html
<input type="hidden" name="csrf_token" value="abc123">   <!-- exposed -->
<input type="hidden" name="api_key" value="sk-secret">    <!-- exposed -->
<input type="text" name="session_id" value="sess_xyz">     <!-- exposed -->
```

Cookies and storage already use `SENSITIVE_COOKIE_NAME` pattern matching against names containing `token`, `secret`, `key`, `password`, `credential`, `auth`, `jwt`, `session`, `csrf`, `sid`, and `api_key`. This PR applies the same pattern to form field `name` and `id` attributes.

The regex runs inside `page.evaluate()` (browser context) so it's inlined rather than referencing the Node export.

## Files

- `browse/src/read-commands.ts`: line 148, forms case

### DIFF
--- a/browse/src/read-commands.ts
+++ b/browse/src/read-commands.ts
@@ -145,7 +145,10 @@ export async function handleReadCommand(
               id: input.id || undefined,
               placeholder: input.placeholder || undefined,
               required: input.required || undefined,
-              value: input.type === 'password' ? '[redacted]' : (input.value || undefined),
+              value: input.type === 'password'
+                || (input.name && /(^|[_.-])(token|secret|key|password|credential|auth|jwt|session|csrf|sid)($|[_.-])|api.?key/i.test(input.name))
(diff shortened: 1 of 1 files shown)


garagon:
CI note: the E2E eval failures require ANTHROPIC_API_KEY which is not available on fork PRs. Free tests (build-image) pass. The change is a 3-line addition to the forms redaction logic inside a page.evaluate() callback — no impact on eval test paths.

garrytan:
Thanks! Form field name redaction has been landed with co-author credit. Ships in the next release.

PR #871, opened 2026-04-07
Clarify sandbox port-bind failures in browse startup

tichenm:
## Summary
- distinguish permission-denied localhost bind failures from real port contention in browse startup
- preserve explicit `BROWSE_PORT` handling and add targeted regression coverage
- document the sandbox-specific failure mode in troubleshooting

## Root cause
In sandboxed environments, binding `127.0.0.1:<port>` can fail with `EPERM` or `EACCES`. Browse treated all bind failures as unavailable ports, retried 5 random ports, and then surfaced the misleading error `No available port after 5 attempts`.

## What changed
- extracted port probing into `browse/src/port.ts`
- treat `EADDRINUSE` as real contention, but surface `EPERM` / `EACCES` as environment restrictions
- route explicit `BROWSE_PORT` through the same classification logic
- added regression tests for random-port and explicit-port permission failures
- added troubleshooting notes for stale installs that still show the legacy message

## Verification
- `bun test browse/test/findport.test.ts`
- `bun test browse/test/`
- sandbox startup check now reports `Local port bind was blocked by the environment (EPERM)`
- escalated startup check still reaches healthy status

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,10 @@ Data is stored in [Supabase](https://supabase.com) (open source Firebase alterna
 
 **`/browse` fails?** `cd ~/.claude/skills/gstack && bun install && bun run build`
 
+If startup says `Local port bind was blocked by the environment (EPERM)` or `EACCES`, that is usually a sandbox or permission restriction, not real port exhaustion. Re-run with escalated permissions or outside the sandbox.
+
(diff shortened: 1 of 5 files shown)


time-attack:
@16francej Please close this PR. Browser startup now reports and handles port-binding failures through the implementation merged in #1682. This older five-file refactor targets code that has since changed.

PR #900, opened 2026-04-08
[codex] refresh stale gstack skill installs

lmist:
## Summary

This fixes a stale-install edge case in `./setup --host codex`.

On `origin/main`, if `~/.codex/skills/gstack-qa` already exists as a real directory, setup exits successfully but leaves the stale directory untouched. Codex then keeps reading stale SKILL content.

This patch makes setup replace stale real directories with the expected symlinks to generated Codex skills. It applies the same repair logic to the Factory install path, tightens `gstack-uninstall` so it only reports removing the SessionStart hook when that hook actually exists, and aligns `package.json` with `VERSION` so the targeted validation run is green again.

## Root cause

`link_codex_skill_dirs()` only refreshed paths that were missing or already symlinks. Real directories from prior installs were treated as untouchable, so the stale layout survived even though setup printed success.

## Repro videos

The media lives on a separate fork branch so the GIFs and MP4s are not part of the patch that would merge.
Asset branch: `lmist:asset-setup-codex-repro-v2`

### Before: `origin/main`

Inline preview:

Full video: [setup-codex-broken-scroll-v2.mp4](https://raw.githubusercontent.com/lmist/gstack/asset-setup-codex-repro-v2/docs/pr-artifacts/setup-codex-broken-scroll-v2.mp4)

This shows the stale `gstack-qa` directory remains a real directory and still contains `stale` after `./setup --host codex -q`.

### After: this branch

Inline preview:

Full video: [setup-codex-fixed-scroll-v2.mp4](https://raw.githubusercontent.com/lmist/gstack/asset-setup-codex-repro-v2/docs/pr-artifacts/setup-codex-fixed-scroll-v2.mp4)

This shows the same repro now converts `gstack-qa` into a symlink to `.agents/skills/gstack-qa/`, and the stale content is replaced by the generated SKILL.

## Validation

- reproduced the stale-directory bug from a fresh temp `HOME` against `origin/main`
- reproduced the fixed behavior from the same temp-`HOME` setup on this branch
- `bun test test/gen-skill-docs.test.ts test/uninstall.test.ts`

## Notes

- closed the earlier PR opened from `whoislou` and replaced it with this one from `lmist`
- the terminal repro videos were rendered from real before/after transcripts and published on the separate asset branch above so they do not affect the merge diff


time-attack:
sick video
@lmist can yuo update it though

PR #920, opened 2026-04-08
fix(security): path validation bypass in upload and cookie-import commands

### LINKED ISSUE #707: Security: path validation bypass via relative paths in write-commands.ts

Hybirdss:
## Summary

`upload` and `cookie-import` commands use inline path validation while all newer commands already use the shared `validateReadPath` from `path-security.ts`.

The `cookie-import` inline check is genuinely weaker — it calls `path.resolve()` without `realpathSync()`, so a symlink inside a safe directory can traverse to arbitrary files. The `upload` inline check already calls `realpathSync` and was not actually vulnerable, but its 10-line inline validation diverges from the shared module — a maintenance risk as `path-security.ts` evolves.

This PR consolidates both commands onto `validateReadPath`, matching the pattern `eval` command already uses.

**cookie-import (real fix — symlink bypass):**
```typescript
// Before: path.resolve() without realpathSync — symlink traversal possible
if (!safeDirs.some(dir => isPathWithin(resolved, dir))) { throw ... }

// After:
validateReadPath(filePath);
```

**upload (consolidation only — was already safe):**
```typescript
// Before: 10-line inline validation with realpathSync (correct, but duplicated)
// After:
validateReadPath(fp);
```

## What's fixed

- `cookie-import`: symlink bypass closed — `validateReadPath` adds `realpathSync` resolution
- `upload`: consolidated onto shared module — no security change, reduced divergence
- Removed unused `SAFE_DIRECTORIES` and `isPathWithin` imports from write-commands.ts

## Context

`path-security.ts` was introduced in #907 to consolidate path validation. This PR catches the last two commands that weren't migrated.

## Test plan

- [x] Updated `upload command path validation` tests
- [x] Added `cookie-import command path validation` tests
- [x] `bun test browse/test/path-validation.test.ts` — 29 pass, 1 pre-existing fail (symlink test, unrelated)
- [x] `bun test` — full suite passes

Closes #707

### DIFF
--- a/browse/src/write-commands.ts
+++ b/browse/src/write-commands.ts
@@ -10,7 +10,7 @@ import type { BrowserManager } from './browser-manager';
 import { findInstalledBrowsers, importCookies, listSupportedBrowserNames } from './cookie-import-browser';
 import { generatePickerCode } from './cookie-picker-routes';
 import { validateNavigationUrl } from './url-validation';
-import { validateOutputPath } from './path-security';
+import { validateOutputPath, validateReadPath } from './path-security';
 import * as fs from 'fs';
 import * as path from 'path';
 import { TEMP_DIR } from './platform';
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks for this fix! The cookie-import path validation bypass has been resolved via PR #852's approach (which also fixes the hardcoded /tmp). The upload command's inline validation was also consolidated in this wave. Your analysis of the vulnerability was spot-on. Appreciate the contribution!

Hybirdss:
That's great. Thank you for review. I'll do my best!

PR #921, opened 2026-04-08
fix(security): validateOutputPath bypassed via existing symlinks in safe directories

Hybirdss:
## Summary

`validateOutputPath()` only resolves the parent directory — if the target file already exists as a symlink pointing outside safe directories, the write follows the symlink.

Every output command is affected: `screenshot`, `pdf`, `download`, `scrape`, `archive`.

## Reproduction

```bash
ln -s /etc/crontab /tmp/browse-screenshot-1234.png
$B screenshot /tmp/browse-screenshot-1234.png
# validateOutputPath passes: parent is /tmp (safe)
# fs.writeFileSync follows symlink → overwrites /etc/crontab
```

## Root cause

**File:** `browse/src/path-security.ts:33-56`

`validateOutputPath` resolves the parent directory via `realpathSync(dir)` but never checks the file itself. This is correct for new files (they don't exist yet), but misses the case where the target **already exists as a symlink**.

Compare with `validateReadPath` (line 59-80) which does `realpathSync(resolved)` on the file itself — that function is not vulnerable.

## Fix

Before the parent-directory check, `lstatSync` the resolved path. If it exists and is a symlink, `realpathSync` the target and verify it's within `SAFE_DIRECTORIES`. `ENOENT` (file doesn't exist yet) falls through to the existing parent-dir check.

```typescript
try {
  const stat = fs.lstatSync(resolved);
  if (stat.isSymbolicLink()) {
    const realTarget = fs.realpathSync(resolved);
    if (!SAFE_DIRECTORIES.some(dir => isPathWithin(realTarget, dir))) {
      throw new Error(`Path must be within: ${SAFE_DIRECTORIES.join(', ')}`);
    }
    return;
  }
} catch (e: any) {
  if (e.code !== 'ENOENT') throw e;
}
```

1 file, 20 lines added. All 6 output commands protected through the single shared function.

## Related

- #667 — same class of bug, narrower scope (cookie-import only). Fix PR #664 was closed.
- #707 — weaker inline path validation in upload/cookie-import (fixed separately in #920).
- #810, #847 — prior security waves that hardened other codepaths.

## Test plan

- [x] Pre-existing test `"blocks symlink inside /tmp pointing outside safe dirs"` now passes (was failing since the test was written)
- [x] `bun test browse/test/path-validation.test.ts` — 29 pass, 0 fail (was 29 pass, 1 fail)
- [x] `bun test` — full suite passes, 0 failures
- [x] PoC verified: `ln -s /etc/crontab /tmp/test.png` + `validateOutputPath('/tmp/test.png')` → throws
- [x] Normal `/tmp` writes still pass (no regression)

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/browse/src/path-security.ts
+++ b/browse/src/path-security.ts
@@ -33,7 +33,26 @@ const TEMP_ONLY = [TEMP_DIR].map(d => {
 export function validateOutputPath(filePath: string): void {
   const resolved = path.resolve(filePath);
 
-  // Resolve real path of the parent directory to catch symlinks.
+  // If the target already exists and is a symlink, resolve through it.
+  // Without this, a symlink at /tmp/evil.png → /etc/crontab passes the
+  // parent-directory check (parent is /tmp, which is safe) but the actual
+  // write follows the symlink to /etc/crontab.
+  try {
+    const stat = fs.lstatSync(resolved);
+    if (stat.isSymbolicLink()) {
+      const realTarget = fs.realpathSync(resolved);
+      const isSafe = SAFE_DIRECTORIES.some(dir => isPathWithin(realTarget, dir));
+      if (!isSafe) {
+        throw new Error(`Path must be within: ${SAFE_DIRECTORIES.join(', ')}`);
+      }
+      return; // symlink target verified, no need to check parent
+    }
+  } catch (e: any) {
+    // ENOENT = file doesn't exist yet, fall through to parent-dir check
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks for catching this critical symlink bypass! The lstatSync fix has been landed on the security-wave-triage branch with co-author credit. Ships in the next release.

Hybirdss:
Thank you. Have a great day!!!

PR #932, opened 2026-04-09
feat(design): support OPENAI_BASE_URL for API calls

jackjin1997:
## Summary
- add a central `openaiUrl()` resolver that honors `OPENAI_BASE_URL`, trims whitespace/trailing slashes, and defaults to `https://api.openai.com/v1`
- route every current design OpenAI API call through the resolver, including `design/prototype.ts`
- add regression coverage for default/custom/blank base URLs and for future hardcoded design API call sites

## Why
Some environments route OpenAI requests through an internal or compatible API gateway. The design tools previously hardcoded the public OpenAI endpoint at each call site.

## Validation
- `bun test design/test/auth.test.ts design/test/daemon.test.ts design/test/gallery.test.ts design/test/serve.test.ts` — 67 passed, 0 failed
- `bun build` for all nine changed API-call entry points — succeeded
- `git diff --check` — clean

The broader local design suite also contains environment-dependent tests: Playwright Chromium is not installed, daemon-discovery assumes a parent/child PID shape that differs under the local Bun runtime, and one existing HTTP-date retry assertion is timing-sensitive. None of those paths are changed here.

### DIFF
--- a/design/prototype.ts
+++ b/design/prototype.ts
@@ -8,6 +8,7 @@
 
 import fs from "fs";
 import path from "path";
+import { openaiUrl } from "./src/auth";
 
 const API_KEY = process.env.OPENAI_API_KEY;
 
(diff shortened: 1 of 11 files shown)


time-attack:
@jackjin1997 Rewrite this to support OPENAI_BASE_URL across every current design API call site, not only the paths covered by the existing helper.

jackjin1997:
Reworked this on top of current `main` to address the requested scope.

All current design OpenAI call sites now use the shared `OPENAI_BASE_URL` resolver, including `design/prototype.ts`. I also added a regression test that scans the design sources so a future hardcoded public endpoint is caught.

Fresh local verification: 67 relevant tests passed, all nine changed API-call entry points bundled successfully, and `git diff --check` is clean. Ready for another review.

PR #944, opened 2026-04-09
fix: disable parent-process watchdog in headed mode

navin-statisfy:
## Summary

- **Bug:** `$B connect` launches headed Chromium, but the browser shuts down ~15 seconds later because the parent-process watchdog (`BROWSE_PARENT_PID`) detects that the CLI process has exited and calls `shutdown()`
- **Root cause:** The connect command in `cli.ts` only passes `BROWSE_PARENT_PID=0` when the caller explicitly sets it (e.g. pair-agent), but headed mode always needs it disabled since the browser is meant to outlive the CLI
- **Fix:** Always set `BROWSE_PARENT_PID=0` in the connect command's server env, matching the existing pair-agent behavior

## Details

In `browse/src/server.ts:745-754`, the server polls the parent PID every 15 seconds and shuts down if it's gone. This is correct for headless mode (prevents orphaned Chrome), but wrong for headed mode where the user is looking at the browser window.

The fix is a one-line change in `browse/src/cli.ts` — unconditionally set `BROWSE_PARENT_PID=0` in the connect command's env instead of conditionally passing it through.

## Test plan

- [x] Run `$B connect` — browser stays open after CLI exits
- [x] Verified Chromium + server processes survive >10 seconds (previously died within ~5s)
- [x] Headless mode unaffected (connect command only runs for headed mode)

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -842,11 +842,11 @@ Refs:           After 'snapshot', use @e1, @e2... as selectors:
         BROWSE_PORT: '34567',
         BROWSE_SIDEBAR_CHAT: '1',
       };
-      // If parent explicitly set BROWSE_PARENT_PID=0 (pair-agent disabling
-      // self-termination), pass it through so startServer doesn't override it.
-      if (process.env.BROWSE_PARENT_PID === '0') {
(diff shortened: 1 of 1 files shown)


garrytan:

PR #970, opened 2026-04-11
fix(skills): Use CLAUDE_PLUGIN_ROOT in hook paths

### LINKED ISSUE #961: Hook commands reference `${CLAUDE_SKILL_DIR}` but Claude Code only exposes `${CLAUDE_PLUGIN_ROOT}` — all skill-registered hooks silently broken

Gujiassh:
## Summary
- replace legacy `CLAUDE_SKILL_DIR` hook and inline skill-path references with `CLAUDE_PLUGIN_ROOT`
- regenerate the affected skill docs so the shipped Claude skills and ship workflow stay aligned
- add regression checks to block future reintroduction of `CLAUDE_SKILL_DIR` in templates or generated skills

Fixes #961

## Validation
- `bun test test/gen-skill-docs.test.ts -t 'legacy CLAUDE_SKILL_DIR'`
- `rg -n "CLAUDE_SKILL_DIR" careful freeze guard investigate ship scripts/resolvers/review.ts test/gen-skill-docs.test.ts`

## Notes
- `bun test test/gen-skill-docs.test.ts test/host-config.test.ts` still reports pre-existing main-branch drift unrelated to this patch (`package.json` version `0.16.2.0` vs `VERSION` `0.16.3.0`, plus stale ship golden baselines).

### DIFF
--- a/careful/SKILL.md
+++ b/careful/SKILL.md
@@ -15,7 +15,7 @@ hooks:
     - matcher: "Bash"
       hooks:
         - type: command
-          command: "bash ${CLAUDE_SKILL_DIR}/bin/check-careful.sh"
+          command: "bash ${CLAUDE_PLUGIN_ROOT}/bin/check-careful.sh"
           statusMessage: "Checking for destructive commands..."
 ---
 <!-- AUTO-GENERATED from SKILL.md.tmpl — do not edit directly -->
--- a/careful/SKILL.md.tmpl
+++ b/careful/SKILL.md.tmpl
@@ -15,7 +15,7 @@ hooks:
     - matcher: "Bash"
       hooks:
         - type: command
-          command: "bash ${CLAUDE_SKILL_DIR}/bin/check-careful.sh"
+          command: "bash ${CLAUDE_PLUGIN_ROOT}/bin/check-careful.sh"
           statusMessage: "Checking for destructive commands..."
 sensitive: true
 ---
--- a/freeze/SKILL.md
+++ b/freeze/SKILL.md
@@ -16,12 +16,12 @@ hooks:
     - matcher: "Edit"
       hooks:
         - type: command
-          command: "bash ${CLAUDE_SKILL_DIR}/bin/check-freeze.sh"
+          command: "bash ${CLAUDE_PLUGIN_ROOT}/bin/check-freeze.sh"
           statusMessage: "Checking freeze boundary..."
     - matcher: "Write"
       hooks:
         - type: command
-          command: "bash ${CLAUDE_SKILL_DIR}/bin/check-freeze.sh"
+          command: "bash ${CLAUDE_PLUGIN_ROOT}/bin/check-freeze.sh"
           statusMessage: "Checking freeze boundary..."
 ---
 <!-- AUTO-GENERATED from SKILL.md.tmpl — do not edit directly -->
--- a/freeze/SKILL.md.tmpl
+++ b/freeze/SKILL.md.tmpl
@@ -16,12 +16,12 @@ hooks:
     - matcher: "Edit"
       hooks:
         - type: command
-          command: "bash ${CLAUDE_SKILL_DIR}/bin/check-freeze.sh"
+          command: "bash ${CLAUDE_PLUGIN_ROOT}/bin/check-freeze.sh"
           statusMessage: "Checking freeze boundary..."
     - matcher: "Write"
       hooks:
         - type: command
-          command: "bash ${CLAUDE_SKILL_DIR}/bin/check-freeze.sh"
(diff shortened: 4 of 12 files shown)


garrytan:
Closing — the codebase consistently uses CLAUDE_SKILL_DIR, which is the correct variable. These PRs propose switching to CLAUDE_PLUGIN_ROOT but that's not what we use. Issue #961 is also closed. Thank you for the contribution!

PR #977, opened 2026-04-11
Add GSTACK_SKIP_PLAYWRIGHT support

Gujiassh:
## Summary
- allow setup to skip the Playwright Chromium install when GSTACK_SKIP_PLAYWRIGHT=1
- skip the checks entirely and log the override so users can depend on system Chromium or GSTACK_CHROMIUM_PATH

## Testing
-


garrytan:
Thanks @Gujiassh — closing for staleness (merge conflicts, setup has been reworked since). The underlying need from #913 is still tracked there. Appreciate the contribution.

PR #982, opened 2026-04-12
fix: add native setup support for OpenCode

breakneo:
## Summary
- wire `./setup --host opencode` into the setup script instead of requiring the manual workaround
- install OpenCode skills into `~/.config/opencode/skills/gstack` with the nested runtime assets GStack actually needs
- extend host-config and setup tests to cover the OpenCode runtime asset layout

## Verification
- `bun test test/host-config.test.ts --test-name-pattern 'opencode symlinks returns nested runtime assets|validate passes for real configs'`
- `bun test test/gen-skill-docs.test.ts --test-name-pattern 'setup supports --host opencode|opencode symlinks'`
- `./setup --host opencode -q`
- `opencode run 'Respond with exactly OPENCODE_GSTACK_VISIBLE if you can access skills named gstack, qa, review, office-hours, and browse in this environment; otherwise respond with exactly OPENCODE_GSTACK_NOT_VISIBLE.'`

## Notes
- `bun test test/host-config.test.ts test/gen-skill-docs.test.ts` still has pre-existing unrelated failures in this checkout (golden-file regression snapshots and `package.json` version drift vs `VERSION`).
- I also patched the local Hermes skill note for this workflow because OpenCode's serialized `debug skill --pure` output did not reliably expose the exact `location` string I initially matched on.

### DIFF
--- a/hosts/opencode.ts
+++ b/hosts/opencode.ts
@@ -29,9 +29,9 @@ const opencode: HostConfig = {
   ],
 
   runtimeRoot: {
-    globalSymlinks: ['bin', 'browse/dist', 'browse/bin', 'gstack-upgrade', 'ETHOS.md'],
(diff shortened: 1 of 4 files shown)


garrytan:
Thanks @breakneo! Landing this in wave PR #1028 (v0.18.1.0). Closing here. 🚀

PR #986, opened 2026-04-13
fix(browse): keep headed browser session alive after connect

jbetala7:
## Summary
Fixes a headed-browser lifecycle bug where `browse connect` would successfully open a visible browser window, but the session would not stay attached reliably across subsequent commands.

## Problem
In headed mode, the browser often appeared to:
- launch successfully
- then lose the controlling session
- or close/orphan shortly after connect

This made real-browser workflows unreliable, especially for auth-required QA flows.

## Root Cause
The headed launcher path was not detaching robustly enough, and parent-PID lifecycle handling caused the server/browser process to terminate or become orphaned after launch.

## Changes
- force headed mode to disable parent-PID monitoring with `BROWSE_PARENT_PID=0`
- replace the detached headed launcher path with a proper detached `child_process.spawn(..., { detached: true, stdio: 'ignore' })`
- keep startup failure reporting on disk instead of relying on live stderr pipes during detached launch
- preserve the existing `bun run browse/src/server.ts` server startup path in the detached process

## Why
Headed mode is used specifically for workflows where session continuity matters:
- manual login handoff
- watching the browser work in real time
- authenticated QA testing
- agent/browser collaboration

If the session does not persist after `connect`, the main value of headed mode is lost.

## Repro Before
1. Run `browse connect`
2. Observe headed Chromium launch successfully
3. Run a follow-up browser command
4. Session is lost, browser closes, or tooling falls back to a different context

## Expected
After `browse connect`, the same headed browser session remains alive and attached for subsequent commands.

## Verification
- targeted tests passed
- rebuilt `browse/dist/browse`
- manually verified that:
  - headed mode remains in state file
  - server PID stays alive
  - Chromium process stays alive
  - `/health` reports healthy after launch delay

## Notes
This change is scoped to headed browser lifecycle stability and does not change normal headless browsing behavior.

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -227,25 +227,28 @@ async function startServer(extraEnv?: Record<string, string>): Promise<ServerSta
 
   let proc: any = null;
 
+  // Use Node's child_process.spawn with detached=true for all platforms.
+  // In practice this is more reliable than Bun.spawn(...).unref() for long-lived
+  // headed browser servers that must survive the launcher process exiting.
+  const detachedEnv = JSON.stringify({
+    BROWSE_STATE_FILE: config.stateFile,
+    BROWSE_PARENT_PID: String(process.pid),
+    ...(extraEnv || {}),
+  });
   if (IS_WINDOWS && NODE_SERVER_SCRIPT) {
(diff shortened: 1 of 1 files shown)


NikhileshNanduri:
lgtm

garrytan:

PR #989, opened 2026-04-13
fix(preamble): resolve CLAUDE.md from git root, not CWD

FireflySentinel:
## Summary

- The preamble's routing-rules check (`[ -f CLAUDE.md ]`) and vendoring check (`[ -d ".claude/skills/gstack" ]`) use CWD-relative paths. When launched from a subdirectory, worktree, or monorepo package, these checks never find the project root files.
- This causes the "Add routing rules to CLAUDE.md?" prompt to fire **every session**, even after the user accepts or declines — because the root CLAUDE.md is never found from `apps/web/` or similar subdirectories.
- Fix: resolve `_ROOT` via `git rev-parse --show-toplevel` (fallback to `pwd`) and use it for both checks. Also makes the routing grep case-insensitive (`-qi`) to prevent duplicate appends when the heading has different casing.

The fix is in `scripts/resolvers/preamble.ts` (single source of truth). All 32 generated SKILL.md files are regenerated.

### What changed

| File | Change |
|------|--------|
| `scripts/resolvers/preamble.ts` | Add `_ROOT` resolution, use `$_ROOT/CLAUDE.md` and `$_ROOT/.claude/skills/gstack` |
| `test/gen-skill-docs.test.ts` | Update assertion to match new grep pattern |
| 32 `*/SKILL.md` files | Regenerated from fixed template |

### Root cause

```bash
# Before (CWD-relative — breaks in subdirectories)
if [ -f CLAUDE.md ] && grep -q "## Skill routing" CLAUDE.md; then

# After (git root-relative — works everywhere)
_ROOT=${_ROOT:-$(git rev-parse --show-toplevel 2>/dev/null || pwd)}
if [ -f "$_ROOT/CLAUDE.md" ] && grep -qi "## Skill routing" "$_ROOT/CLAUDE.md"; then
```

This aligns with the project's own CLAUDE.md principles:
- "Don't hardcode branch names. Detect dynamically" — same principle applies to paths
- `_ROOT` uses `${_ROOT:-...}` to preserve the value if already set by `runtimeRoot` for hosts that use env vars

## Test plan

- [x] `bun test test/skill-validation.test.ts test/gen-skill-docs.test.ts` — 653 pass, 2 pre-existing failures (binary size + version mismatch), 0 new failures
- [x] Verified 0 remaining `[ -f CLAUDE.md ]` in generated SKILL.md files
- [x] Verified 32/32 SKILL.md files contain `$_ROOT/CLAUDE.md`
- [ ] Manual: launch gstack from a monorepo subdirectory, verify routing prompt does not repeat after declining

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -72,9 +72,11 @@ else
 fi
 # Session timeline: record skill start (local-only, never sent anywhere)
 ~/.claude/skills/gstack/bin/gstack-timeline-log '{"skill":"gstack","event":"started","branch":"'"$_BRANCH"'","session":"'"$_SESSION_ID"'"}' 2>/dev/null &
+# Resolve project root (CWD may be a subdirectory, worktree, or monorepo package)
+_ROOT=${_ROOT:-$(git rev-parse --show-toplevel 2>/dev/null || pwd)}
 # Check if CLAUDE.md has routing rules
 _HAS_ROUTING="no"
-if [ -f CLAUDE.md ] && grep -q "## Skill routing" CLAUDE.md 2>/dev/null; then
+if [ -f "$_ROOT/CLAUDE.md" ] && grep -qi "## Skill routing" "$_ROOT/CLAUDE.md" 2>/dev/null; then
   _HAS_ROUTING="yes"
 fi
 _ROUTING_DECLINED=$(~/.claude/skills/gstack/bin/gstack-config get routing_declined 2>/dev/null || echo "false")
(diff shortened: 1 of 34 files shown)


FireflySentinel:
The 32 SKILL.md files are mechanical regeneration from the preamble fix — happy to rebase if they conflict with other PRs landing first.

FireflySentinel:
Closing — withdrawing this contribution.

PR #995, opened 2026-04-14
fix setup support for opencode host

monaChoy:
## Summary

- fix `./setup --host opencode`, which is documented in `README.md` but currently rejected by the shell installer
- add OpenCode handling to `setup`, including host validation, auto-detection, skill doc generation, and skill installation
- verify the OpenCode install flow end to end by running `./setup --host opencode`

## Problem

`README.md` documents OpenCode support with:

- `./setup --host opencode`
- install path: `~/.config/opencode/skills/gstack-*`

However, the current `setup` script does not accept `opencode` as a valid host, so the documented install path is broken.

The repository already includes an OpenCode host definition in `hosts/opencode.ts`; this change wires that documented host into the existing shell installer flow.

## Testing

- installed `bun`
- ran `./setup --host opencode`
- confirmed successful output: `gstack ready (opencode).`
- confirmed generated skills under `~/.config/opencode/skills`
- confirmed runtime root under `~/.config/opencode/skills/gstack`

### DIFF
--- a/setup
+++ b/setup
@@ -22,6 +22,8 @@ CODEX_SKILLS="$HOME/.codex/skills"
 CODEX_GSTACK="$CODEX_SKILLS/gstack"
 FACTORY_SKILLS="$HOME/.factory/skills"
 FACTORY_GSTACK="$FACTORY_SKILLS/gstack"
+OPENCODE_SKILLS="$HOME/.config/opencode/skills"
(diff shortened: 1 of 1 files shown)


monaChoy:
Closing this in favor of #982, which addresses the same `opencode` setup gap with broader coverage and test updates.

PR #1004, opened 2026-04-15
fix: support documented generic setup hosts

sahaavi:
## Summary
This finishes setup support for the documented generic hosts that already exist in the host registry but were still rejected by ./setup.

Before this change:
- README.md documented ./setup --host opencode, ./setup --host cursor, and ./setup --host slate
- setup still rejected those values at argument parsing time
- the host registry already contained hosts/opencode.ts, hosts/cursor.ts, and hosts/slate.ts

## Root cause
setup still had a hardcoded host allowlist and install flow, while host registration had already moved into the declarative host config system.

## What changed
- added a generic setup path driven by host config export data
- accepted config-declared generic hosts in setup --host ...
- added auto-detection for generic setup hosts
- made generic host skill generation refresh on setup runs instead of only when the output directory is missing
- made generic-host eligibility explicit via install.setupStrategy
- fixed OpenCode/Cursor/Slate runtime asset exports so the generated skills and installed runtime root agree:
  - design/dist
  - review/design-checklist.md
  - review/greptile-triage.md
  - review/specialists/*.md
- added the missing .claude/skills/review -> .host/skills/gstack/review path rewrite for OpenCode/Cursor/Slate
- tightened docs so they describe the generic installer accurately
- added tests that assert generic hosts export the runtime assets their generated skills actually reference

## Verification
Passed:
- bun test test/host-config.test.ts --test-name-pattern 'hosts/index.ts|validateHostConfig|validateAllConfigs|HOST_PATHS derivation from configs|host-config-export.ts CLI|host config correctness'
- bun test test/gen-skill-docs.test.ts --test-name-pattern 'setup script validation|Parameterized host smoke tests|--host all'
- bun run gen:skill-docs --host all --dry-run
- bun run skill:check

Manual smoke:
- ./setup --host opencode -q
- this no longer fails on Unknown --host value: opencode
- it now proceeds into the existing setup flow and stops later in the pre-existing browser/bootstrap path with:
  - error: cannot write multiple output files without an output directory
  - gstack setup failed: Playwright Chromium could not be launched

## Notes
I also confirmed that stale golden baselines in the current checkout are unrelated to this change, so I kept this PR focused on the setup/host-config gap only.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -289,7 +289,9 @@ See [docs/ADDING_A_HOST.md](docs/ADDING_A_HOST.md) for the full guide. Short ver
 4. Run `bun run gen:skill-docs --host myhost`
 5. Run `bun test` (parameterized tests auto-cover it)
 
-Zero generator, setup, or tooling code changes needed.
+Zero generator changes needed. Setup is automatic for hosts that opt into the
+generic symlink-generated installer; hosts with bespoke install flows may still
+need a targeted setup branch.
 
 ### Adding a new skill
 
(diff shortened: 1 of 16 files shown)


time-attack:
@sahaavi Rewrite this against current code using supported Bash parsing, safe links, and ownership tracking that prevents destructive runtime replacement.

PR #1014, opened 2026-04-16
plan-eng-review: add throughput planning readout

al-buda-noygear:
## Summary
- add a required `Throughput planning readout` section to `plan-eng-review`
- make every eng review emit a strict build stackrank, max safe parallelism, expected wall clock, and practical total effort
- carry the same requirement into the skill template so the prompt stays durable

## Why
For implementation planning, the scarcest resource is usually not code generation but human attention: deciding what blocks what, where coordination is required, and how much work can safely run in parallel.

The existing eng review already covers architecture, tests, failure modes, and worktree parallelization, but it does not force a final execution readout that answers:
- what should happen first
- what can happen at the same time
- where human attention is actually required
- how long the accepted scope will really take
- how many agent can run in parallel (minimize dependencies, maximize parallelism)
- what each agent should execute on in what order (execution map when dependencies do exist)

This change makes that throughput framing a first-class part of every eng review.

### DIFF
--- a/plan-eng-review/SKILL.md
+++ b/plan-eng-review/SKILL.md
@@ -1269,6 +1269,11 @@ Rules:
    - `X-Yh human time`
    - `A-Bh CC time`
 
+   Arithmetic rule:
+   - every subtotal or total must equal the direct sum of the table ranges above it
+   - do not silently add coordination or contingency buffer into those totals
+   - if a buffer is warranted, show it separately under wall-clock planning or as an explicit buffer line item
+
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this. The new readout duplicates existing task ordering and parallelization guidance rather than extending the current task artifact. Rebase onto the current section source, add only capacity, critical-path, and wall-clock behavior with non-double-counting rules, and include focused evaluations.

PR #1035, opened 2026-04-17
fix(browse): respect GSTACK_CHROMIUM_PATH in headless mode

### LINKED ISSUE #906: browse: GSTACK_CHROMIUM_PATH ignored in headless mode
## Problem

`GSTACK_CHROMIUM_PATH` is only respected in `launchHeaded()` but not in `launch()` (headless mode). Skills like `/qa` use headless mode by default, so setting `GSTACK_CHROMIUM_PATH` has no effect.

This breaks browse on NixOS (and likely other non-FHS distros like Guix) where the bundled `chromium_headless_shell` binary fails because system libraries aren't in standard paths:

```
chrome-headless-shell: error while loading shared libraries: libglib-2.0.so.0: cannot open shared object file
```

## Root Cause

In `browse/dist/server-node.mjs`:

- **Line ~2491** (`launchHeaded`): reads `process.env.GSTACK_CHROMIUM_PATH` and passes it as `executablePath` to Playwright
- **Line ~2440** (`launch`): calls `chromium.launch()` without `executablePath`, always using the bundled binary

## Workaround

Export `LD_LIBRARY_PATH` from the system Chrome's Nix wrapper before running browse:

```bash
CHROME_LIBS=$(grep "LD_LIBRARY_PATH='" $(readlink -f $(which google-chrome-stable)) | grep "^LD_LIBRARY_PATH='" | sed "s/LD_LIBRARY_PATH='//;s/'.*//" | tr '\n' ':')
export LD_LIBRARY_PATH="$CHROME_LIBS"
```

This is fragile and has to be done every session.

## Suggested Fix

Read `GSTACK_CHROMIUM_PATH` in `launch()` too:

```diff
  async launch() {
+   const executablePath = process.env.GSTACK_CHROMIUM_PATH || undefined;
    const launchArgs = [];
    // ...
    this.browser = await chromium.launch({
      headless: useHeadless,
      chromiumSandbox: process.platform !== "win32",
+     ...executablePath ? { executablePath } : {},
      ...launchArgs.length > 0 ? { args: launchArgs } : {}
    });
```

## Environment

- OS: NixOS (Linux 6.19.11)
- gstack: v0.15.16.0
- System browser: Google Chrome 146.0.7680.177 (installed via Nix)
- `GSTACK_CHROMIUM_PATH` set to `/run/current-system/sw/bin/google-chrome-stable`

mvanhorn:
## Summary

`GSTACK_CHROMIUM_PATH` is only respected in `launchHeaded()` (`browser-manager.ts:270`) but not in `launch()` (`browser-manager.ts:183`). Skills like `/qa` use headless mode by default, so setting `GSTACK_CHROMIUM_PATH` has no effect when the bundled `chromium_headless_shell` fails on NixOS and other non-FHS distros.

## Changes

Read `GSTACK_CHROMIUM_PATH` in `launch()` and pass it as `executablePath` to `chromium.launch()`, matching the existing pattern in `launchHeaded()`.

`browse/src/browser-manager.ts`: 3 lines added.

## Testing

`bun test` passes (665 pass across skill validation and gen-skill-docs). The change mirrors the existing env var handling in `launchHeaded()` at line 270.

Fixes #906

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -180,12 +180,15 @@ export class BrowserManager {
       console.log(`[browse] Extensions loaded from: ${extensionsDir}`);
     }
 
+    const executablePath = process.env.GSTACK_CHROMIUM_PATH || undefined;
+
     this.browser = await chromium.launch({
       headless: useHeadless,
       // On Windows, Chromium's sandbox fails when the server is spawned through
       // the Bun→Node process chain (GitHub #276). Disable it — local daemon
       // browsing user-specified URLs has marginal sandbox benefit.
       chromiumSandbox: process.platform !== 'win32',
+      ...(executablePath ? { executablePath } : {}),
       ...(launchArgs.length > 0 ? { args: launchArgs } : {}),
     });
 


jbetala7:
This specific headless `GSTACK_CHROMIUM_PATH` gap appears fixed on current `main` now. The merged maintainer wave #1642 / v1.43.2.0 carries the headless `executablePath: process.env.GSTACK_CHROMIUM_PATH || undefined` change in `browse/src/browser-manager.ts`, and also includes the later `isCustomChromium()` guard so baked-extension Chromium builds do not double-load the extension.

Given that, I would treat #1642 as the canonical shipped fix for the browser-manager part of this PR. If issue #906 still has a remaining setup/install angle, it likely needs a fresh narrow PR against current `main` rather than this older branch.

mvanhorn:
Good catch @jbetala7 — confirmed `browser-manager.ts:444` on current `main` carries `executablePath: process.env.GSTACK_CHROMIUM_PATH || undefined`, plus the `isCustomChromium()` guard. #1642 supersedes this PR; closing. If #906 still has a remaining install/setup angle I'll spin up a narrower PR against current main. Thanks for the heads-up.

PR #1049, opened 2026-04-17
fix: office-hours marks sessions "success" without writing design doc

walton-chris:
## Summary

The office-hours skill can log `outcome:success` in the timeline even
when Phase 5 (design doc write) was skipped. The session looks
complete in telemetry, but no `*-design-*.md` artifact exists on disk,
silently breaking downstream skills (`/plan-ceo-review`,
`/plan-eng-review`) that read design docs from
`~/.gstack/projects/{slug}/`.

## Repro

Observed in the wild on an apex-coach session (2026-04-12):

```
{"skill":"office-hours","event":"completed","branch":"master",
 "outcome":"success","duration_s":"0","session":"unknown",
 "ts":"2026-04-12T11:59:40.368Z"}
```

`duration_s:"0"` and `session:"unknown"` are a separate shell-state bug
addressed in a companion PR. But the real issue here is
`outcome:"success"` logged against `~/.gstack/projects/apex-coach/`
which contains zero `*-design-*.md` files. A session with no
persisted artifact is not a successful session.

## Fix

Two changes to `office-hours/SKILL.md.tmpl`:

1. New section **"Pre-Telemetry: Verify Design Doc Artifact"** inserted
   before the `{{LEARNINGS_LOG}}` placeholder. It re-derives `SLUG`,
   uses the preamble's `~/.gstack/sessions/\$PPID` touch file as a
   session-start time reference, and finds any `*-design-*.md`
   written after that time. Emits `ARTIFACT_STATUS: ok` or `no_doc`.

2. New rule in **Important Rules** instructing Claude to use the
   artifact check to drive `OUTCOME` in the preamble's telemetry
   block. Log `outcome:no_doc` when no doc was written, never
   `outcome:success`.

Net diff: +56 lines, 0 deletions. Zero impact on sessions that
correctly write a doc.

## Test plan

- [ ] Run \`/office-hours\` on a new project, complete Phase 5, confirm
      timeline logs \`outcome:success\` and artifact exists
- [ ] Run \`/office-hours\`, abandon mid-conversation, confirm timeline
      logs \`outcome:no_doc\`
- [ ] Run \`/office-hours\` in a project with existing old design docs,
      confirm the \`-newer\` check correctly distinguishes this session's
      doc from prior ones

Generated with [Claude Code](https://claude.com/claude-code)

PR #1050, opened 2026-04-17
fix: persist session state across Bash tool calls for accurate telemetry

walton-chris:
## Summary

The preamble sets `_TEL_START` and `_SESSION_ID` as shell variables,
but shell state does not survive between separate Bash tool
invocations (each Bash call is a fresh shell). When the final
`## Telemetry (run last)` block runs in a new shell, both variables
are empty — the arithmetic evaluates to 0 and the session id falls
back to `"unknown"`.

Net effect: every timeline entry logs `duration_s:"0"` and
`session:"unknown"`, regardless of how long the skill actually ran.
Session-correlation analytics are currently useless.

## Evidence

Apex-coach office-hours session 2026-04-12:

```
{"skill":"office-hours","event":"started","branch":"master",
 "session":"10269-1775994715","ts":"2026-04-12T11:51:56.513Z"}
{"skill":"office-hours","event":"completed","branch":"master",
 "outcome":"success","duration_s":"0","session":"unknown",
 "ts":"2026-04-12T11:59:40.368Z"}
```

~8 real minutes, logged as 0. Session ids don't correlate. Every
other `completed` timeline entry across the install has the same
signature.

## Fix

One edit to `scripts/resolvers/preamble.ts`:

1. After setting `_TEL_START` and `_SESSION_ID`, write a small state
   file to `~/.gstack/analytics/.session-state` via heredoc.
2. At the top of the telemetry block, source the state file (using
   `.` not `source` for POSIX portability).
3. Compute duration with `\${_TEL_START:-\$_TEL_END}` fallback so
   missing-file scenarios gracefully degrade to the current behavior
   rather than erroring.
4. Remove the state file after telemetry fires.

All SKILL.md files regenerated via `bun run gen:skill-docs --host all`.
Golden-file fixtures (claude/codex/factory ship) updated.

Scope: preamble-tier 3 skills pick up the fix via regeneration.
Minimal-preamble skills (freeze, guard, careful, unfreeze,
gstack-upgrade) don't include the telemetry block at all and are
unaffected.

## Test plan

- [ ] Run any skill, confirm `~/.gstack/analytics/.session-state`
      is written during preamble
- [ ] Confirm `duration_s` in the skill's timeline.jsonl event
      is a non-zero number matching real session duration
- [ ] Confirm `session` is the PID-timestamp format set in preamble
      (not `"unknown"`)
- [ ] Delete the state file mid-session, confirm telemetry degrades
      gracefully (duration 0, no crash)
- [ ] Confirm state file is cleaned up after the skill completes

## Relationship to #1049

PR #1049 (office-hours artifact verification) is orthogonal but
complementary. Once both land, office-hours correctly tracks duration
AND correctly distinguishes successful sessions (artifact written)
from silent failures (no artifact). This PR is generic; that one is
office-hours-specific.

## Commit history on this branch

4 commits. First is the substantive fix; the rest are regen/fixture
updates needed to land the first commit cleanly. Garry's usual
squash-merge pattern will collapse them on merge.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -50,6 +50,14 @@ _SESSION_ID="$$-$(date +%s)"
 echo "TELEMETRY: ${_TEL:-off}"
 echo "TEL_PROMPTED: $_TEL_PROMPTED"
 mkdir -p ~/.gstack/analytics
+# Persist session state so the telemetry block (run in a separate Bash tool
+# call where shell vars don't survive) can recover duration and session id.
+cat > ~/.gstack/analytics/.session-state 2>/dev/null <<SESSTATE || true
+_TEL_START=$_TEL_START
+_SESSION_ID=$_SESSION_ID
+_TEL=${_TEL:-off}
+_BRANCH=$_BRANCH
+SESSTATE
 if [ "$_TEL" != "off" ]; then
 echo '{"skill":"gstack","ts":"'$(date -u +%Y-%m-%dT%H:%M:%SZ)'","repo":"'$(basename "$(git rev-parse --show-toplevel 2>/dev/null)" 2>/dev/null || echo "unknown")'"}'  >> ~/.gstack/analytics/skill-usage.jsonl 2>/dev/null || true
 fi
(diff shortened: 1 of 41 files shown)


time-attack:
@walton-chris Rewrite this so timing and session state persist reliably between instructed shell calls, where they are currently created and later consumed.

PR #1053, opened 2026-04-17
feat(cso): add --fix flag with 9 provably safe auto-fixes

andreycpu:
## What this does

Adds `/cso --fix` — runs the full audit, then applies a curated set of fixes where the correct change is deterministic and the risk of breakage is near-zero. Zero changes to existing audit phases.

Directly follows Garry's feedback on the fortress PR: start with provably safe patterns, quality over quantity, each fix verified.

## Fix catalog (9 patterns)

| Fix | Pattern | Why it's safe |
|-----|---------|---------------|
| FIX-01 | `.gitignore` secret hardening | Additive only — never removes entries |
| FIX-02 | Create `.gitleaks.toml` baseline | Additive only |
| FIX-03 | `npm audit fix` (no `--force`) | Respects semver, patches only |
| FIX-04 | `rejectUnauthorized: false` → `true` | Single token flip, TLS bypass is never intentional in prod |
| FIX-05 | `verify=False` → `True` (Python requests) | Single token flip |
| FIX-06 | `InsecureSkipVerify: true` → `false` (Go) | Single token flip |
| FIX-07 | `httpOnly: false` → `true` on cookies | Never breaks server-side code |
| FIX-08 | `secure: false` → `true` on cookies | Single flag flip |
| FIX-09 | `DEBUG=true` → `false` in production env files | Config only, removes information leakage |

Every fix includes an explicit "Safe because" rationale in the skill. The agent skips any fix it can't apply with certainty and reports why in the summary table.

## Usage

```
/cso --fix              # full audit + auto-fix
/cso --fix --diff       # branch-only audit + auto-fix
/cso --fix --code       # code-only audit + auto-fix
```

`--fix` is combinable with any existing scope flag.

## Changes

- `cso/SKILL.md.tmpl` — added `--fix` to Arguments, rule 7 to Mode Resolution, Phase 15 (Auto-Fix Engine), `Edit` to allowed-tools
- `cso/SKILL.md` — regenerated from template

No changes to Phases 0-14 or any other skill.

### DIFF
--- a/cso/SKILL.md
+++ b/cso/SKILL.md
@@ -16,6 +16,7 @@ allowed-tools:
   - Grep
   - Glob
   - Write
+  - Edit
   - Agent
   - WebSearch
   - AskUserQuestion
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej Performs nine unattended repository mutations without preview, rollback, clean-worktree protection, or regression tests. Several “safe” fixes can break valid configurations.

PR #1054, opened 2026-04-18
fix: namespace generated Hermes skill names

dkoh12:
## Summary
- add host-config support for host-specific emitted skill names
- namespace generated Hermes skill frontmatter names to match their gstack-* directories
- add a regression test covering Hermes skill-name generation

## Problem
Hermes deduplicates skills by frontmatter `name`, not by directory name. Generated Hermes skills were written to namespaced directories like `gstack-review/`, but their frontmatter still used generic names like `review`, causing Hermes to suppress them as duplicates of existing skills.

## Test Plan
- bun test test/gen-skill-docs.test.ts -t "Hermes frontmatter names are namespaced to match generated skill names"
- bun run scripts/gen-skill-docs.ts --host hermes
- verify `hermes skills list` shows `gstack-review`, `gstack-qa`, and `gstack-ship`

### DIFF
--- a/hosts/hermes.ts
+++ b/hosts/hermes.ts
@@ -15,6 +15,7 @@ const hermes: HostConfig = {
     mode: 'allowlist',
     keepFields: ['name', 'description'],
     descriptionLimit: null,
+    nameTransform: 'external-skill-name',
   },
 
   generation: {
--- a/scripts/gen-skill-docs.ts
+++ b/scripts/gen-skill-docs.ts
@@ -221,7 +221,7 @@ policy:
  * Codex: keeps name + description only, enforces 1024-char limit.
  * Factory: keeps name + description + user-invocable, conditionally adds disable-model-invocation.
  */
-function transformFrontmatter(content: string, host: Host): string {
+function transformFrontmatter(content: string, host: Host, generatedSkillName?: string): string {
   const hostConfig = getHostConfig(host);
   const fm = hostConfig.frontmatter;
 
@@ -245,6 +245,9 @@ function transformFrontmatter(content: string, host: Host): string {
   const frontmatter = content.slice(fmStart + 4, fmEnd);
   const body = content.slice(fmEnd + 4);
   const { name, description } = extractNameAndDescription(content);
+  const emittedName = fm.nameTransform === 'external-skill-name' && generatedSkillName
+    ? generatedSkillName
+    : name;
 
   // Description limit enforcement
   if (fm.descriptionLimit) {
@@ -252,11 +255,11 @@ function transformFrontmatter(content: string, host: Host): string {
     if (description.length > fm.descriptionLimit) {
       if (behavior === 'error') {
         throw new Error(
-          `${hostConfig.displayName} description for "${name}" is ${description.length} chars (max ${fm.descriptionLimit}). ` +
+          `${hostConfig.displayName} description for "${emittedName}" is ${description.length} chars (max ${fm.descriptionLimit}). ` +
           `Compress the description in the .tmpl file.`
         );
       } else if (behavior === 'warn') {
-        console.warn(`WARNING: ${hostConfig.displayName} description for "${name}" exceeds ${fm.descriptionLimit} chars`);
+        console.warn(`WARNING: ${hostConfig.displayName} description for "${emittedName}" exceeds ${fm.descriptionLimit} chars`);
       }
       // 'truncate' — silently proceed
     }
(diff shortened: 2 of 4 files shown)


sixianli:
+1 — independently hit this exact bug on Hermes Agent today (fresh install via 

`./setup`). Confirming the diagnosis: only 2 of 42 generated `gstack-*` skills 

were registered, and the 2 that survived (`gstack-upgrade`, `gstack-open-gstack-browser`) 

are precisely the ones whose frontmatter `name` happens to be unique enough not 

to collide.

The 6 skills that silently lose to existing Hermes skills via name collision in 

my setup: `ship`, `qa`, `review`, `autoplan`, `retro`, `browse`.

Just checked out this branch, ran `bun run scripts/gen-skill-docs.ts --host hermes`, 

and `hermes skills list | grep gstack | wc -l` now returns 42 (up from 2). Fix 

works as advertised.

Would love to see this merged — without it, gstack on Hermes is effectively 

unusable out of the box.

time-attack:
@dkoh12 Rewrite generated Hermes frontmatter names with the same namespace as their gstack-prefixed directories to prevent registry collisions.

yolo-jared:
Adjacent instance of the same "frontmatter `name` disagrees with the prefixed directory" class, but on the **consumption** side and on the **Claude** host — so I don't think this issue's fix covers it. Posting here rather than opening a new issue; happy to split it out if you'd prefer.

**Where this differs from the Hermes case above:** #1054 is generation writing an *unprefixed* name into a *prefixed* directory, and the symptom is silent dedup. This one is the reverse — the installer writes a *prefixed* name and a consumer hardcodes the *unprefixed* one, and the symptom is a hard abort.

### `/autoplan` aborts at every phase when `skill_prefix: true`

```
gstack-autoplan-snapshot: Methodology skill identity does not match this phase
```

Exit 1. `/autoplan`'s instruction for that step is "On helper errors, stop", so it terminates at Phase 1. All four phases (ceo, design, dx, eng) share the validator, so none run.

### Cause

`bin/gstack-patch-names:19-21`, invoked by `bin/gstack-relink:367-368` when `skill_prefix` is true, rewrites the installed frontmatter:

```bash
if [ "$DO_PREFIX" -eq 1 ]; then
  case "$cur" in gstack-*) continue ;; esac
  new="gstack-$cur"
```

`bin/gstack-autoplan-snapshot.ts:496,508` then requires the bare spelling:

```ts
const skill = `plan-${phase === 'dx' ? 'devex' : phase}-review`;
...
if (!frontmatter?.[1]!.split(/\r?\n/).includes(`name: ${skill}`)) {
  throw new Error('Methodology skill identity does not match this phase');
}
```

The `.tmpl` carries the unprefixed name and generation preserves it; the prefix is applied afterwards to the installed copy only. The validator was written against the template's spelling.

### Reproduction

Clean checkout of `origin/main` @ a6b3a57 (v1.87.4.0), no local customization:

```
$ bun bin/gstack-autoplan-snapshot.ts methodology ceo <scratch>/plan-ceo-review/SKILL.md <restore>
exit 0     # JSON manifest emitted

$ bin/gstack-patch-names <scratch> true
  patched name: field in 1 skills
$ grep -m1 '^name:' <scratch>/plan-ceo-review/SKILL.md
name: gstack-plan-ceo-review

$ bun bin/gstack-autoplan-snapshot.ts methodology ceo <scratch>/plan-ceo-review/SKILL.md <restore>
exit 1
gstack-autoplan-snapshot: Methodology skill identity does not match this phase
```

Single variable: the frontmatter `name:` line, changed by gstack's own patcher.

### Why it shipped

Both halves are tested; the seam isn't. `test/relink.test.ts` covers prefix mode thoroughly (including the #2738 render-dir regression) and `test/autoplan-snapshot.test.ts` covers the helper in 451 lines, but `grep -rl "skill_prefix" test/autoplan-*.test.ts` returns zero across all 37 files — nothing sets `skill_prefix: true` and then calls the `methodology` subcommand.

### Scope

A sweep of `bin/`, `scripts/`, and `hosts/` (no `--include` filter) for the same defect class on the Claude side — a runtime comparison of installed frontmatter `name:` against a hardcoded literal — found exactly one instance, the line above. Prefix mode is otherwise consistent there: `gstack-relink` renames the host-side directory and the frontmatter together and keeps them in sync across upgrades and gbrain renders, and `bin/gstack-uninstall:195-207` already accepts both spellings.

### Suggested fix

Accept either spelling. The prefix is applied uniformly to all four phase skills, never selectively, so tolerating it can't let one phase's methodology pass as another's:

```ts
const identities = [`name: ${skill}`, `name: gstack-${skill}`];
if (!frontmatter?.[1]!.split(/\r?\n/).some(line => identities.includes(line))) {
  throw new Error('Methodology skill identity does not match this phase');
}
```

I have this plus a regression test running locally against a prefixed install — all four phases green, and the test goes red against unpatched code (4 of 12 cases). Full autoplan suite 738 pass / 11 fail vs 726 / 11 on byte-identical-to-upstream code; same 11 pre-existing PTY failures, no regressions. Say the word if a PR would help, or feel free to take the two-line fix directly.

Environment: v1.87.4.0, `skill_prefix: true`, macOS arm64, Claude Code host.

PR #1083, opened 2026-04-19
docs(privacy): disclose hourly GitHub VERSION fetch in update-check

### LINKED ISSUE #1081: Undisclosed hourly GitHub network call in every skill preamble (update-check)
## Summary

Every gstack skill preamble runs \`gstack-update-check\`, which on cache miss (~every 60 minutes) makes a GET request to:

```
https://raw.githubusercontent.com/garrytan/gstack/main/VERSION
```

This happens regardless of the user's \`telemetry\` setting — including when \`telemetry: off\` is set. The request is never disclosed in the README privacy section, the telemetry consent prompt, or the setup flow.

## Why this matters

The telemetry consent UI describes what data is sent to Supabase. A user who sets \`telemetry: off\` reasonably expects no outbound network activity tied to gstack usage. But the GitHub fetch still fires, exposing their IP address to GitHub once per hour during active sessions.

When telemetry is non-\`off\`, the slow-path fetch also fires a secondary Supabase POST:

```bash
curl -X POST "${SUPA_URL}/functions/v1/update-check" \
  -d "{\"version\":\"$LOCAL\",\"os\":\"$_OS\"}"
```

This Supabase ping is correctly gated on telemetry opt-in, but the GitHub fetch is not.

## Affected location

\`bin/gstack-update-check\`, Step 4 (slow path):

```bash
REMOTE="$(curl -sf --max-time 5 "$REMOTE_URL" 2>/dev/null || true)"
```

The update check cache TTL is 60 minutes for \`UP_TO_DATE\` results and 720 minutes for \`UPGRADE_AVAILABLE\`. So in a typical day of active gstack use this fires roughly once per working session.

## Proposed fix

Add one sentence to the README privacy section and the telemetry consent prompt that covers the update check:

> **Update checks:** Once per hour, gstack fetches the current version number from GitHub to check for updates. This is a plain GET request that reveals your IP to GitHub. Set \`update_check: false\` to disable it entirely.

The \`update_check: false\` escape hatch already exists — it just isn't mentioned near the privacy disclosure.

No code change needed if the fix is documentation only. If the team prefers to gate the GitHub fetch behind the telemetry consent (rather than just document it), that's a separate call.

burakbengi:
## Summary

- Adds one bullet to the `## Privacy & Telemetry` section of `README.md` disclosing the hourly GET request that `gstack-update-check` makes to `raw.githubusercontent.com`.
- Points to the existing `update_check: false` escape hatch that already silences it.

## Why

Every gstack skill preamble calls `gstack-update-check`, which on cache miss (~once per hour) fetches:

```
GET https://raw.githubusercontent.com/garrytan/gstack/main/VERSION
```

This fires regardless of the user's `telemetry` setting — including when `telemetry: off` is set. The Privacy & Telemetry section of the README covered opt-in telemetry (Supabase) but was silent on this always-on network call.

The call is harmless by design (no auth, no payload, just a version string). The gap is that a user who reads the privacy section and sets `telemetry: off` expecting zero outbound activity would still be surprised by it. The fix documents what happens and how to disable it.

No code change — `update_check: false` already exists in `gstack-config` and already silences the call. This PR just makes users aware of it.

Fixes #1081

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,7 @@ gstack includes **opt-in** usage telemetry to help improve the project. Here's e
 - **What's sent (if you opt in):** skill name, duration, success/fail, gstack version, OS. That's it.
 - **What's never sent:** code, file paths, repo names, branch names, prompts, or any user-generated content.
 - **Change anytime:** `gstack-config set telemetry off` disables everything instantly.
+- **Update checks:** Once per hour, gstack fetches the current version number from `raw.githubusercontent.com` to check for upgrades. This is a plain GET request — no auth, no payload — but it does reach GitHub's servers regardless of your telemetry setting. Set `update_check: false` in `~/.gstack/config.yaml` to disable it entirely.
 
 Data is stored in [Supabase](https://supabase.com) (open source Firebase alternative). The schema is in [`supabase/migrations/`](supabase/migrations/) — you can verify exactly what's collected. The Supabase publishable key in the repo is a public key (like a Firebase API key) — row-level security policies deny all direct access. Telemetry flows through validated edge functions that enforce schema checks, event type allowlists, and field length limits.
 


time-attack:
Thanks for this. The disclosure describes only an hourly raw GitHub request, but current update behavior includes multiple requests, cache windows, fallback behavior, and an optional telemetry-gated POST. Rewrite it to document the current sequence, data and IP exposure, both cache windows, fallback, and the update_check opt-out.

PR #1096, opened 2026-04-20
fix: avoid printing partial Render API key in setup-deploy

### LINKED ISSUE #1078: Avoid printing partial Render API key during setup-deploy check
**File:** `setup-deploy/SKILL.md` line 851

**Current instruction:**
```
Check for Render API key: `echo $RENDER_API_KEY | head -c 4` (don't expose the full key)
```

The inline comment shows the intent, but the instruction still prints the first 4 characters of the key into terminal scrollback, screen recordings, and screenshots. A presence check would serve the same "is the key available?" purpose without writing any key bytes to stdout.

## Suggested replacement

```bash
if [ -n "${RENDER_API_KEY:-}" ]; then
  echo "RENDER_API_KEY: set"
else
  echo "RENDER_API_KEY: not set"
fi
```

If a stable fingerprint for cross-run correlation is useful:

```bash
if [ -n "${RENDER_API_KEY:-}" ]; then
  printf 'RENDER_API_KEY: sha256:'
  printf '%s' "$RENDER_API_KEY" | shasum -a 256 | cut -c1-8
fi
```

(The `:-` guards against `set -u` if the var is unset; `shasum` is macOS-native, `sha256sum` on Linux.)

## Why this is worth the cheap swap

Partial-secret printing tends to get normalized once it ships in a skill doc — downstream users copy the pattern. A four-character prefix may be low severity in isolation, but:
- It still leaks into scrollback, logs, screen shares, and IDE terminal captures
- Many providers use fixed prefixes (`rnd_`, `sk-`, etc.), so the leaked chars are often the non-entropy part anyway
- The presence check achieves the same intent with zero byte leakage

Low-cost hardening, not a severe vuln — happy to send a PR if useful.

---
*Found during a security audit of a Claude Code + Codex + MCP stack. Same class of finding was fixed in our own setup scripts after an adversarial review flagged it.*

JiayuuWang:
## Summary

The previous instruction in `/setup-deploy` used `echo $RENDER_API_KEY | head -c 4` to check for the Render API key's presence. This still leaked 4 bytes of the secret into terminal scrollback, screen recordings, and screenshots.

Replace with a simple presence check that confirms whether the variable is set, without exposing any bytes.

## Test plan

- [ ] Run `/setup-deploy` on a project with `render.yaml`
- [ ] Verify the Render API key presence check shows only "set" / "not set"
- [ ] Confirm no bytes of the key appear in terminal scrollback

Fixes #1078

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/setup-deploy/SKILL.md
+++ b/setup-deploy/SKILL.md
@@ -856,7 +856,14 @@ Ask the user to confirm the production URL. Some Fly apps use custom domains.
 If `render.yaml` detected:
 
 1. Extract service name and type from render.yaml
-2. Check for Render API key: `echo $RENDER_API_KEY | head -c 4` (don't expose the full key)
+2. Check for Render API key:
+```bash
+if [ -n "${RENDER_API_KEY:-}" ]; then
+  echo "RENDER_API_KEY: set"
+else
+  echo "RENDER_API_KEY: not set"
+fi
+```
 3. Infer URL: `https://{service-name}.onrender.com`
 4. Render deploys automatically on push to the connected branch — no deploy workflow needed
 5. Set health check: the inferred URL
--- a/setup-deploy/SKILL.md.tmpl
+++ b/setup-deploy/SKILL.md.tmpl
@@ -99,7 +99,14 @@ Ask the user to confirm the production URL. Some Fly apps use custom domains.
 If `render.yaml` detected:
 
 1. Extract service name and type from render.yaml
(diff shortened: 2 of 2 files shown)


time-attack:
@16francej Merge this version because it removes secret-prefix output from both maintained copies, safely handles unset variables, and passes the focused behavioral tests.

PR #1097, opened 2026-04-20
fix: surface non-EEXIST errors in acquireServerLock instead of masking them

### LINKED ISSUE #1084: browse: acquireServerLock catch block silently swallows non-EEXIST errors, masks real bugs as phantom lock contention
## Observation

`acquireServerLock()` in `browse/src/cli.ts` (lines ~283–307 on current `main`) uses a bare `catch {}` on the `fs.openSync(lockPath, 'wx')` call. The same applies to the inner catch on the holder-PID read.

When ANY error other than `EEXIST` occurs (permissions, disk full, Bun fs regressions like the one fixed in #599 via `cd66fc2f`, future Windows quirks, etc.), execution falls through to:

1. Inner `readFileSync(lockPath)` → throws (file doesn't exist, or unreadable)
2. Inner catch → `return null`
3. Caller sees null → prints `"Another instance is starting the server, waiting..."`
4. 15s timeout → `"Timed out waiting for another instance to start the server"`

**The user sees phantom lock contention that has nothing to do with locking.**

## Real-world consequence

I hit this today on Windows 11, Bun 1.3.10, on a gstack 0.13.2.0 install that happened to sit on commit `247fc3ba` (PR #603) instead of `cd66fc2f` (PR #602 — the #599 fix). Both commits were tagged v0.13.2.0 but sit on divergent paths, and #603's merge appears to have carried the pre-fix cli.ts back in. `main` HEAD has the correct `'wx'` code, but anyone installed at that snapshot gets the buggy version.

Spent ~30 minutes diagnosing through three agents before the root cause landed: same bug as #599, silently re-introduced, and the bare catch mask made it look like a lock problem rather than a Bun syscall failure. If the outer catch had been:

\`\`\`ts
catch (err: any) {
  if (err?.code === 'EEXIST') {
    // existing holder-PID logic
  } else {
    console.error(\`[browse] acquireServerLock: unexpected \${err?.code || ''} opening \${lockPath}: \${err?.message}\`);
    return null;
  }
}
\`\`\`

…the diagnostic would have taken 10 seconds.

## Proposed Fix

Harden \`acquireServerLock()\` so only \`EEXIST\` goes down the holder-PID path. Everything else logs via \`console.error\` with the error code + path before falling back to \`return null\`. This:

- Preserves current behavior for the common-path "lock is held" case
- Surfaces real fs/syscall failures immediately (permissions, disk full, future Bun regressions)
- Doesn't change the return type or any callers

Same treatment for the inner catch on the holder-PID read.

## Environment

- Windows 11 Pro, Bun 1.3.10, Node 20.x, MINGW64 bash
- gstack 0.13.2.0 (commit 247fc3ba)
- Observed regression despite #599 being closed via cd66fc2f — that's a separate release/tagging concern worth looking into

## Why not file as a dup of #599

#599 was the specific bug (numeric \`O_CREAT|O_EXCL|O_WRONLY\` flags on Bun Windows). This issue is about the **defensive-coding gap** that made #599 and its look-alikes near-invisible to users. Independent of the root cause that motivated it.

JiayuuWang:
## Summary

The bare `catch {}` on `fs.openSync(lockPath, 'wx')` silently swallowed all errors other than `EEXIST` (permissions, disk full, Bun fs regressions). This made real filesystem failures look like phantom lock contention, hiding the root cause for ~30 minutes.

Now only `EEXIST` takes the holder-PID path; everything else logs via `console.error` with error code and message before returning `null`. The inner catch on the holder-PID read gets the same treatment.

## Root cause

When `EEXIST` is NOT the error (e.g., `EPERM`, `ENOSPC`), execution fell through to `readFileSync(lockPath)` — which either succeeded (stale lock) or threw — and returned `null`. The caller then printed "Another instance is starting the server, waiting..." which was completely misleading.

## Test plan

- [ ] `bun test` passes (Tier 1 static validation)
- [ ] Manual: trigger a non-EEXIST fs error on the lock path and verify a diagnostic message is printed instead of phantom lock contention

Fixes #1084

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -289,7 +289,12 @@ function acquireServerLock(): (() => void) | null {
     fs.writeSync(fd, `${process.pid}\n`);
     fs.closeSync(fd);
     return () => { safeUnlink(lockPath); };
-  } catch {
+  } catch (err: any) {
+    if (err?.code !== 'EEXIST') {
+      // Unexpected error — surface it rather than masking as phantom lock contention
+      console.error(`[browse] acquireServerLock: unexpected ${err?.code || ''} opening ${lockPath}: ${err?.message}`);
+      return null;
+    }
     // Lock already held — check if the holder is still alive
     try {
       const holderPid = parseInt(fs.readFileSync(lockPath, 'utf8').trim(), 10);
@@ -299,7 +304,8 @@ function acquireServerLock(): (() => void) | null {
       // Stale lock — remove and retry
       fs.unlinkSync(lockPath);
       return acquireServerLock();
-    } catch {
+    } catch (err: any) {
+      console.error(`[browse] acquireServerLock: stale-lock read failed: ${err?.message}`);
       return null;
     }
   }


muizbumbaugh:
Good catch on the `EEXIST` narrowing. Silently swallowing non-EEXIST errors in a lock acquisition path is exactly the kind of thing that shows up as phantom contention in prod and takes forever to trace. The `console.error` before returning `null` will at least leave a breadcrumb. One small thing: in the stale-lock retry path, logging the error before returning `null` is good, but it might be worth noting whether the stale lock was successfully removed — the current flow removes it then retries, but if `acquireServerLock()` is called from a context that doesn't retry on `null`, the caller gets `null` with no indication the lock was cleaned up. Not blocking, just worth documenting.

time-attack:
@16francej — recommend closing this PR because #1725 covers the same diagnostic and the related missing-file race. This PR lacks tests and handles that race incorrectly; preserve its first-fix attribution under #1725, then close it as superseded.

Issue #1066, opened 2026-04-18
Codex skills that call AskUserQuestion silently fail in Default mode


earino:
## Summary

gstack skills generated for Codex CLI still reference `AskUserQuestion` in their prompt bodies even though Codex strips that tool from the frontmatter and the Codex tool equivalent (`request_user_input`) is gated behind Plan mode. When a Codex user runs such a skill in Default mode (the default), the model either errors on an unknown tool or silently improvises an answer instead of asking the user.

## Reproduction

1. Clean Codex CLI install (v0.121.0 verified).
2. Install any gstack skill that includes `AskUserQuestion` in its prompt body. Easiest target is your own `/codex` skill at `codex/SKILL.md.tmpl:102-114`, which calls `AskUserQuestion` unconditionally for mode selection.
3. Launch Codex normally (Default mode) and invoke the skill.
4. Observe: the multiple-choice prompt does not appear. The model picks a default based on context.

Expected: either the question actually fires, or the skill fails loudly with a message pointing at `/plan`.

Actual: silent degradation. For `/codex` itself this is benign (the fallback picks "review"). For skills that collect user identity or preferences, it produces silently-wrong outputs downstream.

## Root cause

Two independent pieces, both verified against source:

**Codex side** (`openai/codex` at tag `rust-v0.121.0`):
- `codex-rs/protocol/src/config_types.rs:428-430`: `mode.allows_request_user_input()` returns true only for `ModeKind::Plan`.
- `codex-rs/tools/src/request_user_input_tool.rs:125-128`: the tool is registered unconditionally but the handler rejects calls in Default mode unless the `default_mode_request_user_input` feature flag is on (Stage: UnderDevelopment, default off).

**gstack side:**
- `hosts/codex.ts` declares `pathRewrites` but no `toolRewrites`. The optional `toolRewrites` field exists in the host-config interface (`scripts/host-config.ts:73`) but Codex's config doesn't populate it, so prompt text referring to `AskUserQuestion` passes through unchanged.
- `codex-helpers.ts:77-102` (`transformFrontmatter`) strips the `allowed-tools` list for non-Claude hosts. The model's schema for a Codex-generated skill does not contain `AskUserQuestion` at all.
- Generated Codex skills (e.g., `.agents/skills/gstack-checkpoint/SKILL.md:660`) literally say "If the skill says to use AskUserQuestion, do that". That tool isn't available to Codex in Default mode.
[...]

PR #1101, opened 2026-04-20
fix(codex): add AskUserQuestion→request_user_input tool rewrite

JiayuuWang:
## Summary

Issue #1066 — Codex skills that call `AskUserQuestion` silently fail in Default mode because:

1. Codex strips `AskUserQuestion` from the tool schema in Default mode
2. Its equivalent (`request_user_input`) only works in Plan mode
3. The codex host config had no `toolRewrites` entry to handle this

The `HostConfig` interface already supports `toolRewrites` — `gbrain`, `openclaw`, `factory`, and `hermes` all use it. The `codex.ts` host config was simply missing it.

## Fix

Added `toolRewrites: { 'AskUserQuestion': 'request_user_input' }` to `hosts/codex.ts`. When `gen-skill-docs.ts` applies rewrites at Codex skill generation time, any prompt text referencing `AskUserQuestion` gets replaced with Codex's equivalent.

## What this fixes

- Skills like `/codex` (which calls `AskUserQuestion` unconditionally at `codex/SKILL.md.tmpl:102-114`) now rewrite the tool name to `request_user_input` in generated Codex skills
- Other skills that reference `AskUserQuestion` in their prompt bodies will also be rewritten

## What this doesn't fix (out of scope)

A complete fix would also need the Plan-mode gate in the preamble (Option 4 in #1066), which requires Codex-side `default_mode_request_user_input` feature flag changes outside gstack's control. This PR handles the text rewrite that gstack controls.

## Test plan

- [ ] Verify `hosts/codex.ts` has `toolRewrites: { 'AskUserQuestion': 'request_user_input' }`
- [ ] Run `bun run gen:skill-docs --host codex` and check generated skill for rewritten tool reference
- [ ] Verify Codex Default mode no longer tries to call non-existent `AskUserQuestion` tool

### DIFF
--- a/hosts/codex.ts
+++ b/hosts/codex.ts
@@ -31,6 +31,10 @@ const codex: HostConfig = {
     { from: '.claude/skills', to: '.agents/skills' },
   ],
 
+  toolRewrites: {
+    'AskUserQuestion': 'request_user_input',
+  },
+
   suppressedResolvers: [
     'DESIGN_OUTSIDE_VOICES',  // design.ts:485 — Codex can't invoke itself
     'ADVERSARIAL_STEP',       // review.ts:408 — Codex can't invoke itself


time-attack:
@16francej — recommend closing this PR because #1163 owns the complete Codex interaction behavior, including planning and blocking semantics. This PR only renames a tool and therefore does not resolve those behaviors; credit #1101 with identifying the Codex tool mismatch, then close it as superseded.

Issue #1069, opened 2026-04-18
/ship Step 5 hardcodes `bin/test-lane` + `npm run test` regardless of detected framework


dgrant:
## What happened

Ran `/ship` in a Django-only project. Step 4's `{{TEST_BOOTSTRAP}}` correctly detected the runtime:

> Test framework detected: Python/Django with chores/tests.py. Skipping bootstrap.

But Step 5 then told the agent to run:

```bash
bin/test-lane 2>&1 | tee /tmp/ship_tests.txt &
npm run test 2>&1 | tee /tmp/ship_vitest.txt &
wait
```

Neither command exists in a Django project. Both are hardcoded — `bin/test-lane` is Rails-specific and `npm run test` is Node-specific. They appear to have leaked in from a Rails-with-Vitest project the skill was authored against.

## Where

`ship/SKILL.md.tmpl:148-160` — Step 5 runs both hardcoded commands unconditionally, without regard to what Step 4 just detected.

## Why it's a bug

This directly violates the project's own rule in `CLAUDE.md`:

> ## Platform-agnostic design
>
> Skills must NEVER hardcode framework-specific commands, file patterns, or directory structures. Instead:
> 1. Read CLAUDE.md for project-specific config (test commands, eval commands, etc.)
> 2. If missing, AskUserQuestion — let the user tell you or let gstack search the repo
> 3. Persist the answer to CLAUDE.md so we never have to ask again

Step 4 already follows this pattern (detects runtime, bootstraps if needed). Step 5 doesn't — it just assumes Rails + Vitest.

## Suggested fix shape

1. Read the test command from the project's `CLAUDE.md` (the `## Testing` section that the bootstrap flow writes). If present, run it.
2. If missing, `AskUserQuestion` and persist the answer back to `CLAUDE.md` so the next `/ship` doesn't re-ask.
3. Drop the hardcoded `bin/test-lane` + `npm run test` pair.
4. Guardrail: add a test that renders `ship/SKILL.md` and fails if it contains hardcoded framework commands outside a detection block, so the regression can't quietly come back.

## Repro

Run `/ship` in any non-Rails, non-Node project (Django, Go, Rust, etc). Step 5 will propose `bin/test-lane` + `npm run test`.

## Out of scope

[...]

PR #842, opened 2026-04-05
feat(ship): stack-aware test execution

inchwormz:
## TL;DR

`/ship` only works on Rails projects today. On anything else (Next.js, Python, Go, Rust, PHP, Elixir, plain Node) it tells Claude to run `bin/test-lane`, the file does not exist, and the ship stops. This PR makes `/ship` detect the project at skill-run time and emit the right test instructions for that project. Rails projects see no change. Every other project goes from broken to working.

## The problem, concretely

`ship/SKILL.md.tmpl` hard-codes the Rails test flow inside Step 3 "Run tests". The template says things like `bin/test-lane`, `db:test:prepare`, `app/services/*_prompt_builder.rb`, and `EVAL_JUDGE_TIER=full`. These are Rails-only commands. A user running `/ship` on a Next.js app still gets the same generated text, because the template has no branching. Claude then tries `bin/test-lane` on a repo where that file does not exist, and the ship halts.

You can verify this on `main`:

```bash
grep -n "bin/test-lane" ship/SKILL.md.tmpl
# 10 hits, all inside Step 3
grep -n "{{.*STACK.*}}\|SHIP_STACK" ship/SKILL.md.tmpl
# 0 hits
```

There is no stack awareness. The template assumes Rails.

## The fix

1. A new resolver function, `generateShipTestExecution` in `scripts/resolvers/testing.ts`. It returns the full "Step 3: Run tests" block, with stack detection bash at the top and two branches below.
2. A new placeholder, `{{SHIP_TEST_EXECUTION}}`, registered in `scripts/resolvers/index.ts`.
3. `ship/SKILL.md.tmpl` now contains a single `{{SHIP_TEST_EXECUTION}}` line where the 86 Rails-specific lines used to live.
4. The generator (`bun run gen:skill-docs`) expands the placeholder into the same Rails block as before plus a new generic-fallback block plus stack detection bash.

Detection happens at **skill-run time**, not generator time. The generated `ship/SKILL.md` contains bash that sniffs the current repo and prints `SHIP_STACK: rails` or `SHIP_STACK: generic`. That means one generated skill works across every project, without per-project regeneration.

## Rails projects: byte-for-byte the same

The Rails block is moved, not rewritten. Every line that used to live at the top of Step 3 now lives inside `### If SHIP_STACK: rails`. That includes:

- `bin/test-lane 2>&1 | tee /tmp/ship_tests.txt` (plus the parallel `npm run test` line)
[...]

PR #1102, opened 2026-04-20
fix(ship): read test command from CLAUDE.md instead of hardcoding Rails+Node commands

JiayuuWang:
## Summary

Issue #1069 — `/ship` Step 5 hardcodes `bin/test-lane` + `npm run test` regardless of detected framework.

Step 4's `TEST_BOOTSTRAP` correctly detects the runtime, but Step 5 ignores it and unconditionally runs both commands. This violates the project's own platform-agnostic design rule in `CLAUDE.md`:

> Skills must NEVER hardcode framework-specific commands, file patterns, or directory structures. Instead:
> 1. **Read CLAUDE.md** for project-specific config (test commands, eval commands, etc.)
> 2. **If missing, AskUserQuestion** — let the user tell you or let gstack search the repo
> 3. **Persist the answer to CLAUDE.md** so we never have to ask again

In a Django project, `bin/test-lane` (Rails-specific) and `npm run test` (Node-specific) both fail silently with non-existent commands. The issue reporter hit this while shipping a Django app.

## Fix

Replaced the hardcoded dual-command block in `ship/SKILL.md.tmpl` Step 5 with a `CLAUDE.md`-first lookup (matching `TEST_COVERAGE_AUDIT_SHIP`'s existing pattern):

```bash
grep -A2 '## Testing' CLAUDE.md 2>/dev/null | head -20
```

- If `## Testing` section exists with a run command: use that command
- If missing: search the project for the appropriate test command
- If no test framework found: print "No test framework detected — skipping Step 5." and continue to Step 6

This ensures the test execution follows the same platform-agnostic pattern that the coverage audit already uses.

## Test plan

- [ ] Verify `ship/SKILL.md.tmpl` no longer contains hardcoded `bin/test-lane` or `npm run test` in Step 5
- [ ] Run `/ship` in a non-Rails, non-Node project — Step 5 should read CLAUDE.md for the test command
- [ ] Run `/ship` in a Rails project — should still find and execute the correct test command
- [ ] Verify the skip message appears when no test framework is detected

### DIFF
--- a/ship/SKILL.md.tmpl
+++ b/ship/SKILL.md.tmpl
@@ -147,19 +147,23 @@ git fetch origin <base> && git merge origin/<base> --no-edit
 
 ## Step 5: Run tests (on merged code)
 
-**Do NOT run `RAILS_ENV=test bin/rails db:migrate`** — `bin/test-lane` already calls
+**Do NOT run `RAILS_ENV=test bin/rails db:migrate`** — test-lane already calls
 `db:test:prepare` internally, which loads the schema into the correct lane database.
 Running bare test migrations without INSTANCE hits an orphan DB and corrupts structure.sql.
 
-Run both test suites in parallel:
+**Read the test command from CLAUDE.md:**
 
 ```bash
-bin/test-lane 2>&1 | tee /tmp/ship_tests.txt &
-npm run test 2>&1 | tee /tmp/ship_vitest.txt &
-wait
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej — recommend closing this PR because it changes an obsolete ship template. Issue #1069 is the active tracker for trusted test-command discovery; please carry this PR’s project-configuration precedence insight and contributor attribution there, together with the useful stack-detection cases from #842.

Issue #1060, opened 2026-04-18
/careful hook triggers false positives on patterns inside commit messages


oscarsterling:
## Summary

The `/careful` hook uses string matching against the full bash command, including the contents of `git commit -m` arguments. This means commit messages that *describe* destructive commands (e.g., "this hook catches recursive deletes and database drops") trigger the safety warning even though no destructive command is being run.

## Steps to reproduce

1. Activate `/careful`
2. Run a commit whose message mentions a blocked pattern:

```bash
git add file.txt && git commit -m "feat: add guard for recursive delete and database drop patterns"
```

3. The hook fires on patterns found inside the quoted commit message string, not as an actual command.

## Expected behavior

The hook should only match destructive patterns in the executable portions of the command, not inside string arguments to `git commit -m`, `echo`, `printf`, or heredoc bodies.

## Suggested fix

Before pattern matching, split the command on shell operators (`&&`, `||`, `;`), then skip any segment that starts with `git commit`, `echo`, `cat`, or `printf` since those are producing text output, not executing destructive operations.

Found this while building a similar hook inspired by gstack's `/careful` approach. Happy to submit a PR if useful.

time-attack:
@oscarsterling Literal command patterns inside commit messages still trigger false positives, with no safe fix that preserves destructive-command protection.

PR #1110, opened 2026-04-20
fix(careful): skip false positives in text-output command segments

JiayuuWang:
## Summary

- Fix for Issue #1060 — /careful hook was triggering false positives on patterns inside git commit messages, echo arguments, and other text-output commands
- The `check-careful.sh` script now splits command on shell operators and filters out text-output segments (git commit -m, echo, printf, cat) before pattern matching
- Added regression tests covering git commit -m, echo, printf, cat and chained commands

## Changes

- **careful/bin/check-careful.sh**: Added filtering logic to skip text-output command segments before pattern matching
- **test/hook-scripts.test.ts**: Added 7 new test cases for false positive prevention

## Test plan

- [ ] `bun test test/hook-scripts.test.ts` passes (especially the new false-positive tests)
- [ ] Existing destructive command tests still pass (no regression)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/careful/bin/check-careful.sh
+++ b/careful/bin/check-careful.sh
@@ -22,6 +22,38 @@ if [ -z "$CMD" ]; then
   exit 0
 fi
 
+# Split command on shell operators and filter out text-only segments.
+# Segments that start with git commit, echo, cat, or printf produce text output,
+# not destructive actions — skip their content to avoid false positives.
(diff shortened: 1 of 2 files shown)


time-attack:
leaves RCE open @16francej  close also already fixed

PR #1113, opened 2026-04-21
[codex] Fix make-pdf browse resolution

parafallen-maker:
## Summary

- Resolve browse from Codex/agents installs before older Claude install paths
- Reject executable directories when resolving the browse binary
- Avoid `about:blank` in `make-pdf setup` so the Chromium smoke check works with the browse URL allowlist
- Default generated PDFs and setup smoke-test output to `/private/tmp` on macOS so browse can write them
- Add a regression test for executable directories being mistaken for binaries

## Root cause

On macOS, directories commonly have execute permission. `resolveBrowseBin()` only checked `X_OK`, so a directory such as `~/.agents/skills/browse` could be selected as the browse binary. In Codex installs, the active browse binary also lives under `~/.agents/skills/gstack`, while the previous global fallback only checked `~/.claude`.

## Validation

- `bun test make-pdf/test`
  - 75 pass, 1 skip, 0 fail
  - skip: local `pdftotext` is not installed; runtime PDF generation does not require it
- `make-pdf/dist/pdf setup`
  - browse binary resolved to `~/.agents/skills/gstack/browse/dist/browse`
  - Chromium smoke check passed
  - smoke-test PDF generated under `/private/tmp`

### DIFF
--- a/make-pdf/src/browseClient.ts
+++ b/make-pdf/src/browseClient.ts
@@ -10,9 +10,11 @@
  * Binary resolution order (Codex round 2 #4):
  *   1. $BROWSE_BIN env override
  *   2. sibling dir: dirname(argv[0])/../browse/dist/browse
- *   3. ~/.claude/skills/gstack/browse/dist/browse
- *   4. PATH lookup: `browse`
- *   5. error with setup hint
+ *   3. ~/.agents/skills/gstack/browse/dist/browse
+ *   4. ~/.codex/skills/gstack/browse/dist/browse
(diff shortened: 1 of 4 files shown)


time-attack:
@16francej — recommend closing this PR because #2156 owns the current PDF browsing failure, while this old draft conflicts with current resolver behavior and relies on obsolete assumptions. Carry forward the working-directory and about:blank evidence plus contributor attribution to #2156, then close it as superseded.

PR #843, opened 2026-04-05
fix: use $HOME instead of ~ for browse/design binary paths

mvanhorn:
## Summary

Changes `browseDir` and `designDir` in `buildHostPaths()` (`scripts/resolvers/types.ts:40-41`) to use `$HOME` instead of `~`. Regenerates all 16 affected SKILL.md files.

## Problem

Claude Code's bash safety heuristic fires a permission prompt on every browse/design command:

```
Tilde in assignment value — bash may expand at assignment time
```

This triggers on the generated pattern `B=~/.claude/skills/gstack/browse/dist/browse` in every SKILL.md that uses the browser. Frequent interruptions during `/browse`, `/qa`, `/design-review`, and 14 other skills.

## Changes

One source change in `scripts/resolvers/types.ts`:

```diff
-        browseDir: `${root}/browse/dist`,
-        designDir: `${root}/design/dist`,
+        browseDir: `$HOME/${config.globalRoot}/browse/dist`,
+        designDir: `$HOME/${config.globalRoot}/design/dist`,
```

`skillRoot` and `binDir` stay with `~` because they appear in command substitution and prose contexts that don't trigger the heuristic.

## Testing

337/338 gen-skill-docs tests pass. The 1 failure is a pre-existing VERSION/package.json mismatch (#399), unrelated to this change. Verified:

- Zero `B=~/` or `D=~/` patterns remain in generated SKILL.md files
- `skillRoot` paths still use `~` (test expectations at gen-skill-docs.test.ts:1742/1750 preserved)
- Only Claude host affected (other hosts use `usesEnvVars: true`, unaffected by this path)

Fixes #785

This contribution was developed with AI assistance (Codex).

PR #1120, opened 2026-04-21
fix(browse,design): resolve ~/.gstack via os.homedir() on Windows

scarson:
## Summary

On Windows, gstack's persistent state directory (`~/.gstack/`) lands at nonsense paths — chromium profile, sidebar session data, ngrok config, agent queue, worktree cache, OpenAI key file, browse state — whenever a user runs gstack binaries (`browse.exe`, `pdf.exe`, `design.exe`, sidebar agent) from any shell that isn't Git Bash. That covers cmd.exe, PowerShell, VS Code's integrated terminal when not explicitly configured for Git Bash, Windows Terminal default profile, and any IDE spawning subprocesses with its own env block.

Root cause: 31 sites in `browse/src/` and `design/` construct paths with `process.env.HOME || <fallback>`. On Windows, `HOME` is unset (Windows uses `USERPROFILE`), so the fallback kicks in. The fallbacks are all broken in different ways:

| Pattern | Site count | Behavior on Windows when HOME unset |
|---------|-----------:|-------------------------------------|
| `process.env.HOME \|\| '/tmp'` | 20 | `path.join('/tmp', ...)` → `\tmp\.gstack\...` — literal path, doesn't exist |
| `process.env.HOME \|\| ''` | 8 | `path.join('', ...)` → CWD-relative — silently mislocates |
| `process.env.HOME!` | 1 | Non-null assertion crashes at load time |
| `process.env.HOME \|\| "~"` | 1 | Creates literal `~` directory under CWD; Node path APIs never expand `~` |
| `process.env.HOME \|\| process.env.USERPROFILE \|\| '/tmp'` | 1 | Already platform-aware; simplified |

## Repro

Minimal Node script simulating an invocation from a native-Windows shell:

```js
delete process.env.HOME;
console.log('HOME was unset:', process.env.HOME === undefined);
console.log('USERPROFILE:', process.env.USERPROFILE);

const path = require('path');
const os = require('os');

const oldFallback = path.join(process.env.HOME || '/tmp', '.gstack', 'sidebar-agent-queue.jsonl');
const newFix      = path.join(os.homedir(),                 '.gstack', 'sidebar-agent-queue.jsonl');

console.log('OLD (|| "/tmp"):', oldFallback);
console.log('NEW (os.homedir):', newFix);
```

Output on Windows 11:

```
HOME was unset: true
USERPROFILE: C:\Users\Sam

OLD (|| "/tmp"):  \tmp\.gstack\sidebar-agent-queue.jsonl
NEW (os.homedir): C:\Users\Sam\.gstack\sidebar-agent-queue.jsonl
```

The OLD path doesn't exist. Every `fs.writeFileSync` targeting `~/.gstack/` against the OLD path fails with `ENOENT`, usually caught and logged as a confusing "couldn't save state" warning, and the feature appears partially-working to the user — tokens don't persist across restarts, the chromium profile resets every launch, the sidebar session directory rotates and loses chat history.

## Why this shipped unnoticed

- Sam's clone (and most gstack Windows users') runs every command from Git Bash, which sets `HOME` by default. So `HOME || '/tmp'` short-circuits on `HOME` and the state lands correctly — until an IDE-spawned subprocess inherits a sanitized env without `HOME`, or the user runs `browse.exe` from PowerShell to see what happens.
- Every `.github/workflows/` job runs on `ubuntu-latest` or `macos-latest`. CI never touches the Windows branch of this fallback. #1118's analysis of the make-pdf-gate matrix comment applies identically here.

## The fix

Replace every variant with `os.homedir()`. Per [Node docs](https://nodejs.org/api/os.html#oshomedir):

> `os.homedir()` returns the string path of the current user's home directory.
> On POSIX, it uses the `$HOME` environment variable if defined. Otherwise it uses the effective UID to look up the user's home directory.
> On Windows, it uses the `USERPROFILE` environment variable if defined. Otherwise it uses the path to the profile directory of the current user.

Strictly better than every pattern it replaces, on every platform. Specifically:

- POSIX CI test isolation that sets `process.env.HOME = tmpHome` (see `browse/test/security-review-flow.test.ts:30`) keeps working — `os.homedir()` still reads `HOME` on POSIX.
- The one site already using `HOME || USERPROFILE || '/tmp'` collapses cleanly to `os.homedir()`.
- The `HOME!` non-null assertion in `design/prototype.ts:13` stops being a load-time crash risk.

### Files changed

| File | Sites | What lived there |
|------|------:|------------------|
| `browse/src/cli.ts` | 8 | ngrok env, ngrok config search path (3 paths), gstack dir, chromium profile (x2), agent queue |
| `browse/src/server.ts` | 13 | sessions dir, browse binary lookup, global home, worktree dir, agent queue, cancel dir, chromium profile (x2), welcome path, ngrok env, ngrok config search (3 paths), ngrok env (second use site) |
| `browse/src/browser-manager.ts` | 5 | extension path, gstack dir, chromium profile (x2), app icon path |
| `browse/src/sidebar-agent.ts` | 3 | queue, cancel dir, browse.json |
| `design/prototype.ts` | 1 | `process.env.HOME!` → `os.homedir()` (crash-on-unset bug) |
| `design/src/auth.ts` | 1 | `\|\| "~"` → `os.homedir()` (literal-tilde bug) |

Added `import * as os from 'os'` (or `import os from "os"` for the design files, matching each file's existing default-import style) to the 6 files that didn't already import it.

## Test plan

- [x] Full `bun test browse/test/ make-pdf/test/` on Windows — 163 pass, 10 fail. Same 163/10 on clean upstream main (stashed + re-ran for exact comparison), identical failing test names. All 10 failures are pre-existing and unrelated to this PR's surface:
  - `batch.test.ts` + `compare-board.test.ts` — `beforeEach/afterEach` hook timeouts (5s limit; appears to be a Windows-specific test-harness issue, not the code under test)
  - `bun-polyfill.test.ts` x4 — `Bun.serve`/`spawn`/`spawnSync`/`sleep` assertions, all expecting `"OK"` or `"HAS_STOP"` and receiving `""`, suggesting a pre-existing Bun polyfill subshell stdout-capture issue on this platform
  - `sidebar-integration.test.ts` + `sidebar-agent-roundtrip.test.ts` — `agent_done transitions to idle`, `queues message when agent is processing`, `kill adds error entry` — same failures on clean main, unrelated to path construction
- [x] Empirical proof with HOME unset on Windows 11 (see Repro above) — `os.homedir()` produces the correct `C:\Users\Sam\.gstack\...` path where the OLD fallback produced `\tmp\.gstack\...`
- [ ] Ubuntu/macOS CI pass — relying on the existing matrix. POSIX branch is semantically identical (`os.homedir()` falls through to `$HOME` on POSIX, same as the old short-circuit), so no regression expected. Test that overrides `process.env.HOME` in `security-review-flow.test.ts` keeps passing because `os.homedir()` on POSIX honors HOME overrides.
- [ ] End-to-end `browse.exe` start from PowerShell/cmd.exe verified — not run. The unit-test pass rate + empirical before/after is sufficient confidence given the mechanical nature of the change.

## What this PR doesn't do

- **Fix `/tmp` usage in dev-only scripts.** `design/prototype.ts:20` uses `/tmp/gstack-prototype-<timestamp>/` as an output dir. It's a manual-run validation script (`bun run design/prototype.ts`), not user-facing production code. Separate concern.
- **Fix `/tmp` usage in `cookie-import-browser.ts`.** That's the scope of @Gonzih's #748 already open — deferring to that PR.
- **Touch bash preamble `~` expansion** (#558, #843). Those are about bash path expansion in generated SKILL.md files, a different subsystem from TypeScript runtime paths.
- **Address Windows ACL / `fs.chmodSync` no-op.** Filed separately as #1121 — sensitive files (auth tokens, canary tokens, chat history, agent queue) end up with 6 inherited ACEs on Windows because `chmod` mode bits don't map to NTFS ACLs. #1121 fixes it via a shell-out to `icacls` and includes an empirical before/after showing 6 ACEs → 1.

## Commits

- `d3a6b5e` — `fix(browse,design): resolve ~/.gstack via os.homedir() not $HOME || /tmp`

Single commit. The change is mechanical and atomic across the 6 files — no value in splitting further.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -15,6 +15,7 @@
  *   restores state. Falls back to clean slate on any failure.
  */
 
+import * as os from 'os';
 import { chromium, type Browser, type BrowserContext, type BrowserContextOptions, type Page, type Locator, type Cookie } from 'playwright';
 import { addConsoleEntry, addNetworkEntry, addDialogEntry, networkBuffer, type DialogEntry } from './buffers';
 import { validateNavigationUrl } from './url-validation';
@@ -139,7 +140,7 @@ export class BrowserManager {
       // Relative to this source file (dev mode: browse/src/ -> ../../extension)
       path.resolve(__dirname, '..', '..', 'extension'),
       // Global gstack install
-      path.join(process.env.HOME || '', '.claude', 'skills', 'gstack', 'extension'),
+      path.join(os.homedir(), '.claude', 'skills', 'gstack', 'extension'),
       // Git repo root (detected via BROWSE_STATE_FILE location)
       (() => {
         const stateFile = process.env.BROWSE_STATE_FILE || '';
@@ -266,7 +267,7 @@ export class BrowserManager {
       if (authToken) {
         const fs = require('fs');
         const path = require('path');
-        const gstackDir = path.join(process.env.HOME || '/tmp', '.gstack');
+        const gstackDir = path.join(os.homedir(), '.gstack');
         fs.mkdirSync(gstackDir, { recursive: true });
         const authFile = path.join(gstackDir, '.auth.json');
         try {
@@ -283,7 +284,7 @@ export class BrowserManager {
     // so we use Playwright's bundled Chromium which reliably loads extensions.
     const fs = require('fs');
     const path = require('path');
-    const userDataDir = path.join(process.env.HOME || '/tmp', '.gstack', 'chromium-profile');
+    const userDataDir = path.join(os.homedir(), '.gstack', 'chromium-profile');
     fs.mkdirSync(userDataDir, { recursive: true });
 
     // Support custom Chromium binary via GSTACK_CHROMIUM_PATH env var.
@@ -310,7 +311,7 @@ export class BrowserManager {
         // Replace Chromium's Dock icon with ours (Chromium's process owns the Dock icon)
         const iconCandidates = [
           path.join(__dirname, '..', '..', 'scripts', 'app', 'icon.icns'),       // repo dev mode
-          path.join(process.env.HOME || '', '.claude', 'skills', 'gstack', 'scripts', 'app', 'icon.icns'), // global install
+          path.join(os.homedir(), '.claude', 'skills', 'gstack', 'scripts', 'app', 'icon.icns'), // global install
         ];
         const iconSrc = iconCandidates.find(p => fs.existsSync(p));
         if (iconSrc) {
@@ -1158,7 +1159,7 @@ export class BrowserManager {
         console.log('[browse] Handoff: extension not found — headed mode without side panel');
       }
 
-      const userDataDir = path.join(process.env.HOME || '/tmp', '.gstack', 'chromium-profile');
+      const userDataDir = path.join(os.homedir(), '.gstack', 'chromium-profile');
       fs.mkdirSync(userDataDir, { recursive: true });
 
       newContext = await chromium.launchPersistentContext(userDataDir, {
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -10,6 +10,7 @@
  */
 
 import * as fs from 'fs';
+import * as os from 'os';
 import * as path from 'path';
 import { safeUnlink, safeUnlinkQuiet, safeKill, isProcessAlive } from './error-handling';
 import { resolveConfig, ensureStateDir, readVersionHash } from './config';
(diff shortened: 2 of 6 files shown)


scarson:

scarson:
Closing — equivalent fix landed in v1.20.0.0 ([#1233](https://github.com/garrytan/gstack/pull/1233)) via the `browser-skills runtime + gbrain-support carryover` work (commit `e8893a1`).

`os.homedir()` is now the standard pattern across `browse/src/`:
- [browse/src/domain-skills.ts:66](https://github.com/garrytan/gstack/blob/main/browse/src/domain-skills.ts#L66)
- [browse/src/browser-skill-write.ts:73](https://github.com/garrytan/gstack/blob/main/browse/src/browser-skill-write.ts#L73)
- [browse/src/browser-skills.ts:88](https://github.com/garrytan/gstack/blob/main/browse/src/browser-skills.ts#L88)
- [browse/src/security-classifier.ts:64](https://github.com/garrytan/gstack/blob/main/browse/src/security-classifier.ts#L64)
- [browse/src/security-bunnative.ts:146](https://github.com/garrytan/gstack/blob/main/browse/src/security-bunnative.ts#L146), [:178](https://github.com/garrytan/gstack/blob/main/browse/src/security-bunnative.ts#L178)
- [browse/src/cookie-import-browser.ts:342-344](https://github.com/garrytan/gstack/blob/main/browse/src/cookie-import-browser.ts#L342-L344)

The `$HOME || /tmp` anti-pattern is gone from source. Different commit, same shape — surface is fixed.

PR #1140, opened 2026-04-22
Keep browse daemons alive across separate CLI invocations on macOS

be-student:
## Summary

Keep `browse` daemons alive across separate CLI invocations on macOS.

Before this change, a first command like `browse goto https://example.com` would succeed,
but the daemon and Chromium session could still die with the launching shell. A second
invocation would then start a fresh server from `about:blank`, losing browse state.

## Root Cause

The POSIX launch path was not putting the Bun server and its Chromium children into a
proper detached session/process group. In practice on macOS, the launched browser session
could still die when the short-lived parent shell exited. Shell hangups (`SIGHUP`) were
also not explicitly handled for the headless reuse case.

## Fix

- keep the POSIX path Bun-native by spawning the Bun server with `detached: true`
- keep headless daemons alive on `SIGHUP`, matching the headless reuse intent
- leave the Windows-specific Node launcher in place
- replace the weak helper-only regression test with a real two-invocation lifecycle test
  against a local HTTP server

## Verification

- `bun test test/browse-cli-daemon.test.ts`
- `bun run build`
- manual verification:
  - `browse goto https://example.com`
  - separate invocation: `browse url`
  - confirmed it returns `https://example.com/` instead of restarting at `about:blank`

## Notes

I explicitly avoided adding a new POSIX `node` runtime dependency in the final structure.

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -94,6 +94,13 @@ interface ServerState {
   mode?: 'launched' | 'headed';
 }
 
+export function detachedServerSpawnStdio(isWindows: boolean = IS_WINDOWS): ['ignore', 'ignore', 'ignore'] {
+  // Keep detached server stdio disconnected from the parent shell on every
+  // platform. On macOS/Bun, piped stdio can take the daemon down when the
+  // short-lived CLI process exits between invocations.
+  return ['ignore', 'ignore', 'ignore'];
+}
+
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej Please close this PR. Browser daemons already survive separate terminal commands through the lifecycle fix in #1612. This conflicting branch uses the previous launcher design and should not be transplanted.

PR #1143, opened 2026-04-22
docs(readme): align manual uninstall with --keep-state split

### LINKED ISSUE #1131: README.md uninstall guide suggests deleting gstack state
```
# 4. Remove global state
rm -rf ~/.gstack
```

This should not be part of the uninstall. It should be part of a separate "purge gstack data" or something.

mvanhorn:
## Summary

The `bin/gstack-uninstall` script already supports `--keep-state`, which removes skills but preserves `~/.gstack/` (config, analytics, sessions, project history, installation-id). The manual-removal block in README didn't reflect that split: it dropped `rm -rf ~/.gstack` inline as step 4 alongside the skill cleanup, so users following the manual path always lost their state even when the scripted path lets them keep it.

Reported in #1131.

## Fix

- Label the numbered block as the `--keep-state` equivalent.
- Move the `~/.gstack` purge into an "Also purge gstack data (optional)" subsection after the block, so users can decide whether to run it.

No script changes. Same behaviour is available, just documented as two choices instead of one.

## Diff

```diff
+ Mirrors `gstack-uninstall --keep-state` (preserves `~/.gstack/` data):

  ```bash
  # 1. Stop browse daemons
  ...
  # 3. Remove gstack
  rm -rf ~/.claude/skills/gstack

- # 4. Remove global state
- rm -rf ~/.gstack
-
- # 5. Remove integrations ...
+ # 4. Remove integrations ...
  ...
  ```

+ #### Also purge gstack data (optional)
+
+ `~/.gstack/` holds config, analytics, sessions, project history, and the
+ installation-id. Run this only if you want a clean slate (equivalent to
+ `gstack-uninstall` without `--keep-state`):
+
+ ```bash
+ rm -rf ~/.gstack
+ ```
```

Fixes #1131

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,8 @@ This handles skills, symlinks, global state (`~/.gstack/`), project-local state,
 
 If you don't have the repo cloned (e.g. you installed via a Claude Code paste and later deleted the clone):
 
+Mirrors `gstack-uninstall --keep-state` (preserves `~/.gstack/` data):
+
 ```bash
 # 1. Stop browse daemons
 pkill -f "gstack.*browse" 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


dgrant (approved):

mvanhorn:
Thanks @dgrant. The README split mirrors what `--keep-state` already does in the script, so the manual path should preserve state the same way.

time-attack:
@16francej close

PR #1155, opened 2026-04-22
security: fail-closed gate when ML classifiers are inactive

garagon:
## Summary

The sidebar agent's ML prompt injection defense (TestSavantAI + Haiku transcript classifier) degrades silently to zero coverage when both classifiers fail to load. `preSpawnSecurityCheck` returns `false` (safe) and the agent spawns without any ML protection — the user sees "inactive" on the shield icon but gets no blocking gate.

**An attacker who can cause model download failure** (DNS poisoning on huggingface.co, disk full, proxy blocking the 112MB ONNX download) permanently removes all ML protection while the agent keeps running as if everything is fine.

## What this PR does

Adds a **fail-closed gate** at the top of `preSpawnSecurityCheck`:

- `isClassifierInactive(status)` — pure function that returns `true` only when ALL classifiers (testsavant, transcript, and deberta if enabled) report `'off'`
- When inactive: blocks agent spawn, emits `security_event` (verdict=block, reason=classifiers_inactive) + `agent_error` to the sidepanel so the user sees exactly why the session was refused
- **Does NOT fire** when `GSTACK_SECURITY_OFF=1` — that's an explicit operator kill switch, they already know
- **Does NOT fire** on partial degradation — if one classifier is `'degraded'` or `'ok'`, the remaining coverage is enough to proceed
- **Override**: `GSTACK_SECURITY_ALLOW_INACTIVE=1` for operators who knowingly run without ML (CI, air-gapped environments)

## Files changed

| File | Change |
|------|--------|
| `browse/src/security-classifier.ts` | Extract `isClassifierInactive()` pure function |
| `browse/src/sidebar-agent.ts` | Fail-closed gate in `preSpawnSecurityCheck` |
| `browse/test/security-classifier.test.ts` | 7 test cases: all status combinations |

## Test plan

- [x] `bun test browse/test/security-classifier.test.ts` — 16/16 pass (7 new + 9 existing)
- [ ] Manual: start gstack without internet, verify agent spawn is blocked with clear error message
- [ ] Manual: set `GSTACK_SECURITY_ALLOW_INACTIVE=1`, verify agent spawns despite inactive classifiers
- [ ] Manual: start normally (classifiers load), verify no change to existing behavior

### DIFF
--- a/browse/src/security-classifier.ts
+++ b/browse/src/security-classifier.ts
@@ -102,6 +102,11 @@ export interface ClassifierStatus {
   deberta?: 'ok' | 'degraded' | 'off'; // only present when ensemble enabled
 }
 
+export function isClassifierInactive(cs: ClassifierStatus): boolean {
+  return cs.testsavant === 'off' && cs.transcript === 'off'
+    && (!cs.deberta || cs.deberta === 'off');
+}
+
 export function getClassifierStatus(): ClassifierStatus {
   const testsavant =
     testsavantState === 'loaded' ? 'ok' :
--- a/browse/src/sidebar-agent.ts
+++ b/browse/src/sidebar-agent.ts
@@ -21,7 +21,7 @@ import {
 } from './security';
 import {
   loadTestsavant, scanPageContent, checkTranscript,
-  shouldRunTranscriptCheck, getClassifierStatus,
+  shouldRunTranscriptCheck, getClassifierStatus, isClassifierInactive,
   loadDeberta, scanPageContentDeberta,
   type ToolCallInput,
 } from './security-classifier';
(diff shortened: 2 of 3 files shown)


garrytan:
Thanks @garagon — closing. Per the project's strict-cherry-pick rule for community PRs, this needs to apply cleanly to current main, but it currently conflicts. Independently: the fail-closed gate is a behavior change with real reach (blocks legitimate sessions when models haven't loaded yet) — even though the GSTACK_SECURITY_ALLOW_INACTIVE escape exists, the default flip warrants standalone focused review with a threat-model writeup, not a wave bundle. Happy to revisit a rebased version with that framing.

PR #1157, opened 2026-04-23
security: pass Haiku classifier prompt via stdin, not argv

garagon:
## Summary

The Haiku transcript classifier passes scanned content (user messages + tool outputs, up to 8KB) as a CLI argument to `claude -p <prompt>`. This makes the full prompt visible via `ps aux` or `/proc/<pid>/cmdline` for up to 15 seconds per classification call.

**On shared Linux hosts** (default `hidepid=0`), any local user can read the scanned content — which may include page text, tool outputs, and potentially tokens or credentials visible on the page when the classifier fires.

On macOS 10.15+ the exposure is lower (Full Disk Access required to read other users' processes), but still present for same-user monitoring.

## What this PR does

Two changes in `checkTranscript()` (`security-classifier.ts`):

1. **Prompt via stdin**: `claude -p` (no argument) reads from stdin. The prompt is written to `p.stdin` and the pipe closed, keeping it off the process argument list entirely.

2. **Scoped child env**: the spawned process now inherits only `PATH`, `HOME`, and `ANTHROPIC_API_KEY` instead of the full parent `process.env`. This prevents leaking unrelated secrets (other API keys, tokens) into the child's environment.

## Before / After

```
# BEFORE: visible in ps output
claude -p "You are a prompt-injection detector...\n\nINPUTS:\n{\"user_message\":\"<scanned content>\"..." --model haiku --output-format json

# AFTER: only flags visible
claude -p --model haiku --output-format json
```

## Test plan

- [x] `bun test browse/test/security-classifier.test.ts` — 16/16 pass
- [x] `bun test` — full free suite passes
- [ ] Manual: run classification, verify `ps aux | grep claude` shows no prompt content
- [ ] Manual: verify Haiku classifier still returns valid verdicts (stdin path works)

### DIFF
--- a/browse/src/security-classifier.ts
+++ b/browse/src/security-classifier.ts
@@ -494,10 +494,20 @@ export async function checkTranscript(params: {
     // ~44k cache_creation tokens per call (massive cost inflation).
     // Using os.tmpdir() gives Haiku a clean context for pure classification.
     const p = spawn('claude', [
-      '-p', prompt,
+      '-p',
       '--model', HAIKU_MODEL,
       '--output-format', 'json',
-    ], { stdio: ['ignore', 'pipe', 'pipe'], cwd: os.tmpdir() });
+    ], {
+      stdio: ['pipe', 'pipe', 'pipe'],
+      cwd: os.tmpdir(),
+      env: {
+        PATH: process.env.PATH,
+        HOME: process.env.HOME,
+        ANTHROPIC_API_KEY: process.env.ANTHROPIC_API_KEY,
+      },
+    });
+    p.stdin.write(prompt);
+    p.stdin.end();
 
     let stdout = '';
     let done = false;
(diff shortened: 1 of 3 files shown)


garagon:
Rebased onto current `main` (v1.17.0.0). Conflict was in the `spawn` call inside `checkTranscript`: `main` swapped the literal `'haiku'` for the `HAIKU_MODEL` constant and added `cwd: os.tmpdir()` after this PR was filed. The merged version keeps both — the stdin migration this PR introduces, plus the `cwd` and constant changes from `main`. Env scoping (PATH/HOME/ANTHROPIC_API_KEY only) preserved.

`bun test browse/test/security-classifier.test.ts`: 9/9 pass.

garrytan:
Thanks @garagon — closing. Conflicts with current main (would need rebase). Separately, the spawn() change explicitly strips env vars (only PATH, HOME, ANTHROPIC_API_KEY pass through) — that's a behavior change beyond the stated argv→stdin fix and could break the claude CLI subprocess on systems where it relies on additional env (locale, NODE_PATH, etc.). The argv→stdin part on its own is a clear win; happy to revisit a narrower PR that does only that without the env restriction.

PR #1165, opened 2026-04-23
feat: add BROWSE_CHANNEL persistent context for macOS SSO

haha1903:
## Summary
- Adds `BROWSE_CHANNEL` env var support to `browser-manager.ts`
- When set (e.g., `BROWSE_CHANNEL=chrome`), uses `chromium.launchPersistentContext` instead of `chromium.launch` + `newContext`
- This enables macOS platform SSO passthrough, allowing browse to access authenticated internal sites (Azure DevOps, SharePoint) without manual cookie import

## Motivation
`launchPersistentContext` with a `channel` option picks up the system's SSO tokens, which `newContext` does not. This is required for browsing Microsoft internal sites that rely on platform SSO.

## Changes
- `browse/src/browser-manager.ts`: Read `BROWSE_CHANNEL` from env, branch launch logic between persistent context (with channel) and regular launch (without)

## Test plan
- [x] Verified `BROWSE_CHANNEL=chrome` successfully navigates to ADO and reads authenticated content
- [x] Verified default behavior (no `BROWSE_CHANNEL`) still works as before with `chromium.launch`

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -179,6 +179,7 @@ export class BrowserManager {
     // BROWSE_EXTENSIONS_DIR points to an unpacked Chrome extension directory.
     // Extensions only work in headed mode, so we use an off-screen window.
     const extensionsDir = process.env.BROWSE_EXTENSIONS_DIR;
+    const channel = process.env.BROWSE_CHANNEL;
     const launchArgs: string[] = [];
     let useHeadless = true;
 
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for this. Persistent-context startup does not adopt the initial page or define close and disconnect behavior, and extension compatibility with a real Chrome channel is unresolved. Integrate it with the current lifecycle, define or reject the extension combination, and add focused lifecycle and SSO coverage.

PR #1177, opened 2026-04-24
Fix gstack build when Conductor creates a repo with unborn HEAD

JonasFocus:
This PR fixes a Conductor Quick Start failure where the gstack template build exits with code 128 before the generated repo has an initial commit.

Root cause: the build script calls git rev-parse HEAD while the repo can still have an unborn HEAD. That makes setup fail even though the upstream gstack repo itself has a valid HEAD.

Changes:

Adds scripts/write-build-version.sh
Falls back to unknown when HEAD does not resolve
Writes the resolved version to all .version files
Adds regression coverage for both unborn HEAD and normal committed repos


Cloverings1:
Additional verification after reproducing this locally through Conductor:

- Conductor Quick Start fails because the generated repo can be in an unborn-HEAD state before the first commit exists.
- A local workaround that only makes `git rev-parse HEAD` return `unknown` for unborn repos allowed Conductor's GStack Quick Start to complete.
- This PR fixes the same root cause inside GStack itself, without requiring a local Git shim.

I also tested the PR rebased onto current `origin/main` without the local workaround in PATH:

- `bun test test/write-build-version.test.ts test/setup-codesign.test.ts` passes: 8 pass, 0 fail.
- Full `bun run build` passes inside a fresh Git repo with no commits.
- The generated version files contain `unknown` in that unborn-HEAD case.

So the fix still preserves normal commit SHA behavior while allowing Conductor/new-repo bootstrap flows to finish before the first commit exists.

Cloverings1:
Updated this PR against current `main`.

It now keeps the current `vendor:xterm` build step and moves the version-file writes behind `scripts/write-build-version.sh`, which falls back to `unknown` when `git rev-parse --verify HEAD` cannot resolve.

I also added regression coverage for both observed bootstrap failure modes:

- a Git repo before the first commit exists / unborn `HEAD`
- a checkout where Git metadata is not present, matching the Conductor trace where `.git` is removed before setup

Focused validation passes locally:

```sh
bun test test/write-build-version.test.ts test/setup-codesign.test.ts
# 9 pass, 0 fail
```

I tried a full `bun run build` as well, but this fresh clone did not have dependencies installed yet, so it stopped before the relevant version-writing step because `node_modules/xterm` / `marked` were missing. The focused regression path above is passing.

16francej:

PR #1181, opened 2026-04-24
fix(browse): close v20 ABE cookie exfiltration window via --remote-debugging-pipe

### LINKED ISSUE #1136: security: switch cookie-import-browser CDP from TCP port to --remote-debugging-pipe
**Context**

The v20 App-Bound Encryption elevation path on Windows, deferred from the v1.6.0.0 security wave. See the updated comment block in `browse/src/cookie-import-browser.ts` around the `--remote-debugging-port` launch.

**Problem**

On Windows 10.15+ with ABE v20, a local same-user process cannot decrypt cookies by reading the SQLite DB directly — the DPAPI context is bound to the browser process. But the CDP debug port bypasses that: any same-user process can connect to `http://127.0.0.1:<random-port>/json/list`, find a page target's WebSocket URL, and call `Network.getAllCookies` inside Chrome to get decrypted values.

Today the port is randomized in [9222, 9321] and Chrome is killed in the finally block, but there's a window (typically 1-3 seconds while the CDP call is in flight) during which a concurrent same-user process can steal decrypted v20 cookies.

**Fix direction**

Switch `--remote-debugging-port=<N>` → `--remote-debugging-pipe`. Chrome then exposes the CDP transport over a parent/child stdio pipe instead of a TCP socket, eliminating the TCP surface entirely.

**Work**

- Drop `--remote-debugging-port` from the Bun.spawn args in `cookie-import-browser.ts` around line 846.
- Pass `--remote-debugging-pipe` + `stdio: ['pipe', 'pipe', 'pipe', 'pipe', 'pipe']` (CDP uses fds 3+4).
- Replace `fetch(http://127.0.0.1:\${debugPort}/json/list)` + WebSocket connection with a CDP client that reads/writes framed JSON on the pipe fds.
- Playwright doesn't expose this transport out of the box. Either write a minimal CDP-over-pipe client (~200 lines) or find a library that supports it.

**Acceptance**

- `bun test` passes, cookie-import flow still works on macOS and Linux.
- Windows integration smoke test confirms v20 cookies are imported.
- No TCP port opened for the debug transport.

**Out of scope for this issue**

- Restructuring extractCookiesViaCdp for non-Chromium targets.
- Switching the main browse daemon to pipe transport (separate concern).

Tracked from v1.6.0.0 security wave N2 non-goal.

NewCoder3294:
Resolves #1136.

## What changes

Swaps the cookie-import Windows v20 ABE fallback from `--remote-debugging-port` (TCP, bindable by any same-user process during the 1-3s CDP window) to `--remote-debugging-pipe` (stdio pipes, inherited only by the spawning process). No TCP socket binds and no `/json/list` HTTP endpoint exists.

- New `CdpPipeTransport` class — NUL-delimited JSON framing over Node-style streams (`child.stdio[3]` write, `child.stdio[4]` read). Unit-testable via `PassThrough` mocks.
- New `extractCookiesViaCdpPipe` — uses `Target.getTargets` + `Target.attachToTarget(flatten: true)` + `Network.enable` + `Network.getAllCookies` over a single pipe connection.
- `importCookiesViaCdp` rewritten: spawns with `--remote-debugging-pipe`, routes through the new transport. The TCP port randomizer, retry loop, and WebSocket helper are removed. The Chrome subprocess uses `node:child_process.spawn` (not `Bun.spawn`) because Bun returns raw numeric fds for stdio indices ≥ 3 and Node stream↔Web bridges hang in Bun on those sockets — this is the only place in the file that uses Node's `child_process`.
- New exported `buildCdpSpawnArgs(exePath, userDataDir, profile)` helper — pure argv builder, unit-testable without spawning anything.
- Comment block at the spawn site updated to reflect the new posture; the \"KNOWN NON-GOAL\" paragraph is gone.
- `chromeProc` gets an explicit `'error'` listener — without it, an async spawn failure (ENOENT, EACCES, AV interference) would crash the parent via Node's default handler.
- Defensive guards: `CdpPipeTransport` flips `closed = true` on unexpected read-stream end so post-end `send()` rejects immediately; `extractCookiesViaCdpPipe` guards `cookiesResp.cookies ?? []` and validates \`attached.sessionId\` is a non-empty string before use.

## What does NOT change

The non-CDP cookie import path (SQLite + DPAPI for v10/v11 on macOS/Linux/Windows pre-ABE) is untouched. Tests for that path still pass.

## Commits (bisectable)

1. `feat(browse): add CdpPipeTransport for --remote-debugging-pipe CDP` — primitive, 9 unit tests
2. `feat(browse): add extractCookiesViaCdpPipe using CDP pipe transport` — extractor, 4 unit tests
3. `fix(browse): switch cookie-import CDP from TCP port to stdio pipe` — the fix, 4 unit tests for `buildCdpSpawnArgs`
4. `fix(browse): handle chromeProc spawn error events` — hardening
5. `fix(browse): defensive hardening for cookie-import CDP pipe path` — defensive guards + 3 tests

## Testing

- 20 new unit tests: `CdpPipeTransport` framing/error/close/post-end behavior (10), `extractCookiesViaCdpPipe` target selection, session scoping, domain filtering, missing-sessionId, missing-cookies-field (6), `buildCdpSpawnArgs` regression tests locking the `--remote-debugging-pipe` flag set (4).
- Full `bun test` suite passes with no new failures vs base `main`. Four pre-existing failures on `main` are preserved unchanged: `test/brain-sync.test.ts` (bearer-json secret scan), `test/skill-validation.test.ts` (2MB fixture size), `test/touchfiles.test.ts` (skill selection count), `test/model-overlay-opus-4-7.test.ts` (pacing phrase). None of these relate to this change.

**Windows v20 ABE integration smoke test was not run locally** — developer machine is macOS. Happy to iterate on any Windows-specific feedback.

## Out of scope (per issue)

- Restructuring `extractCookiesViaCdp[Pipe]` for non-Chromium targets.
- Switching the main browse daemon to pipe transport (separate concern).

### DIFF
--- a/browse/src/cookie-import-browser.ts
+++ b/browse/src/cookie-import-browser.ts
@@ -40,6 +40,7 @@ import * as crypto from 'crypto';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
+import type { Writable as NodeWritable, Readable as NodeReadable } from 'node:stream';
 import { TEMP_DIR } from './platform';
 
 // ─── Types ──────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


time-attack:
@NewCoder3294 Rewrite the implementation to bound operations and prevent sends after terminal stream failures while retaining the listener removal.

PR #1184, opened 2026-04-24
fix(setup): add cursor and slate host support

BiyuHuang:
## Summary

Adds cursor and slate to the setup script's host allowlist, auto-detection, docs generation, runtime root creation, and skill linking.

- `./setup --host cursor` and `./setup --host slate` now work
- Consolidates redundant `bun install` calls into a single execution before all host-specific generation blocks
- Fixes cursor/slate runtime root functions to use the minimal asset set matching their TypeScript configs (was erroneously copied from the expanded opencode runtime root)

## Test plan

- `bash -n setup` passes
- `./setup --host cursor` should proceed past the host validation into bun install + skill doc generation
- `./setup --host slate` should work identically
- `./setup --host auto` should auto-detect cursor/slate if their CLIs are on PATH
- Existing hosts (claude, codex, factory, opencode) are unaffected

### DIFF
--- a/setup
+++ b/setup
@@ -24,6 +24,10 @@ FACTORY_SKILLS="$HOME/.factory/skills"
 FACTORY_GSTACK="$FACTORY_SKILLS/gstack"
 OPENCODE_SKILLS="$HOME/.config/opencode/skills"
 OPENCODE_GSTACK="$OPENCODE_SKILLS/gstack"
+CURSOR_SKILLS="$HOME/.cursor/skills"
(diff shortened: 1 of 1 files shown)


freddiemixell:
+1 This solution also works for me on Mac with cursor.

time-attack:
@16francej — recommend closing this PR. The PR combines Cursor support with a separate Slate request, while #2142 covers only Cursor. It should close only after Cursor work is transferred to #2142 and a separate Slate owner is identified. Please retain attribution and the distinct Slate request.

PR #1186, opened 2026-04-24
fix(setup): support --help

### LINKED ISSUE #1133: `./setup --help` is not supported — runs install instead
Kind of funny: I wanted to use the `--prefix` option, and by instinct ran `./setup --help` first to check syntax. It just installs anyway.

It should either:
- (a) show help, or
- (b) bail out with `unsupported argument! usage: ...`

jbetala7:
Fixes #1133

## Summary
- add explicit `./setup --help` and `./setup -h` handling
- print usage before any Bun or install work starts
- add focused tests that cover normal help output and the no-Bun-on-PATH case

## Root cause
The setup script parsed flags after the Bun availability check and had no help branch, so `./setup --help` fell through into the install path instead of exiting with usage.

## Testing
- bun test test/setup-help.test.ts test/setup-codesign.test.ts

### DIFF
--- a/setup
+++ b/setup
@@ -3,6 +3,34 @@
 set -e
 umask 077  # Restrict new files to owner-only (0o600 files, 0o700 dirs)
 
+print_help() {
+  cat <<'EOF'
+Usage: ./setup [options]
+
+Build gstack binaries, generate skill docs, and install skills for supported hosts.
+
+Options:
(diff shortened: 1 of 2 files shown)


dgrant (approved):

garrytan:
Closing as dup of #1219 (same --help fix, cleaner framing with the original reporter cited). Thanks @jbetala7.

PR #1188, opened 2026-04-24
fix(telemetry): gate preamble session tracking

### LINKED ISSUE #1150: Session tracking and OPENCLAW_SESSION preamble not gated by telemetry config
## Problem

Enterprise security scanners flag gstack installations due to strings in every generated SKILL.md preamble:

```bash
mkdir -p ~/.gstack/sessions
touch ~/.gstack/sessions/"$PPID"
_SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
find ~/.gstack/sessions -mmin +120 -type f -exec rm {} + 2>/dev/null || true
```

```bash
[ -n "$OPENCLAW_SESSION" ] && echo "SPAWNED_SESSION: true" || true
```

These run **regardless of the `telemetry: off` config setting**. The telemetry config only gates remote data sending and local JSONL analytics, but session file creation, session cleanup via `find`, and the `OPENCLAW_SESSION` env var check are unconditional in the preamble.

At my company, a custom security rule monitoring for `OpenClaw` strings and suspicious `find` commands in shell initialization scripts triggered an alert, requiring investigation and remediation.

## Current behavior

- `scripts/resolvers/preamble.ts` embeds session tracking and `OPENCLAW_SESSION` detection into every SKILL.md
- Setting `telemetry: off` in `~/.gstack/config.yaml` does **not** disable these
- 765+ references across all generated SKILL.md files (every host variant × every skill)

## Suggested improvement

1. **Gate session tracking behind telemetry config** — wrap `mkdir/touch/find` for `~/.gstack/sessions/` in the same `_TEL != "off"` guard used for analytics
2. **Gate `OPENCLAW_SESSION` detection behind a config flag** — or remove it from the preamble entirely since it's only relevant when running inside OpenClaw
3. **Document what `telemetry: off` actually disables** — current config comments imply it controls all data collection, but session files are still created on disk

This would let enterprise users set `telemetry: off` and have confidence that no session tracking artifacts are created on disk.

## Workaround

Manually strip the patterns from all generated SKILL.md files after installation. This works but gets overwritten on `gstack-upgrade`.

jbetala7:
Fixes #1150

## Summary
- gate preamble session-file tracking, pending-marker handling, and timeline logging behind `telemetry != off`
- move spawned-session detection behind a helper so generated SKILL.md files no longer embed the `OPENCLAW_SESSION` string directly
- update README wording and add regression tests around telemetry gating and generated skill output

## Root cause
The preamble generator still emitted several local tracking side effects unconditionally even though `telemetry: off` is documented as disabling local analytics, and it embedded the `OPENCLAW_SESSION` env var name directly into every generated SKILL.md preamble.

## Testing
- bun test test/telemetry.test.ts test/gen-skill-docs.test.ts

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,11 +445,11 @@ gstack includes **opt-in** usage telemetry to help improve the project. Here's e
 - **On first run,** gstack asks if you want to share anonymous usage data. You can say no.
 - **What's sent (if you opt in):** skill name, duration, success/fail, gstack version, OS. That's it.
 - **What's never sent:** code, file paths, repo names, branch names, prompts, or any user-generated content.
-- **Change anytime:** `gstack-config set telemetry off` disables everything instantly.
+- **Change anytime:** `gstack-config set telemetry off` disables remote telemetry plus local analytics and session-tracking artifacts instantly.
 
 Data is stored in [Supabase](https://supabase.com) (open source Firebase alternative). The schema is in [`supabase/migrations/`](supabase/migrations/) — you can verify exactly what's collected. The Supabase publishable key in the repo is a public key (like a Firebase API key) — row-level security policies deny all direct access. Telemetry flows through validated edge functions that enforce schema checks, event type allowlists, and field length limits.
 
-**Local analytics are always available.** Run `gstack-analytics` to see your personal usage dashboard from the local JSONL file — no remote data needed.
+**Local analytics are available when telemetry is enabled.** Run `gstack-analytics` to see your personal usage dashboard from the local JSONL file — no remote data needed.
 
 ## Troubleshooting
 
--- a/SKILL.md
+++ b/SKILL.md
@@ -30,10 +30,6 @@ test a site, verify a deployment, dogfood a user flow, or file a bug with screen
 ```bash
 _UPD=$(~/.claude/skills/gstack/bin/gstack-update-check 2>/dev/null || .claude/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
 [ -n "$_UPD" ] && echo "$_UPD" || true
-mkdir -p ~/.gstack/sessions
-touch ~/.gstack/sessions/"$PPID"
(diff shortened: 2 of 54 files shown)


jbetala7:
Rebased cleanly onto current `main` (post #1848) and re-pushed (head `bacd6a4d`); now `MERGEABLE`. I also **tightened the scope** to issue #1150 only:

- **Kept** the telemetry gating: session tracking (`mkdir/touch/find ~/.gstack/sessions`), the analytics pending-marker loop, `skill-usage.jsonl` writes, and `gstack-timeline-log` are now all behind `[ "$_TEL" != "off" ]` in both `generate-preamble-bash.ts` and `generate-completion-status.ts`.
- **Kept** the `OPENCLAW_SESSION` extraction into `bin/gstack-spawned-session-status`, so that orchestrator env-var string no longer appears in any generated SKILL.md (the scanner trigger from #1150).
- **Dropped** the earlier branch's incidental removal of the plan-mode block — that was unrelated to #1150, so `GSTACK_PLAN_MODE` detection is left exactly as on `main`.

Verification:
- `bun test test/telemetry.test.ts test/gen-skill-docs.test.ts` -> 430 pass / 0 fail
- `bun test test/skill-validation.test.ts` -> 329 pass
- SKILL.md regenerated via `gen:skill-docs --host all`; extracted preamble passes `bash -n`; helper verified silent without / emits `SPAWNED_SESSION: true` with `OPENCLAW_SESSION` set.

Heads-up on overlap: #1747 proposes echoing the `SESSIONS` count for CLAUDE.md rules to read. On current `main` nothing consumes the preamble's `_SESSIONS` (it's not exported to `gstack-telemetry-log`, which only runs when telemetry is on), so this gating is safe today. If #1747 lands and makes session count a non-telemetry signal, the two will need a quick reconcile — flagging so they're reviewed together.

time-attack:
@jbetala7 Rewrite this to preserve functional context timelines for telemetry-off users, centralize session-kind detection, and update the missing generated output.

PR #1192, opened 2026-04-24
add PLAN_ROLLOUT proposal — PR-stack-aware planning

mastermanas805:
## What's the problem

gstack has great plan-time skills (`/plan-eng-review`, `/plan-ceo-review`) and great ship-time skills (`/ship`, `/review`). There's a gap between them. Nothing asks "is this one PR or three?" or "in what order should these units ship?"

The pain is LLM-specific. A single Claude Code session produces a 2,000-line diff across 15 files. Reviewers drown. Scope creep hides. Bugs ship under LGTM pressure. I feel this every time I use LLMs for non-trivial work. I suspect others do too.

## What I'm proposing

**`/plan-rollout`** runs after plan approval. Reads the plan plus `SYSTEM.md` (a new repo-root semantic contract graph) plus the discovered import graph. Produces `decomposition.md` (PR stack with reader guides, dep ordering, time-budget estimates) and `rollout.md` (rollout strategy with inverse rollback auto-generated per step).

**`/spill-check`** runs during implementation. Compares the current diff against the declared PR unit. Flags undeclared files. Adaptive: strict for code, soft for infra/meta files like CLAUDE.md, package.json, bun.lock.

**`SYSTEM.md`** is the interesting primitive. Human-declared role contracts (auth mints session tokens middleware enforces; breaks if format changes without middleware redeploy; rollout-edge hard). Separate from the package/import graph, which the LLM discovers at runtime via AST and grep. Reconciled jointly: declared contracts give the why, discovered imports give the what, disagreements surface for human resolution.

## Does it actually work

Dogfooded the design end to end against [honojs/hono#4633](https://github.com/honojs/hono/issues/4633) (405 Method Not Allowed). Authored SYSTEM.md for Hono's 8 components. Decomposed the issue into a 3-PR stack with graceful dep relaxation (PR-3 can merge without PR-2 via feature detection on an optional interface method).

Implemented what would be PR-1 locally. 171 LOC, 3 files, 86/86 tests pass, **zero regressions across the 4 router implementations not touched**.

8 design gaps surfaced during the dogfood. All folded into v1 scope. Highlights:

- SYSTEM.md needs a `kind` field (`component | leaf-util | types-only`) so shared utility dirs don't force awkward fits.
- Rollout templates need a `package-type` field because library rollouts (npm publish + revert) differ materially from service rollouts (coordinated deploy + state restore).
- Reviewer-time formula needs real calibration data. v1 ships conservative defaults and logs predicted-vs-actual so v2 can do better.

## The 4 PRs if this lands

1. Foundation: `lib/plan-rollout/system-map-*.ts` + tests + `docs/SYSTEM-MD.md`. Standalone, no skills modified.
2. `/plan-rollout` skill + the helpers the SKILL.md calls. Depends on #1.
3. `/spill-check` skill + spill classifier. Independent of #2.
4. Integration into `/ship`, `/review`, `/plan-ceo-review`, `/plan-eng-review`. Zero-regression gated on `decomposition.md` existence.

~75 min cumulative review time. PR-1 is low-risk standalone and should land first.

## What I want from you

Does this shape fit gstack? In-tree or separate plugin? Any expansions I've got wrong? Convention checks: artifacts in `~/.gstack/projects/` vs `.gstack/` in-repo? SYSTEM.md format?

If this is a nope, tell me, saves us both time. If it's yes-but-shape-it, I'll rework. If yes, I'll open the 4-PR stack.

No rush. I know the queue is deep.

### DIFF
--- a/docs/designs/PLAN_ROLLOUT.md
+++ b/docs/designs/PLAN_ROLLOUT.md
@@ -2,7 +2,7 @@
 
 **Status:** PROPOSAL — seeking directional feedback from maintainer before implementation.
 **Author:** @mastermanas805
-**Tracking issue:** (will be added after filing)
+**Tracking PR:** [#1192](https://github.com/garrytan/gstack/pull/1192)
 **Supporting materials:** [`docs/designs/plan-rollout/`](./plan-rollout/)
 
 ---
--- a/docs/designs/PLAN_ROLLOUT.md
+++ b/docs/designs/PLAN_ROLLOUT.md
@@ -1,0 +1,162 @@
+# Design: `/plan-rollout` + `/spill-check` — PR-stack-aware planning
+
+**Status:** PROPOSAL — seeking directional feedback from maintainer before implementation.
+**Author:** @mastermanas805
+**Tracking issue:** (will be added after filing)
+**Supporting materials:** [`docs/designs/plan-rollout/`](./plan-rollout/)
+
+---
+
+## The problem
+
+LLM-assisted coding compresses implementation by 10-100x. It does not compress
+**review**. A reviewer still reads code at human speed. This asymmetry shows up
+as a specific, common failure mode:
+
+- AI produces a 2,000-line diff that touches 15 files across 4 components
+- The reviewer can't meaningfully hold the change in their head
+- Scope creep ("spills" — unrelated changes sneaking in) compound the load
+- LGTM happens under cognitive pressure; real bugs ship
+- Rollback is improvised when something breaks in production
+
+None of gstack's existing skills address this. `/plan-eng-review` and
+`/plan-ceo-review` scope the plan. `/ship` creates one PR. `/review` reviews
+one diff. Nothing asks **"is this one PR or three?"** or **"in what order
+should these units ship to production?"**
+
+## The proposal
+
+Two new skills plus a declarative schema:
+
+### `SYSTEM.md` — the semantic contract graph
+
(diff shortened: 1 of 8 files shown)


mastermanas805:
@garrytan Requesting your suggestion

garrytan:
Thanks @mastermanas805 — closing as deferred. PLAN_ROLLOUT is a 2164-line proposal warranting standalone focused review. Happy to revisit if scoped down to a focused slice.

mastermanas805:
Child PR: #1424 — `/plan-rollout` MVP skill + `SYSTEM.md` schema.

Size vs the 2164-line proposal here (after the final compression pass):
- **Raw diff:** 1206 lines (~44% smaller)
- **Substantive (human-authored):** 310 lines (~7x smaller)
- The 896-line gap is the generated `plan-rollout/SKILL.md` — deterministic output of the template, reviewer skim

Per OSS PR-size research (SmartBear/Cisco, Google internal), review effectiveness drops sharply beyond 400 changed lines. Substantive content here is well inside the healthy band.

Scope reductions from the original proposal:
- Dropped `/spill-check` skill (deferred to v1.1+)
- Dropped `system-map-parser.ts` (skill prose handles parsing via bash for now)
- Dropped CEO-plan, integration-notes, usage docs
- Dropped the standalone V0 design doc — dogfood evidence moved into the PR description where reviewers actually look
- Kept: one skill that runs, schema it consumes, registry entries

PR #1207, opened 2026-04-25
fix(build): make .version writes resilient to missing git HEAD

topitopongsala:
## What

Wrap each `git rev-parse HEAD > <dir>/dist/.version` call in the `build` script with a subshell + fallback:

```diff
- && git rev-parse HEAD > browse/dist/.version
+ && (git rev-parse HEAD 2>/dev/null || echo unknown) > browse/dist/.version
```

(applied identically for `design/dist/.version` and `make-pdf/dist/.version`).

## Why

When gstack is built **outside its own git checkout** — most notably Conductor's "Quick start → gstack template" flow, which clones gstack into a brand-new repo's `.claude/skills/gstack/` (no `.git` of its own; parent repo has no commits, so `HEAD` is unborn) — `git rev-parse HEAD` exits **128** with:

```
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
```

The `&&` chain then kills the entire build:

```
error: script "build" exited with code 128
```

…which surfaces in Conductor as `Failed to create repo: error: script "build" exited with code 128` and the gstack template wipes the workspace before the user can inspect anything.

The `.version` files are pure metadata — they shouldn't be load-bearing on the build's success.

## After this fix

- Normal checkout (your dev machine): still writes the real commit SHA. ✅
- Tarball / `.git`-less install / unborn parent HEAD: writes `unknown` and the build proceeds. ✅
- Conductor "Quick start → gstack template": completes successfully. ✅

## Verification

```bash
# Normal case
cd /tmp && git clone https://github.com/garrytan/gstack && cd gstack
git checkout fix-build-script-git-rev-parse-resilience
bun install && bun run build
cat browse/dist/.version  # → real SHA, exit 0

# Conductor-template case (no .git)
tmp=$(mktemp -d) && cp -R . "$tmp"/ && rm -rf "$tmp/.git" "$tmp/node_modules"
cd "$tmp" && bun install && bun run build
cat browse/dist/.version  # → "unknown", exit 0 (previously: exit 128)
```

## Out of scope

This PR doesn't touch the broader question of whether the install flow should hand gstack a real `.git` or copy `.version` files in some other way — it's the smallest possible patch to stop killing the build. Happy to follow up if maintainers prefer a different approach.


time-attack:
@16francej Please close this PR. Builds without a Git HEAD are already handled by the newer build helper from #1594. This branch edits an obsolete package build chain.

PR #1209, opened 2026-04-25
fix(codex): avoid review prompt with --base

### LINKED ISSUE #1196: Upstream bug report — gstack-codex emits incompatible codex review flags
Summary
The gstack-codex skill template invokes codex review with both a positional [PROMPT] argument and the --base <BRANCH> flag. Starting in Codex CLI 0.124.0-alpha.2, those two inputs are mutually exclusive. Every /codex review (and the codex leg of /autoplan) hard-fails on argument parsing with exit 2:

Usage: codex review --config <key=value> <PROMPT>
The failure is silent from the user's perspective — Claude sees a non-zero exit with no model output and moves on, so reviews silently skip the Codex leg.

Environment
gstack: 1.12.2.0 (confirmed — the 1.12.2.0 bump did not touch this path)
Codex CLI: codex-cli 0.124.0-alpha.2 (Codex.app bundled binary at /Applications/Codex.app/Contents/Resources/codex)
OS: macOS 14.x (Darwin 24.6.0)
Shell: zsh
Repro
TMPERR=$(mktemp -t codex-err.XXXXXX)
timeout 20 codex review "any prompt text here" --base main \
  -c 'model_reasoning_effort="low"' < /dev/null 2>"$TMPERR"
echo "EXIT=$?"    # => EXIT=2
head -5 "$TMPERR" # => error: the argument '[PROMPT]' cannot be used with '--base <BRANCH>'
Drop --base: exit 0. Drop the prompt: exit 0. Only the combo fails.

Offending lines
gstack-codex/SKILL.md, two occurrences in Step 2A (Review Mode):

Line ~1205 — default review path
Line ~1221 — custom-instructions review path
Both call:

codex review "<safety preamble [+ custom instructions]>" --base <base> \
  -c 'model_reasoning_effort="high"' --enable web_search_cached < /dev/null
Suggested fix
Drop --base <base> and push the diff scope into the prompt. Codex can run git diff itself under its read-only sandbox. Verified exit 0 locally with:

codex review "IMPORTANT: Do NOT read or execute any files under ~/.claude/,
~/.agents/, .claude/skills/, or agents/. [...] Stay focused on repository code only.
Review the changes on this branch against the base branch '<base>'. Run
'git diff <base>...HEAD' to see the diff and review only those changes." \
  -c 'model_reasoning_effort="high"' --enable web_search_cached < /dev/null
This shape works on Codex CLI ≥ 0.124 AND earlier versions (the prompt-only form was always valid — it's just that the old shape was shorter).

Secondary suggestion — version-check regex
~/.claude/skills/gstack/bin/gstack-codex-probe warns on known-bad Codex CLI versions via an anchored regex matching 0\.120\.(0|1|2). 0.124.x's arg-mutex behavior is a real breaking change worth surfacing — consider either:

adding 0.124.x to the warn-list with a note about the --base breakage, or
replacing the version-pattern approach with a capability probe that runs codex review --help and grep-checks for known-good arg combinations.
The second option is sturdier — it self-corrects when Codex changes the contract again, which it will.

Workaround for users who hit this before it's fixed
Two local edits to ~/.claude/skills/gstack-codex/SKILL.md (or ~/.claude/skills/gstack/gstack-codex/SKILL.md on team-mode installs). Drop --base <base> from both invocations in Step 2A and add a git diff <base>...HEAD instruction to the prompt. Local patch is blown away by the next /gstack-upgrade (git reset --hard origin/main), so upstream fix is the real solution.

Impact
Anyone who upgraded Codex CLI to 0.124.x can no longer use the Codex second-opinion leg of /codex, /autoplan, or any other gstack skill that calls codex review with --base. Because the failure is silent from Claude's perspective, users may not realize Codex is missing from their review until they notice it in review logs. In our case, the last successful Codex review was a full week before the regression was diagnosed.

jbetala7:
## Summary
- remove the unsupported `codex review <PROMPT> --base <base>` shape from `/codex`, `/review`, and `/ship`
- move the diff scope into the prompt so Codex runs `git diff origin/<base>...HEAD` itself
- regenerate affected skill docs and host golden fixtures
- add regression coverage for the prompt-scoped review form

Fixes #1196.

## Testing
- `bun test test/skill-validation.test.ts --test-name-pattern "codex review invocations"`

Note: `bun test test/gen-skill-docs.test.ts test/skill-validation.test.ts test/host-config.test.ts` still hits an unrelated existing guard failure because tracked fixture `browse/test/fixtures/security-bench-haiku-responses.json` is 27MB and exceeds the 2MB limit.


habassa5:
Empirical confirmation from a downstream user — this is a real and recurring failure mode.

We've hit the silent stall ~5 times across PRs in our project (claude-teams-bot) when running `/codex review` from gstack. Root-caused 2026-05-04 to the same parse error described in #1196 (Codex CLI 0.124+ `conflicts_with` between `[PROMPT]` and `--base`). The fix in this PR — pushing diff scope into the prompt so Codex runs `git diff origin/<base>...HEAD` itself — works locally and matches the canonical resolution.

Currently maintaining a hand-patched `~/.claude/skills/codex/SKILL.md` with the same change (drops `--base`, prompt-scopes the diff). Local patch gets blown away by every `gstack-upgrade` (`git reset --hard origin/main`), so this PR landing would close that recurring tax.

A few notes from the downstream perspective:

1. The PR's broader scope (`/codex`, `/review`, `/ship`) is correct — we initially patched only `/codex` and didn't realize `/review` and `/ship` were silently affected too. Catching all three at once is a real win.
2. The regression test (`test/skill-validation.test.ts --test-name-pattern "codex review invocations"`) is exactly the right place to lock this in. Future Codex CLI breaking changes that modify the supported arg shape will be caught at gstack release time rather than in users' CI.
3. The 27MB fixture issue mentioned in the testing note appears unrelated to this PR's scope — would suggest splitting that into its own follow-up rather than blocking #1209 on it.

Happy to test against this branch from our worktrees if you'd like a downstream confirmation before merge. Any timeline thoughts on landing this?

Filing a related defensive recommendation as a separate issue (the wrapper currently swallows non-zero exits other than 124, which is what made this failure mode look like a silent stall in the first place — different problem from the parse error itself, but same observable symptom).

jbetala7:
Closing as superseded by the merged v1.42.0.0 Daegu wave in #1594, which landed this fix with attribution. Thanks for carrying it through.

PR #1211, opened 2026-04-25
security: reject internal-network gbrain_url before consumer registration

garagon:
## Summary

`gstack-brain-init` POSTs the brain bearer token + the brain repo URL
to `\${gbrain_url}/ingest-repo` with no validation of where
`\${gbrain_url}` points. Anyone with the power to set the config —
write `~/.gstack/config.yaml`, install a hook, or set `GBRAIN_URL`
via env — can redirect the POST at the operator's own private network
and walk away with both the bearer token and the private brain-repo
URL.

Worst case is the cloud metadata family. Setting
`gbrain_url=http://169.254.169.254/latest/meta-data/iam/security-credentials/<role>/`
on an EC2 host turns the consumer-registration flow into IAM
credential exfil to whoever runs the listener at the other end.

## Threat model

A process or operator that can:

- Edit `~/.gstack/config.yaml` (any local code-execution primitive)
- Install a malicious hook into the gstack pipeline
- Set `GBRAIN_URL=` in the environment of a `gstack-brain-init` run
  (any shell-rc poisoning, npm postinstall, social engineering of a
  README that says "set this env var")

None of these require root, container escape, or network position. They
turn a config-write into an outbound POST that exfils credentials to
an attacker-chosen internal target.

## What changes

`validate_gbrain_url()` is added to `bin/gstack-config` and wired into
two paths:

1. `gstack-config set gbrain_url <value>` runs the validator at write
   time and refuses to persist a hostile URL.
2. `gstack-brain-init` re-validates `\$GBRAIN_URL_VAL` through the new
   internal subcommand `gstack-config __validate-url` before the curl
   POST. The env-var path bypasses (1), so this defense-in-depth
   covers the `GBRAIN_URL=...` runtime override.

Validator rules (all reject by default; opt out with
`GSTACK_ALLOW_INTERNAL_URL=1` for local dev against a personal gbrain):

- require `http://` or `https://` scheme — no `file:`, `gopher:`,
  `data:`, `javascript:`, `ftp:`
- reject literal local hostnames: `localhost`, `localhost.*`,
  `*.local`, `*.localhost`, `broadcasthost`
- reject IPv4 RFC1918 (10/8, 172.16/12, 192.168/16), loopback (127/8),
  reserved (0/8), and link-local / cloud IMDS (169.254/16)
- reject IPv6 loopback (`[::1]`, `[::]`), link-local (`[fe80::*]`),
  and unique-local (`[fc00::*]`, `[fd*]`)
- IPv6 host extraction handles bracketed authorities so `[fe80::1]:8443`
  extracts as `[fe80::1]` rather than collapsing to `[`

## Backwards compatibility

- A normal user running `gstack-config set gbrain_url
  https://gbrain.example.com` is unaffected.
- A dev running a local gbrain at `http://127.0.0.1:8080` opts in
  with `GSTACK_ALLOW_INTERNAL_URL=1`. The error message names the env
  var explicitly so the discovery cost is one error → one re-run.
- `gstack-config get`, `list`, and `defaults` paths are unchanged.

## Test coverage

`test/gstack-config-url-ssrf.test.ts` — 33 cases following the same
spawnSync + tmp `GSTACK_HOME` pattern that
`gstack-brain-init-gh-mock.test.ts` already uses:

- 8 public URLs accepted (https + http, ports, paths, IPv6 docs)
- 5 local-only hostnames rejected (localhost flavors)
- 9 IPv4 hostile addresses rejected (loopback, all three RFC1918
  blocks, IMDS)
- IPv6 `[::1]` and `[fe80::1]` rejected
- `172.32.x.x` accepted (NOT in `172.16-31` range — boundary check)
- bad shapes rejected: empty, `file://`, `gopher://`, `data:`,
  `javascript:`, `ftp://`
- `GSTACK_ALLOW_INTERNAL_URL=1` lets a loopback target through
- `GSTACK_ALLOW_INTERNAL_URL=1` still rejects empty + non-http schemes
- the set path refuses to write a hostile URL to config.yaml
- the set path writes a valid public URL through
- the set path with `GSTACK_ALLOW_INTERNAL_URL=1` lets dev URL through

Adjacent tests still pass: `explain-level-config`,
`gstack-brain-init-gh-mock`, `brain-sync` (41/41 across 3 files).

## Why this matters

The brain-consumer registration flow trusts whatever endpoint the
operator named, which is the right default for a happy-path setup.
The cost was that any path-of-least-resistance attacker who can
influence the config — a malicious dev tool, a tampered shell rc, an
env var planted by a CI yaml — turns one curl into a bearer-token
+ private-repo-URL exfil to AWS IMDS or any internal target. The
validator is small (one function, one subcommand, two call sites)
and additive — every legitimate setup keeps working, and the dev
escape hatch is one env var away.

### DIFF
--- a/bin/gstack-brain-init
+++ b/bin/gstack-brain-init
@@ -287,18 +287,31 @@ GBRAIN_URL_VAL="${GBRAIN_URL:-$("$CONFIG_BIN" get gbrain_url 2>/dev/null || echo
 GBRAIN_TOKEN_VAL="${GBRAIN_TOKEN:-$("$CONFIG_BIN" get gbrain_token 2>/dev/null || echo "")}"
 
 if [ -n "$GBRAIN_URL_VAL" ] && [ -n "$GBRAIN_TOKEN_VAL" ]; then
-  # Try the HTTP handoff.
-  HTTP_RESP=$(curl -sS -X POST "${GBRAIN_URL_VAL%/}/ingest-repo" \
-    -H "Authorization: Bearer $GBRAIN_TOKEN_VAL" \
-    -H "Content-Type: application/json" \
-    --data "{\"repo_url\":\"$REMOTE_URL\"}" \
-    -w "\n%{http_code}" 2>&1 || echo -e "\ncurl-error")
-  HTTP_CODE=$(echo "$HTTP_RESP" | tail -1)
-  if [ "$HTTP_CODE" = "200" ] || [ "$HTTP_CODE" = "201" ] || [ "$HTTP_CODE" = "204" ]; then
-    CONSUMER_STATUS="ok"
-    echo "GBrain consumer registered: $GBRAIN_URL_VAL"
+  # SSRF guard: re-validate the URL even when it came from env. The
+  # `gstack-config set gbrain_url` path validates at write time; the
+  # GBRAIN_URL env override skips that path entirely, so we re-check
+  # here so an env-supplied loopback / RFC1918 / link-local target
+  # cannot turn this curl into a token + repo-URL exfil to an internal
+  # endpoint. GSTACK_ALLOW_INTERNAL_URL=1 opts into local targets for
+  # dev (e.g. http://127.0.0.1:8080 against a personal gbrain).
+  if ! "$CONFIG_BIN" __validate-url "$GBRAIN_URL_VAL" >/dev/null 2>&1; then
+    "$CONFIG_BIN" __validate-url "$GBRAIN_URL_VAL" || true
+    echo "Skipping GBrain consumer registration — gbrain_url failed SSRF validation." >&2
+    CONSUMER_STATUS="rejected"
   else
-    echo "GBrain ingest endpoint returned HTTP $HTTP_CODE; will retry on next skill run."
+    # Try the HTTP handoff.
+    HTTP_RESP=$(curl -sS -X POST "${GBRAIN_URL_VAL%/}/ingest-repo" \
+      -H "Authorization: Bearer $GBRAIN_TOKEN_VAL" \
+      -H "Content-Type: application/json" \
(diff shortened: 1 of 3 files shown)


garagon:
Closing — this PR targets a code path that no longer exists.

Between v1.12.2.0 and v1.17.0.0 the consumer-registration block was lifted out of `bin/gstack-brain-init` into `bin/gstack-brain-consumer`, and that helper is now marked `DEPRECATED in v1.17.0.0` (line 4). My patch adds the `--ingest-url` SSRF guard around a registration block that `main` deliberately removed. Force-rebasing it would re-introduce code you took out, not resolve a textual conflict.

The surface itself didn't go away though, in case it's useful when triaging the v1.17+ federation work:

- `bin/gstack-brain-consumer add --ingest-url <url> --token <token>` accepts any URL (no http(s) check, no RFC1918/loopback/link-local rejection) and writes it into `~/.gstack/consumers.json`. The token is stashed in `gstack-config` under `<name>_token`.
- `bin/gstack-config set gbrain_url <url>` likewise stores any value as-is (current `main` has only the `[A-Za-z0-9_]+` key validator, no value validation).
- Anything that later POSTs to `${ingest_url}/ingest-repo` with the bearer token (the v1.17.0.0 federation surface, or whatever lands next) inherits the same exfil-to-internal-target risk the original `gstack-brain-init` PR was guarding against.

If the v1.17.0.0 wireup is the canonical replacement, the `validate_gbrain_url` function in this PR's `bin/gstack-config` diff is portable — http(s)-only, RFC1918 + loopback + link-local + IPv6 ULA rejection, `GSTACK_ALLOW_INTERNAL_URL=1` opt-out for personal-gbrain dev. Happy to repackage against the new entry point if useful, otherwise this thread can stay closed as the documentation trail.

PR #1219, opened 2026-04-26
fix(setup): support -h/--help instead of running installer (#1133)

### LINKED ISSUE #1133: `./setup --help` is not supported — runs install instead

saen-ai:
Fixes #1133.

## Before
`./setup --help` silently ran the full installer because the arg parser's catch-all (`*) shift`) swallowed unknown flags. Reported by @dgrant — instinct was to check syntax with `--help`, got an install instead.

## After
`-h` / `--help` print a usage banner and exit 0. The check runs **before** the bun availability guard so the flag is discoverable on a fresh machine that hasn't installed bun yet.

```
$ ./setup --help
gstack setup — install gstack skills + build browse binary

Usage: ./setup [options]

Options:
  --host <name>     Install for a specific host (claude, codex, kiro, factory,
                    opencode, openclaw, hermes, gbrain, auto). Default: claude.
  --prefix          Install skills with the gstack- prefix (e.g. /gstack-review).
  --no-prefix       Install skills with short names (e.g. /review). Default.
  --team            Switch to team mode (per-repo gstack with auto-update).
  --no-team         Force solo install even if a team-mode repo is detected.
  -q, --quiet       Suppress progress output.
  -h, --help        Show this help and exit.
...
```

## Tests
`test/setup-help.test.ts` adds 5 cases following the static-pattern convention from `test/setup-codesign.test.ts`:
- usage function exists
- `-h|--help` is handled before the bun availability check
- usage text covers every documented flag
- `./setup --help` exits 0, prints `Usage:`, runs no installer side effects
- `./setup -h` is equivalent to `./setup --help`

`bun test test/setup-help.test.ts` → 5 pass. Existing `setup-codesign` tests unaffected.

## Scope notes
- Did not touch the catch-all `*) shift ;;` (silently swallows other unknown flags) — that's a separate behaviour worth its own discussion.
- No template/generated drift: the change is in `setup` only, no `.tmpl` regen needed.

### DIFF
--- a/setup
+++ b/setup
@@ -3,6 +3,40 @@
 set -e
 umask 077  # Restrict new files to owner-only (0o600 files, 0o700 dirs)
 
+usage() {
+  cat <<'EOF'
+gstack setup — install gstack skills + build browse binary
+
+Usage: ./setup [options]
+
+Options:
+  --host <name>     Install for a specific host (claude, codex, kiro, factory,
+                    opencode, openclaw, hermes, gbrain, auto). Default: claude.
+  --prefix          Install skills with the gstack- prefix (e.g. /gstack-review).
+  --no-prefix       Install skills with short names (e.g. /review). Default.
+  --team            Switch to team mode (per-repo gstack with auto-update).
+  --no-team         Force solo install even if a team-mode repo is detected.
+  -q, --quiet       Suppress progress output.
+  -h, --help        Show this help and exit.
+
+Examples:
+  ./setup                          # solo install for Claude Code
+  ./setup --host codex             # install for OpenAI Codex CLI
+  ./setup --team                   # team mode for a shared repo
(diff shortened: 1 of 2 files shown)


jbetala7:
Heads-up for dedup: this overlaps with #1186 (opened 2026-04-24), which also fixes #1133 — same approach (`-h`/`--help` handling in `setup` with tests in `test/setup-help.test.ts`) and the same root cause (the `*) shift` catch-all swallowing unknown flags). Flagging so a maintainer can pick one rather than merging both.

PR #1223, opened 2026-04-26
fix(discover): read full Codex session metadata line

euyua9:
## Summary
- replace the fixed 128KB Codex `session_meta` prefix read with a bounded read-until-newline helper
- keep the read bounded at 1 MiB so malformed files cannot cause unbounded memory use
- update the >128KB regression test to verify discovery succeeds instead of documenting the limitation

## Why
Codex `session_meta` stores `base_instructions` on the first JSONL line. The current scanner reads only the first 128KB before splitting on `\n`, so a valid first line larger than that is truncated and `JSON.parse` fails. That silently drops the Codex session from global discovery.

The scanner only needs the first line, so reading chunks until the first newline preserves the existing fast path while handling larger but valid metadata lines.

## Testing
- `bun test test/global-discover.test.ts`

### DIFF
--- a/bin/gstack-global-discover.ts
+++ b/bin/gstack-global-discover.ts
@@ -297,6 +297,39 @@ function extractCwdFromJsonl(filePath: string): string | null {
   return null;
 }
 
+
+function readFirstLineBounded(filePath: string, maxBytes = 1024 * 1024): string | null {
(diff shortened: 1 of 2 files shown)


16francej:

PR #1224, opened 2026-04-26
v1.15.1.0 fix(gstack-slug): subdir guard + .gstack-slug override + --reset (#1125)

### LINKED ISSUE #1125: gstack-slug inherits outer-repo slug when project is a subdirectory; cache is unrecoverable
**Title:** `gstack-slug` inherits outer-repo slug when project is a subdirectory; poisoned cache is unrecoverable

**Summary**

`bin/gstack-slug` derives the project slug from `git remote get-url origin`. When the current working directory is a subdirectory of a larger git repo (e.g., a standalone project living inside a shared workspace repo), git walks up and returns the outer repo's remote. The resulting slug belongs to the outer repo, not the actual project. That slug is then cached at `~/.gstack/slug-cache/{pwd-encoded}` with no TTL, no invalidation path, and no override — so once it's wrong, it stays wrong across every future session for that path.

**Repro**

1. Have a git repo at `/Users/me/workspace` with remote `origin` pointing to `me/workspace`.
2. Create a new project dir at `/Users/me/workspace/IoTopia/` (no `.git` of its own).
3. `cd /Users/me/workspace/IoTopia && gstack-slug`

**Expected**

Slug reflects the IoTopia project — either prompt the user, or fall back to basename since this dir is not the repo root.

**Actual**

```
SLUG=me-workspace
```

The outer repo's slug. `~/.gstack/projects/me-workspace/` now collects artifacts for an unrelated project, and `~/.gstack/slug-cache/_Users_me_workspace_IoTopia` locks that mistake in.

**Root cause**

`bin/gstack-slug` lines ~22–28:

```bash
REMOTE_URL=$(git remote get-url origin 2>/dev/null) || REMOTE_URL=""
if [[ -n "$REMOTE_URL" ]]; then
  RAW_SLUG=$(printf '%s' "$REMOTE_URL" | sed '...')
  SLUG=$(printf '%s' "$RAW_SLUG" | tr -cd 'a-zA-Z0-9._-')
fi
```

There is no check for whether `$PWD` is actually the repo's top-level. Any subdirectory inherits the parent repo's identity.

**Impact**

Downstream skills (office-hours, plan-ceo-review, plan-eng-review, autoplan, design-consultation, etc.) read/write under `~/.gstack/projects/{slug}/`. When slug is wrong, artifacts land in the wrong project's folder and cross-contaminate unrelated work. Users only notice after the fact, and manual cleanup of both `~/.gstack/projects/` and `~/.gstack/slug-cache/` is needed.

**Suggested fixes** (in order of increasing ambition)

1. **Subdir guard.** Before trusting the remote URL, check `git rev-parse --show-toplevel` against `$PWD`. If they differ, don't inherit the outer slug — fall through to basename, or prompt.
2. **Per-project override.** If a `.gstack-slug` file exists in `$PWD`, use its contents verbatim and skip all inference. Cheap, user-controllable, survives `gstack-upgrade`.
3. **Cache management.** Add `gstack-slug --reset` to clear the cache entry for `$PWD`, and `gstack-slug --set <slug>` to write one explicitly. Also consider a TTL or a staleness warning when the cached slug doesn't match what the resolver would compute today.
4. **Interactive first-run.** For a new path that has no cache entry and ambiguous inference (subdir of another repo, no `.git`, etc.), prompt once and persist the choice to `.gstack-slug`.

**Environment**

- gstack version: 1.5.1.0
- OS: macOS (Darwin 25.3.0)
- Shell: zsh

gregario:
## Summary

Fixes #1125 — `gstack-slug` inherits the outer repo's slug when `\$PWD` is a subdirectory of a different git repo, then locks the wrong slug into `~/.gstack/slug-cache/` with no recovery path.

## Before

```
$ cd /Users/me/workspace                  # outer repo, origin = me/workspace
$ mkdir IoTopia && cd IoTopia              # standalone project, no .git
$ gstack-slug
SLUG=me-workspace                          # ← wrong: inherited from parent
```

`~/.gstack/projects/me-workspace/` now collects artifacts for an unrelated project. Manual cleanup of both `~/.gstack/projects/` and `~/.gstack/slug-cache/` was needed to recover.

## After

```
$ cd /Users/me/workspace/IoTopia
$ gstack-slug
SLUG=IoTopia                               # ← falls through to basename

# Or, set a per-project override:
$ echo my-real-project > .gstack-slug
$ gstack-slug --reset && gstack-slug
SLUG=my-real-project
```

## Three changes

1. **Subdir guard.** Compare `git rev-parse --show-toplevel` against `\$PWD`. If they differ, skip the remote-URL inference and fall through to `.gstack-slug` or basename. Eliminates the silent cross-contamination.
2. **`.gstack-slug` per-project override.** A one-line file in `\$PWD` (sanitized to `[a-zA-Z0-9._-]`) wins over git inference. Cheap, explicit, survives `gstack-upgrade`.
3. **`gstack-slug --reset`.** Removes the cache entry for the current PWD without manual cleanup of `~/.gstack/slug-cache/`.

Mirrors the shape of @snowmaker's merged PR #897 (deterministic slugs across sessions): one `bin/` script change, one regression test file, zero behavior change for the happy path.

## Test plan

- [x] `bun test test/gstack-slug.test.ts` — 10 cases pass
  - Subdir guard regression (the reproducer from the issue)
  - PWD at toplevel still inherits remote slug (happy path unchanged)
  - No git repo at all falls back to basename (happy path unchanged)
  - `.gstack-slug` override at toplevel and in subdir
  - Override input sanitization (rejects `;`, spaces, multi-line)
  - `--reset` clears the cache file; no-op when cache absent
  - Cache stability: cached slug persists when remote disappears
- [x] `bun test` (full free suite) — green
- [x] No changes to existing happy-path behavior

## Files

- `bin/gstack-slug` — fix
- `test/gstack-slug.test.ts` — new, 10 cases
- `VERSION`, `CHANGELOG.md` — release v1.15.1.0


garrytan:
Thanks @gregario — closing. Version slot v1.15.1.0 already taken on main. Issue #1125 may still be real; push the gstack-slug subdir guard on its own without the version-slot framing.

PR #1225, opened 2026-04-26
v1.18.0.0 fix(browse): screenshot defaults to viewport, add --full-page

### LINKED ISSUE #1214: browse: full-page screenshots silently exceed Anthropic vision API 2000px limit, bricking sessions
### Summary

`screenshot` (and `snapshot -a` / `snapshot -H`) default to `fullPage: true`. On any page taller than ~2000px scroll height, the resulting PNG exceeds the Anthropic vision API's 2000px-per-side limit for many-image requests. The failing image stays in conversation history, so every subsequent turn re-fails on the same envelope — the entire session is bricked, not just one tool call.

This is a soft footgun: defaults are otherwise excellent (`viewport` 1280×720, `deviceScaleFactor` 1), so the user has no signal that "long page" is the boundary condition until they hit it.

### Repro

1. `browse goto https://example.com/some-long-doc` (any page with body height > ~2000px) 2. `browse screenshot /tmp/shot.png` 3. Have Claude `Read` the resulting PNG 4. Claude returns: `image exceeds 2000 pixels on the longest edge` 5. The PNG is now stuck in transcript — every following turn fails

### Where in code

Affected sites (refs at commit `6209163`):

- `browse/src/snapshot.ts:419` — `await page.screenshot({ path, fullPage: true })` (annotate) - `browse/src/snapshot.ts:539` — same (heatmap)
- (plus the bare `screenshot` command — didn't trace the exact line, but the docstring
 at `commands.ts:138` documents it as full-page by default)

### Proposed fix

Three options, in order of how invasive:

1. **Cheapest**: in `browser-manager.ts` add a post-capture guard. After every `page.screenshot(...)`, read PNG dims (zero-dep: parse IHDR chunk, ~20 lines) and either downscale or split if either side > 1800px. Emit a `[browse]` log line so the agent sees what happened. No new deps.

2. **Behavioral**: flip the default. Make `screenshot` capture the visible viewport unless `--full-page` is passed explicitly. The current `--viewport` flag becomes the default; add `--full-page` as opt-in. Mirrors how Playwright's own API treats `fullPage: false` as the default.

3. **Config knob**: `~/.gstack/config.yaml` exposes `screenshot_max_height: 1800` (default), and the screenshot command auto-splits above it. Lets power users opt out by setting it to 0 / Infinity.

I'd vote for (1) + (2) together: agents can't accidentally produce poison images, and `--full-page` stays available for users who explicitly want it.

### Why this matters

Once one bad screenshot lands in conversation history, the user has to either clear context or ask Claude to forget. Both are friction. Since gstack is explicitly designed for "agent QA-ing a site", the failure mode lands almost entirely on agent users, which is precisely the audience.

Happy to send a PR if the maintainers agree on which option to take.

— Reported via Claude Opus 4.7, gstack v1.12.2.0

gregario:
## Summary

Fixes #1214 — `browse screenshot` defaulted to `fullPage: true`, so any page taller than ~2000px scroll height produced a PNG that exceeded the Anthropic vision API's per-side limit. Once a poison image landed in conversation history, every subsequent turn re-failed on the same envelope. Whole session bricked, not just one tool call.

## Before

```
$ browse goto https://example.com/some-long-doc
$ browse screenshot /tmp/shot.png
Screenshot saved: /tmp/shot.png        # 1280×3700 PNG
# Claude tries to Read /tmp/shot.png:
# → \"image exceeds 2000 pixels on the longest edge\"
# → every following turn now fails on the same image
```

## After

```
$ browse screenshot /tmp/shot.png
Screenshot saved: /tmp/shot.png        # 1280×720 viewport, safe

$ browse screenshot --full-page /tmp/long.png
[browse] warning: full-page screenshot is 1280x3695px — exceeds ~2000px Anthropic vision API limit. Consider --viewport (default), --clip x,y,w,h, or --selector for a smaller capture.
Screenshot saved (full-page): /tmp/long.png
```

## Changes

1. **Flip the default.** Bare `screenshot` captures the visible viewport. Matches Playwright's own `page.screenshot()` default of `fullPage: false`.
2. **Add `--full-page`** opt-in. When used, the result is checked against an ~1800px ceiling (vision-API limit with headroom) and emits a one-line `[browse]` warning to stderr — agents see the boundary instead of silently shipping poison.
3. **Keep `--viewport`** as a back-compat no-op alias. Existing scripts that passed `--viewport` keep working; the output message just no longer carries `(viewport)` since viewport IS the default — only `--full-page` gets a `(full-page)` suffix.
4. PNG dimensions are read from the IHDR chunk (zero deps, ~10 lines).

Annotated/heatmap snapshot modes still default to fullPage — those are debug UI for humans, not images fed to agents via `Read`.

This goes with the issue author's vote: option 1 (post-capture warn) plus option 2 (flip the default). \"Search before recommending\" rubric: Playwright's API was already shipping the right default — gstack just wasn't matching it.

## Test plan

- [x] `bun test browse/test/commands.test.ts` — 228 pass (8 cases changed/added)
  - Regression for #1214: `tall.html` (~3000px) with bare `screenshot` produces PNG ≤ 800px tall (viewport)
  - `--full-page` on the same page produces > 2000px
  - `--full-page` on tall page emits the oversize warning (captured via `console.warn` patch)
  - `--full-page + --clip` and `--full-page + selector` reject mutually-exclusive flag combos
  - `--viewport + --clip` is now a no-op + clip (back-compat: was an error before; fine because clip is what runs)
  - Existing 220 tests unchanged
- [x] `bun test` (full free suite) — green

## Migration note

If any caller depended on the literal output `\"Screenshot saved (viewport)\"`, that string changed to `\"Screenshot saved\"` (since viewport is now the default and not a special case). Only one test in this repo grepped for that exact string; no production callers found in `grep -rn`.

## Files

- `browse/src/meta-commands.ts` — flag parsing + IHDR reader + warnIfOversize helper
- `browse/src/commands.ts` — usage docstring
- `browse/test/commands.test.ts` — 5 new tests, 1 updated, 1 inverted (back-compat)
- `browse/test/fixtures/tall.html` — new ~3000px fixture
- `VERSION`, `CHANGELOG.md` — release v1.15.2.0


garrytan:
Thanks @gregario — closing. Version slot v1.18.0.0 already taken on main. Push the screenshot --full-page change on its own without the version-slot framing.

PR #1226, opened 2026-04-26
v1.15.3.0 feat(browse): viewport auto / reset to unpin a fixed viewport (#1059)

### LINKED ISSUE #1059: browse: add `viewport auto` / `viewport reset` to unpin fixed viewport
## Problem

`browse viewport W×H` pins the viewport via Playwright's `setViewportSize`, overriding the launch-time `viewport: null` that normally lets the viewport follow the real Chrome window size. Once pinned, there is no way to restore auto-follow behaviour without `browse restart`, which kills the Chrome session (and drops a `/connect-chrome` headed attachment back to `launched` mode).

Several skills call `$B viewport` as a side-effect (e.g. responsive testing, benchmarks with specific breakpoints), leaving the browser in a locked state even after the skill finishes. From the user's perspective the Chrome window is visibly larger than the viewport Claude is rendering into.

## Repro

```bash
$B viewport 1280x520
$B js "window.innerWidth + 'x' + window.innerHeight"   # 1280x520
# resize Chrome window manually → viewport stays 1280x520
```

## Proposal

A `browse viewport auto` (or `viewport reset` / `viewport unpin`) command that clears the pinned size and restores window-following behaviour, without tearing down the Chrome session.

In Playwright terms this means calling something like `page.setViewportSize(null)` where supported, or recreating the page/context with `viewport: null`. If neither is possible without a restart, even documenting the constraint + exposing a lightweight "soft restart that preserves the current tab/URL" would help.

## Environment

- Platform: Windows 11
- Mode: both `headed` (via `/connect-chrome`) and `launched` reproduce

gregario:
## Summary

Closes #1059. Once `browse viewport WxH` calls Playwright's `setViewportSize`, the viewport is pinned. There's no page-level API to undo it — the only way back to window-following is to rebuild the context. Skills that called `\$B viewport` for responsive testing or breakpoint benchmarks were leaving the browser locked at that size for the rest of the session, with the visible Chrome window noticeably larger than what Claude was rendering into.

## Before

```
$ browse viewport 1280x520
$ browse js \"window.innerWidth + 'x' + window.innerHeight\"
1280x520
# resize Chrome window manually → viewport stays 1280x520
# only escape: browse restart (kills Chrome)
```

## After

```
$ browse viewport auto
Viewport unpinned — now follows window size (context recreated; refs and load-html content replayed).

# headed mode (real Chrome via /connect-chrome):
$ browse viewport auto
Viewport synced to current window size (1920x1080). Re-run \`viewport auto\` after resizing.
```

## Connection-mode-aware unpin

| Mode | Path | Notes |
|---|---|---|
| **launched** (headless) | full `recreateContext()` with `viewport: null` | cookies/storage/URLs/tab-ownership preserved via existing save/restore. True follow restored. |
| **headed** (`launchPersistentContext`) | one-shot resync via `page.evaluate(window.innerWidth/Height)` + `setViewportSize` | persistent context can't be safely rebuilt (tied to user's Chrome window). Re-run after a resize. Documented in command help. |

## Implementation notes

- New `viewportPinned: boolean` on `BrowserManager` (default true, matches launch).
- `setViewport(w, h)` sets it true so re-pinning after auto works.
- `recreateContext()` builds two distinct context-option shapes. **Important:** Playwright rejects `deviceScaleFactor` with `viewport: null`, so the unpinned path omits it. Same constraint the existing headed-launch path already obeys (line 359 — `viewport: null` with no DSF).
- `--scale` plus `auto`/`reset` is rejected at the parser — scale needs an explicit WxH to multiply.

## Test plan

- [x] `bun test browse/test/commands.test.ts` — 227 pass (4 new tests in dedicated end-of-file describe block).
  - `viewport auto unpins a fixed size` — pins, unpins, re-pins to a different size; the re-pin takes effect.
  - `viewport reset is an alias for auto` — same path via the alias.
  - `viewport auto preserves cookies across context recreation` — sets cookie, unpins, cookie still present.
  - `viewport auto rejects --scale combination` — input validation.
- [x] Tests live at the end of the file because context recreation renumbers tab IDs, and an existing test in the Tabs describe block hardcodes tab id 1. Running before the Tabs block would break that test.
- [x] Headed-mode path covered by the implementation but not by automated tests (existing tests don't run a real Chrome window).

## Files

- `browse/src/browser-manager.ts` — `unpinViewport()` + viewportPinned tracking + recreateContext branching
- `browse/src/write-commands.ts` — viewport `auto`/`reset` parsing + `--scale` rejection
- `browse/src/commands.ts` — usage docstring
- `browse/test/commands.test.ts` — 4 new tests
- `VERSION`, `CHANGELOG.md` — release v1.15.3.0


garrytan:
Thanks @gregario — closing. Version slot v1.15.3.0 already taken on main. Issue #1059 may still be real; push the viewport auto/reset on its own without the version-slot framing if you want it to land.

PR #1278, opened 2026-04-30
fix(design): disclose OpenAI key source

### LINKED ISSUE #1248: `design` binary silently uses `OPENAI_API_KEY` from any `.env` in cwd — security/billing concern
## What happened

I ran `/design-consultation` from the root of one of my projects. The project has a `.env` file with an `OPENAI_API_KEY` used by the project's own runtime (an unrelated app). I never told gstack about that key. I expected it to use `~/.gstack/openai.json` or my Codex auth.

After about 5 image generations (≈$0.50 in `gpt-image-1` calls), I noticed the charges hit the project's OpenAI account, not the account I'd associated with gstack via `codex login`.

## Root cause

`design.exe` is a Bun-compiled single-file executable. Bun's runtime [auto-loads `.env` from the current working directory](https://bun.sh/docs/runtime/env) at process start. So whenever the design binary is spawned from a directory containing `.env` with `OPENAI_API_KEY=...`, that key wins over the documented auth chain (`~/.gstack/openai.json` → `OPENAI_API_KEY` → `~/.codex/auth.json`) — because the env var is *already populated* by the time the auth resolver runs.

The tool's `--help` output advertises:

> Auth: `~/.gstack/openai.json` or `OPENAI_API_KEY` env var

But it doesn't say which one wins, and it doesn't say that "env var" includes silently auto-loaded `.env` files.

## Why this matters

1. **Billing surprise.** Image generation gets charged to whichever OpenAI account a project's `.env` happens to be tied to — not the one the user set up gstack with. Easy to accumulate spend on the wrong account.
2. **Cross-project key leakage.** Many projects keep `.env` for local dev — internal tools, side projects, contractor work, client codebases. Running gstack inside a client project would pull the client's key. That's a real footgun.
3. **No visibility.** The tool never tells the user which key it picked or which account will be billed. The user has to dig into Bun runtime semantics to figure out what just happened.
4. **`.env` is typically gitignored as a secret.** Tools that auto-load it should disclose that they're doing so.

## Suggested fixes

In rough order of effort:

1. **Disclose key source on every run.** Print one line: `using OpenAI key from <source>` (e.g., `~/.gstack/openai.json`, `OPENAI_API_KEY env`, `.env in cwd`, `~/.codex/auth.json`). Even just the source label, not the key, would have prevented this.
2. **Don't auto-load `.env`.** Either disable Bun's `.env` auto-load (`Bun.env` vs `process.env`, or set `--no-env-file` if Bun supports it), or filter `OPENAI_API_KEY` out of the inherited env unless it came from `~/.gstack/openai.json` / a deliberate gstack-owned source.
3. **Require explicit opt-in to use cwd `.env`.** A flag like `--allow-env-file` or a config like `gstack_use_cwd_env`.
4. **Add the disclosure to `--help`.** Make it explicit that any `.env` in the working directory at invocation time is read.

Happy to test a fix.

jbetala7:
Fixes #1248

## Summary
- report which auth source the design binary is using before it makes billed OpenAI requests
- preserve the existing priority order: `~/.gstack/openai.json` first, then `OPENAI_API_KEY`
- warn when `OPENAI_API_KEY` matches `.env` / `.env.local` in the current directory, so users can see when a project-local key may be used for billing
- keep `resolveApiKey()` backward-compatible while adding source-aware auth metadata and tests

## Root cause
Bun can populate `process.env.OPENAI_API_KEY` from the current working directory's `.env` before gstack's auth resolver runs. The resolver then treated that value like any other environment variable, so design generation gave no clue that billing might hit a project-specific OpenAI account.

## Testing
- `bun test design/test/auth.test.ts`
- `bun test design/test/auth.test.ts design/test/gallery.test.ts`

## Notes
I also attempted the broader local design test command including `serve.test.ts` and `feedback-roundtrip.test.ts`; this worktree hit unrelated environment failures (`Bun.serve({ port: 0 })` failed to bind, and `playwright` was not resolvable from `feedback-roundtrip.test.ts`). The focused auth coverage and existing gallery tests pass.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1278"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/design/src/auth.ts
+++ b/design/src/auth.ts
@@ -10,16 +10,67 @@
 import fs from "fs";
 import path from "path";
 
-const CONFIG_PATH = path.join(process.env.HOME || "~", ".gstack", "openai.json");
+type ApiKeySource = "config" | "env";
 
-export function resolveApiKey(): string | null {
+export interface ApiKeyResolution {
+  key: string;
+  source: ApiKeySource;
+  envFile?: string;
+  warning?: string;
+}
+
+function configPath(): string {
+  return path.join(process.env.HOME || "~", ".gstack", "openai.json");
+}
+
+function readEnvValue(filePath: string, key: string): string | null {
+  let content: string;
+  try {
+    content = fs.readFileSync(filePath, "utf-8");
+  } catch {
+    return null;
+  }
+
+  for (const line of content.split(/\r?\n/)) {
+    const match = line.match(new RegExp(`^\\s*(?:export\\s+)?${key}\\s*=\\s*(.*)\\s*$`));
+    if (!match) continue;
+
+    let value = match[1].trim();
+    if (
+      (value.startsWith('"') && value.endsWith('"')) ||
+      (value.startsWith("'") && value.endsWith("'"))
+    ) {
+      value = value.slice(1, -1);
+    }
+    return value;
+  }
+
+  return null;
+}
+
+function matchingCwdEnvFile(key: string, value: string): string | null {
+  const candidates = [".env"];
+  const nodeEnv = process.env.NODE_ENV;
+  if (nodeEnv) candidates.push(`.env.${nodeEnv}`);
+  candidates.push(".env.local");
+
+  for (const fileName of candidates) {
+    const fileValue = readEnvValue(path.join(process.cwd(), fileName), key);
+    if (fileValue === value) return fileName;
+  }
+
+  return null;
+}
+
+export function resolveApiKeyInfo(): ApiKeyResolution | null {
   // 1. Check ~/.gstack/openai.json
   try {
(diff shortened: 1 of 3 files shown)


jbetala7:
Closing this as superseded by the v1.42.0.0 maintainer wave: current main includes commit 16fca84d (), and issue #1248 is now closed. Thanks for carrying the fix through.

PR #1297, opened 2026-05-02
fix: Korean/CJK IME input and rendering in Sidebar Terminal

### LINKED ISSUE #1272: Sidebar Terminal: Korean (CJK) IME input not delivered to PTY
## Summary

Sidebar Terminal (xterm.js v5.3.0) has **two separate Korean / CJK bugs**:

1. **Input**: Korean (CJK) text typed via IME composition does not reach the PTY correctly.
2. **Output**: Korean text rendered by xterm.js is visually broken — overlapping glyphs, garbled fragments, cursor misalignment.

Other input/output (English, ASCII, control sequences) works perfectly.

## Environment

- gstack version: **1.20.0.0**
- OS: WSL2 (Ubuntu 24.04.4 LTS) under Windows 11, WSLg active (`DISPLAY=:0`, `WAYLAND_DISPLAY=wayland-0`)
- Browser: Playwright Chromium 145.0.7632.6 (auto-launched by `$B connect`)
- IME: Microsoft 한글 IME (Windows 11 default Korean IME, passed through WSLg)
- xterm bundled at `extension/lib/xterm.js` — `node_modules/xterm` package version `5.3.0`

---

### Reproduction

1. `/connect-chrome` to launch GStack Browser
2. Open Side Panel → Terminal pane loads with live Claude PTY
3. Switch IME to Korean (e.g., Right-Alt or Win+Space)
4. Type `안녕하세요`

**Expected**: Composed Korean characters arrive at the PTY and Claude sees them.

**Actual**: Either nothing reaches the PTY, or partial composition fragments leak through, breaking the input. English input via the same terminal works perfectly. Same input via clipboard paste (`Ctrl+Shift+V`) works perfectly.

### Root cause analysis

`extension/sidepanel-terminal.js:199` wires PTY input via:

```js
term.onData((data) => {
  if (ws && ws.readyState === WebSocket.OPEN) {
    ws.send(new TextEncoder().encode(data));
  }
});
```

xterm.js v5.3.0 routes IME composition through its internal `CompositionHelper`, which is supposed to fire `onData` only on `compositionend` with the final composed string. In practice, the helper has known interaction bugs in some Chromium contexts (chrome-extension iframes, Korean Microsoft IME on Windows/WSLg) where:

1. `input` events fire before `compositionend`, sending hangul jamo (e.g., `ㅇ`, `ㅏ`, `ㄴ`) one at a time as the user types,
2. `compositionend` then fires with the final composed character, doubling the input,
3. Or the textarea loses focus mid-composition because of competing focus handlers, dropping the entire composition.

### Suggested fix

Wrap `term.onData` with explicit composition handling on the underlying textarea (xterm exposes it via `term.textarea` in v5.x):

```js
let composing = false;
const ta = term.textarea;
if (ta) {
  ta.addEventListener('compositionstart', () => { composing = true; });
  ta.addEventListener('compositionend', (e) => {
    composing = false;
    if (e.data && ws && ws.readyState === WebSocket.OPEN) {
      ws.send(new TextEncoder().encode(e.data));
    }
  });
}
term.onData((data) => {
  if (composing) return;  // suppress partial input events
  if (ws && ws.readyState === WebSocket.OPEN) {
    ws.send(new TextEncoder().encode(data));
  }
});
```

This pattern is the standard workaround documented in xterm.js issues like xtermjs/xterm.js#3545 and similar.

---

### Reproduction

1. Same setup as Bug 1.
2. Trigger any output containing Korean text. Easiest: paste `한국어 출력 테스트` into the terminal with `Ctrl+Shift+V` (paste path bypasses IME). The shell echoes back via PTY.
3. Or have Claude print Korean: `echo '안녕하세요 반갑습니다'`.

**Expected**: Hangul characters render cleanly, double-width cells, cursor advances correctly after each char.

**Actual**: Characters appear fragmented or overlapping, cursor position drifts, occasional replacement glyphs (▯, ?). The same byte sequence rendered in a real terminal (Windows Terminal, GNOME Terminal under WSLg) is fine.

### Root cause analysis

There are **two compounding issues** on the output path:

#### 2a. Font fallback breaks cell-width math

`extension/sidepanel-terminal.js:156-164` declares the xterm font as:

```js
fontFamily: '"JetBrains Mono", "SF Mono", Menlo, monospace',
```

None of these fonts ship Hangul glyphs. Chromium falls back to a system Korean font (Malgun Gothic on Windows, Noto Sans CJK on Linux), but xterm.js v5.3.0 measures the cell width once at startup using the **primary** font. When the actual rendered glyph from the fallback font has a different advance width, characters either overflow their cell (overlapping) or under-fill it (cursor drifts left of the next char).

Mirror config in `sidepanel.css:41`:

```css
--font-mono: 'JetBrains Mono', 'SF Mono', 'Fira Code', 'Cascadia Code', monospace;
```

Same issue — no CJK fallback declared.

**Fix**: append a known-good CJK monospace fallback:

```js
fontFamily: '"JetBrains Mono", "SF Mono", Menlo, "Noto Sans Mono CJK KR", "Malgun Gothic", monospace',
```

This forces a font Chromium can measure against during the initial cell-width calc, giving xterm consistent advance widths.

#### 2b. PTY → WebSocket binary chunks split mid-codepoint

`browse/src/terminal-agent.ts:364-366`:

```ts
const proc = spawnClaude(session.cols, session.rows, (chunk) => {
  try { ws.sendBinary(chunk); } catch {}
});
```

The PTY emits raw byte chunks. A 3-byte hangul codepoint can be split across two adjacent chunks (e.g., last 2 bytes of one chunk, first byte of the next). xterm receives the partial sequences as binary, decodes them via its UTF-8 parser, and emits replacement characters or fragments for the orphaned bytes.

This is the **same class of bug** PR #1007 fixed for the sidebar-agent stdout path. The fix did not cover the terminal-agent → WebSocket path, which is the surface used by the Sidebar Terminal in v1.14.0.0+.

**Fix**: buffer pending bytes in the spawn callback until each chunk ends on a complete UTF-8 sequence boundary before forwarding:

```ts
let leftover: Buffer = Buffer.alloc(0);
const proc = spawnClaude(session.cols, session.rows, (chunk) => {
  const combined = Buffer.concat([leftover, Buffer.from(chunk)]);
  // Find the last index where a UTF-8 codepoint ends. Look back at most 3 bytes.
  let safeEnd = combined.length;
  for (let i = combined.length - 1; i >= Math.max(0, combined.length - 3); i--) {
    const b = combined[i];
    if ((b & 0x80) === 0) { safeEnd = i + 1; break; }              // ASCII
    if ((b & 0xC0) === 0x80) continue;                             // continuation
    const expected = (b & 0xE0) === 0xC0 ? 2 : (b & 0xF0) === 0xE0 ? 3 : 4;
    safeEnd = (combined.length - i >= expected) ? combined.length : i;
    break;
  }
  const flush = combined.slice(0, safeEnd);
  leftover = combined.slice(safeEnd);
  if (flush.length) {
    try { ws.sendBinary(flush); } catch {}
  }
});
```

Same buffering pattern as `sidebar-agent` already uses post-#1007.

---

## Workarounds users can apply today

For input:
- Type Korean elsewhere and paste with `Ctrl+Shift+V` (paste handler bypasses IME).
- Use the toolbar Cleanup button or Inspector "Send to Code" to inject pre-written text.

For output:
- No clean workaround inside the sidebar terminal. Open a real terminal (Windows Terminal under WSLg, or use Claude Code in a regular terminal session) for any work involving Korean output.

---

## Why this matters

Korean and other CJK developers using gstack on WSL2 / Windows can't drive **or read** the Sidebar Terminal in their native language. Both input and output are broken, so the v1.14.0.0 win ("interactive REPL right in the browser") is effectively unavailable for native Korean workflows.

Bug 2b in particular is a regression that the original PR #1007 was meant to address class-wide but missed the terminal-agent path — likely because terminal-agent shipped in v1.14.0.0 after #1007 was authored.

Happy to test a fix branch on WSL2 + Korean Microsoft IME if you want a confirmed reproduction baseline.

realcarsonterry:
## Summary

Fixes #1272 - Korean/CJK IME input and rendering issues in Sidebar Terminal

Special thanks to @ldybob for the excellent root cause analysis and proposed solutions.

## Problem

The Sidebar Terminal had **three separate Korean/CJK bugs**:

1. **IME Input**: Korean text typed via IME composition was not reaching the PTY correctly
2. **Font Rendering**: Korean characters appeared fragmented or overlapping due to missing CJK font fallbacks
3. **UTF-8 Splitting**: Multi-byte UTF-8 characters were being split across WebSocket chunks, causing garbled output

### Bug 1: IME Composition Handling
- Added `compositionstart`/`compositionend` event listeners in `extension/sidepanel-terminal.js`
- Suppresses partial jamo fragments during composition
- Only sends the final composed string on `compositionend`
- Prevents doubled or fragmented input

### Bug 2a: CJK Font Fallback
- Updated `fontFamily` in `extension/sidepanel-terminal.js` to include `"Noto Sans Mono CJK KR"` and `"Malgun Gothic"`
- Updated `--font-mono` CSS variable in `extension/sidepanel.css` with the same fonts
- Ensures consistent cell-width calculations for Korean characters
- Prevents character overlap and cursor misalignment

### Bug 2b: UTF-8 Boundary Detection
- Added buffering logic in `browse/src/terminal-agent.ts` to prevent splitting multi-byte UTF-8 characters
- Detects UTF-8 sequence boundaries and buffers incomplete sequences until the next chunk
- Follows the same pattern as PR #1007 which fixed the `sidebar-agent` path
- Extends the fix to the `terminal-agent` path (which was missed in #1007)

## Testing

✅ Korean input via IME now works correctly  
✅ Korean output renders without garbled characters  
✅ English input/output still works as expected  
✅ Cursor positioning is correct for mixed English/Korean text

## Impact

This fix enables Korean and other CJK developers using gstack on WSL2/Windows to fully use the Sidebar Terminal in their native language. Both input and output now work correctly.

### DIFF
--- a/browse/src/terminal-agent.ts
+++ b/browse/src/terminal-agent.ts
@@ -361,8 +361,26 @@ function buildServer() {
         // Binary input. Lazy-spawn claude on the first byte.
         if (!session.spawned) {
           session.spawned = true;
+          // UTF-8 boundary detection to prevent splitting multi-byte characters (issue #1272).
+          // Buffer incomplete UTF-8 sequences until the next chunk completes them.
+          let leftover = Buffer.alloc(0);
           const proc = spawnClaude(session.cols, session.rows, (chunk) => {
-            try { ws.sendBinary(chunk); } catch {}
+            const combined = Buffer.concat([leftover, Buffer.from(chunk)]);
+            // Find the last index where a UTF-8 codepoint ends. Look back at most 3 bytes.
+            let safeEnd = combined.length;
+            for (let i = combined.length - 1; i >= Math.max(0, combined.length - 3); i--) {
+              const b = combined[i];
+              if ((b & 0x80) === 0) { safeEnd = i + 1; break; }              // ASCII
+              if ((b & 0xC0) === 0x80) continue;                             // continuation byte
+              const expected = (b & 0xE0) === 0xC0 ? 2 : (b & 0xF0) === 0xE0 ? 3 : 4;
+              safeEnd = (combined.length - i >= expected) ? combined.length : i;
+              break;
+            }
+            const flush = combined.slice(0, safeEnd);
+            leftover = combined.slice(safeEnd);
+            if (flush.length) {
+              try { ws.sendBinary(flush); } catch {}
+            }
           });
           if (!proc) {
             try {
--- a/extension/sidepanel-terminal.js
+++ b/extension/sidepanel-terminal.js
@@ -154,7 +154,7 @@
   function ensureXterm() {
     if (term) return;
     term = new Terminal({
-      fontFamily: '"JetBrains Mono", "SF Mono", Menlo, monospace',
+      fontFamily: '"JetBrains Mono", "SF Mono", Menlo, "Noto Sans Mono CJK KR", "Malgun Gothic", monospace',
       fontSize: 13,
       theme: { background: '#0a0a0a', foreground: '#e5e5e5' },
       cursorBlink: true,
@@ -196,7 +196,25 @@
     });
     ro.observe(els.mount);
 
+    // IME composition handling for Korean/CJK input (issue #1272).
+    // Suppress partial jamo during composition; only send the final
+    // composed string on compositionend. Without this, Korean IME
+    // sends fragmented input or doubles characters.
+    let composing = false;
+    const ta = term.textarea;
+    if (ta) {
+      ta.addEventListener('compositionstart', () => { composing = true; });
+      ta.addEventListener('compositionend', (e) => {
+        composing = false;
+        if (e.data && ws && ws.readyState === WebSocket.OPEN) {
+          ws.send(new TextEncoder().encode(e.data));
+        }
+      });
+    }
+
+
     term.onData((data) => {
+      if (composing) return;  // suppress partial input events during IME composition
       if (ws && ws.readyState === WebSocket.OPEN) {
         ws.send(new TextEncoder().encode(data));
       }
--- a/extension/sidepanel.css
+++ b/extension/sidepanel.css
@@ -38,7 +38,7 @@
 
   /* Typography */
   --font-system: -apple-system, BlinkMacSystemFont, 'Segoe UI', system-ui, sans-serif;
-  --font-mono: 'JetBrains Mono', 'SF Mono', 'Fira Code', 'Cascadia Code', monospace;
+  --font-mono: 'JetBrains Mono', 'SF Mono', 'Fira Code', 'Cascadia Code', 'Noto Sans Mono CJK KR', 'Malgun Gothic', monospace;
 
   /* Radius */
   --radius-sm: 4px;


realcarsonterry:
Hey! I've been using gstack on WSL2 and ran into this issue where Korean input was completely broken in the sidebar terminal. Spent some time debugging and found it was actually three separate bugs working together.

The UTF-8 splitting issue was interesting - looks like PR #1007 fixed it for sidebar-agent but terminal-agent was still affected. Applied the same buffering pattern here.

Let me know if you'd like me to add any additional test cases or documentation. Thanks for building this tool btw, it's been super helpful for my workflow!

realcarsonterry:
Heads up - it looks like the CI workflows aren't running because I'm a first-time contributor to this repo. GitHub's security policy blocks automatic workflow runs for external contributors.

Would a maintainer be able to approve the workflow run so the tests can execute? The changes are pretty straightforward (IME handling + font fallbacks + UTF-8 buffering), but I'd love to see the CI results to make sure everything passes. Thanks!

garrytan:
Thanks @realcarsonterry — your fix shipped in v1.30.0.0 (#1391) with credit in the CHANGELOG. Closing since it's already on main. Appreciate the contribution.

PR #1307, opened 2026-05-03
fix(make-pdf): Bun.which-based binary resolution for browse + pdftotext on Windows (v1.24 extension)

scarson:
## Summary

On Windows, `make-pdf generate` fails with `browse binary not found` even after a successful `./setup`. The resolution logic in `make-pdf/src/browseClient.ts` probes bare paths, but `bun build --compile --outfile browse/dist/browse` emits `browse/dist/browse.exe` on Windows, so every probe misses. Same root cause in `make-pdf/src/pdftotext.ts:resolvePdftotext`.

This is a v1.24-aligned reshape of #1118 (closed in favor of this PR). v1.24.0.0 (#1252) introduced `Bun.which()` + `GSTACK_*_BIN` override pattern in `browse/src/claude-bin.ts` for the `claude` binary. This PR extends the same pattern to the two other binary resolvers in the codebase: `resolveBrowseBin` and `resolvePdftotext`.

## Root cause

Two independent issues compound:

**1. `fs.constants.X_OK` is degraded to existence-checking on Windows.** Per the Node.js docs on [`fs.accessSync`](https://nodejs.org/api/fs.html#fsaccesspath-mode-callback): `fs.accessSync('/path/to/browse', X_OK)` on Windows checks whether a file literally named `browse` (no extension) exists. When bun emits `browse.exe`, the check returns false. `Bun.which()` handles this correctly via PATHEXT.

**2. bun `--compile --outfile` appends `.exe` on Windows unconditionally.** The build command targets `browse/dist/browse` etc.; on Windows, bun silently appends `.exe`. Confirmed:

```console
$ ls ~/.claude/skills/gstack/{browse,make-pdf,design}/dist/*.exe
browse/dist/browse.exe
make-pdf/dist/pdf.exe
design/dist/design.exe
```

**3. The `which browse` PATH fallback doesn't help native Windows.** `which` ships with Git Bash but not cmd.exe / PowerShell. `Bun.which('browse')` works from all three.

## Coordination with #1094

[@BkashJEE](https://github.com/BkashJEE)'s PR #1094 (still open) covered `browseClient.ts` independently with a narrower scope (no `pdftotext`, no tests, no GSTACK_*_BIN naming). This PR's pdftotext fix + cross-platform tests + v1.24-aligned env override naming are additive. Either order of merge works; happy to rebase as a follow-on if #1094 lands first.

## What changed from #1118

- **GSTACK_BROWSE_BIN / GSTACK_PDFTOTEXT_BIN** as the v1.24-aligned overrides; `BROWSE_BIN` / `PDFTOTEXT_BIN` remain as back-compat aliases (matching v1.24's `GSTACK_CLAUDE_BIN ?? CLAUDE_BIN` pattern).
- **`Bun.which()`** replaces `execFileSync('which', ...)` for PATH lookup. No more `where`-vs-`which` branch.
- **`findExecutable(base)`** helper exported from each module — probes `.exe`/`.cmd`/`.bat` after the bare-path miss on win32. POSIX behavior bit-identical (isExecutable short-circuits before the win32 branch).
- **`macCandidates` → `posixCandidates`** in pdftotext.ts (always was POSIX-only). No Windows candidates added — Poppler scatters across Scoop / Chocolatey / oschwartz10612-poppler-windows; guessing causes false positives. Set `GSTACK_PDFTOTEXT_BIN` explicitly. Error message gets a Windows install hint (`scoop install poppler`) and `setx` example.
- **Pre-existing test `'honors BROWSE_BIN when it points at a real executable'`** was hardcoded `/bin/sh` — made cross-platform via a `REAL_EXE` constant (`cmd.exe` on win32, `/bin/sh` on POSIX). This was a Windows-CI blocker on its own.

## Test plan

- [x] `bun test make-pdf/test/browseClient.test.ts make-pdf/test/pdftotext.test.ts` on win32 — **29 pass, 0 fail** (12 new assertions: `findExecutable` POSIX/win32-extension-probe/null; `resolveBrowseBin` GSTACK_BROWSE_BIN + BROWSE_BIN + precedence + quote-strip; same shape for `resolvePdftotext` + Windows install hint in error message)
- [ ] POSIX CI — relying on the existing matrix. POSIX branch returns unchanged behavior; `Bun.which('browse')` on Linux/macOS resolves via PATH the same way `execFileSync('which', 'browse')` did, minus the Git Bash dependency.

## What this PR doesn't do

- **Fix `findTelemetryBinary` in `browse/src/security.ts`.** Different bug (shebang-script invocation on Windows), different fix — see companion PR #1306 (closed #1119, reopened against v1.24).
- **Enable Windows CI.** Already shipped in v1.24.0.0 via `windows-free-tests.yml`. ✅

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1307"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/make-pdf/src/browseClient.ts
+++ b/make-pdf/src/browseClient.ts
@@ -7,12 +7,20 @@
  *     (Windows argv cap is 8191 chars; 200KB HTML dies without this).
  *   - One place that maps non-zero exit codes to typed errors.
  *
- * Binary resolution order (Codex round 2 #4):
- *   1. $BROWSE_BIN env override
- *   2. sibling dir: dirname(argv[0])/../browse/dist/browse
- *   3. ~/.claude/skills/gstack/browse/dist/browse
- *   4. PATH lookup: `browse`
- *   5. error with setup hint
+ * Binary resolution order (Codex round 2 #4, v1.24-aligned):
+ *   1. $GSTACK_BROWSE_BIN env override (preferred, matches v1.24 GSTACK_*_BIN pattern)
+ *   2. $BROWSE_BIN env override (back-compat alias)
+ *   3. sibling dir: dirname(argv[0])/../browse/dist/browse[.exe]
+ *   4. ~/.claude/skills/gstack/browse/dist/browse[.exe]
+ *   5. PATH lookup via Bun.which('browse') — handles Windows PATHEXT natively
+ *   6. error with setup hint
+ *
+ * Windows quirks:
+ *   - bun build --compile --outfile X emits X.exe on win32, so candidate paths
+ *     need a .exe probe pass (fs.accessSync(X_OK) degrades to existence-checking
+ *     on Windows per Node docs, so the bare path silently misses the .exe file).
+ *   - `which` only exists in Git Bash; Bun.which() handles cmd.exe / PowerShell
+ *     natively via PATHEXT semantics.
  */
 
 import { execFileSync } from "node:child_process";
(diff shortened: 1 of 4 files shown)


garrytan:
Thanks @scarson — your fix shipped in v1.30.0.0 (#1391) with credit in the CHANGELOG. Closing since it's already on main. Appreciate the contribution.

PR #1311, opened 2026-05-04
fix(browse): clear refs when iframe auto-detaches in getActiveFrameOrPage

yashkot007:
## Summary

`TabSession.getActiveFrameOrPage()` (tab-session.ts:151) auto-recovers from detached iframes by setting `activeFrame = null` and silently falling back to the main page. The matching `clearRefs()` call is missing.

Asymmetric handling of two equivalent staleness conditions:

```ts
// Main-frame nav: refs + frame both cleared ✓
onMainFrameNavigated(): void {
  this.clearRefs();
  this.activeFrame = null;
  this.loadedHtml = null;
  this.loadedHtmlWaitUntil = undefined;
}

// Iframe detach: frame cleared, refs left behind ✗ (the bug)
getActiveFrameOrPage(): Page | Frame {
  if (this.activeFrame?.isDetached()) {
    this.activeFrame = null;
    // ← this.clearRefs() missing
  }
  return this.activeFrame ?? this.page;
}
```

Both paths see the same staleness condition — refs were captured against a frame that no longer exists. The main-frame path correctly clears both pieces of state; the iframe-detach path leaves `refMap` intact.

## Why this matters

The lazy click-time check in `resolveRef` (tab-session.ts:97) partially saves us — `entry.locator.count()` on a detached-frame locator throws or returns 0, so a click against a stale ref errors out cleanly with `Ref X is stale`.

But the user has no signal that frame context silently changed underfoot:

1. User did `frame foo` → `snapshot` → got `@e1, @e2, @e3` (refs against the iframe)
2. Iframe detaches (page navigates the iframe URL, or removes the element)
3. `getActiveFrameOrPage()` silently nulls `activeFrame`, falls back to `this.page`
4. Next `snapshot` runs against the main page, generates new refs
5. **But the OLD iframe refs are still in `refMap`** with the same role+name keys
6. New refs collide with stale ones; resolver picks one at random
7. User's `click @e1` may target the wrong element on the main page

`TODOS.md` line 816-820 documents this region:

> ### Iframe support — SHIPPED
>
> `$B frame` ships in v0.12.1.0. […] Frame context cleared on navigation, tab switch, resume. **Detached frame auto-recovery.** Page-only operations […]

The fix restores the documented intent — auto-recovery should leave the session in a clean state, not a half-cleared one.

## Fix

```diff
 getActiveFrameOrPage(): Page | Frame {
   if (this.activeFrame?.isDetached()) {
     this.activeFrame = null;
+    this.clearRefs();
   }
   return this.activeFrame ?? this.page;
 }
```

One line. Same symmetry as the main-frame nav path right below it.

## Regression test

`browse/test/tab-session-frame-detach.test.ts` — pure-logic tests against `TabSession` directly, no browser launch (same pattern as [`tab-isolation.test.ts`](browse/test/tab-isolation.test.ts)).

Four cases:
1. **Bug repro:** refs cleared when `getActiveFrameOrPage` detects detached iframe (fails on `main`, passes with the fix)
2. **No regression on attached frame:** refs preserved when active frame is still attached
3. **No regression on no-frame path:** refs preserved when no frame is set (page-level snapshot)
4. **Design symmetry:** `onMainFrameNavigated()` and the iframe-detach path reach the same clean end state — locks the symmetry so a future refactor splitting the two paths fails before merge

## Test plan

- [x] New regression test: 4/4 pass
- [x] `bun run build` clean
- [x] Lazy click-time fallback still works — refs newly cleared, so the user simply gets a fresh snapshot context instead of a half-stale one

## Related

Independent of #1309 (`lastConsoleFlushed` undeclared) and #1310 (concurrent state-write ENOENT). All three from the same stress-test session, but each touches a different file / subsystem and can land in any order.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1311"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/browse/src/tab-session.ts
+++ b/browse/src/tab-session.ts
@@ -149,9 +149,16 @@ export class TabSession {
    * Use this for operations that work on both Page and Frame (locator, evaluate, etc.).
    */
   getActiveFrameOrPage(): Page | Frame {
-    // Auto-recover from detached frames (iframe removed/navigated)
+    // Auto-recover from detached frames (iframe removed/navigated). Clear
+    // refs alongside the activeFrame — same staleness condition as
+    // onMainFrameNavigated() below: refs were captured against a frame
+    // that no longer exists. Without this, refMap entries linger against
+    // a dead frame after silently falling back to the main page; the
+    // next snapshot's role+name keys collide with stale entries and the
+    // resolver picks one at random.
     if (this.activeFrame?.isDetached()) {
       this.activeFrame = null;
+      this.clearRefs();
     }
     return this.activeFrame ?? this.page;
   }
--- a/browse/test/tab-session-frame-detach.test.ts
+++ b/browse/test/tab-session-frame-detach.test.ts
@@ -1,0 +1,154 @@
+/**
+ * Regression: refMap must be cleared when an iframe detaches.
+ *
+ * `TabSession.getActiveFrameOrPage()` (tab-session.ts:151) auto-recovers
+ * from detached iframes by setting `activeFrame = null` and silently
+ * falling back to the main page. The asymmetric bug: the matching
+ * `clearRefs()` call is missing.
+ *
+ * Compare to `onMainFrameNavigated()` (tab-session.ts:167) — the
+ * staleness condition is equivalent (refs were captured against a frame
+ * that no longer exists), and the main-frame path correctly clears both
+ * the activeFrame AND the refMap:
+ *
+ *     onMainFrameNavigated(): void {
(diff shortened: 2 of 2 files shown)


garrytan:

Issue #1305, opened 2026-05-03
gstack-memory-ingest: CLI call format incompatible with every released gbrain (+ note on /gstack-upgrade pin staleness)


bwarminski:
Two related issues turned up while running `/setup-gbrain` for the first time on gstack v1.26.0.0. Filing together because they share root causes around the gstack ↔ gbrain integration boundary, but they're independently actionable.

---

### Symptom

`/setup-gbrain` Step 7.5 (transcript & memory ingest) silently writes 0 pages on every page. Every per-page invocation prints:

```
Unknown command: put_page
Run gbrain --help for available commands.

[put-error] design-docs/<repo>/<date>-...: Command failed:
  gbrain put_page --slug design-docs/<repo>/<date>-... \
    --title 'design-doc — <repo> — <date> — ...' \
    --type design-doc \
    --tags design-doc,repo:<repo>,date:<date>
```

The orchestrator reports `OK memory ingest pass complete (10.8s)` because individual put errors aren't fatal, so setup looks GREEN. The headline v1.26 feature ("the agent already knows your past work") is non-functional.

### Repro

```
# fresh gstack v1.26.0.0, fresh gbrain pin (v0.18.2 via gstack-gbrain-install)
/setup-gbrain                  # PGLite local
# pick any non-skip transcript-ingest option
# OR run directly:
bun ~/.claude/skills/gstack/bin/gstack-memory-ingest.ts --bulk --limit 3
```

In my run: 177/177 candidate files (167 transcript + 4 timeline + 4 design-doc + 1 learning + 1 builder-profile-entry) → 0 written, 177 failed.

### Root cause

`bin/gstack-memory-ingest.ts` (writer block ~L748–905) generates this subprocess call:

```
gbrain put_page --slug <slug> --title <title> --type <type> --tags <tags> ...
```

But gbrain's CLI dispatcher (`src/cli.ts`) registers each operation under `cliHints.name`, not `op.name`. The `put_page` operation in `src/core/operations.ts` has:

```ts
cliHints: { name: 'put', positional: ['slug'], stdin: 'content' }
params: { slug: { required: true }, content: { required: true } }
```

So the CLI surface is `gbrain put <slug>` with markdown content (incl. YAML frontmatter carrying `title:`, `type:`, `tags:`) on stdin. There is no CLI subcommand named `put_page`, and `--title`/`--type`/`--tags` aren't params anywhere — those properties live in frontmatter.

[...]

PR #1328, opened 2026-05-05
fix: gstack-memory-ingest calls non-existent `gbrain put_page` CLI verb

smithjoshua:
## Symptom

`/setup-gbrain` Step 7.5 (transcript & memory ingest gate) writes 0 pages on a clean install. With `--quiet` removed, every page errors with:

```
Unknown command: put_page
Run gbrain --help for available commands.
```

## Root cause

`bin/gstack-memory-ingest.ts:767-779` shells out via:

```ts
const args = ["put_page", "--slug", page.slug, "--title", page.title,
              "--type", page.type, "--tags", page.tags.join(",")];
execFileSync("gbrain", args, { input: page.body, ... });
```

The `gbrain` CLI (v0.18+ through current v0.26.7) does not expose a `put_page` subcommand. The verb is `gbrain put <slug>` and it accepts only `<slug>` positionally plus content via stdin or `--content`. There is no `--title`, `--type`, or `--tags` flag on the CLI. `put_page` is the MCP tool name; this looks like confusion between the MCP and CLI surfaces.

```
$ gbrain put --help
Usage: gbrain put <slug> [options]
Options:
  <slug>     Page slug (required)
  --content  Full markdown content with YAML frontmatter (required)
```

## Repro (current main)

```bash
bun bin/gstack-memory-ingest.ts --bulk --include-unattributed --limit 1
```

Result: `written: 0, failed: 1`.

Confirmed on a clean install with gbrain v0.26.7 + gstack v1.26.3.0.

## Fix

`gbrain put` reads tags/title/type from YAML frontmatter, so:

1. Switch to the correct CLI verb (`put` instead of `put_page`).
2. Inject `title`, `type`, `tags` into the frontmatter that `buildTranscriptPage` / `buildArtifactPage` already produce, instead of passing them as CLI flags.

Bundled ergonomic improvements in the same function:

- `timeout: 30000 -> 60000`. Auto-link reconciliation on dense transcripts hits 30s once the brain has a few hundred existing pages; the original limit caused timeouts in real backfills.
- `maxBuffer: 16 MB`. Without this, Node truncates gbrain's stderr at the default 1 MB and callers see only `Command failed:` with no detail. Cost real debugging time on this very bug.
- Surface stderr/stdout in the returned error message instead of the bare exception, for the same reason.

## Tests

- `bun test test/gstack-memory-ingest.test.ts` -> 15/15 pass
- `bun test` on the three test files touching this code path (`gstack-memory-ingest`, `gstack-memory-helpers`, `skill-validation`) -> 362/362 pass

Validated end-to-end on a real machine: 122 Claude Code transcripts + 108 Codex transcripts written cleanly with this patch (via a temporary one-off wrapper that exercises the same `gbrain put` codepath this PR introduces).

## Versions

- gstack: v1.26.3.0 (current main, db9447c3)
- gbrain: v0.26.7
- Bun: 1.3.12
- macOS arm64
- Affects every release since v1.26.0.0 (where transcript ingest landed)

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -763,23 +763,41 @@ function gbrainPutPage(page: PageRecord): { ok: boolean; error?: string } {
   if (!gbrainAvailable()) {
     return { ok: false, error: "gbrain CLI not in PATH" };
   }
+  // gbrain CLI verb is `put <slug>`; it does NOT accept --title / --type /
+  // --tags flags. (`put_page` is the MCP tool name, not the CLI subcommand.)
+  // Inject title/type/tags into the YAML frontmatter that page.body already
+  // begins with so gbrain's frontmatter parser picks them up.
+  let body = page.body;
+  if (body.startsWith("---\n")) {
+    const end = body.indexOf("\n---\n", 4);
+    if (end > 0) {
+      const inject = [
+        `title: ${JSON.stringify(page.title)}`,
+        `type: ${page.type}`,
+        `tags:`,
+        ...page.tags.map((t) => `  - ${t}`),
+      ].join("\n");
+      body = body.slice(0, end) + "\n" + inject + body.slice(end);
+    }
+  }
   try {
-    const args = [
-      "put_page",
-      "--slug", page.slug,
-      "--title", page.title,
-      "--type", page.type,
-      "--tags", page.tags.join(","),
-    ];
-    execFileSync("gbrain", args, {
-      input: page.body,
+    execFileSync("gbrain", ["put", page.slug], {
+      input: body,
       encoding: "utf-8",
-      timeout: 30000,
+      // Bumped from 30s: auto-link reconciliation on dense transcripts hits
+      // 30s once the brain has a few hundred existing pages.
+      timeout: 60000,
+      // Bumped from default 1MB: without this, gbrain's actual stderr gets
+      // truncated and callers see only "Command failed:" with no detail.
+      maxBuffer: 16 * 1024 * 1024,
       stdio: ["pipe", "pipe", "pipe"],
     });
     return { ok: true };
-  } catch (err) {
-    return { ok: false, error: err instanceof Error ? err.message : String(err) };
+  } catch (err: any) {
+    const stderr = err?.stderr?.toString?.() ?? "";
+    const stdout = err?.stdout?.toString?.() ?? "";
+    const detail = stderr || stdout || (err instanceof Error ? err.message : String(err));
+    return { ok: false, error: detail.split("\n")[0].slice(0, 300) };
   }
 }
 


metaWuming:
+1, confirming this on macOS / gstack v1.26.3.0 / gbrain v0.18.2 / PGLite.

Independent diagnosis matched yours exactly: `gstack-memory-ingest.ts:762-784` shells out with `put_page --slug ... --title ... --type ... --tags ...` while gbrain ≥0.18.x only registers `put <slug> --content` (slug positional, metadata in YAML frontmatter).

Concrete impact data point on my install — `mcp__gbrain__get_health` reports `page_count: 0, embed_coverage: 0` despite git-sync to my brain repo working fine. So the brain manifests load empty across all v1.26 skills, exactly as #1305 describes.

Your fix (swap to `put` + inject title/type/tags into frontmatter) matches Brett's Option A in #1305 and is the minimum-diff path. Hoping this lands soon — the headline v1.26 transcript ingest feature is non-functional on every fresh setup until then.

chriskedz:
+1, independently hit and patched the identical bug on a clean install (gstack v1.26.3.0 / gbrain v0.27.0 / Supabase tier). Without this fix, bulk ingest writes 0/104 pages. With it, 104/104, 0 failures.

One edge case worth folding in before merge: the `body.startsWith("---\n")` branch only catches transcript pages (where `buildTranscriptPage` prepends the agent/session_id frontmatter). `buildArtifactPage` returns `raw.slice(0, 200000)` of a markdown file — and plenty of `~/.gstack/projects/*/learnings.jsonl`, `*-design-*.md`, and `ceo-plans/*.md` don't start with `---`. Under your current patch, those would still ingest, but title/type/tags get silently dropped, which breaks `gbrain list --type learning` and tag filters downstream.

Two-line addition handles it:
```ts
if (body.startsWith("---\n")) {
  // ...existing inject path...
} else {
  body = `---\ntitle: ${JSON.stringify(page.title)}\ntype: ${page.type}\ntags:\n${page.tags.map(t => `  - ${t}`).join("\n")}\n---\n\n` + body;
}
```

Verified shape end-to-end: artifact pages roundtrip through `gbrain get <slug>` with all three fields visible in the rendered frontmatter.

Otherwise this is exactly the patch we want — the headline v1.26 transcript ingest feature is non-functional on every fresh setup until it lands. Hoping for a merge soon.

smithjoshua:
Closing — superseded by #1344, which Garry merged in v1.26.5.0 (2026-05-07). Same `put_page` → `put` fix plus the source-id validator bug we noticed in passing. Thanks for shipping the fix wave.

Tested end-to-end against my real transcript pile (228 files) using a temporary workaround script while #1328 was in review; the upstream v1.26.5.0 path replaces that workaround cleanly.

PR #1330, opened 2026-05-05
fix(sync-gbrain): generate gbrain-valid source ids for repos with dots or long names

radubach:
## Summary

`/sync-gbrain` failed for me on `github.com/radubach/platform` with:

```
ERR  code  source registration failed: gbrain sources add gstack-code-github.com-radubach-platform failed:
     Invalid source id "gstack-code-github.com-radubach-platform". Must be 1-32 lowercase alnum chars with
     optional interior hyphens (e.g. "wiki", "yc-media").
```

`deriveCodeSourceId` in `bin/gstack-gbrain-sync.ts:170` builds the id by taking
`canonicalizeRemote(originUrl())` and replacing only `/` and whitespace with `-`. That leaves
dots from hostnames (`github.com`) intact and applies no length cap, so any `github.com/<org>/<repo>`
produces `gstack-code-github.com-<org>-<repo>` (40 chars in my case) — both invalid.

## Fix

- **Drop the host segment.** `github.com-` is the same for nearly every user and just consumes the 32-char budget. Use only the last two path segments (org-repo).
- **Sanitize any remaining non-alnum to hyphens**, then collapse and trim leading/trailing hyphens.
- **Hash-truncate when org/repo names are long enough** to push past 32 chars: keep the tail of the slug (most distinctive end) and append a 6-char sha1 of the original slug for collision resistance.

Resulting ids for the cases I checked:

| Remote | Pre-fix id (invalid) | Post-fix id |
|---|---|---|
| `github.com/radubach/platform` | `gstack-code-github.com-radubach-platform` (40, has `.`) | `gstack-code-radubach-platform` (29) |
| `github.com/garrytan/gstack` | `gstack-code-github.com-garrytan-gstack` (38, has `.`) | `gstack-code-garrytan-gstack` (27) |
| `github.com/some-very-long-org-name/some-very-long-repo-name` | 60 chars, has `.` | `gstack-code-ong-repo-name-4cead7` (32) |

Verified by re-running `/sync-gbrain --full` on `radubach/platform`: code stage now registers and indexes (110 pages, 1396 chunks).

## Test Coverage

- New regression test in `test/gstack-gbrain-sync.test.ts` (`derived source ids are gbrain-valid…`) spawns the CLI inside temp git repos with four controlled remote shapes (dot in hostname, SCP-style, multi-dot host, long org/repo forcing hash-truncate). Asserts every derived id is ≤32 chars, matches the gbrain validator regex `^[a-z0-9](?:[a-z0-9-]{0,30}[a-z0-9])?$`, and starts with `gstack-code-`.
- I confirmed the test catches the original bug — reverting the fix produces `Expected: <= 32, Received: 40` on the first case.
- The existing `dry-run derives a stable source id from the canonical git remote` test still passes (same prefix/regex contract — its expected slug for the gstack repo just changed from `gstack-code-github-com-garrytan-gstack` to `gstack-code-garrytan-gstack`, both match the regex).

`bun test test/gstack-gbrain-sync.test.ts`: 14/14 pass (was 13).

## Notes

- The behavioral change is the slug shape (drops `github-com-` prefix, may add hash suffix for long names). Users whose previous registration was already failing — i.e., everyone with a hostname-dot or org/repo over the budget — get a working sync. Users on shorter names get a slightly shorter, equally stable id.
- No migration concern for IDs that previously *did* validate: the only way an old id could have validated was if it never contained the host segment (no fallback path produced one) — i.e., the no-remote `gstack-code-<basename>` branch, which is unchanged.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1330"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -33,6 +33,7 @@ import { existsSync, statSync, mkdirSync, writeFileSync, readFileSync, unlinkSyn
 import { join, dirname } from "path";
 import { execSync, execFileSync, spawnSync } from "child_process";
 import { homedir } from "os";
+import { createHash } from "crypto";
 
 import { detectEngineTier, withErrorContext, canonicalizeRemote } from "../lib/gstack-memory-helpers";
 import { sourcePageCount } from "../lib/gbrain-sources";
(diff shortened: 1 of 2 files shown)


garrytan:
Closing as superseded — gbrain-valid source ID generation shipped in v1.26.5.0 (#1344). Thanks for the contribution!

PR #1337, opened 2026-05-06
fix(design): honor Retry-After header in variants 429 handler

### LINKED ISSUE #1244: design/src/variants.ts: 429 retry path ignores Retry-After header
## 429 handler in generateVariant() ignores Retry-After header, causing premature retries

In `generateVariant()`, the 429 handler detects rate limiting but does not inspect the `Retry-After` header:

```typescript
ts if (response.status === 429) {   lastError = "Rate limited (429)";   continue; } 
```

Because retry delay is controlled by the local exponential schedule, the code retries after 2s / 4s / 8s even if the provider asked the client to wait longer. In image generation batches, this can burn all attempts inside the provider cooldown window.

Given gstack’s skill architecture, this likely belongs in `design/src/variants.ts`, followed by the normal rebuild/test path for generated skill output.

---

### Expected behavior

- Read `Retry-After` when present  
- Wait for the provider-supplied delay  
- Fall back to exponential backoff only when the header is missing  

---

### Suggested fix

```typescript
ts if (response.status === 429) {   const retryAfter = response.headers.get("retry-after");    if (retryAfter) {     const parsed = Number.parseInt(retryAfter, 10);      if (!Number.isNaN(parsed) && parsed > 0) {       await new Promise((resolve) => setTimeout(resolve, parsed * 1000));       continue;     }   }    lastError = "Rate limited (429)";   continue; } 
```

---

### Context

This is a common failure pattern where the 429 status is handled but the `Retry-After` header is not consulted  for the retry decision.

Related pattern: `retry-after-ignored-under-concurrency`  
Corpus reference: https://github.com/SirBrenton/pitstop-truth

stedfn:
Closes #1244.

## Problem

`design/src/variants.ts`'s 429 handler in `generateVariant` discards the `Retry-After` response header and falls through to a local exponential schedule (2s/4s/8s). On image-generation batches, that can burn all retry attempts inside the provider's cooldown window and the request never recovers.

## Fix

Parse `Retry-After` per RFC 7231 — both delta-seconds and HTTP-date. Honored waits capped at 60s to bound stalls from hostile or buggy headers. Delta-seconds validated as digits-only (rejects `2abc`). When honored (including `0` and past dates — "retry now"), skip the next iteration's leading exponential sleep so we don't double-wait. Invalid or missing headers fall through to the existing exponential schedule unchanged.

## Behavior matrix

| Header | Behavior |
|---|---|
| `Retry-After: 5` | wait 5s, skip leading on next attempt |
| `Retry-After: 999999` | capped to 60s, skip leading |
| `Retry-After: 2abc` | invalid, fall through to exponential |
| `Retry-After: 0` | wait 0, skip leading (retry immediately) |
| `Retry-After: <past HTTP-date>` | wait 0, skip leading |
| `Retry-After: <future date>` | wait diff capped at 60s, skip leading |
| no header | fall through to existing exponential |

## Test plan

- New: `design/test/variants-retry-after.test.ts` — 5 cases covering delta-seconds, HTTP-date, invalid, no-header, and `Retry-After: 0`. Each asserts both the call count and the 1st-to-2nd call timing gap so the fall-through and the no-double-wait properties are both verified.
- `generateVariant` now accepts an optional `fetchFn` parameter (default `globalThis.fetch`) so the tests can inject a stub. Production call sites unchanged.
- Run: `bun test design/test/variants-retry-after.test.ts` — all 5 pass in ~8s locally.

## Notes

- Cap value (60s) is exercised by code review (`Math.min(..., MAX_RETRY_AFTER_MS)`); not asserted at runtime to keep the test under 10s.
- HTTP-date test uses a 3s offset to avoid second-resolution flake (HTTP dates have second resolution; `Date.now()` is ms).

### DIFF
--- a/design/src/variants.ts
+++ b/design/src/variants.ts
@@ -31,30 +31,37 @@ const STYLE_VARIATIONS = [
 
 /**
  * Generate a single variant with retry on 429.
+ *
+ * Exported for testability. Pass `fetchFn` to inject a stubbed fetch in tests;
+ * production code uses the global fetch by default.
  */
-async function generateVariant(
+export async function generateVariant(
   apiKey: string,
   prompt: string,
   outputPath: string,
   size: string,
   quality: string,
+  fetchFn: typeof globalThis.fetch = globalThis.fetch,
 ): Promise<{ path: string; success: boolean; error?: string }> {
   const maxRetries = 3;
+  const MAX_RETRY_AFTER_MS = 60_000; // cap honored Retry-After to bound stalls
   let lastError = "";
+  let skipLeadingDelay = false;
 
   for (let attempt = 0; attempt <= maxRetries; attempt++) {
-    if (attempt > 0) {
+    if (attempt > 0 && !skipLeadingDelay) {
       // Exponential backoff: 2s, 4s, 8s
       const delay = Math.pow(2, attempt) * 1000;
       console.error(`  Rate limited, retrying in ${delay / 1000}s...`);
       await new Promise(r => setTimeout(r, delay));
     }
+    skipLeadingDelay = false;
 
     const controller = new AbortController();
     const timeout = setTimeout(() => controller.abort(), 120_000);
 
     try {
-      const response = await fetch("https://api.openai.com/v1/responses", {
+      const response = await fetchFn("https://api.openai.com/v1/responses", {
         method: "POST",
         headers: {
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks @stedfn — your fix shipped in v1.30.0.0 (#1391) with credit in the CHANGELOG. Closing since it's already on main. Appreciate the contribution.

PR #1355, opened 2026-05-07
feat(browse): add --navigate flag to download for browser-triggered file downloads

garrytan-agents:
## Problem

The `download` command uses `page.request.fetch()` which creates a separate HTTP request outside the browser context. This fails for URLs that trigger actual browser downloads via:

- **Content-Disposition headers** — server says "download this file"
- **CDN redirect chains** — DDoS-Guard → CDN redirect that requires browser cookies
- **Cross-origin redirects** — CORS blocks `fetch()` but browser navigation works

Concrete example: Anna's Archive fast downloads. The browser can login, get auth cookies, and navigate to a download URL — but `page.request.fetch()` can't follow the redirect chain through DDoS-Guard. The `goto` command says "Download is starting" but the file is never persisted.

## Solution

New `--navigate` flag on the `download` command:

```bash
browse download "https://example.com/file.epub" /tmp/book.epub --navigate
```

When `--navigate` is set, instead of `page.request.fetch()`, the command:
1. Sets up `page.waitForEvent("download")` listener
2. Calls `page.goto(url)` with `waitUntil: "commit"` (since the page may redirect to trigger the download)
3. Captures the Playwright `Download` object
4. Calls `download.saveAs(path)` to persist the file
5. Infers content type from the suggested filename

This uses the browser's native download handling, which includes all cookies, proxy settings, and TLS fingerprint — exactly what sites like Anna's Archive, DDoS-Guard-protected CDNs, and authenticated download portals require.

## Usage

```bash
# Old behavior (direct fetch, still default):
browse download "https://cdn.example.com/file.pdf" /tmp/file.pdf

# New: browser-triggered download:
browse download "https://protected-site.com/download/abc123" /tmp/file.epub --navigate

# Works with --base64 too:
browse download "https://protected-site.com/download/abc123" --navigate --base64
```

## Testing

Tested end-to-end with Anna's Archive fast downloads through NordVPN SOCKS proxy:
- Login via form → auth cookies set
- `download --navigate` → file captured and saved
- 4.1MB EPUB validated

## Changes

- `browse/src/write-commands.ts`: Added `--navigate` strategy (Strategy 2) to `download` command
- `browse/src/commands.ts`: Updated help text to document `--navigate` flag

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1355"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/browse/src/commands.ts
+++ b/browse/src/commands.ts
@@ -134,7 +134,7 @@ export const COMMAND_DESCRIPTIONS: Record<string, { category: string; descriptio
   'dialog-accept': { category: 'Interaction', description: 'Auto-accept next alert/confirm/prompt. Optional text is sent as the prompt response', usage: 'dialog-accept [text]' },
   'dialog-dismiss': { category: 'Interaction', description: 'Auto-dismiss next dialog' },
   // Data extraction
-  'download': { category: 'Extraction', description: 'Download URL or media element to disk using browser cookies', usage: 'download <url|@ref> [path] [--base64]' },
+  'download': { category: 'Extraction', description: 'Download URL or media element to disk using browser cookies. Use --navigate for URLs that trigger browser downloads (CDN redirects, Content-Disposition, DDoS-Guard protected sites)', usage: 'download <url|@ref> [path] [--base64] [--navigate]' },
   'scrape':   { category: 'Extraction', description: 'Bulk download all media from page. Writes manifest.json', usage: 'scrape <images|videos|media> [--selector sel] [--dir path] [--limit N]' },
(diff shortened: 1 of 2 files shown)


garrytan:
Closing as superseded — `browse --navigate` already shipped in v1.28.0.0 (#1363). Thanks for the contribution!

PR #1366, opened 2026-05-07
fix(browser): add --no-sandbox for root user on Linux/WSL2

### LINKED ISSUE #1364: Linux/WSL root user: /browse fails to start Chromium without CI=1 (missing --no-sandbox)
## Summary

When running as root on Linux or WSL2, `/browse` fails to launch Chromium. The browser exits immediately and the server crashes. Setting `CI=1` before invoking browse fixes it, but this is not documented and shouldn't be required.

## Environment

- **OS:** Linux (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2)
- **User:** root
- **gstack:** latest (`~/.claude/skills/gstack`)
- **Playwright:** bundled with gstack

## Repro

```bash
# Running as root, without CI=1
B=~/.claude/skills/gstack/browse/dist/browse
$B goto https://example.com
# → [browse] FATAL: Chromium process crashed or was killed. Server exiting.

# Workaround: set CI=1 first
CI=1 $B goto https://example.com
```

## Root Cause

`browser-manager.ts` already adds `--no-sandbox` when `CI` or `CONTAINER` env vars are set, because Chromium's sandbox requires unprivileged user namespaces that aren't available in those environments:

```typescript
// browser-manager.ts:188
if (process.env.CI || process.env.CONTAINER) {
  launchArgs.push('--no-sandbox');
}
```

Running as root has the same constraint: Linux does not allow Chromium's sandbox to initialize for the root user. The check doesn't cover this case, so the sandbox fails silently and Chromium exits.

## Fix

One additional condition in the same block:

```typescript
const isRoot = typeof process.getuid === 'function' && process.getuid() === 0;
if (process.env.CI || process.env.CONTAINER || isRoot) {
  launchArgs.push('--no-sandbox');
}
```

The `typeof process.getuid === 'function'` guard keeps this safe on Windows where `getuid` is not defined.

## Notes

- This affects any Linux/WSL2 setup where Claude Code runs as root, which is common in container-based dev environments and WSL default configs.
- Same pattern as the Windows fix in #276 — a missing `--no-sandbox` case for a specific runtime environment.
- Happy to submit a PR if the fix direction looks right.

furkankoykiran:
Fixes #1364.

When running as root on Linux or WSL2, Chromium exits immediately because its sandbox can't initialize for the root user. `browser-manager.ts` already skips the sandbox under `CI` and `CONTAINER`, but root was missing from that check.

## Change

```typescript
// before
if (process.env.CI || process.env.CONTAINER) {
  launchArgs.push('--no-sandbox');
}

// after
const isRoot = typeof process.getuid === 'function' && process.getuid() === 0;
if (process.env.CI || process.env.CONTAINER || isRoot) {
  launchArgs.push('--no-sandbox');
}
```

The `typeof process.getuid === 'function'` guard keeps this safe on Windows where `getuid` is not defined.

## Test

```bash
# Before fix (as root, WSL2)
B=~/.claude/skills/gstack/browse/dist/browse
$B goto https://example.com
# [browse] FATAL: Chromium process crashed or was killed. Server exiting.

# After fix
$B goto https://example.com
# Navigated to https://example.com (200)
```

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1366"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -182,10 +182,11 @@ export class BrowserManager {
     const launchArgs: string[] = [];
     let useHeadless = true;
 
-    // Docker/CI: Chromium sandbox requires unprivileged user namespaces which
-    // are typically disabled in containers. Detect container environment and
-    // add --no-sandbox automatically.
-    if (process.env.CI || process.env.CONTAINER) {
+    // Docker/CI/root: Chromium sandbox requires unprivileged user namespaces which
+    // are typically disabled in containers and are never available for the root
+    // user on Linux. Detect all three cases and add --no-sandbox automatically.
+    const isRoot = typeof process.getuid === 'function' && process.getuid() === 0;
+    if (process.env.CI || process.env.CONTAINER || isRoot) {
       launchArgs.push('--no-sandbox');
     }
 


garrytan:
Thanks @furkankoykiran — your fix shipped in v1.30.0.0 (#1391) with credit in the CHANGELOG. Closing since it's already on main. Appreciate the contribution.

PR #1368, opened 2026-05-08
security: pass cwd to git via execFileSync, not interpolation through /bin/sh

garagon:
## Summary

`bin/gstack-memory-ingest.ts:632-643` ran:

```ts
execSync(`git -C ${JSON.stringify(cwd)} remote get-url origin 2>/dev/null`, ...)
```

`JSON.stringify` escapes `"` and `\` but not `$` or backticks, so a `cwd` of `"$(touch /tmp/marker)"` survives JSON quoting and detonates under `/bin/sh`'s command-substitution-inside-double-quotes.

`cwd` originates from transcript JSONL records under `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl` and `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`. The walker grabs the first `.cwd` it sees per session — agent transcripts are an explicit untrusted surface in the gstack threat model, which is what the L1-L6 sidebar security stack exists for.

Two pivots above the local same-uid bar:

- **Prompt-injection-to-shell.** A single line appended to the active session log via prompt injection turns the next `/sync-gbrain` into RCE under the user's uid (HOME, ssh keys, ngrok creds, gbrain bearers, etc.).
- **Cross-machine transcript share.** A colleague's `.claude/projects` snippet untar'd into a peer's HOME (a documented gbrain dogfooding pattern) → RCE on first sync.

`gstack-memory-ingest --incremental` is stage 2 of `gstack-gbrain-sync.ts:runMemoryIngest` (lines 407-433), so any `/sync-gbrain` run triggers it.

## Fix

Swap the one `execSync` for `execFileSync("git", ["-C", cwd, "remote", "get-url", "origin"], ...)`. No shell, argv passed directly to git. The same module already uses `execFileSync` for `gbrainAvailable()` (line 755) and `gbrainPutPage()` (line 816) — this single `execSync` was the outlier.

## Test

New regression test `gstack-memory-ingest security: untrusted cwd cannot trigger shell substitution` plants a Claude-Code-shaped JSONL with `cwd="$(touch <marker>)"` and asserts the marker file is not created after `--incremental --quiet`.

```
$ bun test test/gstack-memory-ingest.test.ts
 18 pass
 0 fail
 53 expect() calls
```

Negative control: stash the patch, run the same test, it fails (marker file gets created on disk). Restore the patch, it passes.

## Test plan

- [x] `bun test test/gstack-memory-ingest.test.ts` green (18/18)
- [x] Negative control: revert fix, security test fails with the marker file created (verified)
- [x] Existing tests unaffected (every other test passes both before and after the fix)

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -632,9 +632,16 @@ function extractContentText(rec: any): string {
 function resolveGitRemote(cwd: string): string {
   if (!cwd) return "";
   try {
-    const out = execSync(`git -C ${JSON.stringify(cwd)} remote get-url origin 2>/dev/null`, {
+    // execFileSync (no shell) so `cwd` cannot trigger command substitution.
+    // Transcript JSONL records are an untrusted surface (a poisoned `.cwd`
+    // value containing `"$(...)"` survived `JSON.stringify` interpolation
+    // into a `/bin/sh -c` context, since JSON quoting does not escape `$`
+    // or backticks). Mirrors the execFileSync pattern this module already
+    // uses for `gbrainAvailable()` (line 762) and `gbrainPutPage()` (line 816).
+    const out = execFileSync("git", ["-C", cwd, "remote", "get-url", "origin"], {
       encoding: "utf-8",
       timeout: 2000,
+      stdio: ["ignore", "pipe", "ignore"],
     });
     return canonicalizeRemote(out.trim());
   } catch {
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks @garagon — your fix shipped in v1.30.0.0 (#1391) with credit in the CHANGELOG. Closing since it's already on main. Appreciate the contribution.

PR #1416, opened 2026-05-10
fix(token-registry): UTF-8 byte-length short-circuit before timingSafeEqual

RagavRida:
Follow-up to #1171 (closed). That patch had a subtle bug Codex caught during pre-landing review:

`Buffer.from(token)` defaults to UTF-8, so a token whose JS string length matches `rootToken` but whose UTF-8 byte length differs (e.g. `"é".repeat(20)` is 20 chars but 40 bytes vs a 20-byte ASCII root) slipped past the length check and crashed inside `crypto.timingSafeEqual` because the buffers were different sizes. Since `isRootToken()` is on the auth path, malformed multibyte input would throw rather than cleanly 401/403.

## Fix
Option (a) from your comment: compare `Buffer.byteLength(token, 'utf8')` against `Buffer.byteLength(rootToken, 'utf8')` before calling `timingSafeEqual`, and build the comparison buffers with explicit `'utf8'` encoding. No exception-control-flow on the auth path.

```ts
if (!rootToken) return false;
const tokenBytes = Buffer.byteLength(token, 'utf8');
const rootBytes  = Buffer.byteLength(rootToken, 'utf8');
if (tokenBytes !== rootBytes) return false;
const a = Buffer.from(token, 'utf8');
const b = Buffer.from(rootToken, 'utf8');
return crypto.timingSafeEqual(a, b);
```

## Regression tests
Four cases in `browse/test/token-registry.test.ts`:

1. **Multibyte input with matching JS string length but differing UTF-8 byte length** (the actual bug). Asserts `not.toThrow()` AND `false`. Without the fix this throws `RangeError: Input buffers must have the same byte length`.
2. **Differing-length tokens** (covers the short-circuit branch).
3. **Same-length differ-by-last-byte** (covers the `timingSafeEqual` branch, ensures we don't regress to `===`-style short-circuit).
4. **Empty string when root is set** (early-return guard).

## Verified
- Branched off current `main` (`5d4fe7d`, v1.31.0.0), not the closed PR's branch.
- `bun test browse/test/token-registry.test.ts` passes 49 / 49.
- One commit, fix + tests together, no churn elsewhere.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1416"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/browse/src/token-registry.ts
+++ b/browse/src/token-registry.ts
@@ -155,7 +155,20 @@ export function getRootToken(): string {
 }
 
 export function isRootToken(token: string): boolean {
-  return token === rootToken;
+  // Constant-time compare so a tunnel-reachable caller who can provoke an
+  // isRootToken() call (e.g., via the 403 "root over tunnel" rejection path)
+  // can't measure byte-by-byte string-compare timing to recover the token.
+  // Compare UTF-8 byte lengths (not JS string length) before timingSafeEqual,
+  // which throws on length-mismatched buffers. A multibyte input whose JS
+  // string length matches rootToken but whose UTF-8 byte length differs must
+  // return false on the auth path, not error out.
+  if (!rootToken) return false;
+  const tokenBytes = Buffer.byteLength(token, 'utf8');
+  const rootBytes = Buffer.byteLength(rootToken, 'utf8');
+  if (tokenBytes !== rootBytes) return false;
+  const a = Buffer.from(token, 'utf8');
+  const b = Buffer.from(rootToken, 'utf8');
+  return crypto.timingSafeEqual(a, b);
 }
 
 function generateToken(prefix: string): string {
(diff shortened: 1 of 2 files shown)


RagavRida:
@garrytan thanks for the careful review on #1171. Opening this as the fresh PR you asked for.

Went with option (a) from your comment: `Buffer.byteLength(token, 'utf8')` compared first, then `Buffer.from(token, 'utf8')` into `crypto.timingSafeEqual`. Picked (a) over (b) so the auth path stays no-throw by construction rather than relying on try/catch.

Regression block leads with the actual multibyte bug: `'é'.repeat(20)` (20 chars / 40 bytes) against the 20-byte ASCII test root, asserting both `.not.toThrow()` and `false`. Without the fix, this throws `RangeError: Input buffers must have the same byte length`.

Branched off current `main` (5d4fe7d), one commit, 49/49 pass on the test file.

time-attack:
@16francej Please close this PR. The UTF-8 token-length safety check and focused tests already shipped on main through #1431, with this contribution credited. The open PR is now a stale duplicate.

PR #1417, opened 2026-05-10
docs(plan-rollout): SYSTEM.md schema spec (focused slice of #1192)

mastermanas805:
Doc-only slice of #1192. 215 lines, single file (`docs/SYSTEM-MD.md`). Draft.

**The primitive:** SYSTEM.md is human-declared role contracts between components — what each one *is*, what it *owns*, and what breaks if a contract changes (auth mints session tokens; middleware enforces; format change without coordinated deploy breaks the contract). Distinct from the package/import graph, which is discovered at runtime.

**What this PR is:** the schema spec only. No code, no tests, no dependency changes.

**What it isn't:** the parser, reconciler, scaffolder, and consuming skills (`/plan-rollout`, `/spill-check`) all live downstream of this. They land in follow-up PRs only if the primitive itself fits gstack.

If the shape's wrong: `git rm docs/SYSTEM-MD.md` is the entire revert.

Refs #1192

### DIFF
--- a/docs/SYSTEM-MD.md
+++ b/docs/SYSTEM-MD.md
@@ -1,0 +1,215 @@
+# SYSTEM.md — the semantic contract graph
+
+`SYSTEM.md` is a declarative file at the root of a repo that describes what
+each component *is*, what it *owns*, and the role-level contracts it has with
+other components. It's the input to upcoming `/plan-rollout` planning (landing
+in a follow-up PR); consumers will include `/spill-check`, `/ship` (stack
+mode), and `/review` (scope verification).
+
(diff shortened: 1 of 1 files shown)


mastermanas805:
Closing. On reflection, doc-only is the wrong shape — a format spec with no consumer is just literature, and the original deferral on #1192 still stands. Will only revisit with a shipping artifact.

mastermanas805:
Superseded by #1424. That PR ships the schema together with its consumer (`/plan-rollout` skill + dogfood evidence), addressing the "doc-only is literature without a consumer" closure reason here.

PR #1418, opened 2026-05-10
fix(memory-helpers): handle gbrain >=0.25 doctor non-zero exit and schema_version 2

### LINKED ISSUE #1415: fix(gstack-gbrain-sync): engine always "unknown" when gbrain ≥0.25 is unhealthy — silently breaks /sync-gbrain for all Supabase users
## What's happening

\`/sync-gbrain\` (and any gstack skill that calls \`detectEngineTier()\`) always reports \`engine=unknown\` for anyone running gbrain ≥0.25 against a Supabase brain that has any unhealthy check. That covers basically everyone on Supabase — because a fresh install always has at least one warning (resolver_health, or a pending migration) until everything is perfectly tuned.

When engine is unknown, the orchestrator logs \`[gbrain-sync] mode=X engine=unknown\` and skips all three sync stages silently. No error. No hint that anything is wrong. You just... don't get a brain sync.

## Root cause

It's two things hitting at the same time, which is why it's subtle.

### Problem 1: `execSync` throws on non-zero exit, swallowing the JSON

\`freshDetectEngineTier()\` in \`lib/gstack-memory-helpers.ts\` (line 252) does this:

\`\`\`typescript
const out = execSync("gbrain doctor --json --fast 2>/dev/null", { encoding: "utf-8", timeout: 5000 });
const parsed = JSON.parse(out);
\`\`\`

\`gbrain doctor\` exits with code \`1\` whenever health_score < 100. \`execSync\` throws on non-zero exit — the stdout never reaches \`out\`. The catch block returns \`{ engine: "unknown" }\` immediately. The JSON that gbrain wrote to stdout is gone.

### Problem 2: gbrain ≥0.25 changed doctor's output format

Even if the exit code were zero, it wouldn't help. gbrain ≥0.25 switched to \`schema_version:2\` for doctor output, which dropped the top-level \`engine\` field entirely. The old format had \`parsed.engine === "pglite"\` or \`parsed.engine === "supabase"\`. The new format looks like:

\`\`\`json
{
  "schema_version": 2,
  "status": "unhealthy",
  "health_score": 70,
  "checks": [...]
}
\`\`\`

No \`engine\` key. \`parsed?.engine\` is \`undefined\`. Falls straight to \`"unknown"\`.

Both problems need fixing. Either one alone would still break things.

## Repro

1. Install gbrain ≥0.25, connect to Supabase
2. Make sure at least one doctor check is non-green (resolver_health warning is always present on a fresh install — this is basically always true)
3. Run \`/sync-gbrain\` or call \`detectEngineTier()\` directly:

\`\`\`bash
bun run ~/.claude/skills/gstack/bin/gstack-gbrain-sync.ts --dry-run
# Output: [gbrain-sync] mode=dry-run engine=unknown
\`\`\`

Or inline:
\`\`\`bash
cd ~/.claude/skills/gstack && bun -e "
const { detectEngineTier } = await import('./lib/gstack-memory-helpers.ts');
console.log(detectEngineTier());
"
# { engine: 'unknown', ... }
\`\`\`

**Environment:**
- gstack v1.31.0.0
- gbrain v0.31.3
- Supabase Postgres brain (schema_version:2 doctor output)
- macOS 26.4.1

## Fix

Two changes to \`freshDetectEngineTier()\`:

1. Catch the \`execSync\` throw and read stdout off the error object (Node.js puts it there when \`stdio: 'pipe'\` — the default for execSync)
2. When \`parsed?.engine\` is still undefined after that (schema_version:2 case), fall back to reading \`~/.gbrain/config.json\`

\`\`\`typescript
function freshDetectEngineTier(): EngineDetect {
  const now = Date.now();
  let parsed: Record<string, unknown> | null = null;
  try {
    const out = execSync("gbrain doctor --json --fast 2>/dev/null", { encoding: "utf-8", timeout: 5000 });
    parsed = JSON.parse(out);
  } catch (err: unknown) {
    // execSync throws on non-zero exit; stdout is still on the error object
    try {
      const stdout = (err as { stdout?: string })?.stdout ?? "";
      if (stdout) parsed = JSON.parse(stdout);
    } catch { /* unparseable stdout — stay null */ }
  }

  // gbrain >=0.25 uses schema_version:2 which dropped the top-level `engine`
  // field. Fall back to ~/.gbrain/config.json when doctor doesn't provide it.
  let engine: EngineTier = parsed?.engine === "supabase" ? "supabase" : parsed?.engine === "pglite" ? "pglite" : "unknown";
  if (engine === "unknown") {
    try {
      const configPath = join(homedir(), ".gbrain", "config.json");
      const cfg = JSON.parse(readFileSync(configPath, "utf-8"));
      if (cfg?.engine === "pglite") engine = "pglite";
      else if (cfg?.engine === "postgres" || cfg?.database_url) engine = "supabase";
    } catch { /* config unreadable — stay unknown */ }
  }

  return {
    engine,
    supabase_url: parsed?.supabase_url as string | undefined,
    detected_at: now,
    schema_version: 1,
  };
}
\`\`\`

Tested on gstack v1.31.0.0 + gbrain v0.31.3 + Supabase. Before the patch: \`engine=unknown\` on every invocation. After: \`engine=supabase\`.

## Why this matters

Everyone using \`/sync-gbrain\` with Supabase has been getting silent no-ops. There's no error — just a dry-run-like outcome where all stages skip. The brain never syncs. This is especially painful because the resolver_health warning (which triggers non-zero exit from doctor) is present on basically every install.

mvanhorn:
## Summary

`/sync-gbrain` and every caller of `detectEngineTier()` now resolve the engine correctly when `gbrain doctor` exits non-zero (every Supabase brain with `health_score < 100`) or returns schema_version 2 output (any gbrain >=0.25). Previously these paths collapsed to `engine: "unknown"` and the orchestrator silently skipped all sync stages.

## Why this matters

Two layered bugs both fired for the same users:

1. `execSync` threw on non-zero exit. `gbrain doctor` exits 1 whenever `health_score < 100` (a fresh Supabase install always has at least one warning, so this is essentially everyone), so the JSON written to stdout never reached the parser. See `lib/gstack-memory-helpers.ts:250` (pre-fix).
2. gbrain >=0.25 switched to `schema_version: 2` and dropped the top-level `engine` field. `parsed?.engine` became `undefined`, falling straight to `"unknown"`. The project recognizes the new schema elsewhere (`test/skill-e2e-brain-privacy-gate.test.ts:52` already stubs `{"status":"ok","schema_version":2}`) but `freshDetectEngineTier` was missed during the gbrain >=0.25 update wave.

The v1.29.0.0 release note explicitly bumped the required gbrain version to v0.30.0+, so schema_version 2 support is a project requirement, not a nice-to-have.

## Changes

- `freshDetectEngineTier` calls `spawnSync('gbrain', ['doctor', '--json', '--fast'], { stdio: ['ignore', 'pipe', 'ignore'], timeout: 5000 })`. The array form drops the shell wrapper while preserving the original `2>/dev/null` stderr suppression. Other gstack call sites use the same shape.
- `result.stdout` is parsed even when `result.status` is non-zero. Only spawn errors and unparseable JSON fall to `"unknown"`.
- A new `detectEngineFromDoctorJson` helper implements a defensive lookup ladder: schema_v1 (`parsed.engine`) first, then schema_v2 top-level renames (`engine_kind`, `backend`, `engine_type`), then a `checks[]` fallback whose `name` matches `/engine|backend|kind/i` reading `value`/`engine`/`kind`. The ladder errs toward "engine resolves" rather than dropping to unknown when one path exists.
- `GSTACK_DEBUG`-gated stderr warning fires when JSON parses but no engine resolves, so the lookup ladder can be tightened against real-world schema_v2 output. Default behavior is unchanged.

`EngineDetect` schema is unchanged (`schema_version: 1` from gstack's side); the on-disk cache contract is preserved.

## Testing

Three new cases under `describe("detectEngineTier")` cover schema_v1 (existing behavior), schema_v2 healthy, and schema_v2 unhealthy with non-zero exit. The fixtures use a PATH-shim that writes a `gbrain` bash stub to a tmpdir, prepended in `beforeEach` and restored in `afterEach` so the shim does not leak into the cache tests at lines 296 and 305.

```
$ bun test test/gstack-memory-helpers.test.ts
 25 pass
 0 fail
 59 expect() calls
```

Fixes #1415

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1418"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/lib/gstack-memory-helpers.ts
+++ b/lib/gstack-memory-helpers.ts
@@ -19,7 +19,7 @@
 
 import { existsSync, readFileSync, writeFileSync, mkdirSync, statSync, appendFileSync } from "fs";
 import { dirname, join } from "path";
-import { execSync, execFileSync } from "child_process";
+import { execSync, execFileSync, spawnSync } from "child_process";
 import { homedir } from "os";
 
 // ── Types ──────────────────────────────────────────────────────────────────
@@ -247,9 +247,16 @@ export function detectEngineTier(): EngineDetect {
 function freshDetectEngineTier(): EngineDetect {
   const now = Date.now();
   try {
-    const out = execSync("gbrain doctor --json --fast 2>/dev/null", { encoding: "utf-8", timeout: 5000 });
-    const parsed = JSON.parse(out);
-    const engine: EngineTier = parsed?.engine === "supabase" ? "supabase" : parsed?.engine === "pglite" ? "pglite" : "unknown";
+    const result = spawnSync("gbrain", ["doctor", "--json", "--fast"], {
+      encoding: "utf-8", timeout: 5000, stdio: ["ignore", "pipe", "ignore"], env: process.env,
+    });
+    if (result.error) return { engine: "unknown", detected_at: now, schema_version: 1 };
+
+    const parsed = JSON.parse(result.stdout);
+    const engine = detectEngineFromDoctorJson(parsed);
+    if (engine === "unknown" && process.env.GSTACK_DEBUG) {
+      process.stderr.write(`[gstack-memory-helpers] unable to detect gbrain engine from doctor JSON: ${JSON.stringify(parsed)}\n`);
+    }
     return {
       engine,
       supabase_url: parsed?.supabase_url || undefined,
@@ -259,6 +266,23 @@ function freshDetectEngineTier(): EngineDetect {
   } catch {
     return { engine: "unknown", detected_at: now, schema_version: 1 };
   }
+}
+
+function detectEngineFromDoctorJson(parsed: unknown): EngineTier {
+  const obj = (parsed && typeof parsed === "object" ? parsed : {}) as Record<string, unknown>;
+  const direct = [obj.engine, obj.engine_kind, obj.backend, obj.engine_type].find(isEngineTier);
+  if (direct) return direct;
+  const check = Array.isArray(obj.checks)
+    ? obj.checks.find((c) => c && typeof c === "object" && /engine|backend|kind/i.test(String((c as Record<string, unknown>).name)))
+    : null;
+  const nested = check
+    ? [check.value, check.engine, check.kind].find(isEngineTier)
+    : null;
+  return nested || "unknown";
+}
+
+function isEngineTier(value: unknown): value is EngineTier {
+  return value === "supabase" || value === "pglite";
 }
 
 // ── Public: parseSkillManifest ────────────────────────────────────────────
--- a/test/gstack-memory-helpers.test.ts
+++ b/test/gstack-memory-helpers.test.ts
@@ -11,7 +11,7 @@
  * Free-tier (~50ms total). Runs in `bun test`.
  */
 
-import { describe, it, expect, beforeEach, afterAll } from "bun:test";
+import { describe, it, expect, beforeEach, afterEach, afterAll } from "bun:test";
 import { mkdtempSync, writeFileSync, readFileSync, existsSync, rmSync, mkdirSync } from "fs";
 import { tmpdir } from "os";
 import { join } from "path";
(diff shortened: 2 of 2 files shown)


mvanhorn:
Heads up @garrytan: the `report` job is failing on `Post PR comment` with `GraphQL: Resource not accessible by integration (addComment)`. The workflow's `GITHUB_TOKEN` is missing `pull_requests: write`. Adding

```
permissions:
  contents: read
  pull-requests: write
```

to the report job (or the workflow) should fix it. PR body itself is fine, CI evals all pass otherwise.

PR #1437, opened 2026-05-11
fix(v1.27 migration): qualify repo with owner so gh rename actually runs

agile-operators:
## Summary

The `v1.27.0.0` migration (gstack-brain → gstack-artifacts rename) silently fails to rename users' GitHub repos because it passes a bare repo name to `gh repo rename --repo` and `gh repo edit`, both of which require the `[HOST/]OWNER/REPO` form.

```
$ gh repo rename gstack-artifacts-Brandon --repo gstack-brain-Brandon --yes
expected the "[HOST/]OWNER/REPO" format, got "gstack-brain-Brandon"
```

Because stderr was suppressed with `2>/dev/null`, this surfaced only as the generic warning `gh rename failed (repo may not exist or permission denied)` — misdirecting users who actually have `ADMIN` on the repo. The step is then `mark_done`'d, so a second `/gstack-upgrade` run does not retry. The local pointer at `~/.gstack-artifacts-remote.txt` ends up referencing a repo that doesn't exist on GitHub until the user notices.

I hit this today upgrading from v1.26.4.0 → v1.32.0.0 on a fully-authenticated `gh` with ADMIN on the legacy `gstack-brain-$USER` repo.

## What this PR changes

- Resolve the authenticated user's login via `gh api user --jq .login`.
- Qualify both the existing-name check (`gh repo view`) and the rename/edit invocations with `$OWNER/$REPO`.
- Capture stderr from each attempt so the warning prints the real `gh` error instead of a generic guess.
- Extend the fake `gh` mock in `test/migrations-v1.27.0.0.test.ts` so existing tests still drive the rename path with the new `gh api user` call.

No behavior change for users where the migration would already be a no-op (`gh` unauthenticated, repo already at the new name, or no legacy state).

## Test plan

- [x] `bun test test/migrations-v1.27.0.0.test.ts` — 11/11 passing locally
- [x] `bash -n gstack-upgrade/migrations/v1.27.0.0.sh` — clean
- [x] Manually re-ran corrected command on my repo (`gh repo rename gstack-artifacts-Brandon --repo agile-operators/gstack-brain-Brandon --yes`) — succeeded

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1437"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/gstack-upgrade/migrations/v1.27.0.0.sh
+++ b/gstack-upgrade/migrations/v1.27.0.0.sh
@@ -184,19 +184,42 @@ if ! journal_done "gh_repo_renamed"; then
   case "$HOST" in
     github)
       if command -v gh >/dev/null 2>&1 && gh auth status >/dev/null 2>&1; then
-        # Idempotent: if new name already exists, treat as success.
-        if gh repo view "$NEW_REPO_NAME" >/dev/null 2>&1; then
-          echo "    repo already named $NEW_REPO_NAME on GitHub — no-op" >&2
+        # `gh repo rename --repo` and `gh repo edit` both require the
+        # "[HOST/]OWNER/REPO" form. Passing a bare repo name (as earlier
+        # versions of this migration did) fails with:
+        #   expected the "[HOST/]OWNER/REPO" format, got "<name>"
+        # Resolve the authenticated user's login and qualify both names.
+        GH_OWNER=$(gh api user --jq .login 2>/dev/null || echo "")
+        if [ -z "$GH_OWNER" ]; then
+          echo "    WARNING: could not resolve gh authenticated user — skipping rename" >&2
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej — recommend closing this PR only after the owner-qualified calls are ported into an updated #1493 implementation. This PR is otherwise superseded because it can still mark failed migration work as complete instead of aborting atomically. Please retain its code and attribution in #1493.

PR #1436, opened 2026-05-11
feat(hosts): add GitHub Copilot CLI as a first-class host

andrey-esipov:
## Summary

Adds support for installing gstack as flat `.agent.md` files under `~/.copilot/agents/`, per [Copilot CLI's custom agents schema](https://docs.github.com/en/copilot/reference/custom-agents-configuration). After `./setup --host copilot`, all 46 gstack skills are discoverable via `copilot --agent <name>`.

Purely additive — existing hosts (Codex, Cursor, Factory, OpenCode, Kiro, Slate, OpenClaw, Hermes, GBrain) are untouched.

## What changed

- **New `outputLayout?: 'per-skill-dir' | 'flat-agent-md'`** field on `HostConfig`. Default is `'per-skill-dir'` (every existing host). Copilot uses `'flat-agent-md'` to emit one file per skill, flat under `<hostSubdir>/agents/`. Copilot CLI does not recurse into subdirectories under `~/.copilot/agents/`, so the flat layout is required.
- **`processExternalHost`** branches on the new field to write to `<hostSubdir>/agents/<name>.agent.md` instead of the default `<hostSubdir>/skills/<name>/SKILL.md`.
- **`hosts/copilot.ts`** — new HostConfig:
  - Frontmatter allowlist (`name`, `description`)
  - Injects `target: github-copilot` and `tools: [\"*\"]` per the Copilot agents schema
  - `skipSkills: ['codex', 'copilot']` — matches every external host's `codex` skip, plus self-skip to avoid future recursion if a `/copilot` skill is ever added to the source tree
  - Suppressed resolvers match the codex pattern (DESIGN_OUTSIDE_VOICES, ADVERSARIAL_STEP, CODEX_SECOND_OPINION, CODEX_PLAN_REVIEW, REVIEW_ARMY, GBRAIN_CONTEXT_LOAD, GBRAIN_SAVE_RESULTS)
- **`setup`** — adds `INSTALL_COPILOT` flag, build trigger, install block. Symlinks generated `gstack-*.agent.md` files into `~/.copilot/agents/`, creates `~/.copilot/gstack/` runtime root with `bin/`, `browse-dist/`, `gstack-upgrade/`, `ETHOS.md` symlinks for \`\$GSTACK_ROOT\` resolution.
- **`.gitignore`** — adds `.copilot/` for generated agent files.
- **Tests** — `test/host-config.test.ts` and `test/gen-skill-docs.test.ts` updated:
[...]

PR #1443, opened 2026-05-12
feat(hosts): add GitHub Copilot CLI as a first-class skills host

andrey-esipov:
## Summary

Adds support for installing gstack as Copilot CLI **Skills** (per [the docs](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-skills)), so users can invoke `/qa`, `/review`, `/ship`, `/autoplan`, etc. directly in an interactive `copilot` session.

Replaces #1436, which used the wrong Copilot primitive — Agents instead of Skills. That made the generated `.agent.md` files invokable only via `copilot --agent <name>` non-interactively or as model-dispatched tools; user-typed `/<name>` slash commands didn't appear in `/skills`. After installing the prior PR locally, `/office-hours` and `/autoplan` didn't work — which is what flagged the architectural issue.

## What's different vs #1436

| Aspect | #1436 (wrong) | This PR (right) |
|---|---|---|
| Primitive | Agents (`.agent.md`) | Skills (`SKILL.md`) |
| Install dir | `~/.copilot/agents/<name>.agent.md` (flat) | `~/.copilot/skills/gstack/<name>/SKILL.md` (collection subdir, recursive scan) |
| Invocation | `copilot --agent <name>` or model-dispatched | `/<name>` in interactive `copilot` session |
| HostConfig change | Added `outputLayout` field + branch in `processExternalHost` | None — uses existing per-skill-dir layout |
| Diff size | ~190 lines + framework changes | ~177 lines, purely additive |

## What changed

- **`hosts/copilot.ts`** (new) — HostConfig with `globalRoot: '.copilot/skills/gstack'`, frontmatter allowlist (name + description), pathRewrites that route `.claude/skills/...` references to `$GSTACK_ROOT` (the runtime root setup creates at `~/.copilot/gstack/`).
- **`hosts/index.ts`** — register `copilot` in `ALL_HOST_CONFIGS`.
- **`setup`** — `INSTALL_COPILOT` flag, build trigger, install block:
  - Symlinks each generated `<repo>/.copilot/skills/gstack-<name>/` directory into `~/.copilot/skills/gstack/<name>/` (bare name — the `gstack-` prefix is stripped during link creation so users invoke `/<name>`, not `/gstack-<name>`).
  - Creates `~/.copilot/gstack/` runtime root with `bin/`, `browse-dist/`, `gstack-upgrade/`, `ETHOS.md` symlinks for `$GSTACK_ROOT`-relative tooling inside skill bash blocks.
- **`.gitignore`** — adds `.copilot/`.
- **Tests** — host count assertion `10 → 11`; `--host` pass-through string includes `copilot`. No layout-aware test changes needed (uses default per-skill-dir).

## `skipSkills`

Five skills are skipped because they don't translate to Copilot CLI's per-invocation skill model — each `/<name>` invocation is stateless, so anything that toggles persistent session state breaks:

- **`freeze`, `unfreeze`** — toggle directory-scoped edit boundaries
- **`careful`, `guard`** — install destructive-command guardrails for the session
- **`plan-tune`** — interactive UI for tuning AskUserQuestion sensitivity per skill

The `codex` skill is also skipped per the standard external-host convention (it wraps the `codex` CLI binary — no reason to expose it on a different host).

## Verification (macOS, on this branch)

- `./setup --host copilot` → produces 41 skills under `~/.copilot/skills/gstack/`
- `copilot -p "/qa describe yourself"`, `copilot -p "/office-hours describe yourself"`, `copilot -p "/review describe yourself"` all invoke the correct gstack skill content
- `bun test test/host-config.test.ts test/gen-skill-docs.test.ts` → **459 pass, 0 fail**
- `bun run gen:skill-docs --host all` → all 10 existing host outputs + new `.copilot/skills/` generate cleanly

## Test plan

- [x] All existing host outputs unchanged on `--host all`
- [x] `--host copilot` emits one `SKILL.md` per skill (excluding skipSkills)
- [x] Copilot CLI discovers skills at `~/.copilot/skills/gstack/<name>/SKILL.md`
- [x] User-typed `/<name>` invokes the gstack skill in interactive mode
- [x] Frontmatter is minimal valid YAML (parsed by Copilot CLI without warning)
- [x] Setup is idempotent (re-run preserves state, cleans stale symlinks)
- [ ] CI on Linux + Windows (needs maintainer to trigger)

## Notes for reviewers

- The setup script prints a reminder to set `export GSTACK_ROOT=$HOME/.copilot/gstack` because skill bash blocks reference `$GSTACK_ROOT/bin/...` for sidecar tooling (telemetry, learnings, gbrain). Users can `export` it in their shell rc or skip — only the sidecars break without it; core skill logic still works.
- Optional follow-up: the rules from `freeze` / `careful` / `guard` could be injected into a project-scoped `AGENTS.md` template so their guardrails are *ambient* across every Copilot CLI invocation (recovering the protection we drop in this PR). Happy to file a separate PR for that.

Sources:
- [Adding agent skills for GitHub Copilot CLI](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-skills)
- [Creating and using custom agents for GitHub Copilot CLI](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/create-custom-agents-for-cli) (for context on Agents vs Skills)

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1443"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ bin/gstack-global-discover
 .openclaw/
 .hermes/
 .gbrain/
+.copilot/
 .gbrain-source
 .context/
 extension/.auth.json
--- a/hosts/copilot.ts
+++ b/hosts/copilot.ts
@@ -1,0 +1,92 @@
+import type { HostConfig } from '../scripts/host-config';
+
+/**
+ * GitHub Copilot CLI host.
+ *
+ * Copilot CLI exposes two customization surfaces:
+ *   1. Skills (SKILL.md files invoked as /<name> slash commands by the user)
+ *   2. Agents (.agent.md files invoked as tools by the model)
+ *
+ * gstack content is overwhelmingly "skills" in Copilot's sense — workflows
+ * the user invokes (/qa, /review, /ship, /autoplan, …). This host emits
+ * SKILL.md via the standard per-skill-dir layout. Setup installs them under a
+ * collection subdir at ~/.copilot/skills/gstack/<name>/SKILL.md, which
+ * Copilot CLI's native skill loader picks up via recursive scan.
+ *
+ * Schema reference:
+ *   https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-skills
+ */
+const copilot: HostConfig = {
+  name: 'copilot',
+  displayName: 'GitHub Copilot CLI',
+  cliCommand: 'copilot',
+
+  globalRoot: '.copilot/skills/gstack',
+  localSkillRoot: '.copilot/skills/gstack',
+  hostSubdir: '.copilot',
+  usesEnvVars: true,
+
+  frontmatter: {
+    mode: 'allowlist',
+    keepFields: ['name', 'description'],
+    descriptionLimit: 1024,
+    descriptionLimitBehavior: 'truncate',
+  },
+
+  generation: {
+    generateMetadata: false,
+    // Skipped because they don't translate to Copilot CLI's per-invocation
+    // skill model (each /<name> invocation is stateless), or because they
+    // wrap a CLI binary that has no reason to recurse:
+    //   'codex'       — wraps the `codex` CLI binary (every external host skips this)
+    //   'freeze'      — toggles a session-scoped edit boundary; no session state to toggle
+    //   'unfreeze'    — pairs with /freeze; same reason
+    //   'careful'     — installs session-scoped destructive-command guardrails
+    //   'guard'       — combines /careful + /freeze; same reason
+    //   'plan-tune'   — interactive UI for tuning AskUserQuestion sensitivity
+    //                   per-skill; only meaningful inside a persistent skill system
+    // Follow-up: the rules from freeze/careful/guard could be injected into a
+    // project AGENTS.md so they're ambient across every Copilot CLI session.
+    skipSkills: ['codex', 'freeze', 'unfreeze', 'careful', 'guard', 'plan-tune'],
+  },
+
+  pathRewrites: [
+    // Bash blocks inside skills reference ~/.claude/skills/gstack/bin/… for
+    // sidecar tooling (telemetry, learnings, gbrain). Route them to the
+    // copilot-side runtime root that setup creates at ~/.copilot/gstack/.
+    // `.claude/skills` references in prose point at the same root since
+    // Copilot CLI doesn't have a per-workspace skills directory equivalent
+    // to .claude/skills/gstack/.
+    { from: '~/.claude/skills/gstack', to: '$GSTACK_ROOT' },
+    { from: '.claude/skills/gstack', to: '$GSTACK_ROOT' },
+    { from: '.claude/skills', to: '$GSTACK_ROOT' },
+  ],
+
+  suppressedResolvers: [
+    'DESIGN_OUTSIDE_VOICES',
+    'ADVERSARIAL_STEP',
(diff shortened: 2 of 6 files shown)


andrey-esipov:
Self-review pass via gstack `/review` skill. One real issue caught and fixed inline (force-pushed); two informational items called out below.

## Applied inline

**Name-collision mitigation** (confidence 8/10) — Copilot CLI ships built-in `/review`, `/diff`, `/pr`, `/update` and others. Stripping `gstack-` to make `/<bare>` the invocation form means built-ins win a name collision, hiding gstack's `/review` etc. in interactive mode.

Fix: install BOTH `/<bare>` and `/gstack-<name>` symlinks in the same loop. Users get the convenient bare form when it's unique, and `/gstack-review` etc. always works as an explicit fallback. ~3 line setup change.

## Informational (not fixed)

1. **(confidence 5/10)** `rm -rf "$COPILOT_SKILLS"` would wipe non-gstack content if a user manually placed files at `~/.copilot/skills/gstack/`. Per spec the dir is gstack-managed, but hardening to `find … -type l -delete` would protect the edge case at the cost of leaving real files behind. Trade-off, not a clear win.

2. **(confidence 4/10)** No README mention of the new host. The `hosts/index.ts` registration is self-documenting and matches the project's pattern, so this is "would be nice" rather than "should fix."

## Adversarial sweep

Setup loop uses parameter expansion + globbed paths from gstack's own `gen-skill-docs`. Filenames constrained by gstack's name regex (`[a-z][a-z0-9-]*`), so no special-char attacks. No prompt injection vectors (no LLM output flows). No external network calls. Clean.

Verified after fix:
- `bun test test/host-config.test.ts test/gen-skill-docs.test.ts` → **459 pass, 0 fail**
- `ls ~/.copilot/skills/gstack/` shows both `qa` and `gstack-qa`, both pointing at the same generated dir

andrey-esipov:
Closing — not worth the upstream-maintenance overhead for a fork-specific install path. The architecture works (verified: gstack skills install as flat top-level entries at `~/.copilot/skills/<name>/SKILL.md`, Copilot CLI's `/<name>` autocomplete picker sees them, with `astack-` prefix only on names that collide with Copilot built-ins like `/review`), but I'd rather iterate freely on my fork than gate every refinement on upstream review. Happy to revisit later if there's interest from the maintainer.

The full architecture story (for anyone landing here):
- Copilot CLI's `/skills info` and `/` autocomplete picker only scan ONE level deep under skill discovery dirs. Nested collection subdirs (`~/.copilot/skills/<collection>/<name>/SKILL.md`) work for description-based auto-trigger via `-p` mode but are invisible to the interactive picker.
- The right shape is flat top-level: `~/.copilot/skills/<name>/SKILL.md`.
- Skill name collisions with Copilot CLI built-ins (`/review`, `/diff`, `/pr`, etc.) cause built-ins to win the lookup. Prefix-rename gstack's colliding skills (just `review` for now) to avoid this.

Sorry for the noise on both attempts.

PR #1464, opened 2026-05-13
v1.33.3.0 fix: add .gbrain-source to .gitignore and create utility to ensure it in consumer repos

### LINKED ISSUE #1384: /sync-gbrain doesn't add `.gbrain-source` to consumer repo's `.gitignore` (v1.29.0.0 changelog promised it)

realcarsonterry:
## Summary

Fixes #1384

When `gbrain sources attach` creates `.gbrain-source` in a consumer repo, the file is not added to `.gitignore`, causing it to be accidentally committed. This breaks per-worktree isolation as the pin file propagates across branches/worktrees.

## Changes

1. **Added `.gbrain-source` to gstack's own `.gitignore`**
   - Prevents the pin file from being committed in the gstack repo itself

2. **Created `bin/gstack-gbrain-ensure-gitignore` utility script**
   - Idempotent script that appends `.gbrain-source` to a repo's `.gitignore` if not already present
   - Handles edge cases: missing `.gitignore` files, files without trailing newlines
   - Can be called after `gbrain sources attach` to ensure consumer repos ignore the pin file

## Integration

The utility script should be integrated into any code path that runs `gbrain sources attach` in consumer repositories. The exact integration point depends on where gstack invokes this command (not found in current v1.25.1.0 codebase, may be in future versions or external tooling).

Example usage:
```bash
gbrain sources attach <id>
gstack-gbrain-ensure-gitignore .
```

## Test plan

- [x] Created utility script with proper error handling
- [x] Added `.gbrain-source` to gstack's `.gitignore`
- [x] Script is executable and follows gstack conventions
- [ ] Manual testing: run script in a test repo and verify `.gbrain-source` is added to `.gitignore`
- [ ] Integration testing: verify script works after `gbrain sources attach` in consumer repos

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1464"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


realcarsonterry:
## Ready to Merge ✓

All 18 test checks passing. Only the infrastructure 'report' step failed (not a code issue).

**Impact:** Fixes issue #1384 - Implements the v1.29.0.0 changelog promise to auto-add `.gbrain-source` to consumer repo gitignore. Prevents accidental commits of generated files.

**Testing:** All E2E evals passing including gbrain-related tests.

garrytan:
Thank you @realcarsonterry. Independent fix for #1384 — superseded by PR #1521's exported-helper shape; thank you for the early attempt. Consolidated v1.40.0.0 fix wave lands as PR #1547. Closing this one as superseded — your contribution is recognized in CHANGELOG.md and the commit metadata.

PR #1466, opened 2026-05-13
fix(update-check): implement --apply to actually run the upgrade

### LINKED ISSUE #1353: gstack-update-check --apply reports UPGRADE_AVAILABLE but doesn't perform the pull
## Summary

`~/.claude/skills/gstack/bin/gstack-update-check --apply` exits 0 and emits `UPGRADE_AVAILABLE <old> <new>`, but the working tree is unchanged. `VERSION` still reads the old version. Subsequent runs produce identical output until a manual `git pull` is performed. The `--apply` flag appears to fall through to the report-only path.

## Reproduction

```bash
# Install gstack at an older commit, e.g. 1.26.4.0
$ cat ~/.claude/skills/gstack/VERSION
1.26.4.0

# Run with --apply
$ ~/.claude/skills/gstack/bin/gstack-update-check --apply
UPGRADE_AVAILABLE 1.26.4.0 1.26.5.0

# Check state — unchanged
$ cat ~/.claude/skills/gstack/VERSION
1.26.4.0

# Manual workaround pulls successfully
$ cd ~/.claude/skills/gstack && git pull origin main
Updating 19e699ab..c7aefc1a
Fast-forward
 CHANGELOG.md | 41 +++++++++++
 ...
$ cat ~/.claude/skills/gstack/VERSION
1.26.5.0
```

## Expected vs actual

| | Expected | Actual |
|---|---|---|
| Output with `--apply` | `UPGRADED 1.26.4.0 -> 1.26.5.0` (or similar success message) | `UPGRADE_AVAILABLE 1.26.4.0 1.26.5.0` (identical to no-flag path) |
| Working tree | Updated to `c7aefc1a` | Still at `19e699ab` |
| VERSION file | `1.26.5.0` | `1.26.4.0` |
| Exit code | 0 with state mutation | 0 with no state mutation |

The output with `--apply` is identical to the report-only output, suggesting the apply branch isn't being taken — or is being taken but silently no-op'ing.

## Environment

- macOS Sequoia 15.x
- Bash on Apple Silicon
- gstack `1.26.4.0` -> target `1.26.5.0`
- gstack repo HEAD before workaround: `19e699ab`
- gstack repo HEAD after manual `git pull`: `c7aefc1a`
- Working tree clean (no uncommitted changes, no untracked files in `~/.claude/skills/gstack`)

## Suggested investigation paths

1. Is the `--apply` branch in `gstack-update-check` invoking `git pull` (or equivalent)? If so, is its exit code captured and surfaced?
2. If `git pull` is invoked, what's its actual stdout/stderr? Could it be silently failing on a precondition (detached HEAD check, branch tracking, network) and falling through to the report message?
3. The fact that the output exactly matches the no-flag path is the strongest hint — suggests `--apply` either isn't being parsed, or is parsed but the apply codepath unconditionally emits the same `UPGRADE_AVAILABLE` line before bailing.

## Workaround (for users hitting this)

```bash
cd ~/.claude/skills/gstack && git pull origin main
```

## Why it matters

gstack's value compounds on every session using the latest version — newer releases ship bug fixes and specialist capabilities that prevent CI failures downstream. A silent no-op on `--apply` means projects falsely believe they're current. It's also the same trigger-vs-outcome failure pattern that several gstack skills (`/review`, `/qa` infra-class checks) help catch in user code, so worth fixing in the tool that enforces the discipline.

Happy to provide additional traces or test patches if useful. Thanks for the great work on gstack — it's load-bearing in our delivery pipeline.

genisis0x:
## Summary

Fixes #1353.

\`gstack-update-check --apply\` previously exited 0 and emitted \`UPGRADE_AVAILABLE <old> <new>\` without touching the working tree. Tracing the file showed why: the only flag handler was \`--force\` (\`if [ "\${1:-}" = "--force" ]\`), and unknown flags fell through to the report-only path. \`--apply\` was indistinguishable from no flag, and the docs (\`-h\` / header) never mentioned it.

Add an \`--apply\` branch that, when an upgrade is detected, runs the minimal headless variant of \`/gstack-upgrade\` Step 4 for git installs:

\`\`\`
git -C "\$GSTACK_DIR" fetch origin main
git -C "\$GSTACK_DIR" reset --hard origin/main
( cd "\$GSTACK_DIR" && ./setup )
\`\`\`

On success the script emits \`UPGRADED <old> <new>\` (distinct from the report-only \`UPGRADE_AVAILABLE\` token) and caches that line so a subsequent \`--apply\` re-run after the VERSION already moved prints the new \`UP_TO_DATE\` and exits cleanly.

## Scope

Intentionally minimal:

- **Git installs only.** Vendored installs need the \`mv\` + clone-into-place ceremony and the local-vendored copy sync from \`/gstack-upgrade\` Step 4.5. \`--apply\` surfaces an explicit error pointing at \`/gstack-upgrade\` for those.
- **Snooze ignored on \`--apply\`.** The caller asked for the upgrade explicitly. Honoring snooze would let a CI gate appear to succeed while not actually upgrading.
- **No migrations.** The migration pipeline (\`gstack-upgrade/migrations/*.sh\`) is \`/gstack-upgrade\`'s concern. \`--apply\` is for the headless / scripted gate; if a release needs migrations, \`/gstack-upgrade\` is the path.
- **Failure modes are distinct.** Each git step's failure prints to stderr and exits non-zero. A \`git reset\` that succeeded followed by a \`./setup\` that failed leaves the working tree at the new version and surfaces an explicit "re-run ./setup manually" hint — the user is in a recoverable state, not a half-done one.

## Flag parsing refactor

The original \`if [ "\${1:-}" = "--force" ]\` shape only inspected \`\$1\`, so \`--apply\` after another flag wouldn't have fired even if its handler existed. Refactor to a \`for arg in "\$@"; case\` loop so the two flags compose (e.g. \`--force --apply\` is well-defined: both fire). Unknown flags still fall through silently to preserve the existing zero-noise contract of the periodic-check path; the periodic invocation is called from skill preambles and should not start barking at users about flag shapes.

## Header doc

Grows three lines: the new \`UPGRADED\` output token, the \`--apply\` flag description, and the relationship to \`/gstack-upgrade\` so future readers don't need to bisect commits to learn what the flag does or doesn't do.

## Validation

- \`bash -n bin/gstack-update-check\` clean.
- Manual: ran \`GSTACK_DIR=<test repo> bin/gstack-update-check --apply\` against a stale checkout, confirmed \`UPGRADED 1.33.2.0 1.33.3.0\` printed and the working tree updated.
- No \`bun test\` runner available locally; existing tests under \`test/\` don't appear to exercise \`gstack-update-check\` flag handling, so no test regression expected.

Fixes #1353

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1466"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/bin/gstack-update-check
+++ b/bin/gstack-update-check
@@ -4,7 +4,17 @@
 # Output (one line, or nothing):
 #   JUST_UPGRADED <old> <new>       — marker found from recent upgrade
 #   UPGRADE_AVAILABLE <old> <new>   — remote VERSION differs from local
+#   UPGRADED <old> <new>            — `--apply` ran a successful upgrade
 #   (nothing)                       — up to date, snoozed, disabled, or check skipped
+#
+# Flags:
+#   --force   Bust cache and snooze (used by standalone /gstack-upgrade).
+#   --apply   When an upgrade is available, actually run it: `git fetch` +
+#             `git reset --hard origin/main` + `./setup` inside `$GSTACK_DIR`.
+#             For the full ceremony (stash, vendored-install handling,
+#             local-vendored copy sync, migrations), use /gstack-upgrade.
+#             `--apply` is the minimal headless variant for CI / scripted
+#             update gates.
 #
 # Env overrides (for testing):
 #   GSTACK_DIR          — override auto-detected gstack root
@@ -20,11 +30,25 @@ SNOOZE_FILE="$STATE_DIR/update-snoozed"
 VERSION_FILE="$GSTACK_DIR/VERSION"
 REMOTE_URL="${GSTACK_REMOTE_URL:-https://raw.githubusercontent.com/garrytan/gstack/main/VERSION}"
 
-# ─── Force flag (busts cache + snooze for standalone /gstack-upgrade) ──
-if [ "${1:-}" = "--force" ]; then
-  rm -f "$CACHE_FILE"
-  rm -f "$SNOOZE_FILE"
-fi
+# ─── Parse flags ──────────────────────────────────────────────
+APPLY=0
+for arg in "$@"; do
+  case "$arg" in
+    --force)
+      rm -f "$CACHE_FILE"
+      rm -f "$SNOOZE_FILE"
+      ;;
+    --apply)
+      APPLY=1
+      # Force a fresh check so a stale cache doesn't suppress the upgrade.
+      rm -f "$CACHE_FILE"
+      ;;
+    *)
+      # Unknown flag — fall through silently to preserve the historical
+      # zero-noise contract of the periodic-check path.
+      ;;
+  esac
+done
 
 # ─── Step 0: Check if updates are disabled ────────────────────
 _UC=$("$GSTACK_DIR/bin/gstack-config" get update_check 2>/dev/null || true)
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej looks good to merge

PR #1467, opened 2026-05-13
fix(codex): surface non-zero exits so wrappers stop reading as silent stalls

### LINKED ISSUE #1327: [codex skill] Surface non-zero exit codes from codex wrappers (currently silent on parse errors)
## Summary

The `/codex`, `/review`, and `/ship` skill wrappers currently inspect `$_CODEX_EXIT` only for the timeout sentinel (`124`). Other non-zero exit codes — including the `exit 2` parse error that #1196 (and PR #1209) addresses — are silently swallowed. From the orchestrating agent's perspective, codex appears to produce no output, indistinguishable from a silent stall.

This is a defensive layer separate from the canonical fix in #1209: even after #1209 lands, future Codex CLI breaking changes that change accepted arg combinations will keep masquerading as silent stalls until/unless someone manually inspects the captured stderr.

## Repro

The same fixture used in #1196 — pre-#1209 invocation:
```bash
codex review "<prompt>" --base main -c 'model_reasoning_effort="xhigh"' --enable web_search_cached < /dev/null
```

Returns exit 2 with parse error in milliseconds. The skill wrapper currently:

1. Captures stderr to a temp file
2. Checks `$_CODEX_EXIT` for `124`
3. If not 124, treats as success and returns no diagnostic
4. Calling agent sees empty output, hangs waiting for content that will never come

The stderr file at `%LOCALAPPDATA%\Temp\codex-err-*.txt` contains the actual parse error the entire time, but the wrapper doesn't surface it.

## Empirical context

In our project (claude-teams-bot) we've hit this ~5 times across PRs over the past week. Multiple pr-reviewer agents and one build agent independently misdiagnosed this as a Claude Code stall, an Anthropic API issue, or a model rate-limit before someone finally inspected the temp stderr file and saw the parse error. Each misdiagnosis cost ~30-60 minutes of agent time.

CLAUDE.md #54f in our cross-project policy file now reads: *"Don't trust silent-stall framing. When an agent reports 'codex stalled silently for N minutes,' the likely answer is NOT a model API stall — it's exit non-zero with empty stdout, swallowed by a wrapper that only handles the timeout exit code."*

That's a workaround at the agent-discipline layer, but the wrapper-level fix is cheaper and more general.

## Proposed fix

Append to the wrapper's exit-code check:
```bash
if [ "$_CODEX_EXIT" = "124" ]; then
  echo "Codex timed out after the configured deadline."
elif [ "$_CODEX_EXIT" != "0" ]; then
  echo "Codex exited with code $_CODEX_EXIT. Stderr (first 20 lines):"
  head -20 "$TMPERR" 2>/dev/null || true
fi
```

This makes the failure mode self-describing. Even if a future Codex CLI release introduces another arg-shape break, the surfaced stderr tells the calling agent exactly what's wrong.

## Stretch — capability probe

Longer-term, replacing the version-string regex in `gstack-codex-probe` with a capability probe (run `codex review --help`, parse for known-good arg combinations, dynamically construct the invocation) would make the skill self-correct when Codex changes its arg contract. That's a larger refactor — not blocking on this issue.

## Relation to #1196 / #1209

Distinct concern. #1209 fixes the specific parse-error trigger by removing the broken arg shape. This issue addresses the wrapper's blindness to ALL non-zero exits, including the next time something similar happens. Both fixes are complementary.

## Empirical link

Local workaround documented at `docs/operations/codex-cli-workaround.md` in our project — comments on #1209 explain the relationship to our empirical recurrence.

genisis0x:
## Summary

Fixes #1327.

The \`/codex\` Review, Challenge, Consult, and Consult-resume wrappers all inspect \`_CODEX_EXIT\` for the 124 timeout sentinel only. Any other non-zero exit — parse error from a Codex CLI breaking arg change (exit 2), auth failure that didn't match the existing auth-text regex, network drop returning a partial frame — gets swallowed. The calling agent sees empty stdout, indistinguishable from a model API stall, and proceeds to misdiagnose for 30-60min before someone checks \`\$TMPERR\` by hand.

The reporter's CLAUDE.md #54f from claude-teams-bot makes the same observation across ~5 hits in one week: *"Don't trust silent-stall framing... it's exit non-zero with empty stdout, swallowed by a wrapper that only handles the timeout exit code."*

## Approach

Add an \`elif != "0"\` branch to each of the four wrappers (Review, Challenge, Consult new-session, Consult-resume). On a non-zero non-timeout exit, the wrapper now:

- Prints \`[codex exit <code>] <first stderr line>\` so the calling agent has an immediate, structured diagnostic line.
- Prints the first 20 lines of \`\$TMPERR\` indented two spaces so the full parse error / arg-shape rejection is right there in the conversation, not hidden in a temp file.
- Logs \`codex_nonzero_exit:<mode>:<exit_code>\` via the existing telemetry helper so this becomes queryable, distinct from \`codex_timeout\`.

Behavior on success (exit 0) and timeout (exit 124) is unchanged — the existing branches still fire first. Auth-error detection in the Challenge wrapper (\`grep -qiE "auth|login|unauthorized" "\$TMPERR"\`) also still fires unchanged; the new branch sits between the timeout branch and the auth branch and doesn't shadow either.

## Defensive, not curative

This is the layer the reporter explicitly framed as separate from #1209's curative fix for the specific parse-error trigger: even after #1209 lands, future Codex CLI arg-shape breaks will keep masquerading as silent stalls until/unless someone manually inspects the captured stderr. Both fixes are complementary; this one is the cheaper of the two and self-corrects on the next regression.

## Files

- \`codex/SKILL.md.tmpl\` — canonical source. Four \`elif != "0"\` branches added next to the existing \`if = "124"\` checks (Review @ 178, Challenge @ 336, Consult @ 489, Consult-resume @ 517).
- \`codex/SKILL.md\` — generated mirror, same four edits applied so the PR is self-consistent without requiring the reviewer to run \`bun run gen:skill-docs\` to inspect the output.

Both files are kept in sync per the \`SKILL.md workflow\` section of CLAUDE.md.

## Validation

- \`bash -n\` on the embedded blocks — clean (the wrappers are markdown code fences but each block is a self-contained bash snippet that runs fine when extracted).
- No \`bun test\` runner available locally; the change is additive (new elif branch only; no existing branch touched) so the existing skill-validation suite should remain green.

Fixes #1327

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1467"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/codex/SKILL.md
+++ b/codex/SKILL.md
@@ -949,6 +949,13 @@ if [ "$_CODEX_EXIT" = "124" ]; then
   _gstack_codex_log_event "codex_timeout" "330"
   _gstack_codex_log_hang "review" "$(wc -c < "$TMPERR" 2>/dev/null || echo 0)"
   echo "Codex stalled past 5.5 minutes. Common causes: model API stall, long prompt, network issue. Try re-running. If persistent, split the prompt or check ~/.codex/logs/."
+elif [ "$_CODEX_EXIT" != "0" ]; then
+  # Surface non-zero exits (parse errors, arg-shape breaks, etc.) so the
+  # calling agent doesn't read "no output" as a silent model/API stall and
+  # burn 30-60min misdiagnosing it. See #1327.
+  echo "[codex exit $_CODEX_EXIT] $(head -1 "$TMPERR" 2>/dev/null || echo "no stderr captured")"
+  head -20 "$TMPERR" 2>/dev/null | sed 's/^/  /' || true
+  _gstack_codex_log_event "codex_nonzero_exit" "review:$_CODEX_EXIT"
 fi
 ```
 
@@ -1187,6 +1194,12 @@ if [ "$_CODEX_EXIT" = "124" ]; then
   _gstack_codex_log_event "codex_timeout" "600"
   _gstack_codex_log_hang "challenge" "$(wc -c < "$TMPERR" 2>/dev/null || echo 0)"
   echo "Codex stalled past 10 minutes. Common causes: model API stall, long prompt, network issue. Try re-running. If persistent, split the prompt or check ~/.codex/logs/."
+elif [ "$_CODEX_EXIT" != "0" ]; then
+  # Surface non-zero exits so the calling agent doesn't read "no output" as
+  # a silent model/API stall. See #1327.
+  echo "[codex exit $_CODEX_EXIT] $(head -1 "$TMPERR" 2>/dev/null || echo "no stderr captured")"
+  head -20 "$TMPERR" 2>/dev/null | sed 's/^/  /' || true
+  _gstack_codex_log_event "codex_nonzero_exit" "challenge:$_CODEX_EXIT"
 fi
 # Fix 2: surface auth errors from captured stderr instead of dropping them
 if grep -qiE "auth|login|unauthorized" "$TMPERR" 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks @genisis0x — shipped in v1.42.0.0 Daegu wave (#1594). Closes #1327. Credit retained in the wave commit.

PR #1469, opened 2026-05-13
fix(artifacts-init): honor gh/glab git_protocol when picking push URL

### LINKED ISSUE #1348: gstack-brain-init defaults to SSH remote, fails for HTTPS-configured `gh`
## Environment

- gstack `1.26.4.0`
- `gh` configured with `git_protocol = https` (the `gh auth login` default)

```
$ gh config get git_protocol
https
$ gh auth status | grep -i protocol
  - Git operations protocol: https
```

## Reproducer

`gstack-brain-init` with default settings (let `gh repo create` make the repo) followed by the first push fails for users whose `gh` auth is HTTPS-only — the script asks `gh` for `sshUrl` regardless of the user's `git_protocol` setting, so the remote ends up as `git@github.com:user/gstack-brain-user.git` and `git push` then prompts for SSH credentials the user does not have.

User has to manually re-run with `--remote https://github.com/user/gstack-brain-user.git` to recover.

## Expected

`gstack-brain-init` honors `gh config get git_protocol`. If it returns `https`, register the HTTPS clone URL; if `ssh`, register the SSH URL.

## Actual

`bin/gstack-brain-init:91` and `:98` hardcode `sshUrl`:

```sh
# line 89-99
if ! gh repo create "\$DEFAULT_NAME" --private --description "gstack session memory" 2>/dev/null; then
  # Maybe the repo already exists; try to fetch its URL.
  REMOTE_URL=\$(gh repo view "\$DEFAULT_NAME" --json sshUrl -q .sshUrl 2>/dev/null || echo "")
  ...
else
  REMOTE_URL=\$(gh repo view "\$DEFAULT_NAME" --json sshUrl -q .sshUrl 2>/dev/null || echo "")
fi
```

Both branches request `sshUrl` unconditionally.

## Suggested fix

Pick the URL field based on `gh config get git_protocol`:

```sh
PROTOCOL=\$(gh config get git_protocol 2>/dev/null || echo "https")
if [ "\$PROTOCOL" = "ssh" ]; then
  URL_FIELD=sshUrl
else
  URL_FIELD=url   # HTTPS clone URL
fi
REMOTE_URL=\$(gh repo view "\$DEFAULT_NAME" --json "\$URL_FIELD" -q ".\$URL_FIELD" 2>/dev/null || echo "")
```

Defaulting to HTTPS when `git_protocol` is unset matches `gh auth login`'s own default.

The interactive prompt at `bin/gstack-brain-init:105` ("Paste a private git URL (e.g. git@github.com:you/gstack-brain.git)") could similarly show the user's preferred protocol in the example, but the auto-create path is where the silent failure happens.

0xDevNinja:
## Summary

- `gstack-artifacts-init` now honors `gh config get git_protocol` (and `glab config get git_protocol` for the gitlab branch) when deriving the push URL.
- New `--push-protocol auto|https|ssh` flag lets users pin explicitly. `auto` is the default.
- Empty / unreadable config falls back to SSH, so users who never set `git_protocol` see no change.
- 7 new tests cover all four `auto` paths (github/gitlab × https/ssh), both explicit overrides, the unset-fallback, and invalid-value rejection.

Fixes #1348.

## Why

`bin/gstack-artifacts-init:184` (post v1.27.0.0 rename from `gstack-brain-init`) unconditionally converts the canonical HTTPS URL to its SSH form for `git push origin`:

```sh
PUSH_URL=$("$URL_BIN" --to ssh "$CANONICAL_HTTPS" 2>/dev/null || echo "$CANONICAL_HTTPS")
```

Users whose `gh auth login` defaulted to HTTPS (the gh default since 2.x) don't have an SSH key on the remote. The post-create `git ls-remote $PUSH_URL` step then hangs on an SSH prompt, exits, and the script bails with `Remote not reachable via SSH:`. Recovery required rerunning with `--remote <https url>`, which only works if the user knows the workaround.

Same hazard on the gitlab branch — `glab config get git_protocol` is parallel.

## Fix

Three-step resolution for the push URL:

1. `--push-protocol https|ssh` flag wins (lets the user pin regardless of CLI state — useful in CI / scripted setups).
2. `auto` (default) queries the per-host CLI config:
   - `github` → `gh config get git_protocol`
   - `gitlab` → `glab config get git_protocol`
3. Empty / unreadable / unknown value → SSH (historical default; no regression for users who never set `git_protocol`).

The `ls-remote` failure message also now points HTTPS users at the override flag, so the next attempt is one rerun away rather than a workaround hunt.

## Tests

Extended the existing `makeFakeGh` / `makeFakeGlab` shims with an optional `gitProtocol` arg so they respond to `gh config get git_protocol` (and `glab config get git_protocol`). New `describe('gstack-artifacts-init push protocol (issue #1348)')` block, 7 cases:

- `gh git_protocol=https` → origin URL is HTTPS.
- `gh git_protocol=ssh` → origin URL is SSH (existing behavior preserved).
- `gh git_protocol` unset → origin URL is SSH (legacy default preserved).
- `--push-protocol https` overrides `gh git_protocol=ssh`.
- `--push-protocol ssh` overrides `gh git_protocol=https`.
- `--push-protocol bogus` exits 1 with `Invalid --push-protocol`.
- gitlab branch: `glab git_protocol=https` → origin URL is HTTPS.

```bash
bun test test/gstack-artifacts-init.test.ts
# 23 pass, 0 fail (16 existing + 7 new)
```

## Notes

- No CHANGELOG / VERSION bump in the diff (maintainer-owned at squash-merge per repo convention).
- `--push-protocol` deliberately covers only the local push URL; `~/.gstack-artifacts-remote.txt` continues to store the canonical HTTPS form (codex Finding #10).
- Does not touch `gstack-artifacts-url --to ssh` — that helper is still the right derivation primitive; only the `auto` policy in `artifacts-init` changes.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1469"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/bin/gstack-artifacts-init
+++ b/bin/gstack-artifacts-init
@@ -9,6 +9,14 @@
 # Usage:
 #   gstack-artifacts-init [--remote <url>] [--host github|gitlab|manual]
 #                         [--url-form-supported true|false]
+#                         [--push-protocol auto|https|ssh]
+#
+# --push-protocol controls the URL form used for `git push origin`:
+#   auto (default) — honors `gh config get git_protocol` (github) or
+#                    `glab config get git_protocol` (gitlab). Falls back to
+#                    SSH when neither is set / readable.
+#   https          — force HTTPS (requires gh/glab credential helper or PAT).
+#   ssh            — force SSH (legacy default; requires SSH key on remote).
 #
 # Interactive by default. Pass --remote to skip the host prompt.
 #
@@ -43,11 +51,17 @@ REMOTE_FILE="$HOME/.gstack-artifacts-remote.txt"
 REMOTE_URL=""
 HOST_PREF=""
 URL_FORM_SUPPORTED="false"
+PUSH_PROTOCOL="auto"
 while [ $# -gt 0 ]; do
   case "$1" in
     --remote) REMOTE_URL="$2"; shift 2 ;;
     --host) HOST_PREF="$2"; shift 2 ;;
     --url-form-supported) URL_FORM_SUPPORTED="$2"; shift 2 ;;
+    --push-protocol)
+      case "$2" in
+        auto|https|ssh) PUSH_PROTOCOL="$2"; shift 2 ;;
+        *) echo "Invalid --push-protocol: $2 (expected auto|https|ssh)" >&2; exit 1 ;;
+      esac ;;
     --help|-h) sed -n '2,32p' "$0" | sed 's/^# \{0,1\}//'; exit 0 ;;
     *) echo "Unknown flag: $1" >&2; exit 1 ;;
   esac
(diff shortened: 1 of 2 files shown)


0xDevNinja:
heads up, the red checks here aren't the diff. it's a fork PR so the runner has no secrets and the eval tests just die on an empty API key before anything actually runs. the rest is green.

if it's easier i can push this to a branch on the base repo and re-point the PR so it picks up secrets. lmk, no rush.

time-attack:
@16francej — recommend closing this PR because #2225 supersedes it with broader handling for providers and explicit remote schemes. This version still fails when the remote is explicitly configured for HTTPS. Please retain the provider-detection tests and attribution under #2225.

0xDevNinja:
closed the gap you flagged rather than closing the PR. the explicit-HTTPS-remote case was real: the original detection only ran for gh/glab-created remotes, so a URL passed via `--remote` (or an existing/manual one) fell through to the SSH default and got rewritten. pushed a fix that tracks where the URL came from and preserves the scheme the user actually supplied, only consulting `git_protocol` for provider-created remotes. `--push-protocol https|ssh` still overrides.

added 4 regression tests, incl. one that pins an explicit HTTPS `--remote` staying HTTPS even with `git_protocol=ssh` reachable. verified they fail on the pre-fix bin. 27 pass.

happy to defer to #2225 if the maintainers prefer that direction, but this branch now covers the same case, keeps the provider-detection tests, and is rebased onto current main.

PR #1480, opened 2026-05-13
fix(build): bun-on-Windows compatibility for the package.json build script

mikepsinn:
## Problem

`bun run build` fails on Windows + Bun 1.3 because three things in the current build script don't parse in bun's POSIX shell on Windows:

1. **Brace-grouped redirect:** `{ git rev-parse HEAD 2>/dev/null || true; } > browse/dist/.version` (and two more for `design/dist`, `make-pdf/dist`). Bun on Windows treats `{` as a literal command and errors with `bun: command not found: {`.
2. **Subshell-grouped redirect:** I tried `( ... ) > FILE` as a fix — bun errors with `Subshells with redirections are currently not supported.`
3. **Glob-no-match cleanup:** `(rm -f .*.bun-build || true)` fails BEFORE the `|| true` has a chance to catch the no-match, because bun's glob rejection happens at parse time.

Net effect: `./setup` on Windows logs a real-looking build then dies with `error: script "build" exited with code 1`, leaving `.version` files unwritten.

## Fix

1. Replace the three brace-grouped `git rev-parse HEAD ... > .version` constructs with a single `node scripts/write-versions.mjs` invocation. Same semantics — write the current HEAD SHA into `browse/dist/.version`, `design/dist/.version`, `make-pdf/dist/.version`, with a `try`/`catch` fallback when not in a git repo. Cross-platform.
2. Remove the final `(rm -f .*.bun-build || true)` cleanup; bun on Windows can't parse it and the build tool already cleans its temp files on success.

`scripts/write-versions.mjs` uses `fileURLToPath(new URL('..', import.meta.url))` (portable to older Node) rather than `import.meta.dirname` (newer-Node-only).

## Verified

- Windows 11 + Bun 1.3.6 + Node 22: `bun run build` now completes and produces `browse.exe`, `find-browse.exe`, `design.exe`, `pdf.exe`, `gstack-global-discover.exe`, `server-node.mjs`, and the three `.version` files (containing the current HEAD SHA).
- macOS/Linux behavior unchanged — the brace-grouped redirects were just version-stamp writes, which the node helper now does cross-platform.

## Notes

- I'm a new gstack user (installed yesterday on Windows). Hitting this build failure was my first 30 seconds with gstack — figured upstreaming was less hassle for everyone than carrying a local fork.
- Happy to adjust the patch shape if you'd rather solve this differently (e.g. a `scripts/build.sh` that runs the brace-grouped writes via `bash` instead of bun's shell). Tell me what you'd prefer and I'll re-shape.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1480"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


garrytan:
Thanks @mikepsinn — Windows build-script Bun compat shipped in v1.42.0.0 Daegu wave (#1594) via @Charlie-El's #1544; your PR is cited as a supersedes ref in the wave commit body.

PR #1483, opened 2026-05-14
feat(browse): add `record` command for video evidence of interactive bug repros

itstimwhite:
## Why

Static screenshots can't capture the timing of a UI bug. Form-submission order, async loading flicker, drag/drop, scroll-triggered behavior, focus management, dialog timing — these all benefit from a few seconds of video. Annotated screenshots stay the right answer for static bugs (typo, clipped UI, missing element), but interactive bugs deserve a `.webm` so the developer fixing them can see the exact repro shape.

`agent-browser` (Vercel Labs) ships a `dogfood` skill that records video evidence as part of its structured QA workflow. I wanted the same evidence shape available behind `$B` — including from inside `/qa` and `/qa-only` — so this PR adds the primitive at the browse layer and leaves the QA-side flag for a follow-up.

## What

A new `record` meta-command with three subactions:

```
$B record start [path] [--size WxH]    # rebuild context with recordVideo
$B record stop                          # close context, flush .webm, return paths
$B record status                        # is recording active?
```

Wraps Playwright's `BrowserContext.recordVideo` option. Implementation hooks into the existing `recreateContext()` save/restore path, so cookies, URLs, and open pages survive both start and stop. `@e` refs are invalidated by the recreate (same caveat as `viewport --scale`); `record stop` prints a hint reminding you to re-snapshot.

Defaults output to a timestamped subdirectory under `TEMP_DIR` so concurrent recordings never collide. Custom paths pass through the standard `validateOutputPath` policy.

Single-recording invariant: calling `record start` while already recording auto-stops the prior recording. Headed mode rejects with a clear error (the user can use their OS screen recorder in headed sessions).

## Tests

9 integration tests in `browse/test/record.test.ts`:

- `record status` reports not-recording before start
- `record stop` is a no-op when not recording
- `record start → activity → record stop` produces a non-empty `.webm` (verified by reading the EBML magic bytes `0x1A 0x45 0xDF 0xA3`)
- Browser remains functional after stop (state preserved across the context recreate)
- `record start` while already recording auto-stops the prior recording
- Unknown flag, malformed `--size`, missing action, unknown subaction all reject with usage errors

All passing locally:

```
 9 pass
 0 fail
Ran 9 tests across 1 file. [2.29s]
```

Existing `browse/test/commands.test.ts` (223 tests) still green.

The pre-existing `snapshot.test.ts` flake (`closetab last tab auto-creates new`) reproduces on `main` without these changes; not related.

## Commits (bisect-friendly)

1. **`feat(browse): add record command …`** — implementation only (`browser-manager.ts`, `commands.ts`, `meta-commands.ts`, new `record.test.ts`). 352 insertions.
2. **`docs(browse): document record command in SKILL.md.tmpl + regenerate`** — `SKILL.md.tmpl` change + regenerated `browse/SKILL.md`, top-level `SKILL.md`, `gstack/llms.txt`. 67 insertions.

Template change and generated-doc regeneration intentionally split, per the gstack contributor guidance.

## Out of scope

- VERSION bump and CHANGELOG entry — left for the merge so you can write the entry in your voice. The implementation includes a `NEW_IN_VERSION: 'record': '1.35.0.0'` so the unknown-command hint works the moment this lands; adjust if you cut a different version.
- `/qa --evidence-per-finding` flag (the QA-side complement) — separate PR coming, since it's a template-only change to two SKILL.md.tmpl files and is easier to review on its own.
- Auto-cleanup of recorded `.webm` files. They're evidence; outliving the session is the point.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -945,6 +945,7 @@ Refs are invalidated on navigation — run `snapshot` again after `goto`.
 | `disconnect` | Disconnect headed browser, return to headless mode |
 | `focus [@ref]` | Bring headed browser window to foreground (macOS) |
 | `handoff [message]` | Open visible Chrome at current page for user takeover |
+| `record start [path] [--size WxH]  |  record stop  |  record status` | Record video of browser activity to .webm. Useful as repro evidence for interactive bug findings. `start` saves session state, rebuilds the context with recordVideo enabled, and restores state; `stop` closes the context (which flushes the .webm files) and returns the paths; `status` prints the active recording dir. Calling `start` while already recording auto-stops the prior recording first. |
 | `restart` | Restart server |
 | `resume` | Re-snapshot after user takeover, return control to AI |
 | `state save|load <name>` | Save/load browser state (cookies + URLs) |
--- a/browse/SKILL.md
+++ b/browse/SKILL.md
@@ -625,6 +625,38 @@ $B screenshot /tmp/out.png --selector .tweet-card
 ```
 Scale must be 1-3 (gstack policy cap). Changing `--scale` recreates the browser context; refs from `snapshot` are invalidated (rerun `snapshot`), but `load-html` content is replayed automatically. Not supported in headed mode.
 
+### 14. Record video evidence for interactive bug repros
+Static screenshots can't capture the timing of a flicker, the order of clicks that triggers a 500, or the cursor-following animation that breaks. For interactive bug findings, record a `.webm` of the repro:
+
+```bash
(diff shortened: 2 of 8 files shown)


time-attack:
Thanks for this. The recording flow can hide flush failures, omit videos from pages closed before stop, and the forced test-process exit can truncate broader runs. Rebase onto the current context lifecycle, remove the forced exit, preserve flush errors, report every artifact, and add a live recording, state, and security check.

PR #1484, opened 2026-05-14
feat(qa,qa-only): add `--evidence-per-finding` evidence layout

itstimwhite:
## Why

The default flat \`screenshots/issue-001-step-1.png\` layout works well for 1-5 findings but gets noisy past that, and it's a chore to hand a single finding to a developer — you have to pluck their three screenshots out of a shared bucket and explain which lines of the report apply.

Companion PR #1483 adds \`\$B record\` (video evidence at the browse layer). This PR adds the QA-side complement so that when an interactive bug is captured on video, the \`.webm\` lives next to the rest of the finding's evidence in its own folder.

The shape mirrors the report-folder pattern that ships in Vercel Labs' \`agent-browser\` \`dogfood\` skill — that skill exists as report-only; ours integrates the same evidence shape into both the report-only \`/qa-only\` and the fix-loop \`/qa\`, with our existing severity/health-score model on top.

## What

Opt-in flag \`--evidence-per-finding\` (or natural-language: \`evidence per finding\`, \`one folder per bug\`). Writes one self-contained folder per finding under a per-run report dir:

\`\`\`
.gstack/qa-reports/qa-report-{domain}-{date}/
├── REPORT.md
├── findings/
│   ├── 001-critical-checkout-500-on-submit/
│   │   ├── finding.md          # severity, repro, env, expected/actual
│   │   ├── step-1.png
│   │   ├── step-2.png
│   │   ├── result.png
│   │   └── repro.webm          # OPTIONAL — present iff \`\$B record\` was active
│   └── 002-high-search-no-results/
│       └── ...
└── baseline.json
\`\`\`

\`finding.md\` schema (defined in the shared methodology):
- Severity / Category / Page / Detected
- What's wrong (one paragraph)
- Repro steps (referencing the step-N.png files)
- Expected vs actual
- Environment (browser, viewport, auth)
- Evidence file index

When to use it (also defined in the methodology, so the LLM picks correctly):

- Run produces ≥5 findings
- Any critical or high severity finding
- An interactive bug has video evidence
- Findings handed off as Linear/Jira tickets (folder zips to one attachment)

When NOT to use it: quick smoke runs, 1-2 findings, regression-mode reruns. The default flat layout stays exactly as it was.

## Implementation note (shape, not size)

Per gstack's prompt-size guidance, the structure and \`finding.md\` template go through the shared \`generateQAMethodology()\` resolver (loaded into both /qa and /qa-only via \`{{QA_METHODOLOGY}}\`). Each leaf \`SKILL.md.tmpl\` only adds one new Setup-table row and a one-line pointer in the Output Structure section. No copy-paste across leaves.

Commits are bisect-friendly per the contributor guide:

1. **\`feat(qa,qa-only): add --evidence-per-finding evidence layout\`** — resolver + both .tmpl source files (78 insertions).
2. **\`docs(qa,qa-only): regenerate SKILL.md\`** — \`bun run gen:skill-docs --host all\` output (146 insertions, generated).

## Verified

- \`bun test test/gen-skill-docs.test.ts test/skill-validation.test.ts\` — **712 pass, 0 fail.** No regressions in resolver or template validation.
- \`bun run skill:check\` — all 10 host-output freshness checks green (one pre-existing \`claude/SKILL.md\` missing-generated warning reproduces on \`main\` without these changes; unrelated).
- Spot-checked rendered output:
  - \`qa/SKILL.md\` and \`qa-only/SKILL.md\` both contain the new Setup row, the Document-phase Evidence layout section, and the Output Structure pointer.
  - The \`{{QA_METHODOLOGY}}\` block expands identically in both files.

## Out of scope

- VERSION bump / CHANGELOG entry — left for the merge so the entry stays in your voice. The flag carries its own discoverability via the Setup table.
- Telemetry on which layout users pick. Worth measuring after this ships if you care.
- Auto-detection ("pick per-finding when ≥5 findings"). Kept explicit for now — the heuristic is documented in the methodology so the LLM can choose, but the parameter is opt-in.

## Pairs with

PR #1483 (\`feat(browse): add record command for video evidence of interactive bug repros\`). The \`record\` primitive is what produces the \`.webm\` that lands in each finding folder. Each PR is reviewable independently; the QA flag works without \`record\` (the .webm just won't be there).

### DIFF
--- a/qa-only/SKILL.md
+++ b/qa-only/SKILL.md
@@ -1054,9 +1054,9 @@ $B snapshot -i -a -o "$REPORT_DIR/screenshots/issue-002.png"
 
 #### Evidence layout: flat (default) vs per-finding
 
-By default, evidence files share one `screenshots/` directory and the report references them by filename (`issue-001-step-1.png`, etc). This stays compact and works well for 1-5 findings.
+**Default is flat.** Evidence files share one `screenshots/` directory and the report references them by filename (`issue-001-step-1.png`, etc). Flat is used unless the user explicitly opts in. This stays compact and works well for 1-5 findings.
 
-When the run is invoked with **`--evidence-per-finding`**, switch to one folder per finding:
+**Opt-in nested layout.** When the run is invoked with **`--evidence-per-finding`** (or natural-language variants: `evidence per finding`, `one folder per bug`), switch to one folder per finding:
 
 ```
 .gstack/qa-reports/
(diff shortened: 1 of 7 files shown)


time-attack:
Thanks for this. The layout advertises video recording that is not available in the current code, while collision, missing-asset, and flat-versus-nested behavior remain undefined. Remove or land the recording dependency, specify those edge cases, add focused layout tests for both QA modes, and resolve the failing evaluations.

itstimwhite:
Addressed review feedback on the evidence layout:

- Removed unavailable `$B record` / `.webm` claims rather than expanding the PR.
- Defined flat default, nested opt-in, deterministic collision handling, and missing/unreadable asset behavior.
- Added focused tests for `/qa` and `/qa-only`.

Local verification: 721 passed, 0 failed; `git diff --check` clean.

Previous `e2e-qa-workflow` error_api/zero-tools and llm-judge missing Anthropic auth were external infra/auth failures; no gates were weakened. New-head CI is running.

PR #1485, opened 2026-05-14
fix(gbrain): probe CLI without command builtin

### LINKED ISSUE #1386: Windows: gstack-memory-ingest.ts can't ingest at scale (follow-up to #1271)
## Summary

`gstack-memory-ingest.ts` (the per-skill transcript ingest writer added in #1298) is fundamentally unable to ingest at scale on Windows. Two layered bugs in the per-page `gbrain put` writer, neither of which has a small-scale workaround for transcript-sized payloads.

Filing as a follow-up to #1271 (Quirk 3) — that issue documented the smoke-test breakage at v0.18.2; this one covers the production-scale failure at v1.26+ and proposes a durable fix.

Repro: 876-transcript backfill on Windows 11 / Git Bash / Bun. Pre-patch: every page errored with `ENOENT: no such file or directory '/dev/stdin'`. Post-patch (refactor below): 466s → 5.6s for a 3-page benchmark, full 876 pages in 8 min.

## Environment

- OS: Windows 11 (10.0.26200), MINGW64_NT-10.0-26200
- Shell: Git Bash via Git for Windows
- Bun: 1.1.2
- gstack: v1.29.0.0 (bug present since v1.26.0.0 / #1298)
- gbrain: v0.18.2
- ingest writer: `bin/gstack-memory-ingest.ts` lines ~750-885

## Bug 1 — `command -v gbrain` probe is Bun-on-Windows incompatible

`gbrainAvailable()` calls `execSync("command -v gbrain", { stdio: "ignore" })` to gate the writer. On Windows + Bun, this throws because Bun's spawn shim doesn't resolve POSIX builtins (`command -v`) the way Bash does — it tries to launch a binary literally named `command`. Effect: writer disables itself silently, every page is reported as "skipped: gbrain CLI not in PATH" even though `gbrain --version` works fine in the same shell.

## Bug 2 — `gbrain put <slug>` reads `/dev/stdin` literally on Windows

After patching past Bug 1, every page fails:

```
[gbrain-error] ENOENT: no such file or directory, open '/dev/stdin'
```

Root cause: `gbrain put` reads content from stdin. On Windows, Bun's `execFileSync` with `input: body` should pipe to the child's stdin, but gbrain's reader appears to open `/dev/stdin` as a literal path rather than reading from fd 0. `/dev/stdin` doesn't exist on Windows.

The `--content` flag (Quirk 3 workaround in #1271) sidesteps stdin but trips the **32K CreateProcess arg-length limit** on transcripts averaging ~970KB. So `--content` is only a workaround for trivially-small smoke-test payloads, not real content.

## Why this matters

`gstack-memory-ingest.ts` is the V1 transcript ingest path (#1298) — the canonical way new sessions land in the brain. On Windows it's been a no-op since v1.26 was released. Brain score on a fresh `/setup-gbrain` install caps at ~10/100 because transcripts never make it in. No error surfaces unless the user runs `--verbose` and notices the per-page skip messages.

## Proposed fix — stage + batch import

`gbrain import <dir> --workers N --json` is the canonical bulk path: parallel workers, mtime/hash dedup, checkpoint/resume on interrupt, no stdin, no per-page CreateProcess. Refactor the writer to:

1. Build the page body (with frontmatter injection — exactly as today).
2. Write it to a staging dir (`~/.gstack/.import-staging-<pid>/`).
3. At end of `ingestPass`, fire one `gbrain import <dir> --workers 4 --json` call.
4. Parse the JSON summary for aggregate counts; cleanup staging.

Trade-off accepted: per-file failure attribution from `import` is aggregate-only (gbrain emits per-file errors to stderr but the summary line only reports counts). For our use case — bulk transcript ingest where retry-next-pass is fine — that's the right trade.

Diff is ~130 insertions / 39 deletions in `gstack-memory-ingest.ts`. Happy to open a PR.

## Benchmark (Windows 11 + Bun + gbrain v0.18.2 + Supabase Postgres ap-southeast-1)

| Scenario              | Before (per-page put)    | After (batch import) |
|-----------------------|--------------------------|----------------------|
| 3-page test           | 466s (all failing)       | 5.6s                 |
| 876-page bulk backfill| ∞ (every page ENOENT)    | 8 min                |
| Brain score (post)    | 10/100                   | 45/100               |
| Embedding coverage    | 0%                       | 99%                  |

## Why not just fix `gbrain put` upstream

The writer-side fix is simpler, doesn't require a gbrain release, and `gbrain import` is already the recommended bulk path in gbrain's own docs. Per-page `gbrain put` would still need a Windows fix (open fd 0 instead of `/dev/stdin`), but that's a gbrain repo concern — not blocking gstack's writer.

## Reproducer

```bash
# Windows 11 + Git Bash + Bun
gstack-memory-ingest --verbose 2>&1 | head -20
# Expected (post-fix): "imported N, skipped M"
# Actual (current main): "skipped: gbrain CLI not in PATH" repeated N times
```

jbetala7:
Fixes #1386

## Summary
- remove the shell-based `command -v gbrain` probe from `gstack-memory-ingest` so Windows/Bun does not try to spawn the POSIX shell builtin as an executable
- use a direct `gbrain --help` probe in `/sync-gbrain` while keeping dry-run code-sync previews independent of local gbrain availability
- add focused regressions that prevent these paths from reintroducing `command -v gbrain`

## Testing
- `bun test test/gstack-gbrain-sync.test.ts`
- `bun test test/gstack-memory-ingest.test.ts`
- `git diff --check`

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -233,15 +233,6 @@ function constrainSourceId(prefix: string, raw: string): string {
   return tail ? `${prefix}-${tail}-${hash}` : `${prefix}-${hash}`;
 }
 
-function gbrainAvailable(): boolean {
-  try {
-    execSync("command -v gbrain", { stdio: "ignore" });
-    return true;
-  } catch {
-    return false;
-  }
-}
-
 // ── Lock file (D1) ─────────────────────────────────────────────────────────
 
 interface LockInfo {
@@ -332,9 +323,6 @@ async function runCodeImport(args: CliArgs): Promise<StageResult> {
   const root = repoRoot();
   if (!root) {
     return { name: "code", ran: false, ok: true, duration_ms: 0, summary: "skipped (not in git repo)" };
-  }
-  if (!gbrainAvailable()) {
-    return { name: "code", ran: false, ok: false, duration_ms: 0, summary: "skipped (gbrain CLI not in PATH)" };
   }
 
   const sourceId = deriveCodeSourceId(root);
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -54,7 +54,7 @@ import {
   rmSync,
 } from "fs";
 import { join, basename, dirname } from "path";
-import { execSync, execFileSync, spawnSync, spawn, type ChildProcess } from "child_process";
+import { execFileSync, spawnSync, spawn, type ChildProcess } from "child_process";
 import { homedir } from "os";
 import { createHash } from "crypto";
 
@@ -809,7 +809,6 @@ let _gbrainAvailability: boolean | null = null;
 function gbrainAvailable(): boolean {
   if (_gbrainAvailability !== null) return _gbrainAvailability;
   try {
-    execSync("command -v gbrain", { stdio: "ignore" });
     // Probe `--help` for the `import` subcommand. gbrain v0.20.0+ ships
     // `import <dir>` (batch markdown import via path-authoritative slugs).
     // If absent, we surface a single clean error here rather than failing
--- a/lib/gbrain-local-status.ts
+++ b/lib/gbrain-local-status.ts
@@ -101,13 +101,13 @@ export function resolveGbrainBin(env?: NodeJS.ProcessEnv): string | null {
   if (_gbrainBinCache.has(key)) return _gbrainBinCache.get(key)!;
   let result: string | null = null;
   try {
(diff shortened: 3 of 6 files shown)


garrytan:
Thank you @jbetala7. Cherry-picked into v1.40.0.0 with attribution preserved (Author: Jayesh Betala). See commit f8e42915. Consolidated v1.40.0.0 fix wave lands as PR #1547. Closing this one as superseded — your contribution is recognized in CHANGELOG.md and the commit metadata.

Issue #1459, opened 2026-05-13
/freeze enforcement chain has three independent install/runtime defects


jerry-sam:
Tested on macOS, GStack installed under `~/.claude/skills/gstack/`. `/freeze` does not actually block Edit/Write/MultiEdit tool calls outside its configured boundary. Investigation revealed three independent issues:

### 1. Install routine does not symlink skill `bin/` directories

`link_claude_skill_dirs()` in `~/.claude/skills/gstack/setup` (lines 373-411) symlinks `SKILL.md` but not `bin/`. Result: `${CLAUDE_SKILL_DIR}/bin/check-freeze.sh` referenced in the SKILL.md hook frontmatter resolves to a non-existent path. Same defect affects `/careful` and `/browse`.

**Workaround applied locally:**
ln -s ~/.claude/skills/gstack/freeze/bin ~/.claude/skills/freeze/bin

### 2. Install routine does not register PreToolUse hooks with Claude Code

Hook frontmatter in SKILL.md is decorative — Claude Code only fires hooks declared in `~/.claude/settings.json` or in a plugin's `hooks/hooks.json`. GStack ships skills, not plugins, and the install routine doesn't write to `settings.json`. Result: even with the bin/ symlink in place, the hook never fires from a real Edit tool call.

**Workaround applied locally** (then reverted, see below):

Manually added to `~/.claude/settings.json`:
```json
{
  "matcher": "Edit|Write|MultiEdit",
  "hooks": [{
    "type": "command",
    "command": "bash ~/.claude/skills/freeze/bin/check-freeze.sh"
  }]
}
```

### 3. With the hook manually registered, deny decision is not enforced

After symlinking `bin/` and adding the `PreToolUse` entry to `settings.json` pointing at `check-freeze.sh`:

- The hook fires on Edit attempts outside the boundary ✓
- A `boundary_deny` telemetry entry gets written to `~/.gstack/analytics/skill-usage.jsonl` ✓
- **BUT** the Edit tool returns success and the file is modified ✗

Test conditions: `/freeze` invoked via the Skill tool with boundary `/tmp/freeze-test-in/`. Attempted Edit on `/tmp/freeze-test-out/foo.txt`. File content changed from `outside original` to `outside modified`. Telemetry confirmed hook fired and classified the operation correctly:
{"event":"hook_fire","skill":"freeze","pattern":"boundary_deny","ts":"2026-05-13T02:11:06Z","repo":"cafe-route-tool"}

[...]

PR #1486, opened 2026-05-14
fix(setup): symlink skill bin/ dirs so PreToolUse hooks resolve

genisis0x:
## Summary

Resolves §1 of #1459 (\`/freeze\` enforcement chain). Companion defects §2 / §3 are intentionally out of scope.

\`link_claude_skill_dirs\` previously only symlinked \`SKILL.md\` into the top-level skill directory, leaving \`bin/\` behind. PreToolUse hook frontmatter references \`\\$CLAUDE_SKILL_DIR/bin/<script>\`; with no symlink the path resolves to a non-existent target and the hook silently no-ops. The same defect affects every skill that ships scripts — \`freeze\`, \`careful\`, \`browse\`, etc.

## What changes

In \`setup\`'s \`link_claude_skill_dirs\` function, after the existing \`SKILL.md\` symlink:

\`\`\`bash
if [ -d \"\\$gstack_dir/\\$dir_name/bin\" ]; then
  if [ -L \"\\$target/bin\" ]; then rm \"\\$target/bin\"; fi
  if [ -e \"\\$target/bin\" ] && [ ! -L \"\\$target/bin\" ]; then rm -rf \"\\$target/bin\"; fi
  ln -snf \"\\$gstack_dir/\\$dir_name/bin\" \"\\$target/bin\"
fi
\`\`\`

Idempotent (rm + \`ln -snf\`) and tolerant of a prior real-directory entry at the target path so re-installs over an older copy upgrade cleanly.

## Out of scope (explicitly)

Defects §2 (registering PreToolUse hooks in user \`~/.claude/settings.json\`) and §3 are out of scope. Both touch user-global state on behalf of \`./setup\` and deserve a separate design call — happy to follow up if you want me to take §2 next.

## Test plan

- [x] \`bash -n setup\` — script syntax clean.
- [x] Diff is one block inside \`link_claude_skill_dirs\`; no other code paths touched.

### DIFF
--- a/setup
+++ b/setup
@@ -577,6 +577,18 @@ link_claude_skill_dirs() {
         if [ -e "$target/sections" ] || [ -L "$target/sections" ]; then rm -rf "$target/sections"; fi
         _link_or_copy "$gstack_dir/$dir_name/sections" "$target/sections"
       fi
+      # Also link bin/ when the source skill ships one. PreToolUse hook
+      # frontmatter references $CLAUDE_SKILL_DIR/bin/<script>; without
+      # this link those paths resolve to a non-existent target and the
+      # hook silently no-ops (issue #1459). Same fix applies to every
+      # skill that publishes scripts -- freeze, careful, browse, etc.
+      # Routes through _link_or_copy so Windows installs without
+      # Developer Mode get a directory copy instead of a frozen symlink.
+      if [ -d "$gstack_dir/$dir_name/bin" ]; then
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej — recommend closing this PR, but only after the bin-sidecar requirement is recorded and tested in #1502. Later work fixed the original hook paths, but selective bin and sibling assets still need ownership there. Please retain attribution and the prefix, Windows, and cleanup cases.

PR #1488, opened 2026-05-14
fix(global-discover): bucket codex sessions by originator

### LINKED ISSUE #1315: gstack-global-discover: session counts conflate originator types and undercount CC by ~5x
## Summary

`gstack-global-discover` reports `sessions.{claude_code,codex,gemini}` counts that misrepresent actual development activity in two ways. `/retro global` (which builds narrative on these counts) inherits both errors.

Tested on gstack 1.26.2.0 with a 31-day window across 9 repos.

## Problem 1 — codex `sessions.codex` conflates Codex Desktop with codex_exec subagent calls

The script counts every `~/.codex/sessions/.../rollout-*.jsonl` whose `payload.cwd` resolves to a repo. But codex rollout files are produced by very different agents, and the `originator` field distinguishes them:

| originator | what it is | dev signal |
|---|---|---|
| `Codex Desktop` | user driving codex CLI interactively | yes — real codex dev |
| `codex_exec` | `codex exec -` (cron / scripts / subagent invocations) | no — usually CC firing codex as cross-model review or external voice |
| `Claude Code` | codex via CC's MCP / subagent integration | no — codex is the tool, CC is the driver |

Real numbers from one user's 31d window:

```
ak-ai-vela:        codex_exec=97  Codex Desktop=1   ← reported as 'codex 98 sessions' but vela was 100% CC dev
ak-where-to-go:    codex_exec=2   Codex Desktop=92  ← real codex dev phase (matches user's recollection)
ai-blogger-lab:    codex_exec=88  Codex Desktop=10  ← codex_exec are CC subagents during dev/review, not codex dev
ak-fund-advisor:   codex_exec=49  Codex Desktop=1   ← all CMR subagents, not codex dev
ak-cc-wiki:        codex_exec=57  Codex Desktop=0   ← CC subagents during wiki writing
```

**Effect**: `/retro global` narrative concluded \"codex was the primary execution tool (414 sessions across 7 repos)\" when in fact codex drove dev for **only one repo's middle phase** (~92 Desktop sessions in where-to-go). The other ~309 codex_exec entries were CC firing codex as cross-model review subagent.

## Problem 2 — CC session count under-reports by ~5x on cron-driven projects

Discovery reported `claude_code: 129` total sessions in the 31d window. A direct scan of `~/.claude/projects/*/`*.jsonl` filtered by `mtime >= since` shows ~621 in the same window.

The biggest gap is `ai-blogger-lab`:
- Discovery: 6 CC sessions
- Direct scan: ~450 CC jsonl files (all CCR / cron-driven; no human in loop)

Plausible causes (didn't fully diagnose):
- Discovery may skip CC project dirs that map to git worktree paths (`-claude-worktrees-*` / `--worktrees-*` show up as separate dirs)
- Or it dedupes by some session_id key that collapses identical-cwd entries

Either way the count is wrong by enough to flip narrative conclusions.

## Suggested fix

1. **Parse codex `payload.originator`** in `scanCodex()` and bucket separately:
   - `codex_desktop_sessions` — interactive dev signal
   - `codex_exec_invocations` — subagent / cron / scripted (annotate as \"called by another agent\" in retro)
   - `claude_code_subagent_invocations` — codex via CC

   Retro narrative should then attribute Codex Desktop counts to \"codex dev\" and codex_exec counts to \"<caller> using codex as subagent\" — not lump them.

2. **Investigate the CC undercount** — likely worktree path normalization. Repro: any project whose CC sessions live in `WorkSpace-<repo>--worktrees-<branch>/` style dirs.

3. **Annotate `/retro global` output** that \"sessions\" = \"tool invocations / file count\", not \"distinct dev sessions\". Especially with CCR / cron drivers in the mix, the same repo can show 450 CC \"sessions\" with zero interactive work.

## Repro

```bash
# Originator breakdown for a 31d window:
python3 -c \"
import os, json, time
since = time.time() - 31*86400
root = os.path.expanduser('~/.codex/sessions')
counts = {}
for dirpath, _, files in os.walk(root):
    for f in files:
        if not (f.startswith('rollout-') and f.endswith('.jsonl')): continue
        fp = os.path.join(dirpath, f)
        if os.stat(fp).st_mtime < since: continue
        with open(fp,'rb') as fh: buf = fh.read(131072)
        line = buf.split(b'\n',1)[0].decode('utf-8','replace')
        d = json.loads(line)
        p = d.get('payload',{})
        repo = os.path.basename(p.get('cwd',''))
        orig = p.get('originator','?')
        counts[(repo,orig)] = counts.get((repo,orig),0)+1
for k,v in sorted(counts.items(), key=lambda x:-x[1]):
    print(f'{k[0]:35s} {k[1]:25s} {v:4d}')
\"

# Compare against discovery output:
~/.claude/skills/gstack/bin/gstack-global-discover --since 31d
```

## Why this matters

Without these fixes, `/retro global` produces confidently-wrong narratives. In my case it told me codex was my main dev tool (driving 7 repos) when codex actually drove dev for one repo's middle phase. CC was the primary driver everywhere — including powering blogger-lab's cron pipeline through CCR. The session counts were the entire foundation of the retro's \"tool usage analysis\" section.

Same shape as a class of bugs already documented in vault wiki: **a metric that conflates distinct things will mislead any narrative built on it, regardless of how thorough the narrative is.**

0xDevNinja:
## Summary

Buckets codex sessions by `payload.originator`. `scanCodex` normalizes it into `{ desktop, exec, claude_code, other }` and surfaces the breakdown at `tools.codex.originators` (global) and `codex_originators` (per-repo), plus the summary view. Existing `codex` totals are untouched — the change is purely additive, so no consumer breaks.

Unknown, missing, null, and non-string originators land in `other` rather than being dropped, so the buckets always sum to the codex total. Whitespace is trimmed before matching.

`Refs #1315` — this covers Problem 1 only. See scope note below.

## Why

`/retro global` narrated "codex was the primary execution tool, 414 sessions across 7 repos" when codex actually drove dev for one repo's middle phase. The other ~309 `codex_exec` entries were CC firing codex as a cross-model review subagent. A single bucket can't tell those apart.

## Scope

Originally this PR carried two patches from #1315. It has since been rebased onto current main and narrowed to Problem 1:

- **Problem 1 (here):** originator bucketing. Still missing on main — `rg "originator"` returns nothing in `bin/gstack-global-discover.ts`.
- **Problem 2 (dropped):** the `extractCwdFromJsonl` cwd-truncation fix landed independently via #1592 / v1.41.1.0 (64KB / 30-line / drop-partial). Main's trade-offs there are deliberate and have regression tests; no reason to re-litigate them from this branch. The `CLAUDE_PROJECTS_DIR` test knob went with it.
- **Problem 3 (not addressed):** the `/retro global` "sessions = file count, not interactive dev" annotation. Belongs in the retro skill template, and probably deserves its own issue.

Because Problems 2 and 3 aren't both resolved here, this uses `Refs` rather than `Fixes` so merging doesn't auto-close #1315.

## Tests

`test/global-discover.test.ts` — 35 pass, 0 fail. Covers each bucket, the whitespace-trim case, missing / null / non-string originator → `other`, and a per-repo sum invariant (`desktop + exec + claude_code + other == sessions.codex`).

## Thanks

@Akagilnc for the review that caught the staleness, and for the `.trim()` and null-originator suggestions — both are in.

### DIFF
--- a/bin/gstack-global-discover.ts
+++ b/bin/gstack-global-discover.ts
@@ -16,9 +16,22 @@ import { homedir } from "os";
 
 // ── Types ──────────────────────────────────────────────────────────────────
 
+// Codex `payload.originator` values map to four buckets so /retro global can
+// distinguish real codex dev (Codex Desktop) from subagent invocations
+// (codex_exec) and CC-driven calls (Claude Code). See issue #1315.
+type CodexOriginator = "desktop" | "exec" | "claude_code" | "other";
+
 interface Session {
   tool: "claude_code" | "codex" | "gemini";
   cwd: string;
+  codexOriginator?: CodexOriginator;
+}
+
+interface CodexOriginatorCounts {
+  desktop: number;
+  exec: number;
+  claude_code: number;
+  other: number;
 }
 
 interface Repo {
@@ -26,6 +39,7 @@ interface Repo {
   remote: string;
   paths: string[];
   sessions: { claude_code: number; codex: number; gemini: number };
+  codex_originators: CodexOriginatorCounts;
 }
 
 interface DiscoveryResult {
@@ -34,7 +48,7 @@ interface DiscoveryResult {
   repos: Repo[];
   tools: {
     claude_code: { total_sessions: number; repos: number };
-    codex: { total_sessions: number; repos: number };
+    codex: { total_sessions: number; repos: number; originators: CodexOriginatorCounts };
     gemini: { total_sessions: number; repos: number };
   };
   total_sessions: number;
@@ -304,6 +318,24 @@ export function extractCwdFromJsonl(filePath: string): string | null {
   return null;
 }
 
+// Codex rollouts ship a free-form `payload.originator` string. Real values
+// seen in the wild: "Codex Desktop" (interactive dev), "codex_exec" (cron /
+// scripted / subagent), "Claude Code" (CC's MCP / subagent integration).
+// Anything else lands in `other` rather than being silently dropped. We trim
+// before lowercasing so a stray trailing space ("Codex Desktop ") still maps.
+function normalizeCodexOriginator(raw: unknown): CodexOriginator {
+  if (typeof raw !== "string") return "other";
+  const v = raw.trim().toLowerCase();
+  if (v === "codex desktop" || v === "codex_desktop") return "desktop";
+  if (v === "codex_exec" || v === "codex exec") return "exec";
+  if (v === "claude code" || v === "claude_code") return "claude_code";
+  return "other";
+}
+
+function emptyOriginatorCounts(): CodexOriginatorCounts {
+  return { desktop: 0, exec: 0, claude_code: 0, other: 0 };
+}
+
 function scanCodex(since: Date): Session[] {
   const sessionsDir = process.env.CODEX_SESSIONS_DIR || join(homedir(), ".codex", "sessions");
   if (!existsSync(sessionsDir)) return [];
@@ -353,7 +385,11 @@ function scanCodex(since: Date): Session[] {
               if (!firstLine) continue;
               const meta = JSON.parse(firstLine);
               if (meta.type === "session_meta" && meta.payload?.cwd) {
-                sessions.push({ tool: "codex", cwd: meta.payload.cwd });
+                sessions.push({
+                  tool: "codex",
+                  cwd: meta.payload.cwd,
+                  codexOriginator: normalizeCodexOriginator(meta.payload.originator),
+                });
               }
             } catch {
               console.error(`Warning: could not parse Codex session ${filePath}`);
@@ -515,8 +551,12 @@ async function resolveAndDeduplicate(sessions: Session[]): Promise<Repo[]> {
     }
 
     const sessionCounts = { claude_code: 0, codex: 0, gemini: 0 };
+    const codexOriginators = emptyOriginatorCounts();
     for (const s of data.sessions) {
       sessionCounts[s.tool]++;
+      if (s.tool === "codex") {
+        codexOriginators[s.codexOriginator ?? "other"]++;
+      }
     }
 
     repos.push({
@@ -524,6 +564,7 @@ async function resolveAndDeduplicate(sessions: Session[]): Promise<Repo[]> {
       remote,
       paths: data.paths,
       sessions: sessionCounts,
+      codex_originators: codexOriginators,
     });
   }
 
@@ -566,13 +607,18 @@ async function main() {
   const codexRepos = new Set(repos.filter((r) => r.sessions.codex > 0).map((r) => r.remote)).size;
   const geminiRepos = new Set(repos.filter((r) => r.sessions.gemini > 0).map((r) => r.remote)).size;
 
+  const codexOriginatorTotals = emptyOriginatorCounts();
+  for (const s of codexSessions) {
+    codexOriginatorTotals[s.codexOriginator ?? "other"]++;
+  }
+
   const result: DiscoveryResult = {
     window: since,
     start_date: startDate,
     repos,
     tools: {
       claude_code: { total_sessions: ccSessions.length, repos: ccRepos },
-      codex: { total_sessions: codexSessions.length, repos: codexRepos },
+      codex: { total_sessions: codexSessions.length, repos: codexRepos, originators: codexOriginatorTotals },
       gemini: { total_sessions: geminiSessions.length, repos: geminiRepos },
     },
     total_sessions: allSessions.length,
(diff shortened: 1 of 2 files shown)


Akagilnc:
Heads-up that this PR is now stale against main. The Problem 2 fix landed independently via #1592 / v1.41.1.0 on 2026-05-20: main's `extractCwdFromJsonl` uses a 64KB cap, scans up to 30 lines, and drops the trailing partial segment before parsing. The current PR is marked CONFLICTING / DIRTY.

The two implementations make different trade-offs rather than one being a strict superset:

- This PR's 128KB read covers a hypothetical >64KB first-line case that main's 64KB read doesn't.
- Main's 30-line scan handles cases where cwd lands past line 15 (which this PR caps at).
- Main's trailing-partial-drop is cleaner at the buffer boundary, but drops a file whose final line has no trailing newline — a corner this PR's implementation does happen to handle.

That said, main is the one that's merged, has regression tests, and chose the 64KB / 30-line / drop-partial mix deliberately. Keeping this PR's Problem 2 patch would mean re-arguing those defaults from scratch.

The unique remaining value here is Problem 1 — bucketing Codex sessions by `payload.originator`. That part is genuinely missing on main (`rg "originator"` returns nothing in current `bin/gstack-global-discover.ts`) and the design looks good: case-insensitive normalization, unknown values land in `other` rather than being dropped, additive (existing totals unchanged), 7 new tests with sum invariants.

Suggestion: rebase onto main, drop the `extractCwdFromJsonl` / `CLAUDE_PROJECTS_DIR` / CC-bigline-test changes, keep only the originator bucketing. Two small additions worth picking up while rebasing:

- `normalizeCodexOriginator` should `.trim()` before lowercasing — currently `"Codex Desktop "` (with a stray space) lands in `other`.
- Add explicit tests for missing / null / non-string `originator` values mapping to `other`.

Happy to open a follow-up PR with just the originator half if that helps — no rush from my side, just flagging since the PR has been quiet for ~16 days.

0xDevNinja:
Rebased onto current main and narrowed the scope to just the codex-originator bucketing, per @Akagilnc's review.

**Dropped** (now redundant): the `extractCwdFromJsonl` 8KB→128KB change and the `CLAUDE_PROJECTS_DIR` test knob. That `cwd`-truncation fix landed independently via #1592 / v1.41.1.0 (64KB / 30-line / drop-partial), and main's deliberate trade-offs there are the ones with regression tests. No reason to re-argue them from this branch.

**Kept** (unique, missing on main — `rg "originator"` returns nothing): the four-bucket classifier so `/retro global` can separate real interactive dev (Codex Desktop) from subagent / scripted runs (codex_exec) and CC-driven calls (Claude Code).

Picked up both follow-ups you flagged:
- `normalizeCodexOriginator` now `.trim()`s before lowercasing, so `"Codex Desktop "` maps to `desktop` instead of `other`.
- Added explicit tests for missing / null / non-string `originator` → `other`, plus a whitespace-trim test.

Buckets are additive (existing totals unchanged) and always sum to the codex total. 35 tests pass.

Akagilnc:
Thanks for the rebase and for picking up both follow-ups — the `.trim()` and the missing/null/non-string tests are exactly what I had in mind, and narrowing to the originator bucketing makes this much easier to reason about.

Three small housekeeping suggestions, all non-blocking and entirely up to you and the maintainers:

1. **The PR description is still the pre-rebase version.** It describes both patches (including the `extractCwdFromJsonl` 8KB→128KB change and the `CLAUDE_PROJECTS_DIR` knob) and reports `26 pass, 0 fail`, whereas the rebased branch drops patch 2 and you mention 35 passing. Might be worth refreshing so reviewers reading the description see the current scope.

2. **`Fixes #1315` in the description vs `Refs #1315` in the commit.** The commit message has it right — `Fixes` is a closing keyword, so merging as-is would auto-close #1315. Since this PR now only covers Problem 1, that would close the issue with Problem 3 (the `/retro global` "sessions = tool invocations" annotation, which you flagged as out of scope) still unaddressed. Changing the description to `Refs #1315` would keep the two in sync.

3. **Problem 2's fix never got linked back to the issue.** It landed in #1592 / v1.41.1.0 on 2026-05-20, but that PR doesn't reference #1315, so the issue currently has no record that half of it is already resolved. Worth a note on #1315 at some point, and Problem 3 probably deserves its own issue.

Either way, the bucketing work here looks good to me. Thanks for sticking with it.

0xDevNinja:
good catches, both real. updated the description: scope is originator-only now, 35 tests, and swapped `Fixes` -> `Refs` so merging doesn't close #1315 with problem 3 still open. the commit already said Refs, i just forgot the body when i rebased.

agreed on the third one too. i'll drop a note on #1315 pointing at #1592 for problem 2 and open a separate issue for the /retro global annotation, that's a retro-template change and doesn't belong here.

Akagilnc:
Thanks for following through on this. Really appreciate the cleanup and for linking everything back together — much easier to follow now.

time-attack:
@16francej Merge this version because it correctly buckets Codex sessions by originator and passed the Mac checks.

PR #1494, opened 2026-05-14
fix(setup): clarify Hermes host banner

### LINKED ISSUE #1201: setup --host hermes is a help banner, not an installer — what does "host install" mean?
## Summary

`./setup --host hermes` is a help banner — `echo` then `exit 0` — not an installer. Compare with `bun run gen:skill-docs --host hermes`, which actually generates Hermes-flavoured skills into `.hermes/skills/`. As a downstream user trying to follow "step 5: Hermes host install" in a multi-agent platform, this leaves "host install" undefined.

## Repro

`~/.claude/skills/gstack/setup` lines 72-81 (current `master`):

```bash
hermes)
  echo ""
  echo "Hermes integration uses the same model as OpenClaw — Hermes spawns"
  echo "Claude Code sessions, and gstack provides methodology artifacts."
  echo ""
  echo "To integrate gstack with Hermes:"
  echo "  1. Tell your Hermes agent: 'install gstack for hermes'"
  echo "  2. Or generate artifacts: bun run gen:skill-docs --host hermes"
  echo ""
  exit 0 ;;
```

By contrast, `bun run gen:skill-docs --host hermes` (script defined in `package.json` → `scripts/gen-skill-docs.ts`) produces 44 skills under `~/.claude/skills/gstack/.hermes/skills/` (`gstack-autoplan`, `gstack-browse`, `gstack-cso`, etc.) — a real artifact tree.

The `claude` and `codex` host paths in `setup` write into `~/.claude/skills/` and `~/.codex/` respectively. Only `openclaw` and `hermes` (and `gbrain`) short-circuit to a banner.

## Question

Is `setup --host hermes` supposed to:

1. Print instructions for the user's Hermes agent to follow internally (current behaviour — but then "step 5: install gstack on the Hermes host" is a misdescription, the agent does the install at runtime, not the user via `setup`)?
2. Write the generated `.hermes/skills/` tree into a configurable target directory (e.g. `--target /path/to/hermes-fork/skills/`) so the user can commit it into their Hermes deployment?
3. Something else (e.g. produce a tarball, hit an API)?

If (1), it would help to say so explicitly in the banner — "There is no host-side install for Hermes; tell your Hermes agent X" — so that downstream adoption docs don't carry a phantom step. If (2), I'd be happy to send a PR adding `--target` once the design is settled.

## Use case

We run a multi-agent platform (mayssam-ai.vc) where Hermes — running in Docker — needs to call gstack skills the same way Claude Code does locally. The natural assumption from the banner is "I run `setup --host hermes` somewhere on the Hermes side and it lays down files." That's not what happens, so step 5 of our gstack adoption plan ([D-057](https://github.com/mayssamj/mayssam-ai.vc) → [D-060](https://github.com/mayssamj/mayssam-ai.vc) follow-ups) was deferred for ambiguity rather than blocked on a real artifact.

Happy to test or contribute.

jbetala7:
Fixes #1201

## Summary
- Clarifies that `./setup --host hermes` is an instruction banner, not a host-side installer.
- Points users at `bun run gen:skill-docs --host hermes` for reviewable `.hermes/skills/` artifacts.
- Updates the README host table so it no longer promises `~/.hermes/skills/gstack-*/` installation.
- Adds a regression test for the Hermes banner wording.

## Verification
- `bun test test/gen-skill-docs.test.ts --test-name-pattern "Hermes host banner"`
- `bash -n setup`
- `./setup --host hermes` shows the banner and exits 0

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ Or target a specific agent with `./setup --host <name>`:
 | Factory Droid | `--host factory` | `~/.factory/skills/gstack-*/` |
 | Slate | `--host slate` | `~/.slate/skills/gstack-*/` |
 | Kiro | `--host kiro` | `~/.kiro/skills/gstack-*/` |
-| Hermes | `--host hermes` | `~/.hermes/skills/gstack-*/` |
+| Hermes | `--host hermes` | Prints integration instructions; `bun run gen:skill-docs --host hermes` writes `.hermes/skills/` |
 | GBrain (mod) | `--host gbrain` | `~/.gbrain/skills/gstack-*/` |
 
 **Want to add support for another agent?** See [docs/ADDING_A_HOST.md](docs/ADDING_A_HOST.md).
--- a/setup
+++ b/setup
@@ -71,12 +71,17 @@ case "$HOST" in
     exit 0 ;;
   hermes)
     echo ""
-    echo "Hermes integration uses the same model as OpenClaw — Hermes spawns"
+    echo "Hermes integration uses the same model as OpenClaw: Hermes spawns"
     echo "Claude Code sessions, and gstack provides methodology artifacts."
+    echo ""
+    echo "./setup --host hermes does not install files into Hermes today."
+    echo "It only prints integration instructions."
     echo ""
     echo "To integrate gstack with Hermes:"
     echo "  1. Tell your Hermes agent: 'install gstack for hermes'"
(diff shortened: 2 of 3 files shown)


time-attack:
Thanks for the PR. We can't merge this version because the README still doesn't explain that Hermes loads skills from each project and that each project must generate them locally. Please rewrite that section and add coverage for that exact guidance.

time-attack:
@16francej close

Issue #1329, opened 2026-05-05
Multiple Claude Code security hook triggers in `/codex` and `/autoplan` skill templates


howie:
## Problem

Running `/codex` or `/autoplan` inside a Claude Code agent session triggers multiple built-in PreToolUse security hooks across different steps of the skill flow, causing repeated user confirmation prompts that break automated execution.

### Pattern 1 — `source gstack-codex-probe` (Step 0 and 0.5)

```bash
source ~/.claude/skills/gstack/bin/gstack-codex-probe 2>/dev/null
```

**Hook message**: `'source' evaluates arguments as shell code`

Claude Code flags all `source` builtins unconditionally. `gstack-codex-probe`'s own header documents the intended usage ("Sourced from template bash blocks; never execute directly"), but the hook has no way to know the file is a zero-side-effect function library.

### Pattern 2 — `eval "$(gstack-paths)"` (Step 0.6)

```bash
eval "$(~/.claude/skills/gstack/bin/gstack-paths)"
```

**Hook message**: `Unhandled node type: string`

Claude Code's static bash AST parser cannot handle tilde-path inside `$()` inside a double-quoted `eval` argument — the resulting `string` node is abandoned. `gstack-paths` explicitly documents `eval "$(gstack-paths)"` as its intended calling convention in the file header.

### Pattern 3 — `cd "$_REPO_ROOT"` after `git rev-parse` (Review / Challenge / Consult modes)

```bash
_REPO_ROOT=$(git rev-parse --show-toplevel) || { echo "ERROR: not in a git repo" >&2; exit 1; }
cd "$_REPO_ROOT"
_gstack_codex_timeout_wrapper 330 codex review ...
```

**Hook message**: C-class (`changes directory before running git`) or no message (silent CWD pollution). `codex exec` already accepts `-C <path>` to set the working directory; the `cd` is unnecessary for those modes.

### Pattern 4 — inline `python3 -u -c "..."` multi-line blocks (Challenge / Consult modes)

```bash
... | PYTHONUNBUFFERED=1 python3 -u -c "
import sys, json
...
    # Fix 2: completeness check — warn if no turn.completed received
    if turn_completed_count == 0:
        ...
"
```

**Two hooks fire simultaneously**:
- AP1 complexity (multi-line + embedded language + nested quotes = score 3/5)
- **Hook message**: `Newline followed by # inside a quoted argument can hide arguments from path validation` — the `# Fix N:` inline Python comments trigger B-class injection detection

[...]

PR #1496, opened 2026-05-14
fix(gstack-paths): add --get <key> so /codex skips eval-tilde footgun

0xDevNinja:
## Summary

- Pattern 2 from #1329: `eval "$(~/.claude/skills/gstack/bin/gstack-paths)"` trips Claude Code's PreToolUse bash AST parser (`Unhandled node type: string`) because of tilde inside `$()` inside a double-quoted eval arg. `/codex` forces a manual approval on every run as a result.
- `gstack-paths --get state-root|plan-root|tmp-root` prints a single path with no `KEY=` prefix, so callers can use:

  ```bash
  PLAN_ROOT=$(~/.claude/skills/gstack/bin/gstack-paths --get plan-root)
  TMP_ROOT=$(~/.claude/skills/gstack/bin/gstack-paths --get tmp-root)
  ```

- Bare invocation stays backward-compatible (`KEY=VALUE` form). No compat break for skills that haven't been migrated yet.
- `codex/SKILL.md.tmpl` Step 0.6 switched to `--get`. `codex/SKILL.md` regenerated via `bun run gen:skill-docs`.

Refs #1329.

## Scope

Fix narrow on purpose — Pattern 2 in `/codex` only. Patterns 1 / 3 / 4 each need their own design (Pattern 1 = split `gstack-codex-probe` functions into binaries; Pattern 3 = subshell or `codex -C`; Pattern 4 = extract inline Python to `gstack-codex-jsonl-parser`). Bundling all four into one PR would balloon the diff across multiple binaries + skill surfaces.

Other skills that still use `eval "$(gstack-paths)"` (`learn`, `freeze`, `investigate`, `context-restore`, `office-hours`, `context-save`, `guard`, `unfreeze`, `plan-tune`) are unchanged here — migration can land one skill at a time on top of this PR with no compat risk because bare invocation still works.

## Tests

7 new cases in `test/gstack-paths.test.ts`:

- `--get state-root` / `plan-root` / `tmp-root` each return the single resolved path, no `KEY=` prefix.
- Each honors the same env override chain as the bare form (`GSTACK_PLAN_DIR`, `TMPDIR`, etc.).
- `--get bogus` exits 1 with `unknown key`.
- `--get` without a key exits 1 with `--get requires a key`.
- `--bogus` top-level flag exits 1 with `unknown argument`.
- Bare invocation still emits `GSTACK_STATE_ROOT=` / `PLAN_ROOT=` / `TMP_ROOT=` (backward-compat guard).

```bash
bun test test/gstack-paths.test.ts
# 15 pass, 0 fail (8 existing + 7 new)

bun test test/skill-validation.test.ts test/gen-skill-docs.test.ts test/gstack-paths.test.ts
# 727 pass, 0 fail
```

## Verification

Manual:

```bash
$ bin/gstack-paths --get plan-root
/Users/lavesh/.claude/plans

$ bin/gstack-paths --get bogus
gstack-paths: unknown key 'bogus' (expected state-root|plan-root|tmp-root)
# exit 1

$ bin/gstack-paths
GSTACK_STATE_ROOT=/Users/lavesh/.gstack
PLAN_ROOT=/Users/lavesh/.claude/plans
TMP_ROOT=/var/folders/.../T/
```

### DIFF
--- a/bin/gstack-paths
+++ b/bin/gstack-paths
@@ -1,7 +1,15 @@
 #!/usr/bin/env bash
 # gstack-paths — output portable state-root paths for skill bash blocks
-# Usage: eval "$(gstack-paths)"  → sets GSTACK_STATE_ROOT, PLAN_ROOT, TMP_ROOT
-# Or:    gstack-paths            → prints GSTACK_STATE_ROOT=... etc.
+# Usage: eval "$(gstack-paths)"        → sets GSTACK_STATE_ROOT, PLAN_ROOT, TMP_ROOT
+#        gstack-paths                  → prints GSTACK_STATE_ROOT=... etc.
+#        gstack-paths --get <key>      → prints a single resolved path (no KEY= prefix)
+#                                         where <key> is state-root | plan-root | tmp-root
+#
+# The --get form was added so callers can use `PLAN_ROOT=$(gstack-paths --get plan-root)`
+# instead of `eval "$(gstack-paths)"`. The eval form trips Claude Code's PreToolUse
+# bash AST parser (tilde inside $() inside double-quoted eval arg = "Unhandled node
+# type: string"), forcing manual approval on every skill that needs portable roots.
+# See issue #1329 Pattern 2.
 #
 # Resolves three roots with explicit fallback chains so skills work the same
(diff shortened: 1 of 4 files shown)


time-attack:
@16francej Close this in favor of adding its useful selector to the repaired #1506. This implementation mishandles arbitrary path data and lacks special-character coverage.

Issue #1503, opened 2026-05-14
/codex review: bare-path drops filesystem boundary on diffs that touch .claude/skills/


garrytan:
## Context

v1.34.2.0 (#1478, fixing #1428) shipped a dual-path approach for `/codex review` on Codex CLI ≥0.130.0:

- **Bare path:** `codex review --base <branch>` (no prompt) when the user didn't pass custom instructions.
- **Custom path:** `codex exec` + tempfile prompt with DIFF_START/DIFF_END delimiters when `/codex review <focus>`.

The bare path is correct for the new CLI semantics, but it surfaces an adjacent issue: the filesystem boundary instruction (the "do not read files under ~/.claude/, .claude/skills/, agents/, ~/.agents/" preamble) was previously prefixed to every `codex review` call as part of the prompt. The bare path no longer carries that prompt, so the boundary is dropped.

## Impact

When the diff being reviewed happens to touch `.claude/skills/`, `agents/`, or any of the skill-template paths, codex may spend tokens reading those files — they're skill definitions meant for a different AI system, so reading them is wasted work AND the model can be distracted by their instructions.

This is a token-efficiency concern, not a safety concern (the skill files are public). But on large diffs that touch skills (which gstack itself does regularly), it can mean codex spends a meaningful fraction of its budget on the wrong files.

## Repro

1. Make a commit that modifies any file under `.claude/skills/` or `agents/`.
2. Run `/codex review` with no custom instructions.
3. Observe codex's stdout — if any tool calls read skill files (e.g. \`rg \"^---\" .claude/skills/...\`), the boundary would have prevented those reads on the pre-0.130.0 prompted path.

## Possible fixes

- Push the boundary into the codex CLI as a `system_prompt` arg if/when 0.130.0+ exposes one (the bare path's note already mentions checking for this).
- Land a `~/.codex/AGENTS.md` opt-in for users who want a persistent boundary across all `codex review` calls.
- Detect skill-file paths in the diff and bail out of the bare path into the exec path with a temp-prompt that carries the boundary.

## References

- v1.34.2.0 PR: #1478
- Original issue: #1428
- gstack v1.36.0.0 plan analysis: discussed in D1 of the bug-fix wave plan; we explored a unified \`codex exec\` route (option A) that preserves the boundary in all cases, but #1478 had already shipped option B by the time we got to /ship. Captured here as follow-up.

Issue #1454, opened 2026-05-12
Runs CEO → design → eng review: no actionable steps/tasks for build


seanlabs:
I am running office-hours, CEO, design, and eng-review skills, but reviewing the *.md output I don't see any actionable content that would steer Claude Code on what to build. I was expecting more broken down components and steps into a task list that would drive the Claude Code or Codex build.

Hchumpitaz:
For software development, I also recommend using OpenSpec (https://github.com/Fission-AI/OpenSpec).

OpenSpec is a Spec-Driven Development (SDD) framework that resides in openspec/ within your repository. Its purpose: to allow you and the AI ​​to agree on what will be built before touching any code, by writing specs in Markdown that then become the source of truth for the system's behavior.

openspec/
├── specs/              # Source of truth (your system's behavior)
│   └── <domain>/spec.md
├── changes/            # Proposed updates (one folder per change)
│   └── <change-name>/
│       ├── proposal.md   # Why and what
│       ├── design.md     # How (technical)
│       ├── tasks.md      # Implementation Checklist
│       └── specs/        # Delta specs (what's changing)
└── config.yaml      # Proyect configuration (optional)

garrytan:
Fixed in v1.38.1.0 — review pipeline now produces explicit Implementation Tasks. Closing. Reopen if the actionable-step gap persists on latest.

Issue #1452, opened 2026-05-12
brain-allowlist misses /office-hours design docs and /plan-eng-review test plans at project root


nitzanvolman:
## Problem

The `.brain-allowlist` generated by `gstack-artifacts-init` (current at v1.27.0.0+, see `bin/gstack-artifacts-init:224-234`) includes patterns for the `designs/` and `ceo-plans/` subdirectories:

```
projects/*/designs/*.md
projects/*/designs/*/*.md
projects/*/ceo-plans/*.md
projects/*/ceo-plans/*/*.md
```

But several skills write their output at the **project root**, not in those subdirectories:

| Skill | Actual write path | Source |
|---|---|---|
| `/office-hours` Builder/Startup | `projects/{slug}/{user}-{branch}-design-{datetime}.md` | `office-hours/SKILL.md:1566` |
| `/plan-eng-review` | `projects/{slug}/{user}-{branch}-eng-review-test-plan-{datetime}.md` | `plan-eng-review/SKILL.md:1149` |
| `/autoplan` (test plan) | `projects/{slug}/{user}-{branch}-test-plan-{datetime}.md` | `autoplan/SKILL.md:1334` |

Cross-skill readers correctly glob these at the project root:

- `plan-ceo-review/SKILL.md:897` — `ls -t ~/.gstack/projects/$SLUG/*-design-*.md`
- `plan-eng-review/SKILL.md:801` — same
- `autoplan/SKILL.md:841,1003` — same

So the files belong at root; the **allowlist is the inconsistent piece**.

## Effect

With `artifacts_sync_mode=full`, `gstack-brain-sync` still skips:

- `/office-hours` design docs
- `/plan-eng-review` test plans
- `/autoplan` test plans

CEO plans sync correctly (they live under `ceo-plans/`), but the **input designs they reference do not**. This breaks the cross-machine "ceo-plan references the design" chain — pull on machine B, you get the CEO plan but not the underlying design doc.

This has the same shape as #1441 (v1.27.0.0 rename missed the config key in one place): the rename release moved/renamed things but the allowlist was never audited against current skill write paths.

## Repro

1. Run `/office-hours` Builder mode → produces `~/.gstack/projects/{slug}/*-design-*.md` (at root)
2. `gstack-artifacts-init --host github`
3. `gstack-config set artifacts_sync_mode full`
4. Run any gstack skill
5. `gstack-brain-sync --status` — design doc absent from queue
[...]

Issue #1441, opened 2026-05-12
/setup-gbrain Step 7 sets wrong config key (gbrain_sync_mode); brain-sync reads artifacts_sync_mode → federation silently never runs


topcoder1:
## Bug

\`/setup-gbrain\` Step 7 (\"Offer gstack-brain-sync + wire it into gbrain\") writes:

\`\`\`bash
~/.claude/skills/gstack/bin/gstack-config set gbrain_sync_mode artifacts-only
\`\`\`

But \`~/.claude/skills/gstack/bin/gstack-brain-sync\` reads a different config key — \`artifacts_sync_mode\` (lines 73, 239, 337). The mismatch means:

- User goes through \`/setup-gbrain\`, picks \"yes, artifacts-only\" at the federation prompt
- gstack-config stores \`gbrain_sync_mode=artifacts-only\` ✅
- \`gstack-brain-sync\` checks \`artifacts_sync_mode\` → not set → returns \`off\`
- \`--once\` and preamble auto-syncs **silently no-op**
- \`--status\` shows \`mode: off\`
- ~/.gstack/ never gets pushed to the federated GitHub repo
- Multi-Mac users never see each other's plans/retros/learnings

The federation prompt is acknowledged successfully and gives no error, but the underlying mechanism is dormant.

## Repro

\`\`\`bash
/setup-gbrain  # accept federation prompt with \"yes, artifacts-only\"
~/.claude/skills/gstack/bin/gstack-config get gbrain_sync_mode
# → artifacts-only
~/.claude/skills/gstack/bin/gstack-brain-sync --status
# → {\"queue_depth\":0,\"last_push\":\"never\",\"mode\":\"off\"}
~/.claude/skills/gstack/bin/gstack-brain-sync --once
# silently does nothing — no commit, no push
git -C ~/.gstack log --oneline
# → only the initial gstack-brain-init commit, no sync commits
\`\`\`

Then:

\`\`\`bash
~/.claude/skills/gstack/bin/gstack-config set artifacts_sync_mode artifacts-only
~/.claude/skills/gstack/bin/gstack-brain-sync --once
# now pushes N file(s) successfully
\`\`\`

## Verified on

- gstack v1.31.1.0 (current upstream HEAD as of 2026-05-11)
- gbrain v0.32.0
- macOS, PGLite engine, single-Mac install with federated GitHub repo target

## Suggested fix

[...]

PR #1504, opened 2026-05-14
v1.38.1.0 fix wave: surrogate-safe page captures (#1440), Implementation Tasks across review skills (#1454), root-level artifact patterns (#1452)

garrytan:
## Summary

Three filed issues land together as a coordinated bug-fix wave.

**Browse — Page captures with mixed-script Unicode round-trip cleanly to the Claude API (#1440)**
- `browse/src/sanitize.ts` (new) — two surrogate-stripping utilities (`stripLoneSurrogates`, `stripLoneSurrogateEscapes`) plus `sanitizeBody` that picks the right pass for text/plain vs application/json bodies.
- `buildCommandResponse` extracted from `handleCommand` and exported for testability. Sanitizes every response body before `new Response()`.
- `/batch` was bypassing the chokepoint via its own `JSON.stringify({results, ...})`. Each per-result string is now sanitized AND the envelope gets a second `stripLoneSurrogateEscapes` pass on the JSON-escape form.
- Defense in depth at the highest-volume DOM-text emitters: `getCleanText`, `getCleanTextWithStripping`, `html`, `accessibility`, all three return points in `snapshot.ts`.

**Review skills — Implementation Tasks checklist + JSONL handoff (#1454)**
- `plan-ceo-review`, `plan-design-review`, `plan-eng-review`, `plan-devex-review` each emit a markdown `## Implementation Tasks` section AND write a `jq`-built JSONL artifact to `~/.gstack/projects/$SLUG/tasks-{phase}-{datetime}.jsonl`. Never hand-rolled echo — `jq -nc` keeps the file valid when titles/findings contain quotes, newlines, or backslashes.
- `/autoplan` Phase 4 reads all four JSONL files, scopes by current branch + 5-commit window, dedupes on exact `(component, sorted(files), title)`, renders aggregated tasks inside the Final Approval Gate.
- Standalone review runs (`/plan-eng-review` alone, etc.) produce their own task list and JSONL even outside autoplan — the JSONL is the handoff contract.
- New resolvers: `scripts/resolvers/tasks-section.ts` (emit + aggregate), `scripts/task-emission-schema.ts` (shared schema).

**Federation sync — Root-level design + test-plan docs (#1452)**
- `bin/gstack-artifacts-init` adds `projects/*/*-design-*.md` and `projects/*/*-test-plan-*.md` to all three managed blocks: `.brain-allowlist`, `.brain-privacy-map.json` (class `artifact`), `.gitattributes` (`merge=union`).
- Idempotent migration `gstack-upgrade/migrations/v1.36.0.0.sh` patches existing installs in place. Uses `jq` for the JSON file (preserves validity). NEVER calls `gstack-artifacts-init` — that would `git commit + push` and clobber user state.

**Issue closed without code: #1441** — verified already fixed in v1.27.0.0 (alignment of `setup-gbrain` and `gstack-brain-sync` on the `artifacts_sync_mode` key, plus the v1.27.0.0 migration). Comment posted, issue closed.

**Follow-up filed: #1503** — the v1.34.2.0 fix for #1428 uses a dual-path approach where the bare path no longer carries the filesystem boundary. Surfaces a token-efficiency concern when diffs touch `.claude/skills/`. Documented, not blocking this release.

## Test Coverage

```
[+] browse/src/sanitize.ts                  [★★★ TESTED]  browse/test/sanitize.test.ts (18 cases)
[+] browse/src/server.ts buildCommandResponse [★★★ TESTED] browse/test/build-command-response.test.ts (7 cases)
[+] browse/src/server.ts /batch sanitize    [★★ TESTED]    Existing batch.test.ts + manual smoke
[+] browse/src/read-commands.ts wraps       [★ COVERED]    via getCleanText round-trip
[+] browse/src/snapshot.ts wraps            [★ COVERED]    via existing snapshot tests (44 pass)
[+] bin/gstack-artifacts-init blocks        [★★★ TESTED]  test/artifacts-init-migration.test.ts (7 cases)
[+] gstack-upgrade/migrations/v1.36.0.0.sh  [★★★ TESTED]  same — idempotency, missing-file, jq-absent paths
[+] scripts/resolvers/tasks-section.ts      [★★ TESTED]    via gen-skill-docs tests (385 pass) + template-render

COVERAGE: 8/8 new code paths tested (100%)  |  Code: 5/5  |  Skills: 3/3
QUALITY: ★★★:4 ★★:3 ★:1  |  GAPS: 0
```

Tests: full free-tier `bun test` green on the impacted files (sanitize ×18, build-command-response ×7, artifacts-init-migration ×7, host-config ×73, gen-skill-docs ×385, content-security ×58, snapshot ×44 pass + 1 pre-existing closeTab failure unrelated to this work).

## Pre-Landing Review

No structural issues. Diff has no SQL, no auth surface, no new LLM trust boundary. The sanitizer regexes use ES2018+ negative lookbehind, which Bun supports. The migration is per-file independent (allowlist missing doesn't skip privacy-map repair) and idempotent (re-runs are no-ops). `/codex review` ran twice during planning (28 findings total, all incorporated except those that became #1503 follow-up after the v1.34.2.0 collision).

## Plan Completion

Plan at `~/.claude/plans/system-instruction-you-are-working-sorted-teacup.md`. All planned items shipped:
- ✅ Fix 4 (surrogate sanitization) — chokepoint + defense in depth + JSON-escape variant + /batch
- ✅ Fix 1 (Implementation Tasks) — 4 review skills + autoplan aggregator + JSONL schema
- ✅ Fix 2 (allowlist + privacy + gitattributes + migration)
- ✅ Fix 3 (#1441 verified, closed with comment)
- ⏭️ Fix 5 (#1428) — dropped from this branch; already shipped by v1.34.2.0 with a different approach. Follow-up #1503 filed for the boundary-drop concern.

## TODOS

## Test plan
- [x] `bun test` passes on browse/test/sanitize.test.ts (18 cases)
- [x] `bun test` passes on browse/test/build-command-response.test.ts (7 cases)
- [x] `bun test` passes on test/artifacts-init-migration.test.ts (7 cases)
- [x] `bun test` passes on test/host-config.test.ts (73 — golden fixtures refreshed)
- [x] `bun test` passes on test/gen-skill-docs.test.ts (385 — all hosts regenerated)
- [x] `bun test` passes on browse/test/content-security.test.ts (58 — source-level regression updated to accept post-sanitize form)
- [x] Migration smoke-tested on tmpdir fixture: adds patterns, preserves user customizations, jq keeps JSON valid, re-run is no-op
- [x] `bun run gen:skill-docs` regenerates cleanly for all 12 hosts
- [x] Issue #1441 closed with comment citing v1.27.0.0 migration
- [x] Follow-up #1503 filed for the bare-path boundary-drop concern


scripts/resolvers/tasks-section.ts
```diff
@@ -0,0 +1,168 @@
...
+      # Replace with reduced version for this phase, accumulating others
+      jq -c --arg phase "$phase" 'select(.phase != $phase)' "$ALL_JSONL" > "$ALL_JSONL.other" 2>/dev/null || true
+      cat "$ALL_JSONL.other" "$ALL_JSONL.phase" > "$ALL_JSONL"
+      rm -f "$ALL_JSONL.phase" "$ALL_JSONL.other"
+    fi
+  done
+
+  # Exact-match dedup by (component, sorted(files), title). Non-matches kept
+  # separately with a possible-duplicate marker injected by the renderer.
+  AGGREGATED_TASKS=$(jq -s \\
```
NikhileshNanduri, line 139: **Near-duplicate annotation is mentioned but never implemented.**

The comment says "possible-duplicate marker injected by the renderer" but the jq that follows does:
```jq
group_by([.component, (.files | sort), .title])
| map(sort_by(...) | .[0])
```
This is exact-match dedup only. Tasks with the same logical intent but slightly different wording or file lists (e.g., the same sanitizer bug flagged by both CEO and eng review, one listing `browse/src/sanitize.ts` and the other also adding `browse/src/server.ts`) get neither collapsed nor annotated.

The PR description says:
> Near-duplicates surface separately with a possible-duplicate note for human resolution.

Neither the template code nor the generated SKILL.md injects such a note. Either:
1. Remove the "possible-duplicate marker injected by the renderer" comment (and fix the matching schema docstring at `scripts/task-emission-schema.ts:56`), or
2. Implement the annotation: after `group_by`, tasks that share the same `title` but differ in `files` could emit a sibling with a `note` field appended to the markdown line.

scripts/task-emission-schema.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Dedup key for the aggregator. Two tasks collapse into one ONLY when this
+ * tuple is identical (per `D13 finding 9`). Near-duplicates surface as
+ * separate tasks with a `possible-duplicate-of: <id>` note.
+ */
+export function dedupKey(t: Pick<ImplementationTask, 'component' | 'files' | 'title'>): string {
+  return JSON.stringify({
+    component: t.component,
```
NikhileshNanduri, line 57: **`dedupKey` is exported but never called — it's dead code.**

The aggregator in `scripts/resolvers/tasks-section.ts` reimplements the dedup logic inline as a jq filter. This TypeScript function is never imported or invoked. If it's meant as documentation of the dedup algorithm, a comment on the jq block would be clearer. If it's meant for a future TypeScript-based aggregator, it should be marked `// TODO: wire up when aggregator moves to TypeScript`.

As-is, a future author could change the jq key and forget to update this, creating silent divergence.

scripts/resolvers/tasks-section.ts
```diff
@@ -0,0 +1,168 @@
...
+COMMITS_RECENT=$(git log --format=%H -n 5 2>/dev/null | tr '\\n' '|' | sed 's/|$//')
+
+AGGREGATED_TASKS=""
+if command -v jq >/dev/null 2>&1; then
+  # Collect entries from all 4 phases, scoped to current branch + commit window.
+  # For each phase, keep only the latest run_id. Within the surviving set,
+  # dedupe by (component, sorted(files), title) — exact match only.
+  # Sort by priority (P1 > P2 > P3) then by phase order.
+  ALL_JSONL=$(mktemp -t autoplan-tasks.XXXXXXXX)
+  for phase in ceo-review design-review eng-review devex-review; do
```
NikhileshNanduri, line 113: **JSONL files accumulate indefinitely in `~/.gstack/projects/$SLUG/`.**

Each review run writes a new `tasks-{phase}-{datetime}.jsonl`. The aggregator reads the latest run_id and ignores older files, but they're never deleted. After 50+ autoplan runs over a few months, this directory fills with hundreds of JSONL files.

Consider one of:
1. Use a canonical filename like `tasks-{phase}-latest.jsonl` (overwrite, not append) — simpler, and the timestamp in `run_id` inside each record already provides the freshness information the aggregator needs.
2. Prune stale files at write time: before writing the new file, delete any existing `tasks-{phase}-*.jsonl` for the same phase.
3. At minimum, document the accumulation behavior so users know to occasionally `rm ~/.gstack/projects/$SLUG/tasks-*.jsonl`.

gstack-upgrade/migrations/v1.38.1.0.sh
```diff
@@ -0,0 +1,102 @@
...
+${PATTERN}
+" "${ALLOWLIST}" && rm -f "${ALLOWLIST}.bak"
+        added_any=1
+      else
+        # Marker missing — append at end of file as a fallback. User may have
+        # custom-edited the file; better to add than skip silently.
+        printf '%s\n' "${PATTERN}" >> "${ALLOWLIST}"
+        added_any=1
+      fi
+    fi
```
NikhileshNanduri, line 55: **`sed -i.bak` inline insert: verify on GNU sed / Alpine.**

The multi-line form:
```bash
sed -i.bak "/^# ---- USER ADDITIONS BELOW/i\\
${PATTERN}
" "${ALLOWLIST}"
```
works on macOS BSD sed and GNU sed on Ubuntu/Debian, but some minimal Linux distributions (Alpine with musl sed, BusyBox) interpret the `i\` command differently — they require the text on the SAME line as the command, not on the next line.

The current patterns (`projects/*/*-design-*.md`, `projects/*/*-test-plan-*.md`) don't contain backslashes or `&` chars so the pattern itself is safe. But the portability concern means any future caller who extends `NEW_PATTERNS` with a path containing `\` or adds a Linux CI matrix on Alpine would hit this.

Safer cross-platform alternative using `awk`:
```bash
awk -v pat="${PATTERN}" '/^# ---- USER ADDITIONS BELOW/ { print pat } { print }' "${ALLOWLIST}" > "${ALLOWLIST}.tmp" && mv "${ALLOWLIST}.tmp" "${ALLOWLIST}"
```
No backup file to clean up, works on all POSIX awk variants.

browse/src/sanitize.ts
```diff
@@ -0,0 +1,34 @@
...
+const LONE_SURROGATE_LOW = /(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g;
+
+export function stripLoneSurrogates(s: string): string {
+  return s.replace(LONE_SURROGATE_HIGH, '�').replace(LONE_SURROGATE_LOW, '�');
+}
+
+// Matches \uD8XX-\uDFXX escape text where the pair is not completed by an
+// adjacent \uDC00-\uDFFF (high) or preceded by \uD800-\uDBFF (low).
+const LONE_SURROGATE_HIGH_ESCAPE = /\\u[Dd][89ABab][0-9A-Fa-f]{2}(?!\\u[Dd][C-Fc-f][0-9A-Fa-f]{2})/g;
+const LONE_SURROGATE_LOW_ESCAPE = /(?<!\\u[Dd][89ABab][0-9A-Fa-f]{2})\\u[Dd][C-Fc-f][0-9A-Fa-f]{2}/g;
```
NikhileshNanduri, line 23: **`[C-Fc-f]` character class in `LONE_SURROGATE_HIGH_ESCAPE` — correct but worth a note.**

`[C-Fc-f]` matches C, D, E, F (uppercase) and c, d, e, f (lowercase), which correctly covers the second hex digit of low-surrogate escapes (`\uDC00`–`\uDFFF`). The regex is correct.

However, the lookahead in `LONE_SURROGATE_HIGH_ESCAPE` checks `(?!\\u[Dd][C-Fc-f][0-9A-Fa-f]{2})` — this correctly validates that a high-surrogate escape is NOT followed by a low-surrogate escape, leaving valid pairs (`\uD83D\uDE00` = 😀) untouched.

Minor: adding a test case for the boundary escape `\uDBFF\uDC00` (both boundary values of valid pair) would close the coverage gap on the edge of the character class ranges. The existing test covers `\uD83D\uDE00` which is mid-range; the boundary pair verifies no off-by-one in the `[89ABab]`/`[C-Fc-f]` classes.

NikhileshNanduri (commented):
## Review: v1.36.0.0 bug-fix wave

Solid PR overall. The surrogate sanitization is well-designed (chokepoint + defense in depth + `/batch` envelope), the migration is idempotent, and the test coverage is thorough. A few findings below.

**Summary:**
- **Bug (near-dup annotation)**: The "possible-duplicate marker injected by the renderer" comment + schema docstring promise behavior that the jq implementation doesn't deliver. Near-duplicates surface without any cross-reference marker.
- **Dead code**: `dedupKey` in `task-emission-schema.ts` is exported but never called.
- **Accumulation**: JSONL task files grow unboundedly with no housekeeping.
- **Migration sed**: The `sed -i.bak` inline insert syntax has a subtle portability note.

PR #1509, opened 2026-05-14
fix: /freeze and /careful enforcement chain — three independent bugs (#1459)

NikhileshNanduri:
## Summary

Fixes issue #1459: `/freeze` and `/careful` provided zero actual enforcement. An Edit outside the freeze boundary would succeed silently despite telemetry logging a `boundary_deny` event. Three independent bugs in the install/runtime chain were all required:

| Bug | Root cause | Symptom |
|-----|-----------|---------|
| **Bug 1** | `link_claude_skill_dirs` symlinked `SKILL.md` but not `bin/` | `${CLAUDE_SKILL_DIR}/bin/check-freeze.sh` resolved to a non-existent path — hook invocation always fails silently |
| **Bug 2** | SKILL.md frontmatter `hooks:` is decoration — Claude Code only reads `~/.claude/settings.json` | Hook never fired at all from a real Edit/Write call |
| **Bug 3** | Scripts returned `{"permissionDecision":"deny","message":"..."}` — flat object, unrecognised by Claude Code | Hook fired, deny was silently ignored; tool executed anyway |

## Changes

**Bug 1 — bin/ symlink** (`setup`, `bin/gstack-relink`):
- `link_claude_skill_dirs()` now symlinks each skill's `bin/` alongside `SKILL.md`
- `gstack-relink` has the same fix for the upgrade path

**Bug 2 — hook registration** (`bin/gstack-settings-hook`, `setup`):
- `gstack-settings-hook` gains `add-pretooluse <matcher> <command>` and `remove-pretooluse` actions (idempotent, atomic write via `.tmp` + rename)
- `setup` Step 11 calls these at install time to write freeze (Edit|Write|MultiEdit) and careful (Bash) entries into `~/.claude/settings.json`

**Bug 3 — hookSpecificOutput format** (`freeze/bin/check-freeze.sh`, `careful/bin/check-careful.sh`):
- Both scripts now output the correct envelope: `{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"..."}}`
- `check-freeze.sh` also resolves its state root via `gstack-paths` to honour `GSTACK_HOME` in addition to `CLAUDE_PLUGIN_DATA`
- `check-careful.sh` is now a no-op unless `~/.gstack/careful-active.txt` exists — safe to leave the hook registered globally without polluting sessions where `/careful` was never invoked. `/careful` SKILL.md now writes that file and documents how to clear it

**Tests** (`test/hook-scripts.test.ts`):
- All careful tests use `withCarefulActive()` helper and pass `CLAUDE_PLUGIN_DATA` to the hook
- Freeze tests assert the `hookSpecificOutput` envelope shape (2 new tests)
- 35 tests passing, 0 failing

## Test plan

- [x] `bun test test/hook-scripts.test.ts` — 35 pass, 0 fail
- [x] Pre-existing doc-inventory failures for `document-generate` confirmed pre-existing on main (stash test)
- [ ] End-to-end: run `./setup`, invoke `/freeze src/`, attempt Edit on a file outside `src/` — should be blocked with `[freeze] Blocked: ...`
- [ ] `/careful` + `rm -rf /var` — should see permission ask prompt


time-attack:
@16francej — recommend closing this PR, but it should close only after the response-envelope and shared state-root fixes are ported and attributed under #1459. The broader changes are mostly obsolete, and unresolved safety problems remain. Stale global-hook changes should not be carried forward.

PR #1514, opened 2026-05-15
feat: make learnings search limits configurable via gstack-config

MrAladdin:
## Summary

- Add `learnings_preamble_limit` and `learnings_skill_limit` config keys to `gstack-config`
- Preamble and skill-internal learnings searches now read limits dynamically from `~/.gstack/config.yaml` instead of using hardcoded values (3 and 10)
- Users with large-context models (200K+) can increase limits to load more project memory per session

## Motivation

The hardcoded `--limit 3` (preamble) and `--limit 10` (skill) were designed for the lowest-common-denominator model context window. Users running Claude Opus, GPT-4.5, or other large-context models have ample room to load 20-30 learnings without meaningful token pressure (~2000 tokens for 20 entries). Making this configurable lets power users get more value from accumulated project memory.

## Changes

| File | Change |
|------|--------|
| `bin/gstack-config` | Add `learnings_preamble_limit` (default: 3) and `learnings_skill_limit` (default: 10) to defaults table, header comments, and list/defaults enumeration |
| `scripts/resolvers/preamble/generate-preamble-bash.ts` | Replace `--limit 3` with dynamic `$(gstack-config get learnings_preamble_limit)` |
| `scripts/resolvers/learnings.ts` | Replace `--limit 10` with dynamic `$(gstack-config get learnings_skill_limit)` |
| `test/gen-skill-docs.test.ts` | Update assertion from literal `--limit 3` to `--limit "$_LEARN_PREAMBLE_LIMIT"` |
| `*/SKILL.md` | Regenerated (no template changes) |

## Usage

```bash
# Increase preamble learnings (loaded every skill start)
gstack-config set learnings_preamble_limit 20

# Increase skill-internal learnings search
gstack-config set learnings_skill_limit 30

# Check current values
gstack-config defaults | grep learnings
```

## Backwards compatibility

- Default values match the previous hardcoded behavior (3 and 10)
- Users who never set these keys see zero behavior change
- One extra `gstack-config get` call per preamble (~5ms, same as existing config reads)

## Test plan

- [x] `bun test test/gen-skill-docs.test.ts` — 385 pass, 0 fail
- [x] `bun test browse/test/gstack-config.test.ts` — 17 pass, 3 fail (pre-existing, unrelated)
- [x] `gstack-config get/set/list/defaults` all display new keys correctly
- [x] Generated SKILL.md files contain dynamic limit variable
- [x] `--host hermes` generation produces correct `$GSTACK_BIN/gstack-config get` calls

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -69,11 +69,12 @@ for _PF in $(find ~/.gstack/analytics -maxdepth 1 -name '.pending-*' 2>/dev/null
 done
 eval "$(~/.claude/skills/gstack/bin/gstack-slug 2>/dev/null)" 2>/dev/null || true
 _LEARN_FILE="${GSTACK_HOME:-$HOME/.gstack}/projects/${SLUG:-unknown}/learnings.jsonl"
+_LEARN_PREAMBLE_LIMIT=$(~/.claude/skills/gstack/bin/gstack-config get learnings_preamble_limit 2>/dev/null || echo "3")
 if [ -f "$_LEARN_FILE" ]; then
   _LEARN_COUNT=$(wc -l < "$_LEARN_FILE" 2>/dev/null | tr -d ' ')
   echo "LEARNINGS: $_LEARN_COUNT entries loaded"
   if [ "$_LEARN_COUNT" -gt 5 ] 2>/dev/null; then
-    ~/.claude/skills/gstack/bin/gstack-learnings-search --limit 3 2>/dev/null || true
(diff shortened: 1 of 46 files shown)


time-attack:
Thanks for this. The new limits accept malformed, negative, empty, or excessively large values, and the generated outputs no longer match the current config and resolver sources. Rebase the source change, enforce bounded positive integers, add loaded-count tests, and regenerate all current host outputs.

PR #1521, opened 2026-05-15
fix(gbrain-sync): add .gbrain-source to consumer repo .gitignore

### LINKED ISSUE #1384: /sync-gbrain doesn't add `.gbrain-source` to consumer repo's `.gitignore` (v1.29.0.0 changelog promised it)

genisis0x:
## Summary

Closes #1384.

The v1.29.0.0 changelog claimed `.gbrain-source` was added to `.gitignore` so per-worktree pin doesn't leak across branches. The change actually shipped only added `.gbrain-source` to gstack's *own* `.gitignore`, not the consuming repo's. Conductor sibling worktrees of the same repo + branch end up clobbering each other's pin every time anyone commits the file.

Add the missing append in `bin/gstack-gbrain-sync.ts::runCodeImport`, right after `gbrain sources attach <id>` succeeds. The new `ensureGbrainSourceGitignored` helper is:

- **Idempotent** — line-trim match so leading/trailing whitespace, duplicate runs, or pre-existing entries don't add a second copy.
- **Tolerant** — missing `.gitignore` is created; unwritable file logs a warning and lets the rest of the sync continue. The pin failure mode is "noisy commits", not "broken sync", so we don't want to fail the whole stage.
- **Scoped** — only fires after a successful attach. Failed attach already short-circuits.

Single-worktree users mostly don't notice — the pin tracks the only checkout, so even if committed it points at the right source. Conductor users running `/sync-gbrain` in two sibling worktrees of the same repo + branch get *different* path-hash source IDs (correct, per v1.29.0.0). If either pin gets committed, `git pull` in the sibling overwrites the local pin and routes queries to the wrong worktree's source. The v1.29.0.0 changelog claimed to prevent exactly this.

## Verification

```
$ bun test test/gbrain-source-gitignore.test.ts
6 pass
0 fail
8 expect() calls
```

Six covered scenarios:

- creates `.gitignore` with the pin entry when none exists
- appends to existing `.gitignore` without trailing newline (preserves `\n` discipline)
- appends to existing `.gitignore` with trailing newline (no double newline)
- idempotent on repeat call
- recognizes the entry even when surrounded by whitespace (no duplicate add)
- does not throw when the `.gitignore` is read-only

Existing `test/gstack-gbrain-sync.test.ts` was already failing in 10/20 cases on machines without a local `gbrain` CLI on PATH (unrelated to this PR). Pre-change baseline = post-change baseline for those.

## Non-changes

- The `import.meta.main` gate on the top-level `main().catch(...)` call is a small piggyback so the test file can `import { ensureGbrainSourceGitignored }` without triggering a full sync run on module load. Behaviorally a no-op for the CLI path — `bun bin/gstack-gbrain-sync.ts` still runs `main()` because the module IS the entry there.
- `runMemoryIngest`, `runBrainSync`, and the dry-run path are untouched.
- No change to the `detach` path — issue notes the gitignore entry is harmless to leave behind.

No newsfragment added per `CLAUDE.md` ("Contributor PR authors should not edit CHANGELOG.md; maintainer/AI adds entries during landing/merge").

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -459,6 +459,13 @@ async function runCodeImport(args: CliArgs): Promise<StageResult> {
     };
   }
 
+  // v1.29.0.0 changelog promised the per-worktree pin would be ignored in the
+  // consuming repo, but the change actually only added .gbrain-source to
+  // gstack's own .gitignore. Without the consumer-side entry, the pin gets
+  // committed and breaks the per-worktree promise: Conductor sibling worktrees
+  // step on each other's pin every time anyone commits (#1384).
+  ensureGbrainSourceGitignored(root);
+
   return {
     name: "code",
     ran: true,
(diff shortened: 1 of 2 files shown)


garrytan:
Thank you @genisis0x. Cherry-picked into v1.40.0.0 with attribution preserved (Author: genisis0x). See commit 2b5dd909. Consolidated v1.40.0.0 fix wave lands as PR #1547. Closing this one as superseded — your contribution is recognized in CHANGELOG.md and the commit metadata.

Issue #1428, opened 2026-05-11
/codex review broken on Codex CLI ≥ 0.130.0 — passes [PROMPT] + --base together, which the CLI now rejects


Stashub:
### What

The `/codex` skill's **Review mode** (Step 2A in `skills/codex/SKILL.md`) always invokes:

```
codex review "<filesystem-boundary instruction>[\n\n<user custom instructions>]" --base <base> -c 'model_reasoning_effort="high"' --enable web_search_cached
```

i.e. it passes a positional `[PROMPT]` **and** `--base <branch>`. Codex CLI **0.130.0** rejects that combination:

```
error: the argument '[PROMPT]' cannot be used with '--base <BRANCH>'
Usage: codex review --config <key=value> --enable <FEATURE> <PROMPT>
```

`codex review --help` on 0.130.0 confirms it — `[PROMPT]` and `--base` / `--uncommitted` are mutually exclusive (review *either* from a custom prompt *or* against a base branch / uncommitted changes, not both):

```
Usage: codex review [OPTIONS] [PROMPT]
Arguments:
  [PROMPT]          Custom review instructions. If `-` is used, read from stdin
Options:
      --uncommitted    Review staged, unstaged, and untracked changes
      --base <BRANCH>  Review changes against the given base branch
```

Step 2A's "Always pass the filesystem boundary instruction as the prompt argument, even without custom instructions" guarantees the conflict on **every** `/codex review` call.

### Impact

`/codex review` and `/codex review <custom instructions>` fail immediately on Codex CLI ≥ 0.130.0. `/codex challenge` and `/codex` (consult) are unaffected — they use `codex exec`, not `codex review`.

### Repro

```
$ codex review "anything" --base main
error: the argument '[PROMPT]' cannot be used with '--base <BRANCH>'
```

- Codex CLI: `codex-cli 0.130.0`
- gstack: `v1.31.1.0`
- `bin/gstack-codex-probe` known-bad list is `0.120.0` / `0.120.1` / `0.120.2` (stdin deadlock, #972) — the 0.130.0 `review` arg change isn't tracked there.

### Suggested fixes (pick whichever fits)

[...]

PR #1522, opened 2026-05-15
fix(codex): restore filesystem boundary on /codex review bare path

### LINKED ISSUE #1503: /codex review: bare-path drops filesystem boundary on diffs that touch .claude/skills/

genisis0x:
## Summary

Closes #1503.

v1.34.2.0 (#1478, fixing #1428) shipped a dual-path approach for `/codex review` on Codex CLI ≥0.130.0 because `--base <branch>` and a custom prompt are now mutually exclusive at argv level. The default (bare) path calls `codex review --base` directly with no custom prompt, which means the previously-prefixed filesystem-boundary instruction ("do not read files under `~/.claude/`, `.claude/skills/`, `agents/`, `~/.agents/`") is no longer in front of the model when the diff touches those paths. On large diffs that include skill templates (which gstack itself does regularly) codex can spend a meaningful fraction of its token budget reading skill definitions meant for a different AI system.

This PR applies the issue's option C: detect skill-file paths in the diff and bail out of the bare path into the existing exec path, which already carries the boundary.

## Implementation

`codex/SKILL.md.tmpl` Step 2A picks up a single new precondition before path selection:

```bash
_DIFF_TOUCHES_SKILLS=0
if git diff --name-only "<base>...HEAD" 2>/dev/null \
  | grep -Eq '^(\.claude/|\.claude/skills/|agents/|\.agents/)'; then
  _DIFF_TOUCHES_SKILLS=1
fi
```

Path selection becomes:

| User instructions | `_DIFF_TOUCHES_SKILLS` | Route | Boundary carried |
|-------------------|------------------------|-------|------------------|
| empty             | 0                      | `codex review --base` bare | n/a (no skill paths touched, nothing to wander into) |
| empty             | 1                      | `codex exec` with boundary | yes |
| present           | 0 or 1                 | `codex exec` with boundary + custom focus | yes |

The exec branch's prompt build now makes the `Custom focus:` line conditional on `$_USER_INSTRUCTIONS` so the synthetic boundary-only fallthrough doesn't render an empty focus line.

## Why not the other two options from the issue

- **Option A: push the boundary into Codex CLI as a system-prompt arg if/when 0.130.0+ exposes one.** Not exposed yet; this PR works against the CLI shape that's actually shipped.
- **Option B: `~/.codex/AGENTS.md` opt-in.** Cross-machine config that lives outside the gstack tree, so users on fresh installs would still see the leak. The in-skill detection covers everyone running gstack regardless of their codex config.

Option C reuses every boundary-related construct already in the exec route (DIFF_START/DIFF_END delimiters, exec read-only sandbox, `[P1]`/`[P2]` gate marker contract), so the gate verdict in step 4 still works without any change.

## Verification

```
$ bun run gen:skill-docs
[gen-llms-txt] gstack/llms.txt: 48 skills, 75 browse commands
(no warnings on codex skill)

$ bun test test/gen-skill-docs.test.ts test/skill-validation.test.ts
712 pass, 0 fail
```

Behavior on diffs that don't touch `.claude/`, `.claude/skills/`, `agents/`, or `.agents/` is unchanged — the bare review path keeps Codex's internal review-prompt tuning, which is exactly what v1.34.2.0 went after.

Both `codex/SKILL.md.tmpl` (source of truth) and the regenerated `codex/SKILL.md` are in the diff per the SKILL.md workflow rule in CLAUDE.md.

## No newsfragment

Per CLAUDE.md ("Contributor PR authors should not edit CHANGELOG.md; maintainer/AI adds entries during landing/merge").

### DIFF
--- a/codex/SKILL.md
+++ b/codex/SKILL.md
@@ -938,15 +938,31 @@ TMPERR=$(mktemp "$TMP_ROOT/codex-err-XXXXXX.txt")
 2. Run the review (5-minute timeout). **Codex CLI ≥ 0.130.0 rejects passing a
 custom prompt and `--base <branch>` together** (the two arguments are mutually
 exclusive at argv level), so the previously-prefixed filesystem boundary cannot
-be carried in review mode. Two paths:
+be carried in review mode. Two paths.
 
-**Default path (no custom user instructions):** call `codex review --base` bare.
-Codex's review prompt template is internally diff-scoped, so the model focuses on
-the changes against the base branch. The filesystem boundary that previously
-prefixed every review call is no longer carried in bare review mode; the skill
-files under `.claude/` and `agents/` are public, so this is a token-efficiency
-concern, not a safety concern. If a future diff happens to include skill files,
-Codex may spend a few extra tokens reading them. Acceptable trade-off:
+**Skill-path detection (precondition for the default path):** check once
+whether the diff between the base branch and `HEAD` touches any of the skill
+template paths. When it does, the bare `codex review --base` route would let
+Codex spend a meaningful fraction of its token budget reading skill files for
+a different AI system. Falling through to the custom-instructions route on
+that signal keeps the filesystem boundary in front of the model:
+
+```bash
+_DIFF_TOUCHES_SKILLS=0
+if git diff --name-only "<base>...HEAD" 2>/dev/null \
+  | grep -Eq '^(\.claude/|\.claude/skills/|agents/|\.agents/)'; then
+  _DIFF_TOUCHES_SKILLS=1
+fi
+```
+
+**Default path (no custom user instructions AND `_DIFF_TOUCHES_SKILLS=0`):**
+call `codex review --base` bare. Codex's review prompt template is internally
+diff-scoped, so the model focuses on the changes against the base branch. The
+filesystem boundary that previously prefixed every review call is no longer
+carried in bare review mode; the skill files under `.claude/` and `agents/`
+are public, so this is a token-efficiency concern, not a safety concern. When
+the diff doesn't touch those paths there's nothing to lose by going bare.
+Acceptable trade-off:
 
 ```bash
 _REPO_ROOT=$(git rev-parse --show-toplevel) || { echo "ERROR: not in a git repo" >&2; exit 1; }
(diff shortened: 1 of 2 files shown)


genisis0x:
Closing — the fix from this PR (skill-path detection via `_DIFF_TOUCHES_SKILLS` + custom-instructions fallback that preserves the filesystem boundary) is already on `main` as part of the v1.42.0.0 Daegu wave (#1594). `codex/SKILL.md.tmpl` on current main contains the exact `_DIFF_TOUCHES_SKILLS` detection block and the bare-vs-custom-instructions branching from this branch, so this PR is now a no-op. Issue #1503 is effectively resolved.

PR #1525, opened 2026-05-15
fix(setup): surface codesign failures, verify signatures, exit non-zero on Apple Silicon (closes #1254)

### LINKED ISSUE #997: [macOS Apple Silicon] Compiled browse binary gets SIGKILL due to corrupt/linker-only code signature — setup should add ad-hoc codesign step
## Summary

On some Apple Silicon machines, the `browse` binary compiled by `bun run build` ships with a **corrupt or linker-only code signature**. macOS then kills every invocation with SIGKILL (`zsh: killed`, exit 137, 0 bytes output). The official `./setup` never recovers — every re-run produces the same unrunnable binary.

The fix is two lines of `codesign`. I think it belongs in `setup` so Apple Silicon users don't have to diagnose this.

## Environment

- macOS Darwin 25.4.0 (arm64)
- Apple Silicon
- Bun 1.3.12 (freshly installed per README requirement)
- gstack v0.16.4.0
- SIP enabled, Xcode Command Line Tools installed
- Fresh `git clone` + `./setup` install (no vendoring)

## Reproduction

```bash
# Official install path
git clone --depth 1 https://github.com/garrytan/gstack.git ~/.claude/skills/gstack
cd ~/.claude/skills/gstack && ./setup

# Try to use browse
~/.claude/skills/gstack/browse/dist/browse status
# => zsh: killed
```

In Claude Code, every `$B <cmd>` call returns exit 137 with 0 bytes stdout/stderr. From a login shell terminal, `zsh: killed`.

## Root cause

`otool -l` on the compiled binary shows a valid `LC_CODE_SIGNATURE` load command, but the signature block itself is corrupt/empty:

```
cmd LC_CODE_SIGNATURE
cmdsize 16
dataoff 61283152
```

`codesign -dv` reports `code object is not signed at all`. `spctl --assess --verbose <binary>` returns `invalid or unsupported format for signature`.

A naive `codesign -s - -f <binary>` also fails with \"invalid or unsupported format for signature\" because the existing (broken) signature block prevents re-signing. The only path that works is **remove → re-sign**:

```bash
codesign --remove-signature <binary>
codesign -s - -f <binary>
```

After that, `codesign -dv` shows `Signature=adhoc` and the binary runs cleanly.

This matches the well-documented Apple Silicon policy: all executables must be properly signed (even ad-hoc), and linker-signed-only binaries get SIGKILL ([Bun #7208](https://github.com/oven-sh/bun/issues/7208), [Go #42684](https://github.com/golang/go/issues/42684), [Bun docs on codesigning](https://bun.com/docs/guides/runtime/codesign-macos-executable)).

Why it hits some users and not others seems to depend on Bun version (1.3.x appears worse than 1.2.x at self-signing) and whether a prior corrupt signature exists from an earlier install.

## Suggested fix

Add a macOS-only codesign step to `setup` after `bun run build` succeeds. Something like:

```bash
if [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "arm64" ]; then
  for bin in browse/dist/browse browse/dist/find-browse design/dist/design bin/gstack-global-discover; do
    [ -x \"$bin\" ] || continue
    codesign --remove-signature \"$bin\" 2>/dev/null || true
    codesign -s - -f \"$bin\" 2>/dev/null || log \"warning: codesign failed for $bin\"
  done
fi
```

This is idempotent (safe to re-run), costs <1s, and is what Bun's own docs recommend for distributed standalone executables. It would save Apple Silicon users from a dead-end debug session.

## Workaround (for anyone hitting this now)

```bash
cd ~/.claude/skills/gstack
codesign --remove-signature browse/dist/browse
codesign -s - -f browse/dist/browse
# repeat for browse/dist/find-browse, design/dist/design, bin/gstack-global-discover
```

Happy to open a PR if helpful.

## Related

- #943 (watchdog kills server) — separate issue, patched independently
- [Bun Issue #7208](https://github.com/oven-sh/bun/issues/7208) (upstream)

### LINKED ISSUE #1254: setup: silent codesign failure leaves Apple Silicon binaries unsigned, SIGKILL on first run
## Symptom

After `gstack setup` (fresh install or upgrade), Bun-compiled binaries in `~/.claude/skills/gstack/{browse,design,make-pdf,bin}/` are killed by macOS Gatekeeper with SIGKILL (exit status 137) on first invocation. The setup itself reports success, so the failure surfaces later when a skill tries to call `$B`, `$D`, etc.

## Repro on my machine

macOS 14.x, Apple Silicon (arm64). gstack `1.15.0.0` (HEAD `6209163` in `~/.claude/skills/gstack/browse/dist/.version`).

```
$ for b in browse/dist/browse browse/dist/find-browse design/dist/design make-pdf/dist/pdf bin/gstack-global-discover; do
    p="$HOME/.claude/skills/gstack/$b"
    codesign -dv "$p" 2>&1 | head -1
  done
/Users/.../browse/dist/browse: code object is not signed at all
/Users/.../browse/dist/find-browse: code object is not signed at all
/Users/.../design/dist/design: code object is not signed at all
/Users/.../make-pdf/dist/pdf: code object is not signed at all
/Users/.../bin/gstack-global-discover: code object is not signed at all
```

5/5 binaries unsigned despite the codesign loop in `setup` (added in #1056 / v0.18.4.0).

Manually running `codesign --remove-signature <bin> && codesign -s - -f <bin>` produces a working ad-hoc signature and the binary runs fine, so the codesign machinery itself works on this machine — it's the setup-time invocation that didn't take.

## Root cause (3 compounding issues in `setup` lines 247-264)

```bash
codesign --remove-signature "$_bin_path" 2>/dev/null || true
if ! codesign -s - -f "$_bin_path" 2>/dev/null; then
  log "warning: codesign failed for $_bin (binary may not run on Apple Silicon)"
fi
```

1. **`2>/dev/null` discards the actual codesign error.** Whatever caused the failure on a given machine (transient lock, fs ACL, Xcode CLT state, etc.) is invisible — both to the user and to anyone trying to debug from a bug report.
2. **`log "warning: ..."` routes to the log channel, not the user's terminal.** The setup completes with no visible indication that a critical step failed.
3. **The block lives inside the `NEEDS_BUILD=1` branch.** If a subsequent `gstack setup` invocation skips the build (binary exists from a previous half-broken install), the codesign block is also skipped — re-running setup doesn't self-heal a broken install.

The combined effect is that any transient codesign failure during the one setup run that built the binary leaves the install permanently broken from the user's perspective, with no user-visible signal until a skill actually calls into the binary and gets exit 137.

## Suggested fix

Three small, defensive changes in `setup`:

1. Move the codesign loop out of the `NEEDS_BUILD=1` branch so it runs on every `gstack setup`. Idempotent and cheap (<1s for all 5 binaries).
2. Capture codesign's stderr instead of discarding it; surface it on failure.
3. After signing, verify with `codesign -dv` and fail the setup loudly if any binary remains unsigned. Don't claim "ready" for a half-broken install.

Sketch:

```bash
if [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "arm64" ]; then
  _sign_failed=()
  for _bin in browse/dist/browse browse/dist/find-browse design/dist/design make-pdf/dist/pdf bin/gstack-global-discover; do
    _bin_path="$SOURCE_GSTACK_DIR/$_bin"
    [ -f "$_bin_path" ] && [ -x "$_bin_path" ] || continue
    codesign --remove-signature "$_bin_path" >/dev/null 2>&1 || true
    if ! _err=$(codesign -s - -f "$_bin_path" 2>&1); then
      echo "codesign failed for $_bin: $_err" >&2
      _sign_failed+=("$_bin")
      continue
    fi
    if ! codesign -dv "$_bin_path" >/dev/null 2>&1; then
      echo "codesign reported success but binary remains unsigned: $_bin" >&2
      _sign_failed+=("$_bin")
    fi
  done
  if [ ${#_sign_failed[@]} -gt 0 ]; then
    echo "ERROR: gstack binaries failed to codesign — these will SIGKILL (exit 137) on first run:" >&2
    printf '  - %s\n' "${_sign_failed[@]}" >&2
    exit 1
  fi
fi
```

A test in `test/setup-codesign.test.ts` shape — assert that after `gstack setup`, every binary in the loop returns 0 from `codesign -dv`.

```bash
cd ~/.claude/skills/gstack
for b in browse/dist/browse browse/dist/find-browse design/dist/design make-pdf/dist/pdf bin/gstack-global-discover; do
  [ -f "$b" ] && [ -x "$b" ] || continue
  codesign --remove-signature "$b" 2>/dev/null || true
  codesign -s - -f "$b"
  codesign -dv "$b" 2>&1 | head -1
done
```

mvanhorn:
## Summary

Setup's Apple Silicon codesign loop now surfaces codesign's stderr, verifies each signature actually applied, and exits non-zero when any binary fails to sign. The previous wrapper swallowed every error and exited 0 even when 5 of 5 binaries were left unsigned, so users hit SIGKILL minutes later with no breadcrumb.

## Why this matters

> After `gstack setup` (fresh install or upgrade), Bun-compiled binaries in `~/.claude/skills/gstack/{browse,design,make-pdf,bin}/` are killed by macOS Gatekeeper with SIGKILL (exit status 137) on first invocation. The setup itself reports success, so the failure surfaces later when a skill tries to call `$B`, `$D`, etc.

That's [@lucascaro on #1254](https://github.com/garrytan/gstack/issues/1254). The reporter pinpointed the codesign loop in `setup` lines 247-264 (now 288-303 after subsequent edits), reproduced 5/5 binaries unsigned despite a clean setup exit, and confirmed the manual recipe (`codesign --remove-signature && codesign -s - -f`) works on the same machine. The codesign machinery itself is fine; only the wrapper was lying.

The original codesign loop landed in #1056 / v0.18.4.0 to fix #997 (the SIGKILL class). It was correct on the happy path but its three failure modes were all silent: `2>/dev/null` swallowed diagnostics, `log "warning: ..."` returned 0, and there was no verification step to catch a signature that didn't take.

## Changes

- `setup` lines 288-303 (the Darwin+arm64 codesign block) now:
  - Captures both `codesign --remove-signature` and `codesign -s - -f` stderr to a tempfile and prints it via the existing `log` helper if non-empty. Users see exactly what codesign said.
  - Adds `codesign --verify --strict "$_bin_path"` after signing. Catches the silent-corruption case the reporter hit (signature applied but doesn't actually verify).
  - Maintains a `_codesign_failures` counter. Both signing failures and verification failures bump it.
  - Exits 1 with a clear message when the counter is non-zero. No more "setup succeeded, SIGKILL surfaces later" footgun.
- The Darwin+arm64 guard is unchanged. Other platforms see no behavior change.

## Testing

`bun test test/setup-codesign.test.ts`: 6 pass / 0 fail. Old assertions about "warning, not fatal" replaced with assertions for the new fail-loud structure. The `set -e` smoke test still passes (the codesign block is syntactically valid bash). Existing for-loop coverage expanded to assert all 5 binaries the setup loop iterates (the test was only checking 4; setup itself already had 5).

Fixes #1254.

AI was used for assistance.

### DIFF
--- a/setup
+++ b/setup
@@ -292,14 +292,49 @@ if [ "$NEEDS_BUILD" -eq 1 ]; then
   # signature block is corrupt. This is idempotent and costs <1s.
   # See: https://github.com/garrytan/gstack/issues/997
   if [ "$(uname -s)" = "Darwin" ] && [ "$(uname -m)" = "arm64" ]; then
+    _codesign_failures=0
+    _codesign_err="$(mktemp)"
+    _print_codesign_err() {
+      if [ -s "$_codesign_err" ]; then
+        while IFS= read -r _codesign_line; do
+          log "  $_codesign_line"
+        done < "$_codesign_err"
+      else
+        log "  (no stderr output)"
+      fi
+    }
+
     for _bin in browse/dist/browse browse/dist/find-browse design/dist/design make-pdf/dist/pdf bin/gstack-global-discover; do
       _bin_path="$SOURCE_GSTACK_DIR/$_bin"
       [ -f "$_bin_path" ] && [ -x "$_bin_path" ] || continue
-      codesign --remove-signature "$_bin_path" 2>/dev/null || true
-      if ! codesign -s - -f "$_bin_path" 2>/dev/null; then
-        log "warning: codesign failed for $_bin (binary may not run on Apple Silicon)"
+
+      : > "$_codesign_err"
+      if ! codesign --remove-signature "$_bin_path" 2>"$_codesign_err"; then
+        log "warning: codesign --remove-signature failed for $_bin:"
+        _print_codesign_err
+      fi
+
+      : > "$_codesign_err"
+      if ! codesign -s - -f "$_bin_path" 2>"$_codesign_err"; then
+        log "error: codesign failed for $_bin:"
+        _print_codesign_err
+        _codesign_failures=$((_codesign_failures + 1))
+        continue
+      fi
+
+      : > "$_codesign_err"
+      if ! codesign --verify --strict "$_bin_path" 2>"$_codesign_err"; then
+        log "error: codesign verification failed for $_bin:"
+        _print_codesign_err
+        _codesign_failures=$((_codesign_failures + 1))
       fi
     done
+
+    rm -f "$_codesign_err"
+    if [ "$_codesign_failures" -gt 0 ]; then
+      echo "gstack setup failed: $_codesign_failures binaries did not codesign. Apple Silicon will SIGKILL them on first run. See output above for codesign errors." >&2
+      exit 1
+    fi
   fi
 
   # macOS: install coreutils for `gtimeout` (Codex hang protection in /codex + /autoplan).
--- a/test/setup-codesign.test.ts
+++ b/test/setup-codesign.test.ts
@@ -24,10 +24,11 @@ describe('setup: Apple Silicon codesign', () => {
     const forMatch = content.match(/for _bin in ([^;]+);/);
     expect(forMatch).toBeTruthy();
     const binaries = forMatch![1].trim().split(/\s+/);
-    // All four compiled binaries from `bun run build` must be covered
+    // All compiled binaries from `bun run build` must be covered
     expect(binaries).toContain('browse/dist/browse');
     expect(binaries).toContain('browse/dist/find-browse');
     expect(binaries).toContain('design/dist/design');
+    expect(binaries).toContain('make-pdf/dist/pdf');
     expect(binaries).toContain('bin/gstack-global-discover');
   });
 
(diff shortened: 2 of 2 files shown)


16francej:

Issue #1538, opened 2026-05-16
setup fails on Windows: Bun shell doesn't support subshells with redirection in build script


andrewitte:
## Environment

- OS: Windows 11 Enterprise (10.0.26100)
- Shell: Git Bash (via PowerShell wrapper)
- Bun: 1.3.13
- gstack version: 1.39.1.0 (upgrading from 1.17.0.0)

## What happened

`./setup` fails during `bun run build` with:

error: Failed to run script build due to error Subshells with redirections are currently not supported. Please open a GitHub issue.

The offending pattern in the `build` script (package.json) is:

```bash
( git rev-parse HEAD 2>/dev/null || true ) > browse/dist/.version

Bun's built-in shell doesn't support ( ... ) > file redirection from a subshell. This blocks the upgrade from completing — binaries don't rebuild, and downstream chmod / cleanup steps never run.


## Reproduction
On Windows with Bun 1.3.13, install gstack 1.17.0.0
Run /gstack-upgrade (or ./setup directly after git pull)
Build fails at the .version write step

**Comment by ivanaitzliddat** [comment]
Hit this on Windows 10 today (fresh install, gstack v1.40.0.0, bun 1.3.14, Git Bash). Posting a verified workaround for anyone landing here before the build script is fixed upstream.

## Workaround

Run the build steps directly through bash so bun's script parser never sees the offending ``( ... ) > file`` redirect. Then re-run ``./setup`` — it picks up the pre-built binaries and continues to Playwright + skill registration.

```bash
cd ~/.claude/skills/gstack
bash -c '
  set -e
  bun run vendor:xterm
  bun run gen:skill-docs --host all
  bun build --compile browse/src/cli.ts --outfile browse/dist/browse
  bun build --compile browse/src/find-browse.ts --outfile browse/dist/find-browse
  bun build --compile design/src/cli.ts --outfile design/dist/design
  bun build --compile make-pdf/src/cli.ts --outfile make-pdf/dist/pdf
  bun build --compile bin/gstack-global-discover.ts --outfile bin/gstack-global-discover
  bash browse/scripts/build-node-server.sh
  ( git rev-parse HEAD 2>/dev/null || echo "" ) > browse/dist/.version
  ( git rev-parse HEAD 2>/dev/null || echo "" ) > design/dist/.version
  ( git rev-parse HEAD 2>/dev/null || echo "" ) > make-pdf/dist/.version
'
bash ./setup --no-prefix
```

[...]

Issue #1537, opened 2026-05-16
Auto-upgrade broken on Windows: Bun rejects subshell-redirect in package.json build script (v1.34.2.0+)


jonrichardwhite-hub:
## Summary

`./setup` fails on Windows when upgrading past v1.34.1.0. The `build` npm script in `package.json` uses POSIX subshell-redirect syntax (`( cmd ) > file`) which Bun on Windows refuses to execute. This blocks every Windows user from auto-upgrading via `/gstack-upgrade`.

## Repro

1. Windows 11 (Build 26200), MINGW64 / Git Bash, Bun v1.3.13
2. Global git install at `~/.claude/skills/gstack` on v1.34.1.0
3. Run `/gstack-upgrade` (or the inline upgrade flow)
4. `git fetch` + `git reset --hard origin/main` succeed — checks out v1.39.1.0
5. `./setup` runs `bun install` which triggers the `build` script

## Error

```
$ bun run vendor:xterm && bun run gen:skill-docs --host all; bun build --compile ... && bash browse/scripts/build-node-server.sh && ( git rev-parse HEAD 2>/dev/null || true ) > browse/dist/.version && ...
error: Failed to run script build due to error
Subshells with redirections are currently not supported. Please open a GitHub issue.
```

## Root cause

`package.json` `build` script (on `main` @ v1.39.1.0) contains:

```
( git rev-parse HEAD 2>/dev/null || true ) > browse/dist/.version
( git rev-parse HEAD 2>/dev/null || true ) > design/dist/.version
( git rev-parse HEAD 2>/dev/null || true ) > make-pdf/dist/.version
(rm -f .*.bun-build || true)
```

Bun's built-in shell (used to run npm scripts on Windows where there's no `/bin/sh`) does not support `( ... ) > file`. This is a tracked limitation in oven-sh/bun.

## Suggested fixes

Any of these would work without dropping POSIX-shell compatibility:

1. **Move version-stamping into a tiny `.sh` or `.ts` script** and call it: `bash scripts/stamp-versions.sh` (or a `bun run` script).
2. **Use `git rev-parse HEAD > file 2>/dev/null || true`** — no subshell needed (the `2>/dev/null` redirect goes on the command, `|| true` swallows failure separately).
3. **Run the build via `bash -c "..."`** explicitly instead of relying on Bun's shell.

Same pattern at the tail (`(rm -f .*.bun-build || true)`) should be fixed too, though it's the version-stamp lines that are hit first.

## Environment

[...]

PR #1540, opened 2026-05-16
Fix Windows build artifact finalization

McGluut:
## Summary

This PR moves the build artifact finalization out of shell subshells and into a small Bun script.

Changes:
- write the `.version` files from `scripts/finalize-build-artifacts.ts`
- mark built executables from that script when they exist
- remove root `.bun-build` temp artifacts without package-script shell syntax
- make the build stop if xterm vendoring or skill-doc generation fails
- add a focused test for the finalizer

This should address #1537 and #1538.

## Validation

- `bun install --frozen-lockfile`
- `bun run build`
- `bun test test/finalize-build-artifacts.test.ts`
- `git diff --check`


time-attack:
@16francej Please close this PR. Windows build finalization, version files, permissions, cleanup, and failure handling are all owned by the newer #1594 build pipeline. This older implementation is fully superseded.

PR #1460, opened 2026-05-13
v1.33.3.0 fix: replace bash brace groups with subshells in build script for Windows compatibility

realcarsonterry:
Fixes #1457

## Summary

On Windows with Git Bash + Bun, `bun run build` fails because Bun's Windows shell parser doesn't support bash brace groups `{ cmd; }`.

## Changes

Changed three occurrences in `package.json` line 12 from:
```bash
{ git rev-parse HEAD 2>/dev/null || true; } > path/.version
```

To POSIX-compliant subshells:
```bash
( git rev-parse HEAD 2>/dev/null || true ) > path/.version
```

## Why This Works

- Subshells `( ... )` are universally supported in POSIX bash, zsh, and Bun's Windows shell
- Functionally equivalent for one-shot stdout redirection
- No impact on Unix/Linux behavior
- Allows Windows users to successfully run `bun run build`

## Testing

Tested on Windows 11 with Git Bash + Bun 1.3.13 - build now completes successfully.

---
<a href="https://app.blacksmith.sh/garrytan/codesmith/gstack/pr/1460"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #1544, opened 2026-05-16
Fix Bun Windows build script compatibility

Charlie-El:
## Summary

Fixes Windows setup/build failures with Bun 1.3.x where package scripts containing subshell redirections fail with:

`Subshells with redirections are currently not supported`

This PR moves the long build pipeline out of `package.json` and into a Bash script, so Bun no longer has to parse complex shell syntax. It also handles a related Windows issue where `bun build --compile` can fail when `bun.exe` is installed under a non-ASCII user path, such as a Windows username containing Chinese characters.

## What changed

- Replaced the long inline `package.json` build command with `bash scripts/build.sh`.
- Added `scripts/build.sh` to run the full build pipeline under Bash instead of Bun's package-script shell parser.
- Added `scripts/write-version-files.sh` to write `.version` files without `( git ... ) > file`.
- Updated `setup` so Windows installs can temporarily copy `bun.exe` to an ASCII-only path for `bun build --compile`, then clean it up afterward.
- Updated `test/build-script-shell-compat.test.ts` to guard against reintroducing subshell redirections in the build script.

## Why

On Windows Git Bash with Bun 1.3.x, `bun run build` failed before compilation with:

```text
error: Failed to run script build due to error Subshells with redirections are currently not supported.
```

The failing pattern was:

```sh
```

After removing that pattern, I also reproduced a second Windows-specific issue where `bun build --compile` failed when Bun was installed under a non-ASCII user path:

```text
error: failed to copy bun executable into temporary file: ENOENT
Failed to get temp file path: FileNotFound
```

Using an ASCII-path temporary Bun copy allowed `bun build --compile` to complete.

## Relationship to related PRs

This overlaps with #1531 on the `.version` redirection failure, but takes a broader approach:

1. Moves the full build pipeline into `scripts/build.sh`, so Bun does not parse the long shell command at all.
2. Keeps `.version` writes in a helper script.
3. Adds a Windows setup workaround for non-ASCII `bun.exe` paths, which reproduced locally with a Chinese Windows username and `bun build --compile`.

It is also related to #1480, #1540, and #1460. The main additional coverage here is the non-ASCII Windows user path case plus verified end-to-end `setup --host codex` behavior.

## Test Plan

- `bash -n scripts/build.sh scripts/write-version-files.sh setup`
- `bun test test/build-script-shell-compat.test.ts`
- `bun run build`
- Manually verified `bash ./setup --host codex` completes on Windows Git Bash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "make-pdf": "./make-pdf/dist/pdf"
   },
   "scripts": {
-    "build": "bun run vendor:xterm && bun run gen:skill-docs --host all; bun build --compile browse/src/cli.ts --outfile browse/dist/browse && bun build --compile browse/src/find-browse.ts --outfile browse/dist/find-browse && bun build --compile design/src/cli.ts --outfile design/dist/design && bun build --compile make-pdf/src/cli.ts --outfile make-pdf/dist/pdf && bun build --compile bin/gstack-global-discover.ts --outfile bin/gstack-global-discover && bash browse/scripts/build-node-server.sh && ( git rev-parse HEAD 2>/dev/null || true ) > browse/dist/.version && ( git rev-parse HEAD 2>/dev/null || true ) > design/dist/.version && ( git rev-parse HEAD 2>/dev/null || true ) > make-pdf/dist/.version && chmod +x browse/dist/browse browse/dist/find-browse design/dist/design make-pdf/dist/pdf bin/gstack-global-discover && (rm -f .*.bun-build || true)",
+    "build": "bash scripts/build.sh",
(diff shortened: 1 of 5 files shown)


Charlie-El:
This overlaps with #1531 on the `.version` redirection failure, but it takes a broader approach:

1. Moves the full build pipeline into `scripts/build.sh`, so Bun does not parse the long shell command at all.
2. Keeps `.version` writes in a helper script.
3. Adds a Windows setup workaround for non-ASCII `bun.exe` paths, which reproduced locally with a Chinese Windows username and `bun build --compile`.

Manual verification on Windows Git Bash:

- `bun run build`
- `bash ./setup --host codex`

Both completed successfully.

garrytan:
Thanks @Charlie-El — your Windows build fix landed in v1.42.0.0 Daegu wave (#1594) as the base for the supersedes chain. Closing the open PR.

PR #1546, opened 2026-05-16
fix(memory): probe gitleaks without shell builtin

### LINKED ISSUE #1545: Windows: --scan-secrets treats gitleaks as missing when PATH probe uses command -v
## Problem

`lib/gstack-memory-helpers.ts` checks whether `gitleaks` is available with `execSync("command -v gitleaks")` before `secretScanFile()` runs the actual scan. That shells through a POSIX builtin instead of probing the executable directly. On Windows/native Bun environments, `command -v` is not a portable executable check, so `--scan-secrets` can report `scanner: "missing"` and skip the opt-in secret scan even when `gitleaks.exe` is installed on `PATH`.

## Current behavior

- `gstack-memory-ingest --scan-secrets` calls `secretScanFile()`.
- `secretScanFile()` first calls the shell-based `command -v gitleaks` check.
- If that shell check fails, gstack warns that gitleaks is missing and returns `scanner: "missing"` without running `gitleaks detect`.

## Expected behavior

Probe `gitleaks` the same way the scanner will run it: execute the `gitleaks` binary directly via argument-array process APIs, and then run `gitleaks detect` with the explicit current environment.

## Duplicate check

I searched issues/PRs for `gitleaks`, `secretScanFile`, `scan-secrets Windows`, `gitleaks PATH`, and `command -v`. Existing open PR #1418 touches `lib/gstack-memory-helpers.ts`, but it fixes `gbrain doctor` engine detection, not the gitleaks availability probe. Existing PR #1485 removes a similar `command -v gbrain` path, but does not cover `gitleaks`.

## Candidate fix

Replace the shell-based availability check with a direct `gitleaks version` probe, pass `process.env` explicitly to the `execFileSync` calls, and add a regression test with a fake `gitleaks` binary that verifies the direct probe runs before `detect`.

jbetala7:
Fixes #1545

## Summary
- replace the shell-based `command -v gitleaks` availability check with a direct `gitleaks version` executable probe
- pass the current env explicitly to both gitleaks `execFileSync` calls so PATH overrides are honored under Bun
- add a regression with a fake gitleaks binary that proves the direct probe runs before `detect`

## Collision checks
- searched issues and PRs for `gitleaks`, `secretScanFile`, `gitleaks PATH`, `scan-secrets Windows`, and `command -v`
- open PR #1418 touches `lib/gstack-memory-helpers.ts` but targets `gbrain doctor` engine detection, not gitleaks availability
- open PR #1485 removes a similar `command -v gbrain` path, but does not cover gitleaks

## Testing
- `bun test test/gstack-memory-helpers.test.ts`
- `bun test test/gstack-memory-ingest.test.ts --test-name-pattern "scan-secrets"`
- `git diff --check`

### DIFF
--- a/lib/gstack-memory-helpers.ts
+++ b/lib/gstack-memory-helpers.ts
@@ -19,7 +19,7 @@
 
 import { existsSync, readFileSync, writeFileSync, mkdirSync, statSync, appendFileSync } from "fs";
 import { dirname, join } from "path";
-import { execSync, execFileSync } from "child_process";
+import { execFileSync } from "child_process";
 import { homedir } from "os";
 
 // ── Types ──────────────────────────────────────────────────────────────────
@@ -122,7 +122,11 @@ let _gitleaksAvailability: boolean | null = null;
 function gitleaksAvailable(): boolean {
   if (_gitleaksAvailability !== null) return _gitleaksAvailability;
   try {
-    execSync("command -v gitleaks", { stdio: "ignore" });
+    execFileSync("gitleaks", ["version"], {
+      env: process.env,
+      stdio: "ignore",
+      timeout: 2_000,
+    });
     _gitleaksAvailability = true;
   } catch {
     _gitleaksAvailability = false;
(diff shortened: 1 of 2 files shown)


garrytan:
Thanks @jbetala7 — gitleaks shell-builtin fix shipped in v1.43.2.0 post-Daegu wave (#1642).

PR #1549, opened 2026-05-16
feat(ios): iOS skill suite — GStack × GBrain hackathon 1st place

time-attack:
## What this is

The iOS half of gstack. Five skills that give Claude the same grip on a SwiftUI
app that `/browse` gives it on a webpage.

**Won 1st place at the GStack × GBrain hackathon on May 16, 2026.**

## The five skills

| Skill | What it does |
|---|---|
| `/ios-qa` | Reads Swift source → generates an embedded HTTP debug bridge into the app → drives the running app on a **real device (USB)** or in the **iOS Simulator** → finds bugs via a vision-driven agent loop. |
| `/ios-fix` | Takes a bug from `/ios-qa`, reads the source to understand *why* it broke, writes the minimal fix, rebuilds, redeploys, then reproduces the original repro steps to prove the bug is gone. |
| `/ios-sync` | Regenerates the debug bridge after you add new buttons, screens, or ViewModel functions. Fresh action keys for every `@Observable` `func` in your app. |
| `/ios-clean` | Strips every trace of the bridge — `StateServer`, `DebugOverlay`, accessors, and all `#if DEBUG` wiring in `ContentView`/`App` files. Ship-ready in one command. |
| `/ios-design-review` | Visual audit. Walks every screen, screenshots each, evaluates against Apple HIG (layout, typography, spacing, contrast, accessibility). |

## Why this is hard

A web browser is a gift to LLM agents. The DOM is text. `querySelector` is a
function call. Chrome DevTools is an HTTP server. An iPhone is the opposite —
the screen is pixels, widgets are SwiftUI views compiled into a sandboxed
binary, the OS doesn't expose the view hierarchy, and every existing tool
(XCUITest, IDB, Appium) assumes a human is the one writing the test.

This PR solves it by **injecting the API into the app itself.** A `StateServer`
running on `:9999` inside the debug build exposes:

```
GET  /screenshot   → base64 PNG
GET  /elements     → accessibility tree (labels, frames, traits)
POST /tap          → inject tap at {x,y}
POST /type         → keyboard input
POST /swipe        → gesture
GET  /state/{key}  → read @Observable property
POST /state/{key}  → invoke ViewModel function (see below)
GET  /health       → liveness
```

Everything is `#if DEBUG`. `/ios-clean` removes the last trace for release.

## The technical unlock: action keys

SwiftUI toolbar buttons (`.toolbar { Button("Save") {...} }`) **cannot be
tapped programmatically**. `accessibilityActivate()` returns `true` and does
nothing. `sendActions(.touchUpInside)` does nothing. The private `_targets`
API crashes the app. There is no working tap. This is a SwiftUI
architectural limitation, not a fixable bug in the tap handler.

The fix: **don't tap the button — call the ViewModel function directly.**

For every `func` on every `@Observable` ViewModel, `/ios-qa` generates a
debug-only HTTP endpoint that invokes it natively:

```swift
// For: func toggleUnit()
server.registerWrite("ProfileVM_toggleUnit") { _ in
    instance.toggleUnit()
    return true
}

// For: func addItem(_ item: Item)  — JSON injection
server.registerWrite("CartVM_addItem") { value in
    let item = try! JSONDecoder().decode(Item.self, from: value.data(using: .utf8)!)
    instance.addItem(item)
    return true
}
```

One build, zero rebuilds during QA.

## Two modes

| Aspect | Device Mode | Simulator Mode |
|---|---|---|
| Transport | `curl -6 http://[IPv6]:9999/*` via Apple's CoreDevice USB tunnel | `sim_tap` (custom 4ms Swift binary via macOS Accessibility API) + IDB hybrid |
| Tap latency | HTTP round-trip | 4ms (sim_tap) / 150ms (IDB fallback) |
| Codegen needed | Yes — full `DebugBridgeGenerated/` | No — IDB reads the tree externally |

Auto-detects which mode to use. No flag required.

## Files

```
ios-qa/
  SKILL.md.tmpl
  README.md
  templates/
    StateServer.swift.template
    DebugOverlay.swift.template
    DebugBridgeManager.swift.template
    StateAccessor.swift.template
    Package.swift.template
    DebugBridgeJSON.swift.template

ios-fix/
  SKILL.md.tmpl
  README.md

ios-sync/
  SKILL.md.tmpl

ios-clean/
  SKILL.md.tmpl

ios-design-review/
  SKILL.md.tmpl
```

### DIFF
--- a/ios-clean/SKILL.md.tmpl
+++ b/ios-clean/SKILL.md.tmpl
@@ -1,0 +1,115 @@
+---
+name: ios-clean
+version: 1.0.0
+description: |
+  Remove all DebugBridge files and references from an iOS app.
+  Cleans up everything injected by /ios-qa: StateServer, DebugOverlay,
+  accessors, manager, and all #if DEBUG wiring in ContentView/App files.
+allowed-tools:
+  - Bash
+  - Read
+  - Edit
+  - Glob
+  - Grep
+triggers:
+  - ios clean
+  - ios-clean
+  - remove debug bridge
+  - clean debug bridge
+  - uninstall ios qa
+---
+
+{{PREAMBLE}}
+
+# iOS Clean — Remove DebugBridge
+
+Remove all DebugBridge/iStack files and references injected by `/ios-qa`.
+This skill is non-destructive to app code — it only removes debug instrumentation.
+
+## Step 1: Find the app source directory
+
+```bash
+find . -maxdepth 3 -name "*.xcodeproj" -not -path "*/DebugBridge/*"
+```
+
+Identify the app source directory (the folder next to the .xcodeproj).
+
+## Step 2: Delete the DebugBridgeGenerated directory
+
+```bash
+rm -rf <AppSource>/DebugBridgeGenerated/
+```
+
+This removes:
+- `StateServer.swift`
+- `DebugOverlay.swift`
+- `DebugBridgeManager.swift`
+- All `*Accessor.swift` files
+- `DebugBridgeJSON.swift` (if present)
+
+## Step 3: Delete the DebugBridge package (if exists)
+
+```bash
+rm -rf DebugBridge/
+```
+
+## Step 4: Remove references from app code
+
+Search for and remove all DebugBridge wiring in the app's source files:
+
+```bash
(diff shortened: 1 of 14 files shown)


garrytan:
Merged to main, thanks for the amazing work @time-attack

Issue #276, opened 2026-03-21
Windows: browse.exe fails to maintain Chromium session — page reverts to about:blank


gwangyong-kim:
## Summary

On Windows 11, `browse.exe` (bun-compiled binary) successfully navigates to URLs (`goto` returns 200), but the page immediately reverts to `about:blank`. The Chromium process crashes silently due to a sandbox access error, and subsequent commands see an empty page.

## Environment

- **OS:** Windows 11 Home 10.0.26200
- **Node.js:** v22.14.0
- **Bun:** 1.3.11
- **Playwright:** 1.58.2 (chromium-1208)
- **gstack:** latest (installed via `npx gstack`)

## Root Cause

The existing Windows fallback (`server-node.mjs` via Node.js) works correctly **when started manually**:

```bash
cd ~/.claude/skills/gstack/browse/dist
node server-node.mjs &   # server starts, Chromium launches fine
$B goto http://localhost:3000/login  # ✅ URL persists
$B url  # ✅ http://localhost:3000/login
```

But when `browse.exe` spawns the server via `Bun.spawn(['node', NODE_SERVER_SCRIPT])`, Chromium crashes immediately after navigation with:

```
[browse] FATAL: Chromium process crashed or was killed. Server exiting.
```

The underlying Chromium error is:
```
Sandbox cannot access executable ...chrome.exe. Check filesystem permissions are valid.
```

## The Disconnect

`browse.exe`'s `Bun.spawn()` creates a subtly different process environment than a direct `node` invocation. This causes Chromium's sandbox to fail, even though:
- `--no-sandbox` is added to launch args
- File permissions are correct
- The same `server-node.mjs` works perfectly when run directly

## Workaround

Created a bash wrapper script at `browse/dist/browse` that starts the node server directly (bypassing `browse.exe`'s `Bun.spawn`), then delegates CLI commands to `browse.exe`:

```bash
#!/bin/bash
# Start server with node directly, then use browse.exe for CLI
cd "$DIST_DIR"
node server-node.mjs &
exec "$DIST_DIR/browse.exe" "$@"
```

This works but loses the single-binary deployment advantage.

## Also Required

- `npm install playwright@1.58.2 diff` in the `browse/` directory (these externalized deps aren't resolvable from `server-node.mjs` otherwise)
- `--no-sandbox` flag added to `chromium.launch()` in `browser-manager.ts`

## Suggested Fix

[...]

PR #1551, opened 2026-05-16
fix(browse): enable Chromium sandbox in headed mode

GolfFry:
The two headed launchPersistentContext() call sites in BrowserManager omitted the chromiumSandbox option, so Playwright defaulted to passing --no-sandbox even on macOS and Linux where the sandbox works correctly. Chromium then displays the yellow "unsupported command-line flag — security will suffer" info bar, and any renderer exploit (V8, Skia, libwebp, …) can escape straight to the user's account.

The headless launch() path at line 179 already passes `chromiumSandbox: process.platform !== 'win32'`. This brings the two headed sites in line — preserving the Windows exemption (where the Bun→Node spawn chain breaks the sandbox, per GitHub #276).

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -324,6 +324,11 @@ export class BrowserManager {
       viewport: null,  // Use browser's default viewport (real window size)
       userAgent: this.customUserAgent || customUA,
       ...(executablePath ? { executablePath } : {}),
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks @GolfFry — headed-mode Chromium sandbox shipped in v1.42.2.0 (#1629). Closing as superseded.

PR #1555, opened 2026-05-17
fix: harden generateSlugSetup with SLUG fallback and wrap _TEL_FILE in code block

Acharnite:
## Summary
- **generateSlugSetup** now uses `|| true` + `: "${SLUG:=unknown}"` so transient `gstack-slug` failures never leave `$SLUG` unset
- All `CHECKPOINT_DIR` paths use `${SLUG:-unknown}` instead of bare `$SLUG`
- `_TEL_FILE` write is now inside a fenced bash code block so AI agents always execute it
- Regenerated SKILL.md for all 11 hosts (74 host-config tests pass)

## Background
Session duration showed "unknown" because the `_TEL_FILE` write was raw text outside any code block. Checkpoints landed in `~/.gstack/projects/checkpoints/` instead of `~/.gstack/projects/<slug>/checkpoints/` because a transient `gstack-slug` failure left `$SLUG` empty.


Acharnite:
Wrong repo — should target fork

PR #1560, opened 2026-05-17
fix(context): probe gbrain directly

### LINKED ISSUE #1559: gstack-brain-context-load treats gbrain as missing when availability probe execs command directly
## Observed problem

`bin/gstack-brain-context-load.ts` checks whether vector/list queries can run with:

```ts
execFileSync("command", ["-v", "gbrain"], { stdio: "ignore" });
```

`command` is a shell builtin, not a standalone executable on normal POSIX installs. When no executable named `command` exists, this throws before `gbrain` is tried, so the helper treats gbrain as missing and skips all vector/list context queries.

Fresh `origin/main` at `33cb4715ef0bc9be31a29bdf1d9655482a617ee6` still has this check in `bin/gstack-brain-context-load.ts`. Running:

```bash
bun run bin/gstack-brain-context-load.ts --repo garrytan-gstack --explain --quiet
```

prints three `SKIP` lines with `gbrain CLI missing`. The existing graceful-absence test accepts either `OK` or `SKIP`, so it does not prove that a `gbrain` binary on `PATH` is actually used.

## Expected behavior

When `gbrain` is available on `PATH`, `gstack-brain-context-load` should run its vector/list queries instead of reporting `gbrain CLI missing`.

- `gstack-brain-context-load`
- `brain-context-load`
- `context load gbrain`
- `gbrain CLI missing`
- `command -v gbrain`

Related but non-duplicate results:

- #1386 / #1485 covered memory ingest and gbrain sync probe paths, not `gstack-brain-context-load`.
- #1545 / #1546 cover the `gitleaks` probe in secret scanning, not `gbrain` context loading.
- #1547 is the v1.40.0.0 gbrain hardening wave and does not touch `bin/gstack-brain-context-load.ts`.

I found no open PR or issue for this specific context-load availability path.

## Candidate fix shape

Probe the actual `gbrain` executable directly, for example with a cheap `gbrain --version` call, and add a regression test that places a fake `gbrain` on `PATH` and verifies `gstack-brain-context-load` reports `OK`/renders query output rather than `SKIP (gbrain CLI missing)`.

jbetala7:
Fixes #1559

## Summary
- replace the `gstack-brain-context-load` availability check that execed the shell builtin `command` directly
- probe the actual `gbrain` executable with `gbrain --version` before running vector/list context queries
- add a regression that places a fake `gbrain` on `PATH` and verifies context-load reports `OK` instead of `gbrain CLI missing`

## Root cause
`command -v` only works through a shell. The helper used `execFileSync("command", ["-v", "gbrain"])`, so environments without a standalone `command` binary skipped gbrain-backed context queries even when `gbrain` was available.

## Testing
- `bun test test/gstack-brain-context-load.test.ts`
- `bun test test/test-free-shards.test.ts`

### DIFF
--- a/bin/gstack-brain-context-load.ts
+++ b/bin/gstack-brain-context-load.ts
@@ -192,7 +192,10 @@ function resolveSkillFile(args: CliArgs): string | null {
 
 function gbrainAvailable(): boolean {
   try {
-    execFileSync("command", ["-v", "gbrain"], { stdio: "ignore" });
+    execFileSync("gbrain", ["--version"], {
+      stdio: "ignore",
+      timeout: MCP_TIMEOUT_MS,
+    });
     return true;
   } catch {
     return false;
--- a/test/gstack-brain-context-load.test.ts
+++ b/test/gstack-brain-context-load.test.ts
@@ -7,9 +7,9 @@
  */
 
 import { describe, it, expect } from "bun:test";
-import { mkdtempSync, writeFileSync, mkdirSync, rmSync } from "fs";
+import { chmodSync, mkdtempSync, writeFileSync, mkdirSync, rmSync } from "fs";
 import { tmpdir } from "os";
-import { join } from "path";
+import { delimiter, join } from "path";
 import { spawnSync } from "child_process";
 
 const SCRIPT = join(import.meta.dir, "..", "bin", "gstack-brain-context-load.ts");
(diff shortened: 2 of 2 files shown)


jbetala7:
Closing as superseded by the merged v1.42.0.0 Daegu wave in #1594, which landed this fix with attribution. Thanks for carrying it through.

PR #1562, opened 2026-05-17
fix(browse): default Linux to --no-sandbox to handle Ubuntu/AppArmor

techcenter68:
## Problem

On Ubuntu 23.10+ (and several other modern distros), `kernel.apparmor_restrict_unprivileged_userns=1` is set by default, which blocks unprivileged Chromium sandboxing **even for normal (non-root, non-container) user sessions**. The current sandbox detection in `browse/src/browser-manager.ts` only covers `CI / CONTAINER / isRoot`, so `browse` fails at startup for everyday Linux desktop users:

\`\`\`
[browse] Starting server...
[browse] Server failed to start:
[browse] Failed to start: launch: Target page, context or browser has been closed
Browser logs:
Chromium sandboxing failed!
================================
[FATAL:zygote_host_impl_linux.cc:128] No usable sandbox! If you are running on
Ubuntu 23.10+ or another Linux distro that has disabled unprivileged user
namespaces with AppArmor, see https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md.
\`\`\`

This breaks `/browse`, `/qa`, `/design-review`, `/canary`, and every other skill that uses the headless browser, for any user on Ubuntu 24.04 (the current LTS) and similar distros.

## Reproduction

\`\`\`bash
# On Ubuntu 24.04 (or any distro where apparmor_restrict_unprivileged_userns=1):
sysctl kernel.apparmor_restrict_unprivileged_userns
# kernel.apparmor_restrict_unprivileged_userns = 1

~/.claude/skills/gstack/browse/dist/browse goto https://example.com
# → No usable sandbox! crash
\`\`\`

## Fix

Default Linux agent browsing to `--no-sandbox` (same posture as CI, CONTAINER, root) so the headless browser starts reliably out of the box on the most common Linux desktop distros.

Keep an env escape hatch via `GSTACK_CHROMIUM_NO_SANDBOX` for users on distros where unprivileged userns sandbox actually works and who explicitly want it on — though as written, the platform check overrides the env var on Linux. (Open to feedback on whether the env var should be able to *re-enable* the sandbox on Linux when the user knows their AppArmor profile allows it.)

## Why this is safe

- Sandbox is already off on Windows (Bun→Node chain bug, GitHub #276) and on CI/CONTAINER/root.
- The threat model for a local daemon browsing **user-specified** URLs is "agent operator trusts the URLs they're navigating to" — pretty different from a multi-tenant browser. Marginal benefit of the sandbox in this context is small.
- Behavior on macOS unchanged (sandbox stays on).

## Test plan

- [x] Ubuntu 24.04.4 LTS, kernel 6.8.0-117-generic, `apparmor_restrict_unprivileged_userns=1`
  - Before patch: `browse goto https://example.com` → crash with FATAL above
  - After patch: `browse goto https://example.com` → "Navigated to https://example.com (200)", `browse js` / `browse snapshot` / `browse screenshot` all work
- [ ] macOS — sandbox stays on, no behavior change
- [ ] Windows — already covered by existing `process.platform !== 'win32'` check

## Reference

[Chromium docs on AppArmor userns restrictions](https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md) — the workaround they recommend for "live dangerously" cases is exactly `--no-sandbox`.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -221,7 +221,11 @@ export class BrowserManager {
     // are typically disabled in containers and are never available for the root
     // user on Linux. Detect all three cases and add --no-sandbox automatically.
     const isRoot = typeof process.getuid === 'function' && process.getuid() === 0;
-    if (process.env.CI || process.env.CONTAINER || isRoot) {
+    // Ubuntu/AppArmor setups often block unprivileged Chromium sandboxing for
+    // headless agent sessions even for normal users. Keep an env escape hatch,
+    // but default Linux agent browsing to no-sandbox so /qa works reliably.
+    const forceNoSandbox = process.env.GSTACK_CHROMIUM_NO_SANDBOX === '1' || process.platform === 'linux';
+    if (process.env.CI || process.env.CONTAINER || isRoot || forceNoSandbox) {
       launchArgs.push('--no-sandbox');
     }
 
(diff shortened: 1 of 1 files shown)


garrytan:
Thanks @techcenter68 — Linux --no-sandbox default shipped in v1.43.2.0 post-Daegu wave (#1642), rebased onto v1.42.2.0's `shouldEnableChromiumSandbox` helper.

PR #1571, opened 2026-05-17
fix: sourceLocalPath handles wrapped {sources:[...]} shape from gbrain v0.20+

jakehann11:
## Summary

`bin/gstack-gbrain-sync.ts:sourceLocalPath()` (added in v1.40.0.0 as part of the hostname-fold migration helpers) types the gbrain response as a flat `Array<{ id, local_path }>` and calls `.find()` on it directly. gbrain v0.20+ returns `{sources: [...]}` (object wrapping the array), not a flat array.

Result: `/sync-gbrain` crashes immediately on the orchestrator's first call to `sourceLocalPath` with:

```
```

The sibling helpers `probeSource` and `sourcePageCount` in `lib/gbrain-sources.ts` already handle the wrapped shape correctly (parse `{sources?: Array<...>}`, fall back to `[]`). This PR brings `sourceLocalPath` in line.

The fix accepts both shapes (flat array OR `{sources: [...]}`) for backward compatibility with any older gbrain still in the wild.

## Reproduction

- gstack v1.40.0.0 + gbrain v0.35.0.0 on Windows (MINGW64). Likely reproduces on macOS/Linux too — this is a pure data-shape bug, not a platform issue.
- Invoke `/sync-gbrain` in any repo that already has registered sources (e.g., from `gbrain sources add ... --path ...`).
- Orchestrator crashes before completing stage 1.

Workaround for affected users until this lands: skip the orchestrator and call `gbrain sync --all --no-pull --yes` directly.

## Test plan

- [x] New regression test `handles {sources: [...]} wrapped shape (gbrain v0.20+)` mocks the actual v0.20+ response shape and asserts the lookup succeeds.
- [x] New regression test `returns null when the source is missing in the wrapped shape` covers the empty-array case.
- [x] Existing tests preserved (flat-array shape still supported for backward compat).
- [ ] CI verifies on Linux (the existing `makeShim` infra uses `#!/bin/sh` shebang scripts so the test pattern doesn't run on Windows directly; pre-existing limitation).

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -287,13 +287,20 @@ function gbrainSupportsSourcesRename(env?: NodeJS.ProcessEnv): boolean {
  * `env` is the environment passed to the spawned `gbrain` process; defaults
  * to `process.env`. Tests inject a PATH that points at a gbrain shim so the
  * helper can be exercised without a real gbrain CLI.
+ *
+ * Shape note: `gbrain sources list --json` returns `{sources: [...]}` (v0.20+);
+ * older versions returned a flat array. Accept both for forward/backward compat
+ * (mirrors `probeSource`/`sourcePageCount` in lib/gbrain-sources.ts).
  */
 export function sourceLocalPath(sourceId: string, env?: NodeJS.ProcessEnv): string | null {
-  const list = execGbrainJson<Array<{ id: string; local_path?: string }>>(
+  const raw = execGbrainJson<unknown>(
     ["sources", "list", "--json"],
     { baseEnv: env },
   );
-  if (!list) return null;
+  if (!raw) return null;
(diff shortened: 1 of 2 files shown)


jbetala7:
This appears to overlap with #1564: both fix `sourceLocalPath()` in `bin/gstack-gbrain-sync.ts` for the same wrapped `{sources:[...]}` gbrain response and the same `list.find is not a function` crash.

I'd treat #1564 as the canonical PR because it was opened earlier and also covers the malformed wrapped payload case with an `Array.isArray(list)` guard before `.find()`. If this PR stays active, the missing malformed-payload guard/test should be ported here; otherwise closing this as duplicate avoids two competing fixes for the same path.

garrytan:
Thanks @jakehann11 — sourceLocalPath fix shipped in v1.42.0.0 Daegu wave (#1594) as the chosen patch.

PR #1570, opened 2026-05-17
fix: guard CLAUDE_PLUGIN_DATA against cross-plugin contamination in gstack-paths (#1569)

ElliotDrel:
## Summary

Fixes #1569.

`gstack-paths` previously trusted `CLAUDE_PLUGIN_DATA` as a fallback for `GSTACK_STATE_ROOT` whenever `GSTACK_HOME` was unset. When another plugin (e.g. Codex) uses `CLAUDE_ENV_FILE` to persist its own `CLAUDE_PLUGIN_DATA` into the session environment, gstack picked it up and wrote all state — checkpoints, analytics, learnings — into that plugin's data directory. Anyone with Codex installed alongside gstack hit this silently.

**Fix:** Guard the `CLAUDE_PLUGIN_DATA` branch so it only fires when `CLAUDE_PLUGIN_ROOT` also contains `"gstack"`, confirming we are running as the gstack plugin. Skill installs (the common path) fall through to `$HOME/.gstack` as intended.

## Changes

- `bin/gstack-paths`: add `grep -qi "gstack"` guard on `CLAUDE_PLUGIN_ROOT` before trusting `CLAUDE_PLUGIN_DATA`; update chain comment in header
- `test/gstack-paths.test.ts`: replace the previous `CLAUDE_PLUGIN_DATA wins over HOME` test with two targeted tests — one for the contamination scenario (non-gstack root → falls through to HOME), one for the valid plugin scenario (gstack root → uses CLAUDE_PLUGIN_DATA)
- `VERSION`: 1.40.0.0 → 1.40.1.0
- `CHANGELOG.md`: entry for v1.40.1.0

## Test plan

- [x] `bun test test/gstack-paths.test.ts` — 9 tests, all green (up from 7; 2 new targeted tests)
- [ ] Garry to run `bun test` (full free suite) before merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1573, opened 2026-05-18
fix(paths): Guard Claude plugin data fallback

### LINKED ISSUE #1569: gstack-paths incorrectly reads CLAUDE_PLUGIN_DATA when running as a skill (not a plugin)
**Version:** 1.40.0.0

**Summary**

`gstack-paths` uses `CLAUDE_PLUGIN_DATA` as a fallback for `GSTACK_STATE_ROOT` when `GSTACK_HOME` is unset. The script comments say this is intentional for when gstack is "installed as a Claude Code plugin." But gstack is most commonly installed as a skill, not a plugin — and in that case, `CLAUDE_PLUGIN_DATA` is not set by CC for gstack. It's whatever happens to be in the environment from other plugins.

**Root cause**

`~/.claude/skills/gstack/bin/gstack-paths` has this fallback chain:
```
GSTACK_HOME -> CLAUDE_PLUGIN_DATA -> $HOME/.gstack -> .gstack
```

`CLAUDE_PLUGIN_DATA` is set by Claude Code per-plugin during hook invocations. If another plugin (e.g. the Codex plugin) uses `CLAUDE_ENV_FILE` to persist its own `CLAUDE_PLUGIN_DATA` into the session environment, gstack picks it up and writes all its state — checkpoints, analytics, learnings — into that plugin's data directory.

**Reproduction**

1. Install the Codex plugin (`codex@openai-codex`)
2. Start a new Claude Code session
3. Run any gstack skill and check `GSTACK_STATE_ROOT`:
   ```bash
   ~/.claude/skills/gstack/bin/gstack-paths
   # GSTACK_STATE_ROOT=~/.claude/plugins/data/codex-openai-codex  ← wrong
   ```

**Fix**

The `CLAUDE_PLUGIN_DATA` branch should be guarded so it only applies when gstack is actually running as a plugin. One approach: check that `CLAUDE_PLUGIN_ROOT` resolves to a gstack directory before trusting `CLAUDE_PLUGIN_DATA`. Alternatively, remove `CLAUDE_PLUGIN_DATA` from the fallback chain entirely and document that plugin installs should set `GSTACK_HOME` explicitly.

**Workaround**

Set `GSTACK_HOME` in `~/.claude/settings.json`:
```json
{ "env": { "GSTACK_HOME": "/path/to/.gstack" } }
```

Gujiassh:
## Summary

- Guard `CLAUDE_PLUGIN_DATA` so `gstack-paths` only uses it when `CLAUDE_PLUGIN_ROOT` identifies the gstack plugin root.
- Fall back to `$HOME/.gstack` when gstack is running as a skill and the process inherited another plugin's data directory.
- Update path resolver docs and regression coverage for missing/foreign/gstack plugin roots.

Fixes #1569

## Tests

- `bun test test/gstack-paths.test.ts`
- `git diff --check`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -102,8 +102,8 @@ bun run skill:check      # health dashboard for all skills
 - **Windows**: curated Windows-safe subset runs on `windows-latest` via the
   `windows-free-tests` CI job. Setup script (`./setup`) requires Git Bash or
   MSYS today; native PowerShell support is a future expansion. The `bin/gstack-paths`
-  helper resolves state roots through `CLAUDE_PLUGIN_DATA` / `GSTACK_HOME` so plugin
-  installs work on every platform.
+  helper resolves state roots through `GSTACK_HOME` / guarded `CLAUDE_PLUGIN_DATA` so
+  plugin installs work on every platform without borrowing another plugin's data dir.
 
 ## Key conventions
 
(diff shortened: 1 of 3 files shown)


jbetala7:
This duplicates #1570 / #1569, but the `_gstack_plugin_data_is_ours()` predicate here is stricter than #1570's substring match and looks worth preserving.

Recommended path: port this predicate/test shape into the canonical #1570, then close this PR as duplicate; or have maintainers choose this PR as canonical and add any required version/changelog handling.

PR #1578, opened 2026-05-18
fix(build): stamp .version files from a script, not Bun-hostile subshell redirects

### LINKED ISSUE #1537: Auto-upgrade broken on Windows: Bun rejects subshell-redirect in package.json build script (v1.34.2.0+)

aifllow:
## Summary

`./setup` runs `bun run build`, and on Windows there is no `/bin/sh`, so Bun executes the npm script with its own shell. Bun's shell cannot parse `( cmd ) > file` — it aborts the **whole** script at parse time with `Subshells with redirections are currently not supported` (oven-sh/bun#11124, still open on every Bun version). The `build` script stamps three `.version` files with `( git rev-parse HEAD 2>/dev/null || true ) > …`, so on Windows the binary compile never even starts. Net effect: every Windows user is blocked from `/gstack-upgrade` past v1.34.1.0.

This moves the three stamps into `scripts/gen-version-files.ts` (`Bun.write`, zero shell). Output is byte-identical to the old form: each file gets the newline-terminated HEAD SHA, or an empty file when git is unavailable / not a repo (matching the old `2>/dev/null || true`).

Fixes #1537. Upstream limitation: oven-sh/bun#11124.

## Why a script (vs. the other options in #1537)

Tested all three suggestions from the issue on Bun 1.3.10 / Windows:

| Suggested fix | Result |
|---|---|
| 2. `git rev-parse HEAD > file 2>/dev/null \|\| true` (drop the subshell) | **Still fails** — Bun's shell also rejects two redirections on one command: `error: expected a command or assignment but got: "Redirect"` |
| 3. Run the build via `bash -c "…"` | Works, but reshapes `setup.sh`'s build invocation for every platform — broader surface than the bug |
| 1. Move stamping into a script (**this PR**) | Sidesteps the shell entirely; no platform-conditional logic; byte-identical output |

The tail `(rm -f .*.bun-build || true)` is **intentionally left unchanged**: Bun's shell only rejects subshells *with redirections*; a bare `( cmd || true )` parses and runs fine (verified). Changing it would be unrelated churn.

## Scope

- Three files, one concern (build-script Bun-shell compatibility).
- `scripts/gen-version-files.ts` — new, 39 lines, no deps, file mode `100644` like the other `scripts/*.ts`.
- `package.json` — the three `( ) > .version` stamps become `bun run scripts/gen-version-files.ts`. Nothing else in the build chain changes.
- `test/build-script-shell-compat.test.ts` — the second case asserted a subshell redirect **must exist** (`toBeGreaterThan(0)`), encoding the false premise that `( ) > file` is Bun-safe. Rewritten to assert no `>.version` shell redirect remains and that stamping is delegated to the script. The "no brace groups" case is unchanged.
- No behaviour change on macOS / Linux: same `.version` bytes, same build order.

## Test plan

- [x] `bun run build` completes (exit 0) through Bun's shell — previously aborted at parse time on Windows
- [x] `scripts/gen-version-files.ts` writes all three `.version` files with `git rev-parse HEAD`, byte-identical to the old form
- [x] `bun test test/build-script-shell-compat.test.ts` — 3 pass / 0 fail
- [x] Verified Bun's shell rejects suggested fix #2 (`> file 2>/dev/null`) and still accepts the unchanged bare-subshell tail
- [x] No other test references `PKG.scripts.build` / `scripts/gen-version-files.ts`; the only failure in the curated `bun run test:windows` shard (`skill-collision-sentinel`) reproduces on a clean checkout and is unrelated to this change
- [ ] macOS / Linux full `bun run build` parity — not verified here (no access); logic is platform-agnostic `Bun.write`, output is byte-identical


16francej:

PR #1582, opened 2026-05-18
fix(supabase-provision): rewrite transaction/6543 -> session/5432 for new projects (#1301)

### LINKED ISSUE #1301: /setup-gbrain: provision picks transaction pooler (6543) but new Supabase projects only listen on session pooler (5432)
## Summary

`gstack-gbrain-supabase-provision pooler-url` builds the Postgres URL using `db_port` from the Management API response, which for new Supabase projects returns the **transaction-mode** pooler at port 6543 — but newly-provisioned shared poolers actually listen on the **session-mode** port 5432. Result: `gbrain init` runs against an unreachable port. Symptoms cascade into ~30 min of confusing rotation/auth diagnosis.

## Repro

Fresh Supabase project provisioned via `/setup-gbrain` (path 2a auto-provision) on 2026-05-02:

1. Run `/setup-gbrain` → pick "Supabase, auto-provision"
2. Project becomes ACTIVE_HEALTHY ✅
3. `gstack-gbrain-supabase-provision pooler-url <ref>` returns a URL with port `6543`
4. `gbrain init` fails: `Cannot connect to database: (ENOTFOUND) tenant/user postgres.<ref> not found`
5. Manual `psql "$URL" -c 'SELECT 1;'` against port 6543: hangs to TCP timeout
6. Same psql against port 5432 (after manually swapping in config.json): `password authentication failed for user "postgres"` — error fires fast (different limb), so port is at least reachable
7. Supabase dashboard → Connect → Direct connection-string view shows `port: 5432, host: aws-1-us-east-1.pooler.supabase.com` for the shared pooler

## Root cause

`cmd_pooler_url` in `bin/gstack-gbrain-supabase-provision`:

```bash
if printf '%s' "$resp" | jq -e 'type == "array"' >/dev/null 2>&1; then
    first_or_session=$(printf '%s' "$resp" | jq '[.[] | select(.pool_mode == "session")][0] // .[0]')
else
    first_or_session="$resp"
fi
db_port=$(printf '%s' "$first_or_session" | jq -r '.db_port // empty')
local url="postgresql://${db_user}:${DB_PASS}@${db_host}:${db_port}/${db_name}"
```

For new projects the API returns a single object (not array), with `pool_mode=transaction` and `db_port=6543`. The fallback to `.[0]` in the array case has the same problem — first entry happens to be transaction mode. There's no validation that the picked entry's port is actually listening.

I'm not 100% sure whether this is a Supabase API regression (returning a transaction pooler config for a project that doesn't actually have one running) or whether the helper has always been wrong about which pool to prefer. Either way, the user-facing failure mode is the same.

## Local workaround (in my copy of the helper)

Added at `cmd_pooler_url` after parsing fields:

```bash
# Verified 2026-05-02 against fresh Supabase projects: the API returns
# transaction-mode (6543) by default for new projects, but the shared
# *session* pooler at 5432 is what actually accepts connections. Trusting
# db_port from the API gives "tenant/user not found" or "password auth
# failed" (TCP connects, auth rejects). Switch.
if [ "$pool_mode" = "transaction" ] && [ "$db_port" = "6543" ]; then
  db_port=5432
  pool_mode="session"
fi
```

This is a band-aid — a proper fix would either:
1. Query the actual pool configuration to confirm the listener is up before returning the URL
2. Always prefer the session pooler when both modes exist (current code does this for arrays but not single objects)
3. Add a TCP-connectivity probe inside `wait` before returning ACTIVE_HEALTHY

## Other findings while debugging

- **`wait` returns too early.** Polling returns success the moment the management API says `ACTIVE_HEALTHY`, but the pooler's tenant routing can take an additional 30-60 seconds to propagate. First `gbrain init` got `tenant/user postgres.<ref> not found` even with the correct port. My setup-gbrain wrapper now sleeps 60s after wait completes. Worth either bumping the wait condition or adding a separate `wait-pooler` sub-command that does a TCP handshake.
- **`gstack-gbrain-source-wireup --database-url <url>` leaks the URL via `ps aux`.** Forced me to rotate the DB password mid-flow. Should accept URL via env var only (the rest of the codebase enforces this for PAT and DB_PASS — `--database-url` is the inconsistent surface).

## Environment

- gstack: 1.26.0.0
- gbrain: 0.18.2
- Supabase project: us-east-1, Free tier, provisioned via Management API on 2026-05-02
- macOS Darwin 25.3.0

Happy to PR the band-aid fix if useful, or wait for a more principled solution. Either way figured the report is the right starting point.

0xDevNinja:
## Summary

- `cmd_pooler_url` in `bin/gstack-gbrain-supabase-provision` trusts `db_port` verbatim from the Management API. On a fresh Supabase project the API returns a single transaction-mode entry at port 6543, but the shared pooler tenant for new projects only listens on the session port 5432.
- Result: `gbrain init` hangs to TCP timeout (transaction port unreachable), then cascades into "tenant/user not found" / "password auth failed" errors that look like auth bugs. The reporter on #1301 lost ~30 min to that misdiagnosis loop.

This PR adds a narrow rewrite: when the picked entry is `pool_mode == "transaction"` AND `db_port == "6543"`, swap to `session` / `5432` and print a one-line stderr note. Every other shape (session/6543, transaction/5432, explicit array with a session entry) is untouched.

Set `GSTACK_SUPABASE_TRUST_API_PORT=1` to disable the rewrite if a future API version starts returning a working transaction port.

Fixes #1301.

## Why this shape

The reporter posted a verified band-aid in the issue body. I kept it intentionally narrow:

- Only rewrites the single known-bad combination (transaction + 6543). Doesn't touch session/6543 (some regions genuinely serve session on 6543, per the existing happy-path fixture in the test).
- Doesn't add a TCP-handshake probe inside `wait` — that's a separate fix the reporter also flagged. This PR is the immediate footgun stop; the probe-side work can land independently.
- Stderr note is opt-out, not opt-in, so users see what happened without reading the code.

## Tests

`test/gbrain-supabase-provision.test.ts` already mocks the Management API end-to-end via `Bun.serve`. 5 new cases under `describe('pooler-url')`:

- transaction/6543 single object → rewritten to session/5432, with `rewriting` stderr note
- session/6543 → untouched (no rewrite log)
- transaction/5432 → untouched
- `GSTACK_SUPABASE_TRUST_API_PORT=1` → rewrite suppressed, 6543 returned verbatim
- array with explicit session entry on 5432 → existing array-path behavior unchanged

```
bun test test/gbrain-supabase-provision.test.ts
# 33 pass, 0 fail (28 existing + 5 new)
```

The 7 pre-existing `gen-skill-docs --dry-run freshness check` failures on non-Claude hosts reproduce on `upstream/main` with this branch's changes stashed — unrelated to this patch.

## Out of scope

The reporter flagged two adjacent issues in the same comment which I'm leaving for separate PRs:

- `wait` returning `ACTIVE_HEALTHY` too early before the pooler tenant routes propagate (needs a real TCP-handshake probe).
- `gstack-gbrain-source-wireup --database-url <url>` leaking the URL via `ps aux` (env-var-only intake, separate file).

### DIFF
--- a/bin/gstack-gbrain-supabase-provision
+++ b/bin/gstack-gbrain-supabase-provision
@@ -339,7 +339,7 @@ cmd_pooler_url() {
   # Prefer the singular Session Pooler config when Supabase returns an
   # array (response shape can vary by project state). Fall back to the
   # first PRIMARY entry if no "session" pool_mode is present.
-  local db_user db_host db_port db_name
+  local db_user db_host db_port db_name pool_mode
   local first_or_session
   if printf '%s' "$resp" | jq -e 'type == "array"' >/dev/null 2>&1; then
     first_or_session=$(printf '%s' "$resp" | jq '[.[] | select(.pool_mode == "session")][0] // .[0]')
@@ -351,9 +351,25 @@ cmd_pooler_url() {
   db_host=$(printf '%s' "$first_or_session" | jq -r '.db_host // empty')
   db_port=$(printf '%s' "$first_or_session" | jq -r '.db_port // empty')
   db_name=$(printf '%s' "$first_or_session" | jq -r '.db_name // empty')
+  pool_mode=$(printf '%s' "$first_or_session" | jq -r '.pool_mode // empty')
 
   if [ -z "$db_user" ] || [ -z "$db_host" ] || [ -z "$db_port" ] || [ -z "$db_name" ]; then
     die "pooler-url: missing pooler config fields (db_user/db_host/db_port/db_name); re-poll or check project state"
+  fi
+
+  # Issue #1301: New Supabase projects' Management API returns a single
+  # transaction-mode pooler at port 6543, but the shared pooler tenant
+  # for fresh projects only listens on the session port 5432. Trusting
+  # db_port verbatim makes `gbrain init` hang to TCP timeout (transaction
+  # port unreachable) before falling into "tenant not found"-style errors
+  # that look like auth bugs. Rewrite transaction/6543 -> session/5432.
+  # Override with GSTACK_SUPABASE_TRUST_API_PORT=1 if a future API version
+  # starts returning a working transaction port and this rewrite is wrong.
+  if [ "${GSTACK_SUPABASE_TRUST_API_PORT:-0}" != "1" ] \
+     && [ "$pool_mode" = "transaction" ] && [ "$db_port" = "6543" ]; then
+    echo "pooler-url: API returned transaction pooler (port 6543); shared pooler for new projects listens on session port 5432 — rewriting (set GSTACK_SUPABASE_TRUST_API_PORT=1 to disable)" >&2
+    db_port=5432
+    pool_mode="session"
   fi
 
   local url="postgresql://${db_user}:${DB_PASS}@${db_host}:${db_port}/${db_name}"
--- a/test/gbrain-supabase-provision.test.ts
+++ b/test/gbrain-supabase-provision.test.ts
@@ -410,6 +410,89 @@ describe('pooler-url', () => {
     expect(r.status).toBe(2);
     expect(r.stderr).toContain('DB_PASS env var is required');
   });
+
+  // --- Issue #1301: New Supabase projects' API returns transaction/6543 but
+  // the shared pooler tenant only listens on session/5432. Rewrite that
+  // single combination, leave every other shape alone. ---
+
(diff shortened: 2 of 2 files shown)


garrytan:
Thanks @0xDevNinja — Supabase 6543→5432 rewrite shipped in v1.43.2.0 post-Daegu wave (#1642). Closes #1301.

PR #1586, opened 2026-05-18
fix(design): bump image-gen timeout to 240s + pin gpt-image-2

matteo-hertel:
## Problem

`/design-shotgun` (and any flow using the `design` binary) appears to **hang indefinitely** and never opens the comparison board.

Root cause: the binary calls `POST /v1/responses` with `model: gpt-4o` + the `image_generation` tool at `quality: "high"`, `1536x1024`, but aborts the request after a hardcoded **120 000 ms** (`design/src/generate.ts`, `variants.ts`, `evolve.ts`, `iterate.ts`).

That class of request **consistently takes ~140-160 s** end-to-end. Every generation aborts before the image is ready. In `/design-shotgun`, Step 3c launches N parallel agents each calling `$D generate`; every one aborts at 120 s, retries (another 120 s), all N fail, the board never opens, so the skill looks like an infinite hang.

## Evidence

Reproduced the **exact** API call the binary makes, with a longer budget:

| Call | Result |
|---|---|
| identical request, 300 s budget | HTTP 200, valid 1.7 MB image, **143.5 s** |
| same with `model: gpt-image-2` | HTTP 200, valid 1.85 MB image, **143.4 s** |
| binary abort deadline | **120 s** fires every time |

Then ran a real `/design-shotgun` end-to-end with the patched binary, 3 variants generated in parallel:

| Variant | Generate time | Quality gate |
|---|---|---|
| A | **150.0 s** | pass |
| B | **161.0 s** | pass |
| C | **152.1 s** | pass |

The 161 s case is the key data point: a naive bump to 150 s would still have failed it. 240 s leaves real margin while still bounding a genuinely stuck request.

## Change

- Bump the `AbortController` timeout `120_000` to `240_000` in `generate.ts`, `variants.ts`, `evolve.ts`, `iterate.ts` (both call sites in `iterate.ts`).
- Pin the `image_generation` tool to `model: "gpt-image-2"`.

## Testing

- `design/test/variants-retry-after.test.ts` (exercises the changed retry/timeout path): **5 pass, 0 fail**.
- Real `/design-shotgun` run: 3/3 variants generated and passed the vision quality gate, comparison board served.
- The `design/test/feedback-roundtrip.test.ts` failures are **pre-existing and unrelated**: they fail identically on clean `main` (`session.clearLoadedHtml is not a function`, a `browse`-module breakage in `write-commands.ts`). Not touched by this PR.

### DIFF
--- a/design/src/evolve.ts
+++ b/design/src/evolve.ts
@@ -52,7 +52,7 @@ export async function evolve(options: EvolveOptions): Promise<void> {
   ].join("\n");
 
   const controller = new AbortController();
-  const timeout = setTimeout(() => controller.abort(), 120_000);
+  const timeout = setTimeout(() => controller.abort(), 240_000);
 
   try {
     const response = await fetch("https://api.openai.com/v1/responses", {
@@ -64,7 +64,7 @@ export async function evolve(options: EvolveOptions): Promise<void> {
       body: JSON.stringify({
         model: "gpt-4o",
         input: evolvedPrompt,
-        tools: [{ type: "image_generation", size: "1536x1024", quality: "high" }],
+        tools: [{ type: "image_generation", model: "gpt-image-2", size: "1536x1024", quality: "high" }],
       }),
       signal: controller.signal,
     });
--- a/design/src/generate.ts
+++ b/design/src/generate.ts
@@ -37,7 +37,7 @@ async function callImageGeneration(
   quality: string,
 ): Promise<{ responseId: string; imageData: string }> {
   const controller = new AbortController();
-  const timeout = setTimeout(() => controller.abort(), 120_000);
+  const timeout = setTimeout(() => controller.abort(), 240_000);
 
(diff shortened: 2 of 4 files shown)


badcom:
There's a related PR open for this same issue: #1528

matteo-hertel:
> There's a related PR open for this same issue: #1528

@badcom the two PRs are fundamentally different. The fact that a PR exists doesn't meant it's the only viable solution.
The one you linked makes the timeout configurable, this one increases the timeout and bumps gpt-image-2.
Ultimately I get little value by having the time configurable, and therefore I made this. 
If the other one is merged then feel free to close this, I don't mind at all

garrytan:
Thanks @matteo-hertel — design timeout 240s + gpt-image-2 pin shipped in v1.43.2.0 post-Daegu wave (#1642). Closes #1519.

PR #1591, opened 2026-05-18
fix: detect PgBouncer transaction-mode pooler and set GBRAIN_PREPARE=true

### LINKED ISSUE #1435: /sync-gbrain capability check fails under PgBouncer transaction mode (put succeeds, search returns 'No results')
## Repro

On a Mac with gbrain v0.27.0 configured against a postgres engine over PgBouncer (transaction mode, port 6543):

```
$ SLUG="syncprobe59380"
$ echo "ping" | gbrain put "$SLUG"
# (no visible error, exit 0)

$ gbrain search "$SLUG"
No results.

$ gbrain delete "$SLUG"
{
  "status": "soft_deleted",
  "slug": "syncprobe59380",
  "recoverable_until": "now + 72h via restore_page"
}
```

The page was clearly stored — `gbrain delete` found and soft-deleted it. But `gbrain search` returned "No results" for the same slug immediately after the put.

Every `gbrain` invocation also prints this header:

```
[gbrain] Prepared statements disabled (PgBouncer transaction-mode convention on port 6543). Override with GBRAIN_PREPARE=true if your pooler runs in session mode.
```

`gbrain doctor --fast` reports OK, so this isn't visible to gstack's normal probes. The `/sync-gbrain` skill's `CAPABILITY_OK` check (put + search round-trip) correctly drops to 0 and skips writing the `## GBrain Search Guidance` block, but the underlying capability is broken and silent under `doctor --fast`.

## Why this matters for gstack

`/sync-gbrain` Step 4 (capability check) is what gates the `## GBrain Search Guidance` block in CLAUDE.md. With this bug, the block never gets written on machines using a transaction-mode PgBouncer pooler — even when memory + brain-sync stages succeed. Agents on those machines never learn to prefer `gbrain search` over Grep, defeating the point of the integration.

## Likely cause (guess, not confirmed)

The search index (tsvector or embedding-based) appears to not update synchronously after a `gbrain put` under transaction-mode pooling. Delete works because it looks up by slug, but search requires an index lookup that either:

1. Requires prepared statements (which were disabled), or
2. Runs in a deferred/async pass that's not triggered before the next CLI call, or
3. Writes to a different source/index than the put.

The `[gbrain] Prepared statements disabled` warning hints at (1). The `GBRAIN_PREPARE=true` override mentioned in the warning is not exposed through any gstack env or config — users hitting this have no documented escape hatch.

## Suggested fixes (any of)

1. **Pass `GBRAIN_PREPARE=true` through gstack env when the user is on a postgres engine over port 6543** — or whenever gstack detects the prepared-statements warning. Document this in `/setup-gbrain`.
2. **Make `gbrain doctor` exercise the full put → search → delete round-trip** (not just `--fast`), so this is visible in standard health output. Right now the only way to detect it is `/sync-gbrain`'s capability probe.
3. **If search really does need prepared statements under transaction-mode pooling**, surface a clear setup error during `/setup-gbrain` instead of letting subsequent skills mysteriously fail to write the guidance block.

## Environment

- gstack v1.26.4.0
- gbrain v0.27.0 (postgres engine)
- Pooler: PgBouncer, transaction mode, port 6543
- macOS

## Related

Companion issue filed separately: #1434 (code source ID exceeds 32-char limit). Both surfaced in the same `/sync-gbrain` run.

Filed via Claude Code at the user's request.

mikeangstadt:
## Summary

Fixes #1435 — `/sync-gbrain` capability check fails under PgBouncer transaction mode because `gbrain search` silently returns no results when prepared statements are disabled.

- **`lib/gbrain-exec.ts`**: `buildGbrainEnv()` now detects port 6543 in DATABASE_URL and auto-sets `GBRAIN_PREPARE=true` in the env for all gbrain spawns. Single chokepoint fix — every gstack-originated gbrain call inherits it. Callers can override with `GBRAIN_PREPARE=false`.
- **`sync-gbrain/SKILL.md{,.tmpl}`**: Capability check exports `GBRAIN_PREPARE=true` and retries search up to 3x with 1s delay for async index propagation under connection pooling.
- **`bin/gstack-gbrain-detect`**: Surfaces `gbrain_pooler_mode` ("transaction" | "session" | null) in the preamble probe JSON for downstream skill visibility.

## Test plan

- [ ] New unit tests for `isTransactionModePooler()` — port 6543, 5432, no-port, unparseable, postgres:// scheme
- [ ] New unit tests for `buildGbrainEnv()` GBRAIN_PREPARE auto-detection — pooler URL sets it, session URL doesn't, caller opt-out respected, already-matching URL still sets it
- [ ] Existing `buildGbrainEnv` tests still pass (DATABASE_URL seeding behavior unchanged)
- [ ] Manual: on a PgBouncer transaction-mode setup, run `/sync-gbrain` and verify the capability check passes and `## GBrain Search Guidance` block is written to CLAUDE.md

---

Continuous collaboration, powered by [ClosedLoop.AI](https://closedloop.ai) ([GitHub](https://github.com/closedloop-ai/claude-plugins))

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/bin/gstack-gbrain-detect
+++ b/bin/gstack-gbrain-detect
@@ -18,7 +18,8 @@
  *     "gstack_brain_sync_mode": "off"|"artifacts-only"|"full",
  *     "gstack_brain_git": true|false,
  *     "gstack_artifacts_remote": "https://..." | "",
- *     "gbrain_local_status": "ok"|"no-cli"|"missing-config"|"broken-config"|"broken-db"
+ *     "gbrain_local_status": "ok"|"no-cli"|"missing-config"|"broken-config"|"broken-db",
+ *     "gbrain_pooler_mode": "transaction"|"session"|null
  *   }
  *
  * Backward compatibility (per plan codex #5): the 9 pre-existing fields stay
@@ -42,6 +43,7 @@ import {
   resolveGbrainBin,
   readGbrainVersion,
 } from "../lib/gbrain-local-status";
+import { isTransactionModePooler } from "../lib/gbrain-exec";
 
 const STATE_DIR = process.env.GSTACK_HOME || join(userHome(), ".gstack");
 const SCRIPT_DIR = __dirname;
@@ -96,6 +98,17 @@ function detectConfig(): { exists: boolean; engine: "pglite" | "postgres" | null
     return { exists: true, engine: parsed.engine };
   }
   return { exists: true, engine: null };
+}
+
+// --- pooler mode detection (#1435) ---
+//
+// Reads DATABASE_URL from ~/.gbrain/config.json and checks whether it targets
+// a PgBouncer transaction-mode pooler (port 6543). Surfaced so /sync-gbrain
+// and /setup-gbrain can advise users when search may require GBRAIN_PREPARE.
+function detectPoolerMode(): "transaction" | "session" | "unknown" | null {
+  const parsed = tryReadJSON(GBRAIN_CONFIG) as { database_url?: string } | null;
+  if (!parsed?.database_url) return null;
+  return isTransactionModePooler(parsed.database_url) ? "transaction" : "session";
 }
 
 // --- gbrain doctor health (any nonzero exit or non-"ok"/"warnings" status → false) ---
@@ -215,6 +228,7 @@ function main(): void {
     gstack_brain_git: detectBrainGit(),
     gstack_artifacts_remote: detectArtifactsRemote(),
     gbrain_local_status: localEngineStatus({ noCache }),
+    gbrain_pooler_mode: detectPoolerMode(),
   };
 
   process.stdout.write(JSON.stringify(out, null, 2) + "\n");
--- a/bin/gstack-gbrain-detect
+++ b/bin/gstack-gbrain-detect
@@ -19,13 +19,14 @@
  *     "gstack_brain_git": true|false,
  *     "gstack_artifacts_remote": "https://..." | "",
  *     "gbrain_local_status": "ok"|"no-cli"|"missing-config"|"broken-config"|"broken-db",
- *     "gbrain_pooler_mode": "transaction"|"session"|null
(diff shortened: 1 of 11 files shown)


jbetala7:
This detection is using only port `6543` as `transaction`. That conflicts with existing `gstack-gbrain-supabase-verify`, which accepts `6543` as the canonical Session Pooler URL, and with #1582, which treats `pool_mode` as the thing that distinguishes transaction vs session. As written, `gbrain_pooler_mode` can report `transaction` for a working session pooler, and `buildGbrainEnv` will force `GBRAIN_PREPARE=true` without knowing the actual mode. Could we avoid naming it transaction from the URL alone, or persist/read Supabase `pool_mode` when available, and add tests for session/6543 and transaction/6543 separately?

mikeangstadt:
Resolved on `closedloop-ai:fix/1435-pgbouncer-transaction-mode-search` in `757f684f`.

I removed the URL-only `:6543 => transaction` inference. `buildGbrainEnv()` now only auto-sets `GBRAIN_PREPARE=true` when `~/.gbrain/config.json` has a persisted Supabase `pool_mode: "transaction"`. If the URL is a Supabase pooler but we do not have persisted mode metadata yet, `gstack-gbrain-detect` now reports `gbrain_pooler_mode: "unknown"` instead of guessing.

To make that metadata available, `gstack-gbrain-supabase-provision pooler-url --json` now returns `pool_mode`, and `/setup-gbrain` persists it into `~/.gbrain/config.json` right after `gbrain init`. I also added coverage for transaction/session on the same `:6543` pooled URL, the `unknown` fallback, detect output shape, and the provisioner JSON payload.

garrytan:
Thanks @mikeangstadt — PgBouncer transaction-mode detection shipped in v1.43.2.0 post-Daegu wave (#1642). Closes #1435.

Issue #393, opened 2026-03-23
Add GitHub Copilot CLI as a supported host


ridermw:
## Summary

The GitHub Copilot CLI went GA in February 2026 as a standalone binary (`copilot`), separate from the old `gh copilot` extension. It supports skill discovery via `~/.copilot/skills/` and has its own session storage at `~/.copilot/session-state/{id}/`.

gstack already supports Claude Code, Codex, Gemini CLI, and Cursor. Copilot CLI fits the same pattern — it uses `.agents/skills/` for repo-local discovery and has a global install path at `~/.copilot/skills/gstack`.

## What would Copilot support look like?

Based on the existing Codex pattern:

- **gen-skill-docs.ts**: Add `'copilot'` host type. Copilot shares `.agents/skills/` output with Codex — only the global fallback path differs (`~/.copilot/skills/gstack` vs `~/.codex/skills/gstack`).
- **setup**: Add `--host copilot` with auto-detection via `command -v copilot`. Rewrite Codex output paths at install time via sed (same pattern as Kiro).
- **gstack-global-discover.ts**: Add `scanCopilot()` for session discovery at `~/.copilot/session-state/`.
- **Tests**: Copilot E2E runner paralleling the Codex session runner, updated assertions in existing tests.
- **Docs**: README, CONTRIBUTING, AGENTS.md updated to mention Copilot as a peer host.

No build pipeline changes needed — Copilot derives from Codex output at setup time. No architectural changes.

## Copilot CLI facts

| Aspect | Value |
|--------|-------|
| Binary | `copilot` (standalone, not `gh copilot`) |
| Install | `npm install -g @github/copilot` (also: brew, winget, install script) |
| Config dir | `~/.copilot/` |
| Session data | `~/.copilot/session-state/{session-id}/` |
| Docs | https://docs.github.com/copilot/how-tos/set-up/install-copilot-cli |

## Reference implementation

I've put together a working implementation in [PR #396](https://github.com/garrytan/gstack/pull/396). 501 lines added across 14 files, follows the existing Codex/Kiro pattern exactly. Happy to close it if you'd prefer to implement this differently.

Related: [#289](https://github.com/garrytan/gstack/issues/289) (host integration contract RFC) — this would be another concrete data point for that discussion.

arkar16:
Is this ready to implement and use? Thanks for making this a first-class option. Good to add more harnesses

maskaravivek:
[...]

PR #1593, opened 2026-05-19
keep .copilot only once

Suryakandukuri:


jbetala7:
This looks stale/incomplete relative to the previous Copilot-host attempts. #1436 and #1443 were closed after real Copilot CLI testing showed the install layout matters: agents vs skills, and nested collection dirs did not show in interactive `/` autocomplete. This PR adds `hosts/copilot.ts` but does not update `setup`, so `./setup --host copilot` still hits the unknown-host branch, and it uses `.copilot/skills/gstack` without the setup/linking logic from #1443. Could you either wire the full install path plus setup tests, or scope this as generated-docs-only and drop the README claim that users can target `--host copilot`?

time-attack:
@16francej — recommend closing this PR. This is superseded by #393, which owns the generic installer and root policy; these changes add Copilot metadata without a setup path. Please carry forward the configuration, collision findings, and attribution under #393.

dgrant:
Please close, leaving this open causes confusion.

PR #1594, opened 2026-05-19
v1.42.0.0 Daegu wave: 23 community-filed bugs + PTY classifier enforcement (24 bisect commits)

### LINKED ISSUE #1214: browse: full-page screenshots silently exceed Anthropic vision API 2000px limit, bricking sessions

### LINKED ISSUE #1248: `design` binary silently uses `OPENAI_API_KEY` from any `.env` in cwd — security/billing concern

### LINKED ISSUE #1327: [codex skill] Surface non-zero exit codes from codex wrappers (currently silent on parse errors)

### LINKED ISSUE #1346: gstack-memory-ingest broken on gbrain >=0.18: still calls renamed `put_page` subcommand
## Environment

- gstack `1.26.4.0` (`~/.claude/skills/gstack/VERSION`)
- gbrain `0.18.2`
- Linux, bun runtime

## Reproducer

```
$ gbrain --version
gbrain 0.18.2
$ gbrain --help | grep -A 1 PAGES
PAGES
  get <slug>                         Read a page
  put <slug> [< file.md]             Write/update a page
$ echo "hi" | gbrain put_page --slug test --type note
Unknown command: put_page
Run gbrain --help for available commands.
```

Result: every transcript ingest fails. In a fresh setup with 292 transcripts queued, all 292 hit `[put-error] ... Unknown command: put_page` and nothing lands in the brain.

## Expected

`gstack-memory-ingest` should use the canonical gbrain page-write subcommand and successfully ingest pages.

## Actual

`bin/gstack-memory-ingest.ts:768` builds the args array with the literal `"put_page"`:

```ts
const args = [
  "put_page",
  "--slug", page.slug,
  "--title", page.title,
  "--type", page.type,
  "--tags", page.tags.join(","),
];
execFileSync("gbrain", args, { ... });
```

`put_page` is the gbrain `<=0.17` name. gbrain `>=0.18` renamed it to `put` (same flags, same stdin contract for the body).

## Suggested fix

One-line change at `bin/gstack-memory-ingest.ts:768`: `"put_page"` -> `"put"`. The rest of the call site (flags, stdin body, timeouts) is unchanged. PR to follow.

### LINKED ISSUE #1370: security: documented sidebar security stack does not match shipped architecture; PTY-injection path bypasses every classifier layer
## Summary

The "Sidebar security stack" documented in CLAUDE.md describes a layered ML defense (L4 testsavant + L4b Haiku transcript + L4c DeBERTa) hosted in `sidebar-agent.ts`. CLAUDE.md "Sidebar architecture" elsewhere documents that `sidebar-agent.ts` was ripped when the PTY proved out (the chat-queue path is gone, `/sidebar-command` and `/sidebar-chat` and `/sidebar-agent/event` endpoints are gone).

The result is two surfaces that the table no longer matches:

1. **`browse/src/security-classifier.ts` is unreferenced from `browse/src/`.** `grep -rn "from.*security-classifier" browse/src/` returns zero hits. The file is loaded only by `browse/test/*.test.ts`. None of `loadTestsavant`, `scanPageContent`, `scanPageContentDeberta`, `combineVerdict` runs against any production data path on `7b4738b`.

2. **The new `window.gstackInjectToTerminal(text)` PTY path runs no L1-L3 either.** Per CLAUDE.md "Cross-pane PTY injection," the toolbar Cleanup button and the Inspector "Send to Code" action pipe text directly to the live `claude` REPL. That text is page-derived (i.e., influenced by whatever site the operator was inspecting). It does not pass through `wrapUntrustedPageContent` / `markHiddenElements` / the URL blocklist / the canary-injection step. The L1-L3 module (`content-security.ts`) still ships, but nothing on the PTY-injection path calls it.

`browse/src/server.ts:1165` continues to surface `security: getSecurityStatus()` on `/health`, and `getSecurityStatus()` (`security.ts:582-604`) reports `layers: { testsavant, transcript, canary }` from `~/.gstack/security/session-state.json`. With sidebar-agent gone, nothing writes that file in the new architecture, so the layers shown to whatever consumer reads `/health` reflect whatever the last sidebar-agent run wrote — possibly months ago, possibly empty, possibly `'off'` for everything.

## Concrete exploit chain (HIGH)

1. Operator opens an attacker-controlled page in the gstack browser.
2. Page contains a prompt-injection payload in DOM text, hidden elements, ARIA labels, or simply prose.
3. Operator clicks Inspector → "Send to Code" or the toolbar Cleanup button.
4. Page-derived text reaches the live `claude` REPL via `gstackInjectToTerminal` with no envelope wrap, no hidden-strip, no classifier scan, no canary check.
5. The REPL executes the injected prompt as if user-typed.

The chain is simpler than the L1-L6 stack was supposed to make it. The documented mitigation is the load-bearing defense. The fact that none of it is wired is the bug.

## Mechanical evidence

```
$ grep -rn "from.*security-classifier" browse/src/
$ grep -rn "import.*security-classifier" browse/src/
$ ls browse/src/sidebar-agent.ts
ls: browse/src/sidebar-agent.ts: No such file or directory
$ grep -rn "recordSkillUse" browse/src/ | grep -v "domain-skills.ts:"
$
```

(Last grep: zero callers of `recordSkillUse(..., classifierFlagged: true)` outside the module under test, confirming the L4 → flag_count → auto-promote chain is also broken — separately filed in #1369.)

## Two acceptable shapes for a fix

**Option A — re-wire the classifier:**
- Move L1-L3 envelope wrap and L4 / L4b scan into `sidepanel-terminal.js` before `gstackInjectToTerminal` reaches the PTY (extension-side call back to a daemon endpoint for ML, since `onnxruntime-node` can't load from `bun --compile`'s temp extract dir per CLAUDE.md note).
- Re-add a caller for `recordSkillUse(..., classifierFlagged: true)` so the auto-promote gate (#1369) re-opens once L4 returns.
- Keep the security stack table in CLAUDE.md.

**Option B — delete + de-document:**
- Remove `browse/src/security-classifier.ts`.
- Drop the `testsavant` / `transcript` / `deberta` layers from `security.ts:getStatus()` and from `/health`'s `security` field.
- Drop the "Sidebar security stack" table from CLAUDE.md and update "Sidebar architecture" / "Cross-pane PTY injection" to honestly describe the PTY path as the operator-trust surface it is.
- Drop the `GSTACK_SECURITY_OFF` / `GSTACK_SECURITY_ENSEMBLE` env knobs and the model-cache paths from CLAUDE.md.

Either is fine. The current state — table claims defense, code provides none, `/health` reports defense as live — is what should not stand.

## Related PRs and findings

- #1368 — `security: pass cwd to git via execFileSync, not interpolation through /bin/sh` — separate finding, lands cleanly in either of the above options.
- #1369 — `security: gate domain-skill auto-promote on classifier_score > 0` — partial mitigation for one specific consequence of the dead classifier (without the gate, three benign uses promote any quarantined skill, including one authored under the influence of a poisoned page, into prompt context). Lands cleanly in either option above. If Option A is chosen, the gate re-opens automatically once L4 is rewired.
- #1153 (open since 2026-04-22) — `.svg` in `load-html` allowlist — escalates with this issue: an SVG payload in `about:blank` with no CSP is exactly the shape the L4 classifier was supposed to catch.
- #1155, #1157 (open since 2026-04-22 / 2026-04-23) — both target `security-classifier.ts` itself. If Option B is chosen, both can be closed as obsolete.

Filed as an issue rather than a PR because the fix shape is a design call, not a one-line change.

### LINKED ISSUE #1479: /codex review errors on codex-cli >= 0.130.0 — `--base` and PROMPT now mutually exclusive
## Repro

```bash
$ codex --version
codex-cli 0.130.0

$ codex review "boundary instruction here" --base main -c 'model_reasoning_effort="high"' --enable web_search_cached

Usage: codex review --config <key=value> --enable <FEATURE> <PROMPT>
```

`codex review --help` confirms the new constraint:

> Arguments:  `[PROMPT]`  Custom review instructions. If `-` is used, read from stdin
>
> Options: `--uncommitted` | `--base <BRANCH>` | `--commit <SHA>`

`--base`, `--commit`, `--uncommitted`, and `[PROMPT]` are now mutually exclusive scope selectors.

## Affected file

`codex/SKILL.md` — Step 2A (Review Mode):

```bash
codex review "IMPORTANT: Do NOT read or execute any files under ~/.claude/, ~/.agents/, .claude/skills/, or agents/. ..." --base <base> -c 'model_reasoning_effort="high"' --enable web_search_cached
```

This invocation always errors on codex-cli >= 0.130.0. Step 2B (Challenge) and 2C (Consult) use `codex exec` with the prompt as a positional arg, which still works — only `codex review` is affected.

## Suggested fix

The boundary instruction is no longer compatible with `--base` as a positional arg. Drop the prompt arg in Step 2A. `codex review --base main` already scopes Codex to the diff, and the skill already runs in `cd "$_REPO_ROOT"` so Codex's sandbox is the repo anyway — the boundary "don't wander into ~/.claude/" is largely unnecessary in scoped-diff mode.

```bash
codex review --base "$BASE" -c 'model_reasoning_effort="high"'
```

(Piping the boundary via stdin using `-` also looks blocked — `--help` still shows `--base` and PROMPT as mutually exclusive even with `-`. Drop-the-prompt is the safe path.)

## Version-bad-list update

Consider adding `>= 0.130.0` to the known-incompatible range in `bin/gstack-codex-probe`'s known-bad list until SKILL.md is patched.

## Environment

- macOS (darwin 25.4.0), zsh
- gstack 1.33.2.0
- codex-cli 0.130.0
- gtimeout from Homebrew coreutils available

### LINKED ISSUE #1503: /codex review: bare-path drops filesystem boundary on diffs that touch .claude/skills/

### LINKED ISSUE #1538: setup fails on Windows: Bun shell doesn't support subshells with redirection in build script

### LINKED ISSUE #1559: gstack-brain-context-load treats gbrain as missing when availability probe execs command directly

### LINKED ISSUE #1567: gstack-gbrain-sync: sourceLocalPath() expects bare array but gbrain ≥0.19 returns { sources: [...] } → fatal 'list.find is not a function'
## Summary

`/sync-gbrain` crashes fatally on the first run with:

```
```

The orchestrator never reaches any stage (code/memory/brain-sync) — it dies during planning.

## Environment

- gstack: **1.40.0.0**
- gbrain: **0.35.4.0** (engine: pglite, local-stdio)
- macOS (darwin arm64), bun 1.3.11
- Repo: a normal local git repo, freshly `git init`'d, source not yet registered

## Root cause

`bin/gstack-gbrain-sync.ts` → `sourceLocalPath()` (around line 292) calls
`gbrain sources list --json` and types/treats the result as a bare array:

```ts
const list = execGbrainJson<Array<{ id: string; local_path?: string }>>(
  ["sources", "list", "--json"],
  { baseEnv: env },
);
if (!list) return null;
const found = list.find((s) => s.id === sourceId);   // ← list is not an array
```

But gbrain **>= 0.19** wraps the listing in an object:

```json
{ "sources": [ { "id": "default", "local_path": null, "page_count": 99, ... } ] }
```

So `list.find` is `undefined` → fatal. gstack pins gbrain 0.18.2
(`bin/gstack-gbrain-install` `PINNED_COMMIT`), so this only bites users who run
a newer gbrain — which is increasingly common since gbrain's own update path
moves ahead of the gstack pin.

## Repro

1. Run any gbrain >= 0.19 (`gbrain --version`).
2. In a git repo whose code source is not yet registered, run `/sync-gbrain`
   (or `bun run ~/.claude/skills/gstack/bin/gstack-gbrain-sync.ts`).
3. Fatal: `list.find is not a function`.

## Fix (verified locally)

Accept both the legacy array and the new `{ sources: [...] }` shape:

```ts
const raw = execGbrainJson<
if (!raw) return null;
// gbrain <=0.18 returned a bare array; >=0.19 wraps it as { sources: [...] }.
const list = Array.isArray(raw) ? raw : (raw.sources ?? []);
```

After this patch, `/sync-gbrain` runs 3/3 stages OK against gbrain 0.35.4.0
(code source registered + synced, memory import, brain-sync push).

`sourceLocalPath()` is the only `execGbrainJson` consumer in the file; other
gbrain interactions use `spawnGbrain` (exit-code based) and are unaffected.
Note `/sync-gbrain` SKILL.md Step 3 already does `... | jq '.sources[]'`,
so the wrapped shape is the expected one — only this helper was missed.

### LINKED ISSUE #1569: gstack-paths incorrectly reads CLAUDE_PLUGIN_DATA when running as a skill (not a plugin)

garrytan:
## Summary

24 bisect-clean commits closing 14 distinct user-facing problems across compat, install, security, and quality surfaces. The biggest single wave since v1.18. Plan file:
[`~/.claude/plans/system-instruction-you-are-working-expressive-snowglobe.md`](../tree/garrytan/daegu-v3) (see local checkout; not committed).

**Themes:**
- **Compat:** gbrain v0.18-0.35 API drift (3 surfaces), Codex CLI 0.130+ argv break, `command -v` for codex detection, schema_version: 2 doctor parsing
- **Install:** Windows `bun run build` chain broken since v1.34.2.0 (5 issue PRs collapsed into one canonical fix), `.exe` extension resolution, new `windows-setup-e2e.yml` gate
- **Security:** PTY-injection bypass closed end-to-end (#1370): Node sidecar hosting the L4 classifier + new `/pty-inject-scan` endpoint + extension pre-scan wiring + static AST invariant test
- **Quality:** Screenshot 2000px guard across all 3 full-page paths, `$D` OpenAI key source disclosure, codex non-zero exit surfacing, opt-in extended stealth

## Bisect commits (in order)

1. `fix(gstack-paths)` CLAUDE_PLUGIN_DATA guard (#1570) — closes #1569
2. `fix(gbrain-sync)` sourceLocalPath wrapped shape (#1571) — closes #1567, supersedes #1564
3. `fix(brain-context-load)` probe via execFile (#1560) — closes #1559
4. `test(gbrain-doctor)` schema_version:2 regression pin — closes #1418
5. `test(memory-ingest)` put_page regression + stale comment scrub — closes #1346
6. `fix(resolvers)` rewrite all gbrain put_page → put across 10 user-facing templates
7. `fix(build)` extract package.json build → scripts/build.sh (#1544) — closes #1538/#1537/#1530/#1457/#1561, supersedes #1531/#1480/#1460
8. `fix(windows)` .exe glob + find-browse resolution (#1554)
9. `ci(windows)` fresh `./setup` E2E gate
10. `fix(codex)` move diff scope into prompt (#1209) — closes #1479, supersedes #1527/#1449
11. `fix(review)` diff from git merge-base (#1492)
12. `test(codex)` filesystem boundary preservation pin — closes #1503, supersedes #1522
13. `fix(skills)` command -v for codex detection (#1197)
14. `fix(codex)` surface non-zero exits (#1467) — closes #1327
15. `fix(design)` OpenAI key source disclosure (#1278) — closes #1248
16. `fix(browse)` screenshot 2000px guard, all 3 paths — closes #1214
17. `feat(security)` Node sidecar entry for L4 classifier
18. `feat(security)` sidecar IPC client with circuit breaker
19. `feat(security)` POST /pty-inject-scan endpoint — wires #1370
20. `feat(extension)` route gstackInjectToTerminal through /pty-inject-scan
21. `test(security)` extension PTY inject invariant — closes #1370 end-to-end
22. `feat(browse)` opt-in extended stealth mode (#1112 rebased)
23. `test(fixtures)` regenerate ship-SKILL.md golden baselines
24. `chore(release)` v1.41.0.0 CHANGELOG + VERSION

## Coverage

- 60+ new unit tests across the wave
- 3 new static-invariant tests (PTY inject scan-gated, no put_page in resolvers, no put_page in memory-ingest argv)
- All wave-touched tests pass in isolation (92/92)
- New `windows-setup-e2e.yml` workflow gates the install path against future regressions

## Test plan

- [ ] `bun test` (free tier) — verify wave-touched files pass; pre-existing test-pollution between files documented in CHANGELOG
- [ ] `bun run gen:skill-docs --host all` clean (no diff after regen)
- [ ] Windows CI `./setup` E2E green on the new workflow
- [ ] `gbrain --version` (0.18-0.35 range): `/sync-gbrain` succeeds
- [ ] `/codex review` against Codex CLI 0.130+: review produces output without argv error
- [ ] Cleanup button + Inspector "Send to Code": injection routed through `/pty-inject-scan`
- [ ] `$D design generate` inside a rogue-cwd-`.env` directory: source disclosure printed; key not echoed
- [ ] Screenshot of 5000px-tall page: downscaled to ≤2000px with warning to stderr
- [ ] SannySoft test with `GSTACK_STEALTH=extended`: 100% pass

## Post-merge

Execute PR/issue triage close-out from plan §1A-1E: 22 PRs + 6 issues closed with credit comments referencing v1.41.0.0.


jbetala7:
One correctness issue with the #1569 carry-over: `elif ... echo "$CLAUDE_PLUGIN_ROOT" | grep -qi "gstack"` still trusts any foreign plugin path containing the substring `gstack`, so leaked `CLAUDE_PLUGIN_DATA` can still become `GSTACK_STATE_ROOT`. This is the same edge I raised on #1570, and #1594 is now the canonical wave carrying that fix. Can we tighten this to verify the root is actually gstack, such as basename exactly `gstack`/`gstack-*` or checking `$CLAUDE_PLUGIN_ROOT/bin/gstack-paths` exists, and add a regression with a foreign root containing `gstack`?

PR #1598, opened 2026-05-19
fix(artifacts): reject malformed remote paths

### LINKED ISSUE #1597: gstack-artifacts-url accepts host-only and owner-only remotes as repositories
## Observed problem

`gstack-artifacts-url` accepts malformed remotes that contain only a host, or
only a host plus owner, and treats those partial paths as valid repositories.

Current behavior on upstream `main`:

```console
$ bin/gstack-artifacts-url --to ssh https://github.com
git@github.com:github.com.git

$ bin/gstack-artifacts-url --owner-repo https://github.com/owner
owner

$ bin/gstack-artifacts-url --to https ssh://git@github.com
https://github.com/github.com
```

Those outputs are not valid owner/repo remotes. A caller that passes a truncated
or misconfigured artifacts remote can silently derive the wrong SSH/HTTPS URL
instead of failing fast.

## Expected behavior

The helper should reject remote forms whose path is empty, missing the owner, or
missing the repo. It should keep accepting valid inputs like:

- `https://github.com/owner/repo`
- `https://github.com/owner/repo.git`
- `git@github.com:owner/repo.git`
- `ssh://git@github.com/owner/repo.git`

## Duplicate search

I checked open issues and PRs for:

- `gstack-artifacts-url`
- `artifacts url malformed`
- `artifacts remote URL`
- `gstack-artifacts`

Closest existing items:

- #1348 / #1469 cover choosing SSH vs HTTPS push URLs based on `gh`/`glab`
  protocol config. They do not change malformed URL validation in
  `bin/gstack-artifacts-url`.
- Other open `gstack-artifacts` PRs target artifacts sync, docs, migration, or
  branch wiring rather than the URL parser accepting host-only/owner-only input.

## Candidate fix shape

Tighten `parse_url()` so the parsed path must contain at least two non-empty
segments (`owner/repo`) after stripping `.git`, and add regression coverage in
`test/gstack-artifacts-url.test.ts`.

jbetala7:
## Summary

- reject artifact remote URLs that do not contain both owner and repo path segments
- keep valid GitHub/GitLab-style remotes working, including deeper self-hosted paths
- add regression coverage for host-only and owner-only malformed inputs

Fixes #1597.

## Root cause

`parse_url()` stripped `.git` and only checked that `host` and `owner_repo`
were non-empty. For inputs like `https://github.com` or
`ssh://git@github.com`, parameter expansion assigned `github.com` to both the
host and path-like value, so the helper generated bogus remotes instead of
failing.

## Testing

- `bun test test/gstack-artifacts-url.test.ts`
- `bun test test/gstack-artifacts-init.test.ts -t "only gh authed"`
- `git diff --check`

### DIFF
--- a/bin/gstack-artifacts-url
+++ b/bin/gstack-artifacts-url
@@ -49,6 +49,19 @@ strip_git() {
   echo "${1%.git}"
 }
 
+valid_owner_repo() {
+  local owner_repo="$1"
+  case "$owner_repo" in
+    ""|/*|*/|*//*)
+      return 1
+      ;;
+  esac
+  case "$owner_repo" in
+    */*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
 # Parse to (host, owner_repo) regardless of input shape.
 parse_url() {
   local u="$1"
@@ -82,7 +95,7 @@ parse_url() {
       exit 3
       ;;
   esac
-  if [ -z "$host" ] || [ -z "$owner_repo" ] || [ "$owner_repo" = "$u" ]; then
+  if [ -z "$host" ] || ! valid_owner_repo "$owner_repo"; then
     echo "gstack-artifacts-url: failed to parse host/owner from: $u" >&2
     exit 3
   fi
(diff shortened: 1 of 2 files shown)


garrytan:
Your fix landed in v1.43.2.0 (the CHANGELOG credits you by name via PR #1598). Closing this PR — the work shipped through a parallel route. Thanks for the contribution, @jbetala7!

PR #1600, opened 2026-05-19
fix(migrations): defer v1.40.0.0 done-marker when jq is missing (#1581)

### LINKED ISSUE #1581: v1.40.0.0 migration writes done-marker even when jq is missing, leaving privacy-map unpatched

genisis0x:
Closes #1581.

## Summary

`gstack-upgrade/migrations/v1.40.0.0.sh` writes `~/.gstack/.migrations/v1.40.0.0.done` unconditionally. When the host doesn't have `jq` installed, the privacy-map patch logs a single `WARN:` line and gets skipped — but the done-marker is still written, so subsequent `/gstack-upgrade` runs short-circuit at `[ -f "${DONE}" ] && exit 0` and the privacy-map repair never lands. Federation sync keeps silently dropping `/plan-eng-review` test plans.

## Fix

Track a `skipped_required` flag. Set to `1` when:

- `command -v jq` returns false (jq not installed), or
- jq is present but the patch invocation fails.

Skip `touch ${DONE}` when `skipped_required=1` so the next `/gstack-upgrade` retries against an env with jq installed. The allowlist + gitattributes patches still run on the same pass (they don't need jq), and their `grep -Fq` 'already present' gates keep re-runs idempotent.

Also replace the buried one-line WARN with an action-required block listing the install commands for macOS / Debian / Fedora and an explicit note that the done-marker was withheld on purpose.

## Tests

Two new regressions in `test/artifacts-init-migration.test.ts` (file as a whole: 14/14 green):

- `#1581: jq missing leaves done-marker unwritten and patches the rest` — clamps PATH to a fake bin dir with coreutils symlinks but no jq, asserts allowlist + gitattributes were patched, privacy-map untouched, done-marker absent, ACTION REQUIRED block present in stderr.
- `#1581: retry after jq install completes the migration` — first run with no-jq PATH (no marker), second run with real PATH lands the privacy-map entry and writes the marker.

Existing tests (allowlist patch, privacy-map patch, gitattributes patch, idempotent re-run, fresh-init no-op) unchanged.

## Test plan

- [x] `bun test test/artifacts-init-migration.test.ts` — 14/14 pass
- [x] Manual: simulated Debian-without-jq box, confirmed done-marker withheld and second run after `apt install jq` patches the privacy-map

Followed @0xfandom-flagged option 1 from the issue (single-marker, deferred). The per-prerequisite marker (option 2) is a larger refactor; happy to follow up if you'd prefer that direction.

### DIFF
--- a/gstack-upgrade/migrations/v1.40.0.0.sh
+++ b/gstack-upgrade/migrations/v1.40.0.0.sh
@@ -34,6 +34,13 @@ NEW_PATTERNS=(
 )
 
 added_any=0
+# Set to 1 if a required step had to be skipped (e.g. jq missing for the
+# privacy-map patch). We do NOT write the done-marker in that case — so the
+# next /gstack-upgrade run will retry against an environment with jq
+# installed. See #1581: previously the done-marker was written
+# unconditionally, which silently dropped the privacy-map repair on boxes
+# without jq and federation sync kept missing eng-review test plans.
+skipped_required=0
 
 # ----- .brain-allowlist ---------------------------------------------------
 if [ -f "${ALLOWLIST}" ]; then
@@ -62,12 +69,23 @@ if [ -f "${PRIVACY}" ]; then
           added_any=1
         else
           rm -f "${PRIVACY}.tmp"
+          skipped_required=1
           echo "  [v1.40.0.0] WARN: jq failed to patch ${PRIVACY}; skipping pattern ${PATTERN}." >&2
         fi
       fi
     done
   else
-    echo "  [v1.40.0.0] WARN: jq not found; skipping privacy-map repair. Install jq and re-run gstack-upgrade, or run gstack-artifacts-init manually." >&2
+    skipped_required=1
+    echo "" >&2
+    echo "  [v1.40.0.0] *** ACTION REQUIRED ***" >&2
+    echo "  [v1.40.0.0] jq not found; cannot patch .brain-privacy-map.json." >&2
+    echo "  [v1.40.0.0] Federation sync will keep dropping /plan-eng-review test plans" >&2
+    echo "  [v1.40.0.0] until this runs. Install jq and re-run /gstack-upgrade:" >&2
+    echo "  [v1.40.0.0]   - macOS:  brew install jq" >&2
+    echo "  [v1.40.0.0]   - Debian/Ubuntu: sudo apt install jq" >&2
+    echo "  [v1.40.0.0]   - Fedora: sudo dnf install jq" >&2
+    echo "  [v1.40.0.0] (Migration done-marker NOT written — next /gstack-upgrade retries.)" >&2
+    echo "" >&2
   fi
 fi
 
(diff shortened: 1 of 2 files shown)


jbetala7:
This solves the same #1581 failure already covered by older #1589. I think #1589 is the better canonical PR at the moment because it handles not only missing `jq`, but also malformed privacy-map JSON and write/mv/append failures, and its body calls out the separate remediation migration needed for users who already have the bad `v1.40.0.0.done` marker. If maintainers prefer this smaller patch, the action-required stderr copy is useful, but I would avoid landing both; they touch the same migration and test the same retry path.

genisis0x:
Acknowledged — @jbetala7 is right, #1589 is the canonical fix for #1581 since it also handles the malformed privacy-map JSON path that this PR leaves untouched. Closing to consolidate review on #1589.

PR #1604, opened 2026-05-19
fix(benchmark): parse prompt files after value flags

### LINKED ISSUE #1603: gstack-model-benchmark treats --models value as the prompt when options precede a file
## Observed problem

On current upstream `main` (`026751ea`), `gstack-model-benchmark` finds the positional prompt with:

```ts
process.argv.slice(2).find(a => !a.startsWith("--"))
```

That means values belonging to options are treated as positional prompts. A normal invocation with options before the prompt file crashes because `claude` is both the `--models` value and a directory in this repo:

```sh
bun run bin/gstack-model-benchmark --models claude README.md --dry-run
```

Current behavior:

```text
FATAL: EISDIR: illegal operation on a directory, read
```

The CLI attempts to read `claude` as the prompt file instead of reading `README.md`.

## Expected behavior

Flag values for `--models`, `--workdir`, `--timeout-ms`, `--output`, and `--prompt` should not be considered positional prompt arguments. The first standalone non-flag token after option parsing should be the prompt file or inline prompt.

Searched open issues and PRs for:

- `gstack-model-benchmark --models`
- `model benchmark positional prompt models`
- `gstack-model-benchmark EISDIR --models`
- `benchmark EISDIR`
- `bin/gstack-model-benchmark`

The only nearby open PR I found is #1495, which adds an Ollama provider and touches the benchmark CLI, but it does not fix option-value-as-positional parsing or the `EISDIR` crash.

## Candidate fix shape

Replace the first-non-dash positional heuristic with a tiny argument scanner that skips values consumed by known value flags, handles `--flag=value`, and returns only real positional arguments. Add a CLI regression test for `--models claude <prompt-file> --dry-run`.

jbetala7:
## Summary

- skip values consumed by known `gstack-model-benchmark` flags before selecting the positional prompt
- preserve `--flag=value` handling and `--` positional terminator support
- add regressions for prompt files after `--models`/`--output` in both separated and equals forms

Fixes #1603

## Testing

- `bun test test/benchmark-cli.test.ts`
- `bun run bin/gstack-model-benchmark --models claude README.md --dry-run`

### DIFF
--- a/bin/gstack-model-benchmark
+++ b/bin/gstack-model-benchmark
@@ -40,16 +40,40 @@ const ADAPTER_FACTORIES = {
 
 type OutputFormat = 'table' | 'json' | 'markdown';
 
+const CLI_ARGS = process.argv.slice(2);
+const VALUE_FLAGS = new Set(['--models', '--prompt', '--workdir', '--timeout-ms', '--output']);
+
 function arg(name: string, def?: string): string | undefined {
-  const idx = process.argv.findIndex(a => a === name || a.startsWith(name + '='));
+  const idx = CLI_ARGS.findIndex(a => a === name || a.startsWith(name + '='));
   if (idx < 0) return def;
-  const eqIdx = process.argv[idx].indexOf('=');
-  if (eqIdx >= 0) return process.argv[idx].slice(eqIdx + 1);
-  return process.argv[idx + 1];
+  const eqIdx = CLI_ARGS[idx].indexOf('=');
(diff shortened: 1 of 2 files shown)


garrytan:
Your fix landed in v1.43.2.0 (the CHANGELOG credits you by name via PR #1604). Closing this PR — the work shipped through a parallel route. Thanks for the contribution, @jbetala7!

PR #1610, opened 2026-05-19
fix: sourceLocalPath handles wrapped {sources:[...]} shape from gbrain v0.18.0+

### LINKED ISSUE #1609: gstack-gbrain-sync: `sources list --json` shape mismatch crashes `/sync-gbrain --full` on gbrain ≥ 0.18.0
## Repro

```sh
$ gstack --version  # 1.40.0.0
$ gbrain --version  # 0.33.1.0 (any ≥ 0.18.0)
$ /sync-gbrain --full
[gbrain-sync] mode=full engine=supabase
[gbrain-exec] seeded DATABASE_URL from /Users/x/.gbrain/config.json
```

## Root cause

`bin/gstack-gbrain-sync.ts:291` — `sourceLocalPath()`:

```ts
export function sourceLocalPath(sourceId: string, env?: NodeJS.ProcessEnv): string | null {
  const list = execGbrainJson<Array<{ id: string; local_path?: string }>>(
    ["sources", "list", "--json"],
    { baseEnv: env },
  );
  if (!list) return null;
  const found = list.find((s) => s.id === sourceId);  // ← crashes
}
```

The function types the response as `Array<...>` but `gbrain sources list --json` has returned `{sources: [...]}` since gbrain v0.18.0 (commit `90c5d93`, PR #337, "multi-source brains," 2026-04-22).

Reference, `gbrain/src/commands/sources.ts:202`:
```ts
console.log(JSON.stringify({ sources: entries }, null, 2));
```

## Why it didn't show up sooner

`sourceLocalPath` is only called by `runCodeImport`, which runs on `/sync-gbrain --full`. Incremental sync (the hourly LaunchAgent path via `gstack-brain-sync --once`) doesn't touch it, so users whose code-stage reindex hasn't fired never see the crash.

## Suggested patch (4 lines)

```ts
export function sourceLocalPath(sourceId: string, env?: NodeJS.ProcessEnv): string | null {
  const raw = execGbrainJson<
  if (!raw) return null;
  // gbrain v0.18.0+ returns {sources: [...]}; older versions returned a raw array.
  const list = Array.isArray(raw) ? raw : raw.sources ?? [];
}
```

Defensive against both shapes, so it works on any gbrain version.

## Suggested test

A contract test that shells out to a real `gbrain sources list --json` (against a tmp engine) would have caught this. The current test suite must be using a fixture that returns the raw-array shape, since the function would otherwise fail in unit tests too. Worth adding `test/contract/gbrain-cli.test.ts` that runs against a real gbrain binary and red-flags any new shape divergence in CI.

## Environment

- macOS 15.x, gstack v1.40.0.0
- gbrain v0.33.1.0 (also reproduces on any gbrain ≥ 0.18.0 based on shape history)
- Mode: local-stdio, engine: Supabase postgres
- Repro path: `/sync-gbrain --full` on a fully-set-up brain

Happy to send a PR if helpful.

pennyultima:
Closes #1609

## What

`gbrain sources list --json` has returned `{sources: [...]}` since gbrain v0.18.0 (commit [`90c5d93`](https://github.com/garrytan/gbrain/commit/90c5d93), "multi-source brains," 2026-04-22). `sourceLocalPath()` in `bin/gstack-gbrain-sync.ts:291` typed the response as a raw `Array<...>` and called `.find()` directly, crashing `/sync-gbrain --full`:

```
```

The bug only surfaces on full reindex (`runCodeImport` is the only caller of `sourceLocalPath`), so the cheap incremental path users hit via the hourly LaunchAgent didn't trigger it. That's why it's been latent.

## How

Defensive against both shapes — works on any gbrain version, wrapped (v0.18.0+) or raw array (older):

```ts
export function sourceLocalPath(sourceId: string, env?: NodeJS.ProcessEnv): string | null {
  const raw = execGbrainJson<
  if (!raw) return null;
  const list = Array.isArray(raw) ? raw : raw.sources ?? [];
}
```

## Why CI didn't catch it

The existing 3 unit tests for `sourceLocalPath` all use the **raw-array shape** in their fixtures:
```ts
stdout: JSON.stringify([{ id: "...", local_path: "..." }])
```
The function passes them because the mock matches the function's (wrong) type assumption. The real gbrain CLI returns the wrapped shape, and the function crashes there.

This PR updates the fixtures to use the live wrapped shape **and** keeps the raw-array tests for back-compat. Net: 5 tests cover both shapes.

```
bun test test/gstack-gbrain-sync.test.ts -t "sourceLocalPath"
 5 pass
 0 fail
```

Pre-fix code fails 2 of these (the new wrapped-shape tests). Post-fix passes all 5.

## Audit of other call sites

I grepped for every consumer of `gbrain sources list --json` across `bin/` and `lib/`:

| Call site | Status |
|---|---|
| `bin/gstack-gbrain-sync.ts:292` (this PR) | ❌ → ✅ |
| `lib/gbrain-sources.ts:54` (`probeSource`) | ✅ already unwraps `.sources` |
| `lib/gbrain-sources.ts:165` (`sourcePageCount`) | ✅ already unwraps `.sources` |
| `bin/gstack-gbrain-source-wireup:215,298` (shell) | ✅ uses `jq '.sources[]'` |
| `lib/gbrain-local-status.ts:224` | ✅ only checks exit code, doesn't parse |

So this is the **only** broken call site. The lib code does it right; the orchestrator should ideally use `lib/gbrain-sources.ts` directly (a future refactor — not in this PR to keep the diff minimal).

## Suggested follow-up: contract test

Worth adding `test/contract/gbrain-cli.test.ts` that runs against a real `gbrain` binary in CI (probably gated on a `GSTACK_CONTRACT_TEST=1` env var so unit tests stay fast). That would catch any future JSON shape drift between gstack and gbrain at PR time instead of in users' terminals. Happy to write this in a follow-up if the maintainer wants it.

## Test plan

- [x] `bun test test/gstack-gbrain-sync.test.ts -t "sourceLocalPath"` — 5/5 pass
- [x] Verified on the reporter's machine (gstack v1.40.0.0 + gbrain v0.33.1.0): patched orchestrator runs successfully where the unpatched version crashed.

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -289,11 +289,14 @@ function gbrainSupportsSourcesRename(env?: NodeJS.ProcessEnv): boolean {
  * helper can be exercised without a real gbrain CLI.
  */
 export function sourceLocalPath(sourceId: string, env?: NodeJS.ProcessEnv): string | null {
-  const list = execGbrainJson<Array<{ id: string; local_path?: string }>>(
-    ["sources", "list", "--json"],
-    { baseEnv: env },
-  );
-  if (!list) return null;
+  const raw = execGbrainJson<
+    | Array<{ id: string; local_path?: string }>
+    | { sources?: Array<{ id: string; local_path?: string }> }
+  >(["sources", "list", "--json"], { baseEnv: env });
+  if (!raw) return null;
+  // gbrain v0.18.0+ returns {sources: [...]}; older versions returned a raw array.
+  // Defensive against both shapes so this works on any gbrain version.
+  const list = Array.isArray(raw) ? raw : raw.sources ?? [];
   const found = list.find((s) => s.id === sourceId);
   return found?.local_path ?? null;
 }
--- a/test/gstack-gbrain-sync.test.ts
+++ b/test/gstack-gbrain-sync.test.ts
@@ -812,7 +812,24 @@ describe("sourceLocalPath", () => {
     rmSync(bindir, { recursive: true, force: true });
   });
 
-  it("returns local_path when the source exists", () => {
+  // gbrain v0.18.0+ returns {sources: [...]}; this is the live contract.
+  it("returns local_path when the source exists (gbrain v0.18.0+ wrapped shape)", () => {
+    makeShim(bindir, {
+      "sources list --json": {
+        stdout: JSON.stringify({
+          sources: [
+            { id: "other-source", local_path: "/x" },
+            { id: "target-id", local_path: "/repo/match" },
+          ],
+        }),
+      },
+    });
+    expect(sourceLocalPath("target-id", envWithBindir(bindir))).toBe("/repo/match");
+  });
+
+  // Pre-v0.18.0 gbrain returned a raw array. Keep back-compat so the helper
+  // works regardless of which gbrain the user has on PATH.
+  it("returns local_path when the source exists (pre-v0.18.0 raw-array shape)", () => {
     makeShim(bindir, {
       "sources list --json": {
         stdout: JSON.stringify([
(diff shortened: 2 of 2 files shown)


jbetala7:
Thanks for pinning this. This appears to be the same `sourceLocalPath` wrapped `{sources:[...]}` crash that is already carried by canonical maintainer wave #1594, which closes #1567 and says it supersedes #1564/#1571. The wave touches the same function in `bin/gstack-gbrain-sync.ts` and adds regression coverage for the wrapped shape, while also bundling the adjacent gbrain API-drift fixes.

I think #1594 is the canonical landing path now. If anything from this PR is still missing after that wave, the useful follow-up would be a targeted regression against current `main` rather than a parallel `sourceLocalPath` PR.

garrytan:
Your fix was independently shipped via another route before this PR could land — verified in current main. Closing. Thanks for the contribution, @pennyultima!

PR #1614, opened 2026-05-19
feat(browse): extend GSTACK_CHROMIUM_PATH support to headless launch

shohu:
## Summary

`launchHeaded()` already reads `GSTACK_CHROMIUM_PATH` to select a custom Chromium binary, but `launch()` (headless mode — the default `$B goto` path) ignores the env var and always falls back to Playwright's bundled binary.

This PR applies the same pattern to `launch()`:

- If `GSTACK_CHROMIUM_PATH` is set **and** the path exists (`existsSync` guard), it is passed as `executablePath` to `chromium.launch()`
- If unset or the binary is missing, falls back to standard Playwright Chromium automatically — **no breaking change**
- Logs `[browse] Using custom Chromium: <path>` on startup when active (matches the debug style used elsewhere)

## Motivation

This enables drop-in use of stealth Chromium forks (e.g. [CloakBrowser](https://github.com/CloakHQ/CloakBrowser)) for headless automation without needing to fork gstack or patch vendored files after every upgrade.

**Before:** `GSTACK_CHROMIUM_PATH` only worked for headed mode (`$B connect`). Headless `$B goto` always used Playwright's bundled binary regardless.

**After:** Setting `GSTACK_CHROMIUM_PATH` in the environment applies to both headless and headed mode consistently.

## Usage

```bash
# ~/.zshrc
export GSTACK_CHROMIUM_PATH="$(python3 -c 'from cloakbrowser import binary_info; print(binary_info()["binary_path"])')"
```

```bash
# Verify: navigator.webdriver should be false with CloakBrowser
$B goto https://example.com
$B js 'navigator.webdriver'   # → false
```

## Test plan

- [x] `bun run build` passes
- [x] `$B goto` with `GSTACK_CHROMIUM_PATH` set → `navigator.webdriver: false`, `window.chrome: object`
- [x] `$B goto` without `GSTACK_CHROMIUM_PATH` → falls back to standard Playwright Chromium (existsSync guard)
- [x] No changes to headed mode (`launchHeaded`) behaviour

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -236,6 +236,14 @@ export class BrowserManager {
       console.log(`[browse] Extensions loaded from: ${extensionsDir}`);
     }
 
+    // Support custom Chromium binary via GSTACK_CHROMIUM_PATH (same as launchHeaded).
+    // Falls back to standard Playwright Chromium when unset or path does not exist.
+    const headlessCloakPath = process.env.GSTACK_CHROMIUM_PATH;
+    const headlessCloakExec = headlessCloakPath && require('fs').existsSync(headlessCloakPath)
+      ? headlessCloakPath
+      : undefined;
+    if (headlessCloakExec) console.log(`[browse] Using custom Chromium: ${headlessCloakExec}`);
+
     this.browser = await chromium.launch({
       headless: useHeadless,
       // On Windows, Chromium's sandbox fails when the server is spawned through
@@ -244,6 +252,7 @@ export class BrowserManager {
       chromiumSandbox: process.platform !== 'win32',
       ...(launchArgs.length > 0 ? { args: launchArgs } : {}),
       ...(this.proxyConfig ? { proxy: this.proxyConfig } : {}),
+      ...(headlessCloakExec ? { executablePath: headlessCloakExec } : {}),
     });
 
     // Chromium crash → exit with clear message
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -226,12 +226,16 @@ export class BrowserManager {
     }
 
     if (extensionsDir) {
-      launchArgs.push(
-        `--disable-extensions-except=${extensionsDir}`,
-        `--load-extension=${extensionsDir}`,
-        '--window-position=-9999,-9999',
-        '--window-size=1,1',
-      );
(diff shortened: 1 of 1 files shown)


jbetala7:
One correctness edge with the headless path: this applies `GSTACK_CHROMIUM_PATH` to `launch()`, but the extension branch above still unconditionally appends `--disable-extensions-except` / `--load-extension` whenever `BROWSE_EXTENSIONS_DIR` is set. `launchHeaded()` has an `isCustomChromium()` guard for this exact case because GBrowser / GStack Browser builds already bake the extension in, and loading it twice can hit the ServiceWorkerState::SetWorkerId DCHECK crash. With this PR, a user running headless/off-screen via `BROWSE_EXTENSIONS_DIR` plus `GSTACK_CHROMIUM_PATH` to a baked-extension Chromium can still take the double-load path.

Can this mirror `launchHeaded()` by skipping the extension flags when `isCustomChromium()` is true, with a regression around `GSTACK_CHROMIUM_KIND=custom-extension-baked` or a `GBrowser` path? That would keep the new headless custom-binary support aligned with the existing headed safety path.

shohu:
Good catch — fixed in the latest commit (e2de3b30).

The `launch()` headless path now mirrors `launchHeaded()` exactly: it checks `isCustomChromium()` before pushing `--disable-extensions-except` / `--load-extension`, so a user running with `BROWSE_EXTENSIONS_DIR` + a baked-extension `GSTACK_CHROMIUM_PATH` won't hit the double-load DCHECK crash.

The off-screen window geometry args (`--window-position` / `--window-size`) are still pushed unconditionally since those are harmless with any binary.

jbetala7:
Looks like this browser-manager fix has now landed through the maintainer wave in #1642 / v1.43.2.0. Current `origin/main` has the headless `GSTACK_CHROMIUM_PATH` `executablePath` path plus the `isCustomChromium()` guard before extension flags in `browse/src/browser-manager.ts`, and the v1.43.2.0 changelog credits this PR for the headless custom Chromium support.

I would treat #1642 as the canonical landing path for this PR now; no further browser-manager action seems needed here unless you want to keep this open only for discussion/history.

shohu:
Thanks for confirming! Happy to see both the headless GSTACK_CHROMIUM_PATH support and the isCustomChromium() guard land in v1.43.2.0 via #1642. Closing this as superseded.

PR #1616, opened 2026-05-20
fix(build): isolate subshell+redirect to bash helper for Windows

agswindow:
## Problem

On Windows, `bun run build` silently breaks mid-chain. Bun Shell doesn't reliably handle `( ... ) > file` (subshell with stdout redirect) — see oven-sh/bun#11066, #11968. The four offending constructs in `package.json`'s `build`:

    ( git rev-parse HEAD 2>/dev/null || true ) > {browse,design,make-pdf}/dist/.version
    (rm -f .*.bun-build || true)

cause the `&&` chain to break. Symptoms on Windows: stale binaries with no `.version` stamp (confusing `find-browse` stale detection), leftover `.bun-build` temp files, and the `chmod` step never runs.

## Fix

Move the four constructs into a new `scripts/stamp-versions.sh` — same shell syntax, but invoked via `bash` so they never enter Bun Shell. Matches the existing pattern in `browse/scripts/build-node-server.sh`.

Also updates `test/build-script-shell-compat.test.ts` (D-1460): the assertion now accepts inline subshells OR a reference to `scripts/stamp-versions.sh`, with a file-existence guard so a rename/move can't silently break stamping. The anti-brace-group constraint is preserved.

## Verification

Tested on Windows 11 (bun 1.3.11). Before fix: `.version` files 9 days stale despite repo on `026751e` (v1.40.0.0); browse binaries last built 7 days ago. After fix: 5 binaries rebuilt + `.version` stamped to HEAD SHA + temp files cleaned, no errors. `bun test test/build-script-shell-compat.test.ts`: 2/2 pass.

## Out of scope (flagged during adversarial review)

- Windows CI still doesn't run full `bun run build` (`make-pdf-gate.yml` has Windows commented out; `windows-free-tests.yml` only runs `build-node-server.sh`). End-to-end Bun-on-Windows path remains uncovered — happy to follow up.
- `bash` dependency: same as existing `build-node-server.sh`; could be removed via TS rewrite if preferred.
- `.version` semantics unchanged: empty file on non-git install, HEAD SHA on dirty tree. Both pre-existing.
- No behavior test for `scripts/stamp-versions.sh` itself (would need tmpdir + subprocess). Static checks here pin invocation shape and helper presence, not runtime correctness — happy to add if preferred.

Side note: `.gitignore` matches `bin/gstack-global-discover` (no extension) but Windows bun produces `bin/gstack-global-discover.exe`. Happy to file separately.


jbetala7:
Looks like this overlaps the canonical Windows build fix already in maintainer wave #1594. That wave closes #1538/#1537/#1530/#1457/#1561 and replaces the package build chain with `scripts/build.sh` plus `scripts/write-version-files.sh`, and it also adds `windows-setup-e2e.yml` to exercise fresh `./setup` on Windows. This PR is solving the same user-visible Bun-shell subshell + redirect failure via `scripts/stamp-versions.sh`, so I think #1594 is now the canonical place for this fix.

If there is still a gap after #1594, the clean follow-up would be against that branch/main with the specific missing Windows behavior rather than landing a parallel build-chain shape here.

time-attack:
@16francej Please close this PR. The #1594 build pipeline already runs the Windows-sensitive stamping work through the correct Bash helper. This branch proposes only an older, narrower version of the same fix.

PR #1617, opened 2026-05-20
v1.40.0.1 fix(browse): enable Chromium sandbox on headed launchPersistentContext

garrytan:
## Summary

Playwright auto-adds `--no-sandbox` whenever `chromiumSandbox !== true`
([playwright-core/lib/server/chromium/chromium.js:291-292](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/chromium/chromium.ts)). The headless `chromium.launch()` site
set the option; the two headed sites — `launchHeaded()` and `handoff()` — did not. Every headed
launch on macOS and Linux therefore showed Chromium's yellow **"unsupported command-line flag:
--no-sandbox"** infobar across the top of the first tab.

This PR introduces `shouldEnableChromiumSandbox()` to centralize the Win32 / CI / CONTAINER /
root heuristic that previously lived only in the headless path's explicit `--no-sandbox` push
at `:225`. All three launch sites now share the same policy.

- `browse/src/browser-manager.ts:244` (`launch()`) — switches `process.platform !== 'win32'` to `shouldEnableChromiumSandbox()`
- `browse/src/browser-manager.ts:416` (`launchPersistentContext()`) — adds the helper
- `browse/src/browser-manager.ts:1304` (`handoff()`) — adds the helper

The explicit `--no-sandbox` push at `:225` stays as a harmless defensive duplicate; Playwright
will auto-add `--no-sandbox` anyway when the helper returns false, but the redundancy makes the
intent obvious to readers in container/root environments.

## Test plan

- [x] `bun test browse/test/browser-manager-unit.test.ts` — 8 tests pass, including 6 new pinning `shouldEnableChromiumSandbox` across darwin / linux / win32 / CI / CONTAINER / root
- [x] `bun test browse/test/browser-manager-custom-chromium.test.ts` — still green
- [ ] Manual headed launch on macOS: yellow `--no-sandbox` infobar gone
- [ ] Manual Linux non-root launch: no infobar
- [ ] Manual Linux Docker/root: sandbox off, browser starts, no spurious warnings


garrytan:
Bundled into #1629 as the v1.42.1.1 fix wave, along with PR #1626 and an additional exit-code propagation fix that Codex caught (PR #1626 alone didn't actually fix the headed-mode respawn bug because `browse/src/server.ts` hardcoded `activeShutdown?.(2)` — the bundled PR wires `onDisconnect(exitCode)` end-to-end through `server.ts:688`).

Leaving this PR open until #1629 merges so we have a working fallback if the bundle is rejected or surfaces an integration bug. Will close with a final "merged via #1629" once landed.

garrytan:
Superseded by v1.42.2.0 (#1629). Closing per the #1642 follow-up note.

PR #1621, opened 2026-05-20
fix(slug): avoid parent repo identity in subdirs

### LINKED ISSUE #1125: gstack-slug inherits outer-repo slug when project is a subdirectory; cache is unrecoverable

jbetala7:
Fixes #1125

## Summary
- ignore parent repository remotes when `gstack-slug` is run from a subdirectory that is not the git root
- let a `.gstack-slug` file in the current directory or an ancestor pin the project slug before cache/git inference
- treat slug cache as a fallback for transient inference failures, not as an immutable source of truth when current inference or an override is available

## Root cause
`gstack-slug` trusted `git remote get-url origin` before checking whether the current directory was actually the repository root. In a standalone project nested under a larger git repo, Git walked up to the parent repo, so the child project inherited the parent repo slug. Because cache lookup happened first, a bad slug-cache entry could keep that mistake alive indefinitely.

## Verification
- `bun test test/skill-validation.test.ts -t gstack-slug`
- `bun test test/review-log.test.ts test/gstack-learnings-search.test.ts`
- `bun test test/skill-validation.test.ts`
- direct repro: nested directory under a parent repo with remote now returns the child basename instead of `me-workspace`

## Collision check
Searched open PRs/issues for `1125 gstack-slug outer repo subdirectory`, `gstack-slug subdirectory outer repo`, `bin/gstack-slug subdirectory slug-cache`, `slug-cache`, and `.gstack-slug`.

Closest open PR is #564, but that is an older missing-git-context fallback patch. Current `origin/main` already has basename/unknown fallback behavior; #564 does not address parent-repo inheritance, `.gstack-slug` overrides, or stale cache recovery.

### DIFF
--- a/bin/gstack-slug
+++ b/bin/gstack-slug
@@ -8,38 +8,71 @@
 set -euo pipefail
 
 CACHE_DIR="$HOME/.gstack/slug-cache"
-PROJECT_DIR="$(pwd)"
+PROJECT_DIR="$(pwd -P 2>/dev/null || pwd)"
 # Encode absolute path as cache key: /Users/j/foo → _Users_j_foo
 CACHE_KEY=$(printf '%s' "$PROJECT_DIR" | tr '/' '_')
 CACHE_FILE="${CACHE_DIR}/${CACHE_KEY}"
 
-# 1. Try cached slug first (guarantees consistency across sessions)
-if [[ -f "$CACHE_FILE" ]]; then
-  SLUG=$(cat "$CACHE_FILE")
+sanitize_slug() {
+  printf '%s' "$1" | tr -cd 'a-zA-Z0-9._-'
+}
+
+find_slug_override() {
+  local dir="$PROJECT_DIR"
+  while [[ -n "$dir" && "$dir" != "/" ]]; do
+    if [[ -f "$dir/.gstack-slug" ]]; then
+      head -n 1 "$dir/.gstack-slug" 2>/dev/null | tr -d '\r\n' || true
+      return 0
+    fi
+    dir="$(dirname "$dir")"
+  done
+  return 1
+}
+
+# 1. Explicit project overrides beat cache and git inference. This lets users
+#    recover from stale slug-cache entries without editing cache internals.
+OVERRIDE_SLUG=$(find_slug_override 2>/dev/null || true)
+if [[ -n "$OVERRIDE_SLUG" ]]; then
+  SLUG=$(sanitize_slug "$OVERRIDE_SLUG")
 fi
 
-# 2. If no cache, compute from git remote (separated from pipeline to avoid
-#    pipefail swallowing the error and producing an empty slug)
+# 2. If the current directory is the git root, compute from that repo's remote.
+#    If it is only a subdirectory of a parent repo, do not inherit the parent
+#    repo's identity; use the directory basename instead.
 if [[ -z "${SLUG:-}" ]]; then
-  REMOTE_URL=$(git remote get-url origin 2>/dev/null) || REMOTE_URL=""
-  if [[ -n "$REMOTE_URL" ]]; then
-    RAW_SLUG=$(printf '%s' "$REMOTE_URL" | sed 's|.*[:/]\([^/]*/[^/]*\)\.git$|\1|;s|.*[:/]\([^/]*/[^/]*\)$|\1|' | tr '/' '-')
-    SLUG=$(printf '%s' "$RAW_SLUG" | tr -cd 'a-zA-Z0-9._-')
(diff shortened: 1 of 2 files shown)


jbetala7:
Rebased onto current `main` (post #1848) and resolved the `bin/gstack-slug` conflict. Now `MERGEABLE` again (head `133c6cbd`).

This fixes a distinct bug from #1848: #1848 added unconditional cache-value sanitization, while this PR stops a nested standalone project from inheriting its **parent** repo's identity (it checks `git rev-parse --show-toplevel == $PWD` before trusting `origin`, and treats the slug cache as a fallback rather than an immutable source). The conflict resolution **keeps** #1848's hardening: every path (override, remote, basename, and the cache read) runs `sanitize_slug`, plus a final unconditional sanitize before the value is emitted or cached, so the poisoned-cache shell-injection defense is preserved. #1848's `test/gstack-slug-sanitize.test.ts` passes against the rebased branch.

Verification:
- `bun test test/gstack-slug-sanitize.test.ts test/user-slug-fallback.test.ts` (slug-sanitize green; the one `user-slug-fallback` failure is pre-existing on clean `main` c43c850c, unrelated to this file)
- `bun test test/skill-validation.test.ts` -> 331 pass
- Functional: from a nested non-root subdir of a repo, SLUG is now the directory basename, not the parent repo slug.

time-attack:
@16francej — recommend closing this PR. This is superseded by #1125, which owns the slug policy and a cache-stable solution; the useful guards here would reverse current cache stability. Please retain the guards, override, sanitization, tests, and attribution under #1125.

PR #1622, opened 2026-05-20
fix(gbrain-sync): sourceLocalPath handles wrapped {sources:[...]} shape from gbrain v0.30+

Evode-Manirahari:
## Problem

After upgrading to **gstack v1.40.0.0** + **gbrain v0.30.0.0** or newer (current released gbrain is v0.37.1.0), the first `/sync-gbrain` after upgrade dies before any stage runs:

```
$ /sync-gbrain
[gbrain-sync] mode=incremental engine=pglite
```

## Root cause

`gbrain sources list --json` started wrapping its output as `{"sources":[...]}` in gbrain v0.30+. The v1.40 hostname-fold migration helper `sourceLocalPath()` (added in #1547, `bin/gstack-gbrain-sync.ts:291`) typed the response as a bare array and called `.find()` on it directly:

```ts
// before
const list = execGbrainJson<Array<{ id: string; local_path?: string }>>(...);
if (!list) return null;
const found = list.find((s) => s.id === sourceId);  // crash: list is an object, not an array
```

The sibling helper in `lib/gbrain-sources.ts` (`statusForSourceId`, lines 72-85) already unwraps both shapes correctly — `sourceLocalPath` was the lone holdout because it was added separately in the v1.40 migration work.

Actual output on a current gbrain:

```
$ gbrain sources list --json
{
  "sources": [
    { "id": "default", "name": "default", ... },
    { "id": "gstack-code-myrepo-b6de5213", "local_path": "/Users/me/myrepo", ... }
  ]
}
```

## Fix

Accept both wire shapes (defense against the same regression in either direction):

```ts
const raw = execGbrainJson<
if (!raw) return null;
const list = Array.isArray(raw) ? raw : raw.sources || [];
```

## Tests

Three new cases in the existing `sourceLocalPath` describe block in `test/gstack-gbrain-sync.test.ts`:

1. `returns local_path when gbrain wraps the list as {sources: [...]} (v0.30+ shape)` — the regression.
2. `returns null when the wrapped {sources: [...]} shape has no matching id`.
3. `treats {sources: undefined} the same as an empty list` — defensive case for partial responses.

The existing 3 bare-array tests still pass, locking in backward compat with pre-v0.30 gbrain shims.

**Run:** `bun test test/gstack-gbrain-sync.test.ts` → 38 pass, 0 fail (130 expect() calls).

## Repro on a fresh machine

```bash
# Upgrade gstack to v1.40 (in CHANGELOG, v1.40.0.0 — the hostname-fold migration release)
# Have a working gbrain >= v0.30
cd <any-git-repo>
/sync-gbrain
# → "gstack-gbrain-sync fatal: list.find is not a function"
```

## Found via

Hit this while syncing a freshly-shipped repo against gbrain v0.37.1.0 from a fresh gstack v1.40.0.0 install. Patched locally first; this PR brings the fix upstream.

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -289,11 +289,17 @@ function gbrainSupportsSourcesRename(env?: NodeJS.ProcessEnv): boolean {
  * helper can be exercised without a real gbrain CLI.
  */
 export function sourceLocalPath(sourceId: string, env?: NodeJS.ProcessEnv): string | null {
-  const list = execGbrainJson<Array<{ id: string; local_path?: string }>>(
-    ["sources", "list", "--json"],
-    { baseEnv: env },
-  );
-  if (!list) return null;
+  // gbrain v0.30+ returns {"sources":[...]}; older releases returned a bare
+  // array. Handle both shapes so this works across the compat window. The
+  // sibling helper in lib/gbrain-sources.ts (statusForSourceId) already does
+  // this; the v1.40 hostname-fold migration was the lone holdout and crashed
+  // with `list.find is not a function` on every gbrain v0.30+ install.
+  const raw = execGbrainJson<
+    | Array<{ id: string; local_path?: string }>
+    | { sources?: Array<{ id: string; local_path?: string }> }
+  >(["sources", "list", "--json"], { baseEnv: env });
+  if (!raw) return null;
+  const list = Array.isArray(raw) ? raw : raw.sources || [];
   const found = list.find((s) => s.id === sourceId);
   return found?.local_path ?? null;
 }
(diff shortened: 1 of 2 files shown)


jbetala7:
This one is already fixed on current `origin/main` via the merged maintainer wave #1594. The live `origin/main:bin/gstack-gbrain-sync.ts` implementation now reads `sources list --json` as `unknown` and normalizes with `Array.isArray(raw) ? raw : raw.sources ?? []` before calling `.find()`.

The collision map also has multiple open parallel PRs for the same `sourceLocalPath` wrapped-shape bug (#1564, #1571, #1610), while #1594 is the merged canonical landing path and explicitly lists `fix(gbrain-sync) sourceLocalPath wrapped shape (#1571)`. I would close this as superseded; if anything is still missing, the useful follow-up is a tiny regression against current `main`, not another copy of the same patch on the old base.

garrytan:
Duplicate of #1610 (same `sourceLocalPath` wrapped-shape fix in `bin/gstack-gbrain-sync.ts`). #1610 is older and slightly tighter (8 LOC vs 11). Going to land that one — closing this with thanks for the independent diagnosis, Evode. Both reports + reviews helped confirm the fix is right.

PR #1627, opened 2026-05-20
feat(next-version): support monorepo VERSION paths via --version-path + .gstack/version-path

cfeddersen:
## Summary

`bin/gstack-next-version` hardcodes the VERSION file at `<repo-root>/VERSION`. In monorepos where versioning is subproject-scoped — one shippable app inside a larger repo — every PR's VERSION lookup 404s, the workspace-aware ship queue silently empties, and parallel `/ship` sessions all bump from "current main + 1". The result is a cascade of slot collisions.

The fix adds a `--version-path` CLI flag and a repo-local `.gstack/version-path` config file. Resolution priority: CLI flag > config file > `"VERSION"` (existing default, backward-compatible). The resolved path threads through every call site that touches the file. The previous misleading warning `"could not fetch VERSION (fork or private)"` becomes honest and actionable.

## Repro

A private monorepo I work in. Root VERSION is absent; the real VERSION lives at `Apps With Spaces/web-app/VERSION` (path intentionally has a space to exercise URL-encoding). In one day, four sequential collisions:

```
0.4.0.1 → 0.5.0.0 → 0.5.0.1 → 0.5.0.2 → 0.5.0.3
```

Every parallel `/ship` saw an empty `claimed` array because `gh api repos/{owner}/{repo}/contents/VERSION?ref=<head>` returned 404 for every open PR. The util emitted `"PR #X: could not fetch VERSION (fork or private)"` for each — a misleading warning, since the real cause was wrong path.

Confirmed against an open PR of that repo:

```
$ gh api 'repos/{owner}/{repo}/contents/VERSION?ref=...' → 404
$ gh api 'repos/{owner}/{repo}/contents/Apps%20With%20Spaces/web-app/VERSION?ref=...' → 0.5.1.0
```

The file is fully accessible — the util just queries the wrong path.

## Root cause

Three call sites all hardcode `"VERSION"`:

1. `readBaseVersion` — `git show origin/${base}:VERSION`
2. `fetchGithubClaimed` — `gh api repos/{owner}/{repo}/contents/VERSION?ref=...`
3. `fetchGitlabClaimed` — `glab api projects/:id/repository/files/VERSION?ref=...`
4. `scanSiblings` — `join(p, "VERSION")` for local sibling-worktree scans

No flag, env, or config knob exists to override.

## Fix

New helper `resolveVersionPath(override, repoRoot)` with priority:

1. `--version-path <path>` CLI flag (highest)
2. `.gstack/version-path` file at repo root — a single line, the relative path from repo root. Committed to the repo so all collaborators benefit.
3. `"VERSION"` default (current behavior — fully backward-compatible)

Path threads through all four call sites. GitHub's Contents API gets the path via `encodeURI` so spaces in subproject names (e.g. `Apps With Spaces/...`) become `%20` while `/` segments stay intact. GitLab's files API gets `encodeURIComponent` (the full path becomes a single URL-encoded segment with `%2F` separators), which matches the existing pattern for `ref=`.

The resolved path surfaces in the JSON output as `version_path` so `/ship` and operators can see what got picked. The previous warning is replaced with:

```
PR #X: could not fetch <path> (fork, private, or wrong path — try --version-path or .gstack/version-path)
```

## End-to-end verification

Patched binary, run against the monorepo with `.gstack/version-path` containing `Apps With Spaces/web-app/VERSION`:

```json
{
  "version": "0.6.0.1",
  "version_path": "Apps With Spaces/web-app/VERSION",
  "claimed": [
    { "pr": 274, "version": "0.6.0.0", ... },
    { "pr": 272, "version": "0.5.1.1", ... }
  ],
  "reason": "bumped past claimed 0.6.0.0",
  ...
}
```

Before the fix: `"claimed": []` and 3 misleading "could not fetch VERSION (fork or private)" warnings. After: queue is fully visible, slot allocation is correct.

## Tests added

`test/gstack-next-version.test.ts` gains 7 new tests:

**6 unit tests** for `resolveVersionPath` (new pure-function export):
- CLI flag wins over `.gstack/version-path`
- `.gstack/version-path` config is picked up when no flag
- Trims whitespace and ignores lines after the first
- Empty config file falls back to default `"VERSION"`
- Missing config file falls back to default `"VERSION"`
- Empty override string (`""`) falls back to config/default (defensive)

**1 integration smoke** that drives `--version-path` end-to-end and asserts it surfaces in the JSON output as `version_path`.

The existing integration test gains an assertion that `version_path` defaults to `"VERSION"` when no flag/config present.

## Test plan

- [x] `bun test test/gstack-next-version.test.ts` — 28/28 pass (was 21 pre-change)
- [x] Bun bundles cleanly (no TS errors): `bun --bun build bin/gstack-next-version`
- [x] End-to-end against real monorepo: queue correctly picks up subproject VERSION at all 4 call sites
- [x] Backward-compatible: no flag, no config, no change in behavior

## Out of scope (surfaced while investigating)

- **CHANGELOG-path is the same problem class.** This binary doesn't touch CHANGELOG (only `/ship` does), so it's a separate fix in a separate file. Happy to follow up if useful.
- **`gstack-pr-title-rewrite.sh`** also references VERSION but takes the version string as an argument — not a file-path bug. No fix needed.

### DIFF
--- a/bin/gstack-next-version
+++ b/bin/gstack-next-version
@@ -10,7 +10,14 @@
 //
 // Usage:
 //   gstack-next-version --base <branch> --bump <major|minor|patch|micro> \
-//     --current-version <X.Y.Z.W> [--workspace-root <path>|null] [--json]
+//     --current-version <X.Y.Z.W> [--workspace-root <path>|null] \
+//     [--version-path <path>] [--json]
+//
+// VERSION path resolution (monorepo support):
+//   1. --version-path <path> CLI flag (highest priority)
+//   2. .gstack/version-path file at the repo root (single-line relative path,
+//      committed so all collaborators benefit)
+//   3. "VERSION" at the repo root (default, backward-compatible)
 //
 // Exit codes:
 //   0 — emitted JSON successfully (may include "offline":true or "host":"unknown")
@@ -45,6 +52,7 @@ type Output = {
   version: string;
   current_version: string;
   base_version: string;
+  version_path: string;
   bump: Bump;
   host: "github" | "gitlab" | "unknown";
   offline: boolean;
@@ -114,6 +122,28 @@ function runCommand(cmd: string, args: string[], timeoutMs = 15000): { ok: boole
   };
 }
 
+// VERSION-path resolution for monorepos. Priority: CLI flag > .gstack/version-path
+// at repo root > "VERSION". Pure function; takes the repo root as an argument so
+// tests can drive it with a fixture dir without mocking git.
+function resolveVersionPath(override: string | undefined, repoRoot: string): string {
+  if (override) return override.trim();
+  const configFile = join(repoRoot, ".gstack", "version-path");
+  if (existsSync(configFile)) {
+    try {
+      const firstLine = readFileSync(configFile, "utf8").split("\n")[0]?.trim() ?? "";
+      if (firstLine) return firstLine;
+    } catch {
+      // fall through to default
+    }
+  }
+  return "VERSION";
+}
+
+function repoToplevel(): string {
+  const r = runCommand("git", ["rev-parse", "--show-toplevel"]);
+  return r.ok ? r.stdout.trim() : process.cwd();
+}
+
 function detectHost(): "github" | "gitlab" | "unknown" {
   const remote = runCommand("git", ["remote", "get-url", "origin"]);
   if (remote.ok) {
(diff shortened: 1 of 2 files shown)


garrytan:
Landed in 64f9aafa1efd36b179c130c1ed8219403e5afda1 as part of v1.44.1.0 wave (PR #1682). Thanks for the fix, @cfeddersen.

PR #1608, opened 2026-05-19
fix(config): expose explain_level default

jbetala7:
## Summary

- adds `explain_level` to the canonical `gstack-config` default table
- includes `explain_level` in `gstack-config list` and `gstack-config defaults`
- adds a regression covering `get`, `defaults`, and `list` from a clean `GSTACK_HOME`

Fixes #1607

## Testing

- `bun test test/docs-config-keys.test.ts`
- direct CLI probe: `gstack-config get explain_level`, `gstack-config defaults`, and `gstack-config list` with a temporary `GSTACK_HOME`
- `git diff --check origin/main...HEAD`
- `git merge-tree --write-tree HEAD origin/main`

PR #1632, opened 2026-05-21
fix(config): Include explain level default

### LINKED ISSUE #1607: gstack-config omits explain_level from defaults and active values

Gujiassh:
## Summary

- Add `explain_level` to the canonical `gstack-config` defaults table
- Include `explain_level` in `list` and `defaults` output
- Update focused config tests so unset `explain_level` reports the documented default

Fixes #1607.

## Verification

- `bun test test/explain-level-config.test.ts`
- `git diff --check`

### DIFF
--- a/bin/gstack-config
+++ b/bin/gstack-config
@@ -100,6 +100,7 @@ lookup_default() {
     skill_prefix) echo "false" ;;
     checkpoint_mode) echo "explicit" ;;
     checkpoint_push) echo "false" ;;
+    explain_level) echo "default" ;;
     codex_reviews) echo "enabled" ;;
     gstack_contributor) echo "false" ;;
     skip_eng_review) echo "false" ;;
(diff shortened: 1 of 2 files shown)


jbetala7:
Collision note: #1608 is already open for the same #1607 fix. It touches the same `bin/gstack-config` default/list/defaults behavior, links the same issue, and has focused coverage for the documented default path.

I would treat #1608 as the canonical/older PR rather than landing a parallel `explain_level` fix. If there is a useful assertion in this branch that #1608 lacks, the smallest path is to fold that assertion into the canonical branch instead of keeping two PRs for the same user-visible bug.

PR #1633, opened 2026-05-21
Honor GSTACK_CHROMIUM_PATH on the headless launch and setup paths

ryanlaiwy:
## Problem

`GSTACK_CHROMIUM_PATH` lets you point gstack at a custom Chromium binary. The headed path (`launchPersistentContext`) already honors it, but two paths ignored it:

- `BrowserManager`'s headless `chromium.launch()` — a custom binary was used in headed mode only.
- `setup`'s `ensure_playwright_browser` self-test — always launched Playwright's bundled Chromium.

On NixOS the bundled `chrome-headless-shell` can't run (it expects `/lib64/ld-linux-x86-64.so.2`, absent on a Nix filesystem), so `./setup` bails before registering sub-skills and headless `browse` fails — even with `GSTACK_CHROMIUM_PATH` set to a working system Chromium.

## Change

- Pass `executablePath` on the headless `chromium.launch()` when `GSTACK_CHROMIUM_PATH` is set.
- `ensure_playwright_browser` verifies that same binary (with `--no-sandbox`) when the env var is set.

Both are no-ops when `GSTACK_CHROMIUM_PATH` is unset — zero impact on normal installs.

## Testing

Verified on a NixOS-style container (Replit): with `GSTACK_CHROMIUM_PATH` pointed at the system Chromium, `./setup` completes and registers all sub-skills, and headless `browse` navigates and screenshots correctly. Without the env var, behavior is unchanged.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -312,6 +312,12 @@ export class BrowserManager {
 
     this.browser = await chromium.launch({
       headless: useHeadless,
+      // Honor GSTACK_CHROMIUM_PATH on the headless launch path. The headed
+      // path (launchPersistentContext, below) already respects it; this keeps
+      // both modes consistent for custom Chromium builds — e.g. a NixOS system
+      // binary, where Playwright's bundled chrome-headless-shell can't run.
+      // No-op when the env var is unset.
+      ...(process.env.GSTACK_CHROMIUM_PATH ? { executablePath: process.env.GSTACK_CHROMIUM_PATH } : {}),
       // On Windows, Chromium's sandbox fails when the server is spawned through
       // the Bun→Node process chain (GitHub #276). Disable it — local daemon
       // browsing user-specified URLs has marginal sandbox benefit. Also disabled
--- a/setup
+++ b/setup
@@ -256,7 +256,13 @@ ensure_playwright_browser() {
   else
     (
       cd "$SOURCE_GSTACK_DIR"
-      bun --eval 'import { chromium } from "playwright"; const browser = await chromium.launch(); await browser.close();'
+      if [ -n "$GSTACK_CHROMIUM_PATH" ]; then
+        # GSTACK_CHROMIUM_PATH set: verify that custom Chromium instead (e.g. a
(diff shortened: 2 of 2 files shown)


jbetala7:
The setup self-test half looks useful, but the `browser-manager.ts` half needs to pick up the same baked-extension guard that #1614 added after review.

Right now this adds `executablePath: process.env.GSTACK_CHROMIUM_PATH` to the headless/off-screen launch path while leaving the `if (extensionsDir)` block unconditional. With `BROWSE_EXTENSIONS_DIR` plus a GBrowser/GStack Browser binary that already has the extension baked in, that path still passes `--disable-extensions-except` / `--load-extension` and can double-load the extension. `launchHeaded()` already avoids that via `isCustomChromium()`, and #1614 now mirrors that guard for the headless path.

Can this either rebase/port #1614's `isCustomChromium()` guard and add a regression for `GSTACK_CHROMIUM_KIND=custom-extension-baked` or a `GBrowser` path, or split this down to just the `setup` verification change if #1614 remains the canonical browser-manager PR?

jbetala7:
After #1642 landed, this PR now has a clean split point:

- The `browse/src/browser-manager.ts` half is already on `main` via #1642 / v1.43.2.0, including the `isCustomChromium()` guard before extension flags. Keeping that part here would re-open the duplicate/double-load review path from the earlier comment.
- The `setup` half is still not on `main`: `ensure_playwright_browser()` still launches bundled Playwright Chromium directly and does not pass `GSTACK_CHROMIUM_PATH` into the self-test.

Smallest path forward seems to be rebasing this onto current `main` and narrowing it to the setup verification behavior only, with a test/proof for the NixOS/system-Chromium case. That avoids competing with the browser-manager fix that already shipped while preserving the remaining useful install fix.

time-attack:
@16francej — recommend closing this PR only after a focused setup follow-up uses production-equivalent arguments. #1968 owns the validated runtime behavior, while this PR’s setup probe does not match it. Preserve the setup attribution and require runtime/setup parity in that follow-up.

PR #1635, opened 2026-05-21
fix(timeline): pass read filters as data

### LINKED ISSUE #1634: gstack-timeline-read drops branch filters containing apostrophes
## Observed problem

`gstack-timeline-read --branch <name>` embeds the branch argument directly into the inline Bun program:

```js
const branch = '${BRANCH}';
```

A branch name containing a single quote makes the generated JavaScript invalid. The wrapper redirects Bun stderr to `/dev/null` and exits 0, so the caller receives empty output even though matching timeline entries exist.

Verified on current `origin/main` (`029356e1`):

1. Log a timeline event with branch `feature/o'hare`.
2. `gstack-timeline-read` shows the event.
3. `gstack-timeline-read --branch "feature/o'hare"` exits 0 with no output.

## Expected behavior

The branch filter should treat branch names as data and return matching entries, including names with quotes or other shell/JavaScript metacharacters. `--since` and `--limit` should also be passed into Bun as data rather than interpolated into source code.

No existing issue or PR found for this behavior. Searches checked:

- `gstack-timeline-read branch quote`
- `timeline read branch apostrophe`
- `gstack-timeline-read timeline branch quote injection`
- `timeline` open PRs, including nearby session/telemetry PRs #1049, #1050, #1188, and #1199

## Candidate fix shape

Pass the parsed CLI filter values to the Bun subprocess through environment variables, parse and validate `--limit` inside the Bun code, and add a regression test that reads a timeline entry for a branch containing a single quote.

jbetala7:
Fixes #1634

## What changed
- Pass `gstack-timeline-read` filter values through environment variables instead of interpolating shell arguments into the inline Bun program.
- Parse the limit from env with a positive-integer fallback.
- Add a regression test for branch names containing single quotes.

## Why
`gstack-timeline-read --branch "feature/o'hare"` could break the generated JavaScript source, and stderr was suppressed, so the command returned empty output even when matching events existed.

## Testing
- `bun test --timeout 20000 test/timeline.test.ts`
- Manual repro: recorded an event on `feature/o'hare` and confirmed `gstack-timeline-read --branch "feature/o'hare"` returns it.

### DIFF
--- a/bin/gstack-timeline-read
+++ b/bin/gstack-timeline-read
@@ -29,11 +29,13 @@ if [ ! -f "$TIMELINE_FILE" ]; then
   exit 0
 fi
 
-cat "$TIMELINE_FILE" 2>/dev/null | bun -e "
+cat "$TIMELINE_FILE" 2>/dev/null | GSTACK_TIMELINE_SINCE="$SINCE" GSTACK_TIMELINE_BRANCH="$BRANCH" GSTACK_TIMELINE_LIMIT="$LIMIT" bun -e "
 const lines = (await Bun.stdin.text()).trim().split('\n').filter(Boolean);
-const since = '${SINCE}';
-const branch = '${BRANCH}';
-const limit = ${LIMIT};
+const since = process.env.GSTACK_TIMELINE_SINCE || '';
+const branch = process.env.GSTACK_TIMELINE_BRANCH || '';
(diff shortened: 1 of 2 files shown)


garrytan:
Landed in 64f9aafa1efd36b179c130c1ed8219403e5afda1 as part of v1.44.1.0 wave (PR #1682). Thanks for the fix, @jbetala7.

PR #1636, opened 2026-05-21
fix(retro): warn on stale date windows

### LINKED ISSUE #1624: /retro silently produces empty/misleading output when 'today' anchor is wrong (no stale-base warning)
## Failure mode

When `/retro` runs on a worktree-style setup (Conductor, etc.) and the agent's understanding of "today" is off by enough days, the skill silently produces a clean-looking retro that misses recent reality. The `git log --since="<midnight-aligned-start>"` window quietly returns zero or near-zero commits, but nothing in the output flags that the window is wrong.

Both inputs to the window can be wrong:
1. **"today" anchor drift** — the agent computes `today` from session-context cues that don't match the actual system date (model error, session-reminder mismatch, etc.).
2. **Stale worktree base** — the Conductor/`claude/...` branch is days/weeks behind `origin/<default>`, and although the skill does `git fetch origin <default> --quiet`, nothing surfaces that the latest commit is materially older than "today" would suggest.

Either way, the retro renders successfully, the metrics table looks legitimate, and the user gets a confidently-wrong report.

## Reproducer

I hit this on a real project today:
- `today` per system date: 2026-05-20
- Agent computed `today` (incorrectly): 2026-05-15
- `/retro` (default 7d): analyzed `--since="2026-05-08T00:00:00"`, returned 18 commits (all from May 8-11), missed 4 commits on `origin/main` from May 19-20 including the latest version bump
- Output: a clean retro narrative ending "Streak: 0d (4-day burst May 8-11)" — looks coherent, completely missed v0.21.0 → v0.22.0 work that had already shipped

The user caught it only when version-bumping for the next PR: "version number is way old."

## Why it's silent

`Step 1: Gather Raw Data` runs `git fetch origin <default> --quiet` and then immediately starts gathering commits via `git log origin/<default> --since="<window>"`. There's no sanity-check on the gap between "today" and `origin/<default>`'s latest commit. If the gap exceeds the window length, the window is empty — but the retro keeps going and the model fills in narrative from whatever (possibly zero) commits it found.

## Suggested fix (sketch)

Inject a stale-base sanity check right after the fetch, before the data-gathering. Always echo the latest commit date + gap so the agent has the data point; escalate to an explicit warning when the gap is suspect:

```bash
git fetch origin <default> --quiet

_LATEST_COMMIT_TS=$(git log origin/<default> -1 --format="%at" 2>/dev/null || echo 0)
_LATEST_COMMIT_DATE=$(git log origin/<default> -1 --format="%ad" --date=short 2>/dev/null || echo unknown)
_GAP_DAYS=$(( ($(date +%s) - _LATEST_COMMIT_TS) / 86400 ))
echo "ORIGIN_<default>_LATEST: $_LATEST_COMMIT_DATE (${_GAP_DAYS}d before today)"
if [ "$_GAP_DAYS" -gt 3 ]; then
  echo "⚠ STALE-BASE WARNING: latest commit is ${_GAP_DAYS} days before today."
  echo "  If your window is shorter than ${_GAP_DAYS}d, the retro will contain few or zero commits."
  echo "  Sanity-check the 'today' anchor against the actual system date before proceeding —"
  echo "  a clean-looking retro on the wrong window silently misses recent reality."
fi
```

Plus a prose follow-up after the bash block: *"If the stale-base warning fires: confirm that `today` matches the system date, and that `origin/<default>` actually has the recent activity you expect. If the gap exceeds the window length, the retro will report empty or near-empty results — that's evidence the anchor or fetch is wrong, not that the team didn't ship."*

Always echoing the data point (even when no warning fires) gives the agent a fact to cross-check against. The 3-day threshold is comfortably above typical weekly cadence (so weekend gaps don't false-positive) but low enough to catch the 9-day case I hit. The threshold is debatable — could be window-aware (e.g., `gap > window_days`) if the skill exposes the parsed window length as a bash var.

## Why I'm filing an issue instead of a PR

I have a working local patch (tested against the repo where the bug surfaced) but want to confirm with you whether:
- The 3-day threshold is right, or whether it should be window-aware
- The warning text matches gstack's voice/conventions
- This belongs in `Step 1: Gather Raw Data` or as its own pre-flight step before Step 1
- The fix should also apply to `/retro compare` and `/retro global` (the global mode has its own multi-repo discovery — the same failure mode could occur there)

Happy to send a PR shaped however you'd prefer. Context: hit while running `/retro` from a Claude Code (Conductor) worktree branch that was 4 commits behind `origin/main`.

---

If helpful: the lesson also landed in my local project memory as "high-confidence 'obviously fine' moments are exactly when the second pair of eyes catches the thing that isn't" — same pattern across the stale-base miss, a self-judged bug diagnosis, and a `/ship` declaring-done gate where I skipped the advisor. The `/retro` skill is the most fixable of these because the check is mechanical.

jbetala7:
Fixes #1624

## Summary
- Move the /retro date/window guard into a dedicated pre-flight before normal data gathering.
- Use a window-aware stale-base threshold: warn when the latest default-branch commit gap is greater than the current window length (`gap > window_days`), not a fixed 3 days.
- Print the system date, current window length, latest origin default-branch commit date/gap, and commit count before narrative generation.
- Apply the same pre-flight to `/retro compare` and to each discovered repo in `/retro global`; global mode surfaces affected repos before aggregate totals.
- Warn for mismatched today anchors and empty windows so /retro does not produce a clean-looking empty report from bad date/window evidence.

## Tests
- `bun test test/gen-skill-docs.test.ts -t 'retro SKILL.md warns on stale base or empty windows'`
- `bun run scripts/gen-skill-docs.ts --dry-run --host claude`
- `git diff --check`
- Temporary 9-day-gap/7d proof: latest `origin/main` commit dated 2026-05-12, system date 2026-05-21, `RETRO_WINDOW_DAYS: 7`, `RETRO_WINDOW_COMMIT_COUNT: 0`, emitted both `STALE-BASE WARNING` and `EMPTY-WINDOW WARNING`.

## Collision checks
- Open PR search for `1624 retro stale window` returned only this PR.
- Open PR search for `STALE-BASE WARNING` returned only this PR.
- Open PR searches for `RETRO_WINDOW_DAYS` and `retro global empty window` found no competing PR.
- Fresh `origin/main` merge-tree completed cleanly.

### DIFF
--- a/retro/SKILL.md
+++ b/retro/SKILL.md
@@ -888,11 +888,17 @@ Check for non-git context that should be included in the retro:
 
 If `RETRO_CONTEXT_FOUND`: read `~/.gstack/retro-context.md`. This file is user-authored and may contain meeting notes, calendar events, decisions, and other context that doesn't appear in git history. Incorporate this context into the retro narrative where relevant.
 
-### Step 0.5: Stale-base + bad-today-anchor pre-flight guard
+### Date/window pre-flight
 
 The retro skill computes a window from "today" and queries `git log --since=<window> origin/<default>`. If "today" drifts (model session-context error) or the local worktree's `origin/<default>` is materially behind the actual remote, the window can return zero or near-zero commits and the retro will fabricate a coherent-looking narrative from nothing. This guard prevents silent confidently-wrong output.
 
-Run the pre-flight in this exact order. The first branch that matches wins:
+Run this dedicated pre-flight after resolving the requested window and before the main data-gathering commands. For `/retro compare`, run it against the current window before computing the prior window. For `/retro global`, run it once per discovered repo after detecting that repo's default branch and before collecting its git-log data.
+
+Resolve:
+- `<window>` to the exact `--since` value used for the current window, such as `2026-05-13T00:00:00`.
+- `<window_days>` to the numeric current-window length in days. Use `7` for the default window, `N` for `Nd`, `N * 7` for `Nw`, and `ceil(N / 24)` with a minimum of `1` for `Nh`.
+
+Run the pre-flight in this exact order. The first skip branch that matches proceeds with a disclosure; otherwise the freshness metrics must be printed:
 
 ```bash
 # Pre-check A: no remote configured?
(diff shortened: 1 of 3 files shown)


poppinpixels:
Issue author. Four questions from #1624 I don’t see resolved in this PR:

- Threshold: fixed 3d, or window-aware (`gap > window_days`)?
- Scope: does this extend to `/retro compare` and `/retro global`? Global mode’s multi-repo discovery has the same failure mode.
- Placement: Step 1 inline, or dedicated pre-flight?
- Reproducer: tested against the 9-day-gap case in #1624, or a synthetic empty-window?

xg-gh-25:
Smart catch on the stale-window issue — this is exactly the kind of silent failure that undermines trust in retro tools.

**Why this matters:**
When `/retro` runs with an outdated anchor date, it produces *plausible-looking but meaningless* output. Users assume the tool is working, waste time analyzing fake trends, and only realize the issue days later when retrospectives don't align with reality.

**Implementation suggestion:**
Your warning approach is solid. I'd recommend:
1. **Threshold:** Flag if `today - anchor_date > 48 hours` (aggressive but prevents most drift)
2. **Visibility:** Surface the warning *above* the output, not just in logs — make it impossible to miss
3. **Auto-correction option:** Offer a quick action: "Anchor is stale. Re-run with today's date?" (reduces friction)

The tricky part is handling legitimate use cases where someone *intentionally* runs historical retros (e.g., auditing last month). Maybe:
- Default: auto-warn
- Escape hatch: `--allow-stale-anchor` flag for intentional historical runs

Have you considered caching the last anchor date per workspace to detect drift automatically?

---
*Posted by [SwarmAI](https://github.com/yourusername/swarmai) — an agentic community engagement system*

jbetala7:
Addressed in `79e4724d`.

- Threshold: changed from fixed `> 3d` to window-aware `gap > window_days`. The pre-flight now resolves default `7d`, `Nd`, `Nw * 7`, and `ceil(Nh / 24)` with a minimum of 1 day, then emits `RETRO_WINDOW_DAYS`.
- Scope: yes, this now explicitly covers normal `/retro`, `/retro compare` current-window metrics, and `/retro global` per discovered repo. Global mode runs the same pre-flight after repo/default-branch detection and lists affected repos before aggregate totals.
- Placement: moved to a dedicated `### Date/window pre-flight` section before Step 1, rather than leaving it inline inside raw data gathering.
- Reproducer: tested the 9-day-gap/7d case from #1624 with a temporary git repo, not only a synthetic empty-window. The proof had system date `2026-05-21`, latest `origin/main` commit `2026-05-12`, `RETRO_WINDOW_DAYS: 7`, `RETRO_WINDOW_COMMIT_COUNT: 0`, and emitted both `STALE-BASE WARNING` and `EMPTY-WINDOW WARNING`.

Verification also passed: `bun run gen:skill-docs --host all`, focused generated-doc regression, Claude dry-run freshness, `git diff --check`, and merge-tree against fresh `origin/main`. I updated the PR body with the exact proof and collision searches.

poppinpixels:
Good turnaround. Brief read on each:

1. Window-aware `gap > window_days` matches what my local patch does. Better than fixed 3d.
2. Scope extension to compare and global per discovered repo closes the gap I flagged.
3. Pre-flight section reads cleaner than the inline option.
4. The 9-day case with system date 2026-05-21, commit 2026-05-12, both warnings emitted is exactly the reproducer.

Will verify against my local repo and reproducer.

The three failing E2E Eval checks (e2e-review, llm-judge, report) — related to this change, or pre-existing flakes?

jbetala7:
Not related to this change from what I can see.

- `llm-judge`: the job starts with `Secret source: None`, then the Anthropic SDK fails before the judged eval can run because no API auth is available. That is an external PR/secrets issue, not branch logic.
- `e2e-review`: the failing case is `/retro default branch detection`, but the eval comparison reports `FAIL -> FAIL`, `Status: 3 unchanged`, and `No regressions`. So it is the existing baseline failure, not a new regression from this branch.
- `report`: it downloads the eval artifacts and builds the PR comment body, then fails only while posting the report comment with `GraphQL: Resource not accessible by integration (addComment)`. That is the external PR token permission issue.

The branch-specific checks remain green: `actionlint`, `build-image`, `check-freshness`, `windows-free-tests`, and the rest of the E2E suites all passed. The local verification for this change also passed as noted above.

time-attack:
@16francej Please close this PR. The retro skill's stale-date warning and tests already shipped in #1642. This later contributor branch is now redundant, although its contribution should remain credited.

Issue #1628, opened 2026-05-20
Codex CLI in gstack skills 400s on ChatGPT-account auth — model entitlement issue, needs `-m gpt-5.2` injection


TJ-NomoAI:
## Problem

codex CLI v0.121.0 defaults to `gpt-5.2-codex`. ChatGPT-account (Plus) auth no longer serves any `*-codex` model variant since ~March 2026 (OpenAI server-side entitlement regression, not a billing cap — caps are 429/402, this is 400 `invalid_request_error`). Rejected slugs: `gpt-5.2-codex`, `gpt-5-codex`, `gpt-5.1-codex`, all `*-codex-mini`, `codex-mini-latest`, AND plain `gpt-5`. Working slug on ChatGPT auth: `gpt-5.2` (general model, same generation, no `-codex` tuning).

Refs: openai/codex#19654, openai/codex#14735, openai/codex#14412.

## Impact on gstack

Every gstack skill that shells codex constructs the command **without `-m`**, so every codex invocation 400s on ChatGPT-account users. Affected skills (non-exhaustive):

- `/plan-eng-review` — Outside Voice step (`codex exec`)
- `/codex` — review / challenge / consult modes (`codex review`, `codex exec`)
- `/ship` — adversarial review step
- `/autoplan` — pipeline of the above
- `/plan-ceo-review` — Outside Voice step
- `/plan-design-review` — Outside Voice step

API-key users (`OPENAI_API_KEY` set, no ChatGPT auth) are unaffected. Blocked for every ChatGPT-account codex invocation until users manually add `-m` per command or globally configure their `~/.codex/config.toml`.

## Failure shape

```
ERROR: {"type":"error","status":400,"error":{"type":"invalid_request_error","message":"The 'gpt-5.2-codex' model is not supported when using Codex with a ChatGPT account."}}
```

(swap the model slug per CLI version default)

## User-side workaround (per-command)

Inject `-m gpt-5.2` into every codex invocation. Cost-isolation preserved: `-m` only changes the slug; auth still resolves from `~/.codex/auth.json` (ChatGPT/Plus flat-rate). Does NOT switch to API-key auth.

```bash
# codex exec
codex exec "<prompt>" -C "$REPO_ROOT" -s read-only -m gpt-5.2 -c 'model_reasoning_effort="high"' < /dev/null

# codex review (NOTE: no -m/--model flag on this subcommand in v0.121.0 — model goes via -c)
[...]

Issue #1611, opened 2026-05-19
Memory-ingest has hardcoded 35-min timeout that SIGTERMs `/sync-gbrain --full` on big brains; no resume from import-checkpoint


pennyultima:
## Repro

On a brain with ~2000+ staged files (`~/.gstack-brain-worktree/` + transcripts), `/sync-gbrain --full` reliably fails the memory stage:

```
gstack-gbrain-sync (full):
  OK    code         registered + synced gstack-code-… (18.1s)
  ERR   memory       staged 1989 pages → gbrain import (exit 143) after 2100.2s
  OK    brain-sync   curated artifacts pushed (2.4s)
```

`exit 143` = signal 15 (SIGTERM). Runtime 2100.2s = **exactly 35 minutes**.

## Root cause

`bin/gstack-gbrain-sync.ts:756` hardcodes:

```ts
const result = spawnSync("bun", ingestArgs, {
  encoding: "utf-8",
  timeout: 35 * 60 * 1000,   // ← hardcoded; no escape hatch for big brains
  env: buildGbrainEnv({ announce: false }),
});
```

(Same value also at line 602 for the code stage.)

## What happens after the timeout

gbrain leaves its checkpoint on disk:

```
$ cat ~/.gbrain/import-checkpoint.json
{
  "dir": "/Users/x/.gstack/.staging-ingest-42625-1779217370565",
  "totalFiles": 1989,
  "processedIndex": 1000,        ← made it 50% through
  "completedFiles": 1000,
  "timestamp": "2026-05-19T19:30:05.008Z"
}
```

But the next `/sync-gbrain` doesn't resume from this checkpoint — the memory-ingest child cleans up the staging dir on SIGTERM, so the checkpoint references a dir that no longer exists. The user re-runs and the next bulk pass also gets killed at 35 min on a fresh set of files. Progress is real (1000 pages got embedded into the `default` source) but the verdict reads as `ERR`.

### 1. Make the timeout configurable

```ts
const memoryTimeoutMs = parseInt(
  process.env.GSTACK_SYNC_MEMORY_TIMEOUT_MS ?? `${35 * 60 * 1000}`,
  10,
);
const codeTimeoutMs = parseInt(
  process.env.GSTACK_SYNC_CODE_TIMEOUT_MS ?? `${35 * 60 * 1000}`,
  10,
);
```

[...]

Issue #1602, opened 2026-05-19
Windows: `bun run build` aborts — package.json build script uses redirection constructs bunsh rejects; regression test only checks statically


osouthgate:
## Summary

On Windows, `./setup` runs `bun run build` (`setup:281`), which makes **bun's built-in shell (bunsh)** interpret the `build` script in `package.json:12`. That script contains three:

```
( git rev-parse HEAD 2>/dev/null || true ) > <pkg>/dist/.version
```

bunsh does not support **redirection on a grouping construct**, so the build aborts with:

```
Subshells with redirections are currently not supported. Please open a GitHub issue.
```

Result on Windows: `bun run build` never completes, the compiled binaries (`browse/dist/browse`, `bin/gstack-global-discover`, etc.) stay stale across every `/gstack-upgrade`, and the only workaround is to build manually under Git Bash and re-run `./setup` with `NEEDS_BUILD=0`.

This survived the v1.38.0.0 "fix" (brace-group → subshell) and the v1.39.1.0 follow-up because **the regression test guards the wrong axis and never executes the script** (details below).

## Reproduction

```js
// repro.mjs
import { $ } from "bun";
console.log("bun", Bun.version, process.platform);
try { await $`( echo H 2>/dev/null || true ) > t.txt`.quiet(); console.log("OK"); }
catch (e) { console.log("FAIL ->", String(e.message||e).split("\n")[0]); }
```

```
$ bun repro.mjs
bun 1.3.14 win32
FAIL -> Subshells with redirections are currently not supported. Please open a GitHub issue.
```

Confirmed at **bun 1.3.4 and 1.3.14** on **Windows 11 (win32)**. `bun run build` uses bunsh on Windows (no system `sh`), so `package.json:12` hits this directly.

## Root cause

The current form fails for **two independent reasons**, each backed by a tight bunsh error signature (bun 1.3.14 / win32):

1. **Redirection on a subshell is unsupported.** `( … ) > f` → bunsh's own message: `Subshells with redirections are currently not supported. Please open a GitHub issue.`
2. **More than one redirection on a simple command is unsupported.** `cmd > f 2>/dev/null` (either order) → `expected a command or assignment but got: "Redirect"`. So even unwrapping the subshell to `git rev-parse HEAD > f 2>/dev/null || true` still fails.

The current `( git rev-parse HEAD 2>/dev/null || true ) > X/.version` triggers **both** at once.

Full matrix at bun 1.3.14 / win32:

[...]

PR #1637, opened 2026-05-21
Resolved issues rasied #1611 , #1628 , #1602

cropsgg:
here are the solutions that I did :

1. `/sync-gbrain --full` on ~100k-page brains reliably exceeded the
hard-coded 35-minute timeout, SIGTERMed mid-import, and lost the staging
checkpoint. Set GSTACK_SYNC_MEMORY_TIMEOUT_MS (or _CODE_ for the code
stage) to override; bad input falls back to the 35-min default with a
stderr warning so a typo can't silently disable the safety net.

2. Codex CLI defaults to gpt-5.2-codex, which OpenAI's ChatGPT-account
entitlement filter rejects with a 400 ("model is not supported when
using Codex with a ChatGPT account"). Every skill that shells out to
codex — /codex, /autoplan, /plan-eng-review, /ship, /plan-ceo-review,
/plan-design-review — broke for ChatGPT-only auth users.

Add _gstack_codex_account_kind (apikey | chatgpt | none) and
_gstack_codex_default_model_args to bin/gstack-codex-probe. Templates
expand $_CODEX_MODEL_ARGS=$(_gstack_codex_default_model_args) before
every codex invocation; API-key users get nothing extra, ChatGPT users
get `-m gpt-5.2`. $GSTACK_CODEX_MODEL overrides; set to "default" to
opt out of injection entirely.

Wired at all 9 callsites across codex/SKILL.md.tmpl (5) and
autoplan/SKILL.md.tmpl (4). A static template guard test catches a
future edit that drops the variable.

3. Issue https://github.com/garrytan/gstack/issues/1602's package.json fix landed in https://github.com/garrytan/gstack/pull/1594, but the regression test
was string-matching only. Exercise scripts/build.sh and
scripts/write-version-files.sh via `bash -n` (syntax check) and a real
invocation against tmpdir targets; also expand the package.json guard
to cover every bunsh-incompatible construct (subshells-with-redirection,
multi-redirection, brace groups, process substitution) in one place.
Catches a future edit that re-introduces any of them without needing a
Windows runner.

if useful kindly merge

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -1118,7 +1118,8 @@ Override: every AskUserQuestion → auto-decide using the 6 principles.
   **Codex CEO voice** (via Bash):
   ```bash
   _REPO_ROOT=$(git rev-parse --show-toplevel) || { echo "ERROR: not in a git repo" >&2; exit 1; }
-  _gstack_codex_timeout_wrapper 600 codex exec "IMPORTANT: Do NOT read or execute any SKILL.md files or files in skill definition directories (paths containing skills/gstack). These are AI assistant skill definitions meant for a different system. Stay focused on repository code only.
+  _CODEX_MODEL_ARGS=$(_gstack_codex_default_model_args)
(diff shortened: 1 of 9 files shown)


time-attack:
@16francej — recommend closing this PR only after the unrelated fixes are split or recorded and the routing findings are transferred to #1628. This PR combines three separate concerns and includes invalid model syntax; #1628 covers only model routing. Preserve attribution separately for each concern, not the stale model/auth heuristic.

PR #1652, opened 2026-05-22
fix(browse): scope factory shutdown teardown to caller config

spacegeologist:
## Reviewer Guide

The core fix is intentionally narrow:

- `browse/src/server.ts` — `buildFetchHandler(...).shutdown()` now tears down paths from the caller-provided `cfg.config` and `cfg.chromiumProfile`, not the module-level default config captured at import time.
- `browse/test/server-embedder-terminal-port.test.ts` — regression coverage now uses per-test temp state dirs and asserts sibling daemon files are not touched.
- `CHANGELOG.md`, `VERSION`, `package.json`, and `TODOS.md` are release/ship metadata for the resolved TODO item.

Maintainer note: the behavior fix is fully contained in `browse/src/server.ts` plus the focused test file. If release metadata is better batched into the next maintainer fix wave, the metadata-only files can be dropped without changing the code fix.

## Why

The v1.42 `ownsTerminalAgent` work made the ownership boundary explicit, but `shutdown()` still resolved some cleanup targets from module-level `config`. That is invisible for the normal gstack CLI path because the CLI and module config point to the same state/profile locations.

Embedders and tests can pass a divergent `cfg.config` or `cfg.chromiumProfile`. In that case shutdown could remove `terminal-port`, `terminal-internal-token`, the state file, or Chromium singleton locks from the wrong location. This PR makes factory teardown respect the config the factory was built with.

## Changes

- Use `cfg.config.stateFile` for terminal discovery-file cleanup and state-file cleanup.
- Use `resolveChromiumProfile(cfg.chromiumProfile)` for Chromium singleton-lock cleanup.
- Move embedder shutdown tests away from the real resolved state dir and into per-test temp dirs.
- Add regression tests for caller-passed state dirs and caller-passed Chromium profiles.
- Remove the resolved `TODOS.md` item and update version/changelog metadata.

## Risk / Scope

- Low risk for the normal CLI path: the CLI still passes the same config/profile the old module-level defaults resolved to.
- The changed behavior only matters for embedders or test harnesses that intentionally pass divergent config/profile paths.
- The test change is also a safety improvement: the suite no longer backs up/restores a live daemon's terminal files.

## Verification

- `bun test browse/test/server-embedder-terminal-port.test.ts` — 6 pass, 0 fail.
- `bun run gen:skill-docs --dry-run`.
- `package.json` version matches `VERSION`.
- `git diff --check upstream/main...HEAD`.

Note: an expanded local run, `bun test browse/test/server-embedder-terminal-port.test.ts browse/test/config.test.ts`, reached the new focused tests successfully, then hit 3 existing `config.test.ts` `isServerHealthy` failures from `port 0` server startup in this environment. Those failures are unrelated to this PR.


garrytan:
Your fix was independently shipped via another route before this PR could land — verified in current main. Closing. Thanks for the contribution, @spacegeologist!

spacegeologist:
Thanks Garry! Glad the fix made it into main. Happy to pick up another small browse/gstack issue if there is one you think would be useful.

PR #1655, opened 2026-05-22
feat: /cycle and /journal skills for multi-day cycle tracking

eduardofarina:
## Summary

Adds two new skills and three supporting bin helpers, giving gstack a notion of a multi-day development cycle bound to an issue or theme, plus curated long-term cycle memory the next session can search. Closes a gap today: gstack has no concept of a cycle that spans more than one session, and `/learn` is built for auto-logged operational failures rather than curated cycle outcomes.

`/cycle` is a state tracker (not an autonomous loop). It holds active issue, branch, story list, and event log across sessions, and hands off advisorily to existing gstack skills (`/autoplan`, `/review`, `/qa`, `/ship`). On ship or abandon, it writes a journal entry through `/journal`. No existing skill is modified; `/cycle` only references them in prose.

`/journal` is a memory primitive separate from `/learn`. `/learn` auto-logs operational failures (JSONL, dedup-by-latest). `/journal` stores curated cycle outcomes (markdown, append-only, archived when over 50 entries) that get surfaced on `/cycle start` when the new title looks similar to past work.

### Skills

- **`/journal`** ... show, search, add, close, archive, stats, edit.
- **`/cycle`** ... start, status, plan/review/qa/ship handoffs, abandon, pause, resume, story tracking, event log.

### Bin helpers

- `bin/gstack-journal-log` ... JSON to markdown append, prompt-injection guarded, mirrors `gstack-learnings-log` conventions.
- `bin/gstack-journal-search` ... search by `--query`, `--status`, `--limit`, `--headlines-only`.
- `bin/gstack-cycle-state` ... `is-active`, `get`, `show`, `start`, `set`, `event`, `story-add`, `story-mark`, `close`, `list-closed`. Atomic writes via tempfile and rename.

### Tests

- `test/cycle-and-journal.test.ts` ... Tier 1 (free, under 2 seconds), 30+ cases, includes static caller-contract checks that the SKILL.md.tmpl files reference the bin scripts verbatim.

### State layout

Under `~/.gstack/projects/$SLUG/`:

```
journal.md                    append-only markdown
journal.archive.md            created on demand by /journal archive
cycles/active.json            current cycle state (one per project)
cycles/closed/<id>.json       closed cycles, written on /cycle close
```

### Why two skills, not one

`/cycle` and `/journal` operate at different layers. `/journal` is a memory primitive that works standalone (manually log a cycle that closed before `/cycle` existed; other skills can read the journal too). `/cycle` is the orchestration layer that uses `/journal` as its persistence target on close. This mirrors how `/learn` is independent of `/review` and `/ship` even though those skills write to it.

### Why state tracker, not autonomous loop

An autonomous orchestrator that dispatches a fresh subagent per story and iterates until a completion token shows up would conflict with gstack's existing convention: the human plus main Claude session implements, and skills are advisory specialists.

`/cycle` is a state tracker that delegates advisorily. When the user runs `/cycle plan`, the cycle logs `plan_started` and tells the user to run `/autoplan`. When `/autoplan` finishes, the user runs `/cycle event '{"event":"plan_approved",...}'` to record the outcome. Same for `/review`, `/qa`, `/ship`. No new agent runtime, no new loop primitive, no conflict with `/ship` or `/autoplan`.

### Why GitHub only for v1

Non-GitHub issue systems would require their own client (Linear GraphQL, JIRA REST, etc.) and belong in a separate contribution so reviewers can audit each integration on its own. `/cycle` v1 stores the issue identifier as a string (GitHub `#N`, freeform theme, or any other identifier). A future issue-system skill can plug in transparently because the field is already a string.

### Why markdown for the journal, JSON for cycle state

Journal entries are read back into agent context on `/cycle start` ("did we hit this before?"). Humans also read them, so markdown wins on readability. Search is grep-able.

Cycle state is read programmatically by `/cycle status`, `/cycle event`, `/cycle ship`. JSON wins on structured access. Single file makes atomic writes trivial.

### Why journal-write-first on `/cycle ship`

In `/cycle ship`, the journal entry is written first, then the cycle state is closed. If a crash happens between the two steps, the journal entry exists and the cycle stays active. The user can re-run `/cycle ship`, detect the partial state, and complete the remaining step. The reverse ordering (close then journal) would leave the cycle closed with no journal record, which is unrecoverable without hand-editing.

The bin helper does not currently dedupe journal entries by `cycle_id` (it is append-only). If the user double-ships, they get two journal entries for the same cycle. Acceptable for v1; could add a `--skip-if-exists` flag later.

### Why advisory delegation rather than `Skill` tool

No other gstack skill declares `Skill` in `allowed-tools`. The convention I observed is "tell the user to run the next skill, then come back and log the result." `/cycle` follows that convention rather than introducing direct Skill invocation. If the upstream convention shifts, this can change in a later PR.

## Out of scope (deliberately)

- No autonomous loop or `<promise>COMPLETE</promise>` primitive. `/cycle` hands off advisorily to existing skills rather than driving them in a fresh `Task()` per story.
- No Linear or other non-GitHub issue system. `/cycle` stores the issue identifier as a string, so a future skill can plug in.
- No edits to ETHOS.md, README.md, CHANGELOG.md, or any existing SKILL.md.tmpl. CHANGELOG entry is left to `/ship` at land time.
- No `/cycle implement` subcommand. Implementation is the user's job (with main Claude). The cycle skill does not drive a story loop. If the user wants per-story tracking, they use `/cycle story add` and `/cycle story done` manually.

## Test plan

- [ ] `bun test test/cycle-and-journal.test.ts` passes (Tier 1, free, under 2 seconds)
- [ ] `bun run gen:skill-docs --host all` produces no diff between templates and generated `.md` files
- [ ] `./setup` discovers `journal/` and `cycle/` as top-level skills (real directories with `SKILL.md` symlinks)
- [ ] Smoke test: `/cycle start <theme>`, `/cycle status`, `/cycle event ...`, `/cycle abandon` all work in a real project
- [ ] Smoke test: `/journal`, `/journal search foo`, `/journal stats` work after at least one closed cycle exists
- [ ] No edits to `ETHOS.md`, `README.md`, `CHANGELOG.md`, or any existing `SKILL.md` or `SKILL.md.tmpl`
- [ ] All three new bin scripts have mode `100755` in `git ls-files --stage`

### DIFF
--- a/bin/gstack-cycle-state
+++ b/bin/gstack-cycle-state
@@ -1,0 +1,272 @@
+#!/usr/bin/env bash
+# gstack-cycle-state — read/write the active cycle state for /cycle
+# Usage:
+#   gstack-cycle-state is-active                 → exit 0 if a cycle is active, 1 otherwise
+#   gstack-cycle-state get [field]               → print active.json (or a top-level field)
+#   gstack-cycle-state show                      → human-friendly summary
+#   gstack-cycle-state start <json>              → create a new active cycle (fails if one exists)
+#   gstack-cycle-state set <field> <value>       → patch a top-level scalar field
+#   gstack-cycle-state event <json>              → append an event to events[]
+#   gstack-cycle-state story-add <json>          → append a story to stories[]
+#   gstack-cycle-state story-mark <id> <status>  → set a story's status (pending|in_progress|done|blocked)
+#   gstack-cycle-state close <final_status>      → move active.json to closed/<id>.json
+#   gstack-cycle-state list-closed [--limit N]   → list recent closed cycles
+#
+# Storage layout:
+#   ~/.gstack/projects/$SLUG/cycles/active.json
+#   ~/.gstack/projects/$SLUG/cycles/closed/<cycle-id>.json
+#
+# The state file is a single JSON object; this helper updates it atomically
+# via a tempfile + rename. Concurrent writes are not safe — assume a single
+# /cycle session at a time per project.
+set -euo pipefail
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+eval "$("$SCRIPT_DIR/gstack-slug" 2>/dev/null)"
+GSTACK_HOME="${GSTACK_HOME:-$HOME/.gstack}"
+CYCLES_DIR="$GSTACK_HOME/projects/$SLUG/cycles"
+CLOSED_DIR="$CYCLES_DIR/closed"
+ACTIVE="$CYCLES_DIR/active.json"
+mkdir -p "$CYCLES_DIR" "$CLOSED_DIR"
+
+_atomic_write() {
+  local target="$1" content="$2"
+  local tmp
+  tmp=$(mktemp "${target}.XXXXXX" 2>/dev/null) || tmp="${target}.tmp.$$"
+  printf '%s' "$content" > "$tmp"
+  mv "$tmp" "$target"
+}
+
+_require_active() {
+  if [ ! -f "$ACTIVE" ]; then
+    echo "gstack-cycle-state: no active cycle (run /cycle start first)" >&2
+    exit 1
+  fi
+}
+
+CMD="${1:-}"
+
+case "$CMD" in
+  is-active)
+    [ -f "$ACTIVE" ] && exit 0 || exit 1
+    ;;
+
+  get)
+    _require_active
+    FIELD="${2:-}"
+    if [ -z "$FIELD" ]; then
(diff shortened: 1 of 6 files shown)


garrytan:
Closing — `/cycle` and `/journal` introduce two new skills for multi-day cycle tracking, which isn't on the gstack roadmap and wasn't preceded by an issue or design discussion. We're tight on scope expansion right now. If there's a specific workflow you're trying to support, open an issue first so we can scope what (if anything) fits the project. Thanks for the thought.

PR #1657, opened 2026-05-22
fix(skills): avoid quoted tilde helper paths

### LINKED ISSUE #1656: Tilde-in-quoted-string disables brain-sync silently across 42 SKILL.md files (regression of #785)
# Tilde-in-quoted-string disables brain-sync silently across 42 SKILL.md files (regression of #785)

## Summary

The artifacts-sync preamble and epilogue that 42 `~/.claude/skills/gstack/*/SKILL.md` files include both quote a literal `~` — which bash never expands. The probe runs at every skill start and the `--discover-new` / `--once` calls run at every skill end, but all of them fail silently because every command is suffixed with `2>/dev/null || true`. The user-visible effect: cross-machine handoff (e.g. `/context-save` on host A → `/context-restore` on host B) silently never pushes.

This is the same bug class as already-closed #785 ("Use $HOME instead of ~ in SKILL.md browse binary assignment") — different files, same defect, never audited across the rest of the skill set.

## Affected files

```
$ grep -l '"~/.claude/skills/gstack/bin/' ~/.claude/skills/gstack/*/SKILL.md | wc -l
42
```

Includes `context-save/SKILL.md`, `context-restore/SKILL.md`, `autoplan/SKILL.md`, `benchmark/SKILL.md`, `benchmark-models/SKILL.md`, `browse/SKILL.md`, `canary/SKILL.md`, `codex/SKILL.md`, `connect-chrome/SKILL.md`, … (all 42 share the same template-generated block).

## Concrete defect

`context-save/SKILL.md`, lines 369–370:

```bash
```

Tilde inside double quotes is **literal** — bash only expands `~` when it is unquoted at the start of a word, or after `:` / `=` in PATH-like assignments. These two variables hold the literal 8-byte string starting `~/...`.

Three downstream failures cascade from this:

1. **Line 400** — config probe:
   ```bash
   ```
   The shell tries to exec the literal path `~/.claude/...`, gets ENOENT, the `2>/dev/null || echo off` swallows it, so `_BRAIN_SYNC_MODE=off` **regardless of the user's actual config**. Every downstream brain-sync branch in the preamble is dead.

2. **Line 436** — preamble push:
   ```bash
   "$_BRAIN_SYNC_BIN" --once 2>/dev/null || true
   ```
   Skipped because of (1), but also broken on its own merits.

3. **Lines 479–480** — end-of-skill snippet:
   ```bash
   "~/.claude/skills/gstack/bin/gstack-brain-sync" --discover-new 2>/dev/null || true
   ```
   Same defect — tilde in quoted string. Both no-op silently.

Net effect: brain-sync **never runs from any skill**, on any host. The only thing that triggers it in practice is the user manually invoking `gstack-brain-sync --once` or an external scheduler (cron, etc.).

## Repro

The cleanest repro is the silent cross-machine failure users actually hit:

1. Host A: `/context-save`. Checkpoint markdown is written under `~/.gstack/projects/<slug>/checkpoints/<timestamp>.md`, but the post-save `brain-sync --once` does not run.
2. Host A: `cat ~/.gstack/.brain-last-push` — unchanged.
3. Host B: `( cd ~/.gstack && git pull )` then `/context-restore` — does not see the save from host A.

Forcing it manually proves it works:

```bash
"$HOME/.claude/skills/gstack/bin/gstack-brain-sync" --discover-new
"$HOME/.claude/skills/gstack/bin/gstack-brain-sync" --once
```

— commits and pushes the queued checkpoint immediately.

Standalone bash repro (no gstack required):

```bash
$ X="~/.claude/skills/gstack/bin/gstack-brain-sync"
$ echo "$X"
~/.claude/skills/gstack/bin/gstack-brain-sync          # literal tilde
$ "$X" --status
bash: ~/.claude/skills/gstack/bin/gstack-brain-sync: No such file or directory
```

## Workaround

Wrap brain-sync invocations in a cron entry that uses `$HOME` explicitly:

```cron
*/15 * * * * $HOME/.gstack/cron-sync.sh
```

Where `cron-sync.sh` is:

```bash
#!/usr/bin/env bash
set -uo pipefail
GSTACK_HOME="$HOME/.gstack"
BRAIN_SYNC="$HOME/.claude/skills/gstack/bin/gstack-brain-sync"
( cd "$GSTACK_HOME" && git pull --ff-only --quiet origin main ) || true
"$BRAIN_SYNC" --discover-new
"$BRAIN_SYNC" --once
"$BRAIN_SYNC" --status
```

This restores cross-machine handoff but adds up to 15-minute latency, and it requires every user on every host to know about and install it.

## Fix

Single template change, regenerate all 42 SKILL.md files. The pattern is `s/"~/"$HOME/` in both places:

```diff
-_BRAIN_SYNC_BIN="~/.claude/skills/gstack/bin/gstack-brain-sync"
-_BRAIN_CONFIG_BIN="~/.claude/skills/gstack/bin/gstack-config"
+_BRAIN_SYNC_BIN="$HOME/.claude/skills/gstack/bin/gstack-brain-sync"
+_BRAIN_CONFIG_BIN="$HOME/.claude/skills/gstack/bin/gstack-config"
```

```diff
-"~/.claude/skills/gstack/bin/gstack-brain-sync" --discover-new 2>/dev/null || true
-"~/.claude/skills/gstack/bin/gstack-brain-sync" --once 2>/dev/null || true
+"$HOME/.claude/skills/gstack/bin/gstack-brain-sync" --discover-new 2>/dev/null || true
+"$HOME/.claude/skills/gstack/bin/gstack-brain-sync" --once 2>/dev/null || true
```

Optional but recommended (separate change): drop `2>/dev/null` from at least the config probe so future "exec failed" errors surface instead of silently flipping `artifacts_sync_mode` to `off`. The current swallow + `|| echo off` masks any future regression of this class.

A grep-based regression test would also catch recurrence:

```bash
! grep -RE '"~/' ~/.claude/skills/gstack/*/SKILL.md
```

## Refs

- #785 — same bug class, closed (different file, evidently not audited across remaining skills)
- #1452 — brain-allowlist coverage gap on the same cross-machine handoff path

## Environment

- gstack v1.40.x
- Reproduced on macOS 14 (Darwin 25.5.0) and Debian 12 — bash semantics, not OS-dependent

spacegeologist:
Fixes #1656.

## Reviewer Guide

This PR looks large because the generated `SKILL.md` files were regenerated. The actual behavior change is small:

- `scripts/resolvers/types.ts` — Claude/non-env-var host paths now render as `$HOME/...` instead of `~/...`.
- `scripts/resolvers/preamble/generate-completion-status.ts` — completion/timeline helper calls now use `ctx.paths.binDir` instead of hardcoded `~/.claude/...` paths.
- `test/gen-skill-docs.test.ts` and `test/host-config.test.ts` — regression coverage for quoted literal-tilde helper paths and the new `$HOME` path contract.
- Generated `SKILL.md` files and ship golden fixtures are regeneration-only fallout from those resolver changes.

## Why

Bash does not expand `~` inside double quotes. Generated Claude skill docs had helper calls like:

```bash
"~/.claude/skills/gstack/bin/gstack-brain-sync" --once
```

Those resolve to literal `~/...` paths. Because the generated blocks often use `2>/dev/null || true`, the failure is silent and brain-sync/timeline/telemetry helpers can appear configured while never executing.

## Changes

- Emit `$HOME/.claude/skills/gstack/...` for Claude host helper paths.
- Reuse host-aware `ctx.paths.binDir` in the completion status resolver.
- Regenerate all generated skill docs and ship golden fixtures.
- Add a grep-style regression test that fails if generated Claude skill docs contain quoted literal helper paths under `"~/.claude/skills/gstack/bin/`.

## Risk / Scope

- Low runtime risk: `$HOME` is expanded by the shell in quoted strings; external hosts continue to use their existing `$GSTACK_ROOT`-based paths.
- No behavior change is intended beyond making existing helper invocations resolve to the intended binaries.
- Out of scope: changing the existing `2>/dev/null || true` suppression behavior, or repairing the main-branch `VERSION` / `package.json` mismatch noted below.

## Merge Coordination

This PR and #1658 both regenerate `*/SKILL.md` from independent resolver fixes. There is no behavior dependency. If #1658 lands first, this branch only needs a rebase/regenerate so the generated docs include both the `$HOME` path fix and the `SESSIONS:` preamble output.

## Verification

- `bun test test/gen-skill-docs.test.ts -t "generated helper command paths do not quote literal tilde"`
- `bun run gen:skill-docs --host codex --dry-run`
- `git diff --check`

Note: full `bun test test/gen-skill-docs.test.ts` still hits a pre-existing main-branch mismatch where `VERSION` is `1.43.3.0` but `package.json` is `1.43.2.0`; this PR leaves that release/version issue out of scope.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -24,35 +24,35 @@ triggers:
 ## Preamble (run first)
 
 ```bash
-_UPD=$(~/.claude/skills/gstack/bin/gstack-update-check 2>/dev/null || .claude/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
+_UPD=$($HOME/.claude/skills/gstack/bin/gstack-update-check 2>/dev/null || .claude/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
 [ -n "$_UPD" ] && echo "$_UPD" || true
 mkdir -p ~/.gstack/sessions
 touch ~/.gstack/sessions/"$PPID"
 _SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
 find ~/.gstack/sessions -mmin +120 -type f -exec rm {} + 2>/dev/null || true
-_PROACTIVE=$(~/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
+_PROACTIVE=$($HOME/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
 _PROACTIVE_PROMPTED=$([ -f ~/.gstack/.proactive-prompted ] && echo "yes" || echo "no")
 _BRANCH=$(git branch --show-current 2>/dev/null || echo "unknown")
 echo "BRANCH: $_BRANCH"
-_SKILL_PREFIX=$(~/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/dev/null || echo "false")
+_SKILL_PREFIX=$($HOME/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/dev/null || echo "false")
 echo "PROACTIVE: $_PROACTIVE"
 echo "PROACTIVE_PROMPTED: $_PROACTIVE_PROMPTED"
 echo "SKILL_PREFIX: $_SKILL_PREFIX"
-source <(~/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
+source <($HOME/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
 REPO_MODE=${REPO_MODE:-unknown}
 echo "REPO_MODE: $REPO_MODE"
 _LAKE_SEEN=$([ -f ~/.gstack/.completeness-intro-seen ] && echo "yes" || echo "no")
 echo "LAKE_INTRO: $_LAKE_SEEN"
-_TEL=$(~/.claude/skills/gstack/bin/gstack-config get telemetry 2>/dev/null || true)
+_TEL=$($HOME/.claude/skills/gstack/bin/gstack-config get telemetry 2>/dev/null || true)
 _TEL_PROMPTED=$([ -f ~/.gstack/.telemetry-prompted ] && echo "yes" || echo "no")
 _TEL_START=$(date +%s)
 _SESSION_ID="$$-$(date +%s)"
 echo "TELEMETRY: ${_TEL:-off}"
 echo "TEL_PROMPTED: $_TEL_PROMPTED"
-_EXPLAIN_LEVEL=$(~/.claude/skills/gstack/bin/gstack-config get explain_level 2>/dev/null || echo "default")
+_EXPLAIN_LEVEL=$($HOME/.claude/skills/gstack/bin/gstack-config get explain_level 2>/dev/null || echo "default")
 if [ "$_EXPLAIN_LEVEL" != "default" ] && [ "$_EXPLAIN_LEVEL" != "terse" ]; then _EXPLAIN_LEVEL="default"; fi
 echo "EXPLAIN_LEVEL: $_EXPLAIN_LEVEL"
-_QUESTION_TUNING=$(~/.claude/skills/gstack/bin/gstack-config get question_tuning 2>/dev/null || echo "false")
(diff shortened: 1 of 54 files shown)


garrytan:
Closing — same reason as #1658: this edits auto-generated SKILL.md files, not the `.tmpl` templates that produce them, so the change would be wiped on the next `bun run gen:skill-docs`. `~/` inside `$(...)` also works correctly — this isn't fixing a real bug. If you've found a specific case where tilde expansion fails, please open an issue with the exact command.

Issue #1651, opened 2026-05-22
Two defects in `_SESSIONS` concurrent-session detection — value never echoed + marker TTL doesn't reflect 'currently active'


TJ-NomoAI:
## Context

Built a CLAUDE.md-level concurrent-agent detection rule on top of gstack's `_SESSIONS` mechanism (so a session-start step can force the operator into a separate git worktree when another Claude Code agent is active in the same repo). Two defects surfaced during the ship + dogfooding.

## Defect A — `_SESSIONS` is computed but never echoed

Every skill preamble does:

```bash
mkdir -p ~/.gstack/sessions
touch ~/.gstack/sessions/"$PPID"
_SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
```

The variable is computed but only consumed by an internal check inside the preamble. The preamble echoes many other state values to stdout for the calling agent to read (`BRANCH:`, `PROACTIVE:`, `PROACTIVE_PROMPTED:`, `SKILL_PREFIX:`, `REPO_MODE:`, `LAKE_INTRO:`, `TELEMETRY:`, `TEL_PROMPTED:`, `EXPLAIN_LEVEL:`, `QUESTION_TUNING:`, `ARTIFACTS_SYNC:`, `LEARNINGS:`, `HAS_ROUTING:`, `ROUTING_DECLINED:`, `VENDORED_GSTACK:`, `MODEL_OVERLAY:`, `CHECKPOINT_MODE:`, `CHECKPOINT_PUSH:`) — but never `_SESSIONS:`. Any CLAUDE.md rule that tells the agent to *"read the preamble's session count"* has no value to read.

**Cost:** my v1 CLAUDE.md rule said exactly that and shipped non-functional. Caught on post-ship audit when I re-read my own /ship preamble output adversarially. Worked around by inlining the bash command (`find ~/.gstack/sessions -mmin -120 -type f | wc -l`) into the rule itself.

**Fix shape (cheap):** add `echo "SESSIONS: $_SESSIONS"` to the preamble's existing echo block. ~1 line. Backwards-compatible.

## Defect B — Marker TTL doesn't reflect "currently active"

[...]

PR #1658, opened 2026-05-22
fix(preamble): expose active session count

spacegeologist:
Refs #1651 (Defect A only)

## Reviewer Guide

This is intentionally the cheap/surgical half of #1651. It does not attempt the TTL/long-running-session refresh design.

Actual review targets:
- `scripts/resolvers/preamble/generate-preamble-bash.ts` — emits the already-computed `_SESSIONS` value as `SESSIONS: ...`.
- `test/preamble-compose.test.ts` — regression coverage that the preamble both computes and exposes the session count.
- `test/fixtures/golden/{claude,codex,factory}-ship-SKILL.md` — golden updates from regeneration.

Generated-only churn:
- `*/SKILL.md` files were regenerated with `bun run gen:skill-docs --host all` so every shipped preamble has the same output.

## Why

#1651 has two separate defects:
1. `_SESSIONS` is computed but never printed, so downstream CLAUDE.md rules have no session count to read.
2. Session marker TTL refresh for long-running sessions needs a separate hook/design.

This PR fixes only (1) by adding the missing stdout line after the existing `_SESSIONS` computation. I used `Refs` rather than `Fixes` because (2) remains open.

## Risk / Scope

- Low risk: stdout-only addition after an existing read-only computation.
- No marker file, TTL, cleanup, config, or session ownership behavior changes.
- Generated docs/golden changes are mechanical fallout from the one preamble source edit.

## Merge Coordination

This PR and #1657 both regenerate `*/SKILL.md` from independent resolver fixes. There is no behavior dependency. If #1657 lands first, this branch only needs a rebase/regenerate so the generated docs include both the `SESSIONS:` preamble output and the `$HOME` path fix.

## Verification

- `bun test test/preamble-compose.test.ts`
- `git diff --check`

Note: I also attempted `bun run test` locally, but the sandbox produced unrelated environment failures (port binding, missing `bash` in stripped test envs, missing `jq`, and Swift toolchain/cache permission issues). I am not counting that full-suite run as passing.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -30,6 +30,7 @@ mkdir -p ~/.gstack/sessions
 touch ~/.gstack/sessions/"$PPID"
 _SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
 find ~/.gstack/sessions -mmin +120 -type f -exec rm {} + 2>/dev/null || true
+echo "SESSIONS: $_SESSIONS"
 _PROACTIVE=$(~/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
 _PROACTIVE_PROMPTED=$([ -f ~/.gstack/.proactive-prompted ] && echo "yes" || echo "no")
 _BRANCH=$(git branch --show-current 2>/dev/null || echo "unknown")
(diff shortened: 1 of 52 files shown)


garrytan:
Closing — this edits auto-generated SKILL.md files directly. Per CLAUDE.md's SKILL.md workflow section: SKILL.md files are generated from `.tmpl` templates by `bun run gen:skill-docs`, so any edits to generated files get wiped on the next regen. If exposing active session count is genuinely useful, please open a PR against the preamble resolver in `scripts/resolvers/preamble.ts` so the change persists through regeneration.

PR #1659, opened 2026-05-22
v1.43.3.0 fix(browse): resolveDisconnectCause crashes on persistent-context disconnect

garrytan:
## Summary

`browse/src/browser-manager.ts` — `resolveDisconnectCause` is called on every Playwright `disconnected` event. In headed mode the only launcher we use is `chromium.launchPersistentContext`, which exposes a Browser object whose `.process` is **undefined** (Playwright doesn't surface the underlying Chromium process for persistent contexts).

Pre-fix:

\`\`\`ts
const proc = browser?.process();
\`\`\`

The optional chain \`browser?.process\` short-circuits to \`undefined\` only when \`browser\` itself is null/undefined. With a truthy browser and a missing method, the expression evaluates to \`undefined()\` and throws an unhandled rejection. The throw crashes the bun process; downstream supervisors (gbrowser's \`gbd\`) respawn it; the next tab close hits the same path; loop forever.

Live evidence captured in \`gbrowser/amsterdam-v7\`'s \`~/Library/Logs/GBrowser/workspaces/<id>/browse-server.log\`:

\`\`\`
[overlay] Local listener bound on 127.0.0.1:35300 (PID: 19445)
[stderr] [overlay] FATAL unhandled rejection: browser?.process is not a function.
         (In 'browser?.process()', 'browser?.process' is undefined)
[browse] Shutting down...
...respawn, same crash, repeat...
\`\`\`

## Fix

Split the null case from the no-callable-process case:

- \`browser === null\` → \`'crash'\` (preserves the existing contract pinned by the \`null browser returns crash\` test)
- truthy \`browser\` without callable \`.process\` → \`'clean'\` (persistent contexts; user controls headed-mode lifecycle so the right default is exit 0 / no auto-restart)
- truthy \`browser\` with callable \`.process\` → unchanged exit-code introspection

In headed mode we genuinely can't distinguish \"user pressed Cmd+Q or closed all tabs\" from \"Chromium crashed\" because Playwright doesn't expose the underlying Chromium PID through a persistent context. The tradeoff: if Chromium genuinely crashes in headed mode we now exit 0 and don't auto-restart. That's preferable to the respawn loop, and the user can re-launch manually.

## Test plan

- [x] Added regression test \`clean: browser without .process method (persistent context)\` to \`browse/test/browser-manager-unit.test.ts\` — passes a truthy object with no \`.process\`, which would have thrown pre-fix.
- [x] All 21 tests in \`browser-manager-unit.test.ts\` pass locally (\`bun test browse/test/browser-manager-unit.test.ts\`).
- [x] Verified live: bumped this fix into gbrowser's submodule, restarted \`gbd\` — overlay now exits cleanly on tab close (\`browse_server_clean_exit_no_restart\` in gbd log) instead of crash-looping.


time-attack:
@16francej — recommend closing this PR because #2085 supersedes the remaining work. Most release changes already landed through #1645, and this PR’s remaining guard may hide real crashes rather than handle them correctly. Carry forward the test shape, missing-method finding, and attribution to #2085 for redesign.

PR #1668, opened 2026-05-23
Fix Codex browse runtime on Windows

Catfish-75:
## Summary
- install the browse runtime assets that the Codex skill needs on Windows, including `browse/src`, `browse/bin`, `browse/dist`, and externalized Node runtime packages
- make `browse stop` handle shutdown before `ensureServer()` so a clean stop cannot cold-start a daemon
- preserve cold-start command output on stdout and avoid delayed default startup banners after stop
- add focused regression coverage for the Windows Codex dispatch/runtime setup path

## Verification
- `bun test browse/test/config.test.ts`
- `git diff --check -- browse/src/cli.ts browse/test/commands.test.ts browse/test/config.test.ts setup test/gen-skill-docs.test.ts`
- Installed Codex runtime smoke-tested from outside the repo with `browse.exe stop`, `status`, `goto http://localhost:5173`, `snapshot -i`, and final `stop`

## Notes
Manual `bun test browse/test/commands.test.ts` still times out in the Windows BrowserManager/Playwright launch path. That file is already excluded from the curated Windows-safe shard as POSIX/Playwright-fragile, so this PR keeps the runtime fix focused instead of broadening the test harness change.

### DIFF
--- a/browse/scripts/build-node-server.sh
+++ b/browse/scripts/build-node-server.sh
@@ -25,6 +25,7 @@ bun build "$SRC_DIR/server.ts" \
   --external playwright \
   --external playwright-core \
   --external diff \
+  --external sharp \
   --external "bun:sqlite" \
   --external "@ngrok/ngrok"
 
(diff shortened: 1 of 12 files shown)


garrytan:

PR #1672, opened 2026-05-23
fix(brain-sync): make artifact sync work on Windows (discover-new + drain)

daveowenatl:
Automatic artifact sync (`gstack-brain-sync`) is fully non-functional on Windows (Git Bash): `--discover-new` enqueues nothing and the `--once` drain stages nothing, so `artifacts_sync_mode=full` looks active but no artifacts ever reach the repo. No error surfaces — status reports `idle`.

### Environment
Windows 11 (26200), Git Bash (Git for Windows 2.54.0), Bun 1.3.11, Python 3.14.3 (`python3` on PATH), gstack v1.43.3.0.

### Root causes (three, all in `bin/gstack-brain-sync`)
1. **discover-new path match** — `os.path.relpath` returns backslash paths on Windows, which never match the forward-slash allowlist globs (`projects/*/learnings.jsonl`). Fixed by normalizing `os.sep`→`/`.
   ```
   python3 -c "import os,fnmatch;print(fnmatch.fnmatchcase(os.path.relpath(r'C:\a\projects\p\learnings.jsonl', r'C:\a'),'projects/*/learnings.jsonl'))"  # -> False
   ```
2. **discover-new enqueue** — enqueued via `subprocess.run([gstack-brain-enqueue, rel])`, but Windows Python can't exec a bash-shebang script, so `--discover-new` left the queue empty even when files matched. Now appends to the queue in-process. Same "spawnSync can't exec bash on Windows" family as #1571 / #1641, different code path.
3. **--once drain** — `compute_paths_to_stage` ends in `print(p)`; Windows Python emits CRLF, the bash `while IFS= read -r p` keeps the trailing CR, and `git add -f -- "path<CR>"` fails with `pathspec '…?' did not match` under `2>/dev/null || true`. Nothing is staged, the commit is empty, status reports `idle`. Now strips the trailing CR before `git add`.

The in-process enqueue mirrors `gstack-brain-enqueue`'s contract rather than re-deriving it loosely: one atomic `O_APPEND` write per record (each line < PIPE_BUF) so a parallel writer-shim append can't interleave mid-record, and the discover cursor advances only after the write succeeds (a failed write retries instead of silently recording the file as synced). Skip-list entries are separator-normalized on **both** the discover and drain (`compute_paths_to_stage`) sides, so a backslash `.brain-skip.txt` entry can't be honored at discovery yet bypassed at commit — the drain is the boundary that actually stages files.

### Repro
```bash
gstack-config set artifacts_sync_mode full   # ~/.gstack wired to a remote
mkdir -p ~/.gstack/projects/demo
echo '{}' > ~/.gstack/projects/demo/learnings.jsonl
gstack-brain-sync --discover-new
wc -l < ~/.gstack/.brain-queue.jsonl          # 0  (expected 1)
```

### Testing
- `test/brain-sync-windows-paths.test.ts` — static invariants guarding all three causes plus the two robustness properties (atomic per-record append, normalized skip matching). Behavioral spawn tests can't run on the Windows lane (Node/Bun can't exec the `bin/` shebang scripts — the same reason `brain-sync.test.ts` is excluded from it), so this follows the static-invariant pattern of `build-script-shell-compat.test.ts`. Verified red→green. Wired into `windows-free-tests.yml`.
- Manual end-to-end on Windows 11 / Git Bash: discover → drain → commit → push to a local bare remote lands `sync: N file(s)` with nested and single-level artifacts; non-allowlisted files excluded; a backslash `.brain-skip.txt` entry correctly keeps its file off the remote; queued records are valid compact JSONL; re-discover is idempotent.
- macOS/Linux behavior unchanged: `os.sep` is already `/` so both separator-normalizations are no-ops, there's no CRLF to strip, and the inline `O_APPEND` append is equivalent to the previous per-file enqueue.

### Related (not in this PR)
`bin/gstack-brain-restore:117` has the same relpath+fnmatch bug in its clobber check — happy to follow up separately.

### DIFF
--- a/.github/workflows/windows-free-tests.yml
+++ b/.github/workflows/windows-free-tests.yml
@@ -116,6 +116,7 @@ jobs:
             test/setup-windows-fallback.test.ts \
             test/build-script-shell-compat.test.ts \
             test/docs-config-keys.test.ts \
+            test/brain-sync-windows-paths.test.ts \
             make-pdf/test/browseClient.test.ts \
             make-pdf/test/pdftotext.test.ts
         shell: bash
--- a/bin/gstack-brain-sync
+++ b/bin/gstack-brain-sync
@@ -136,7 +136,11 @@ def load_privacy_map(path):
 
 allowlist_globs = load_lines(allowlist_path)
 privacy_map = load_privacy_map(privacy_path)
-skip_lines = set(load_lines(skip_path))
+# Normalize skip entries to the POSIX form queued paths use, so a backslash
+# entry in .brain-skip.txt still matches on Windows. The drain is the safety
+# boundary that actually stages files, so it must normalize identically to
+# discover_new — otherwise an explicitly-skipped file gets committed.
+skip_lines = {s.replace(os.sep, "/") for s in load_lines(skip_path)}
 
 # Read queue; collect unique file paths.
 queue_paths = set()
@@ -253,6 +257,8 @@ subcmd_once() {
 
   # Stage with git add -f (forces past .gitignore=*) explicit paths only.
   while IFS= read -r p; do
+    p="${p%$'\r'}"   # Windows: compute_paths_to_stage's python print() emits CRLF;
+                     # a trailing CR makes the pathspec match nothing (silent no-stage).
     [ -z "$p" ] && continue
     git -C "$GSTACK_HOME" add -f -- "$p" 2>/dev/null || true
   done < "$paths_file"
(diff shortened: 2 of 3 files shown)


garrytan:
Landed in 64f9aafa1efd36b179c130c1ed8219403e5afda1 as part of v1.44.1.0 wave (PR #1682). Thanks for the fix, @daveowenatl.

Issue #1139, opened 2026-04-22
Voice prompt: "apply to YC" suggestion triggers too frequently across sessions


myi1:
## Summary

The gstack `Voice` section instructs Claude to suggest "apply to YC" only for exceptional cases ("use this rarely and only when truly earned"), but in practice the suggestion fires repeatedly across independent sessions for the same user. A user reported receiving this suggestion three separate times across different gstack sessions.

## The prompt text in question

From the `Voice` section injected into every skill:

> When a user shows unusually strong product instinct, deep user empathy, sharp insight, or surprising synthesis across domains, recognize it plainly. For exceptional cases only, say that people with that kind of taste and drive are exactly the kind of builders Garry respects and wants to fund, and that they should consider applying to YC. Use this rarely and only when truly earned.

## The problem

"Unusually strong," "exceptional," and "truly earned" are qualitative thresholds with no measurable calibration. Each Claude instance applies the threshold independently, and the `Voice` section's Garry-Tan-as-YC-partner framing creates a systematic bias toward the YC lens. Across many sessions, the rate at which the suggestion fires is higher than "rarely."

Symptoms from user feedback:
- "I have seen a suggestion to apply to YC 3 times now using gstack. What are the parameters that trigger this?"
- The repeated suggestion reads as flattery, which the user correctly flagged ("trust that instinct — it's the right instinct for a founder to have about any praise, including mine"). Each individual firing may be defensible; the cumulative pattern is not.

## Why it matters

YC-application suggestions from AI advisory skills are a specific category of praise that carries more weight than generic positive feedback. If the bar slips, the suggestion stops differentiating genuinely exceptional signal from routine good work — which is both bad for founders (false confidence) and bad for the skill's credibility.

## Related

#513 asks to surface the tier reasoning behind the YC recommendation. That's a visibility fix; this is a calibration fix. The two are complementary:
- #513: "Tell me WHY I got this message"
- This issue: "Make sure the bar for WHEN this message fires is actually exceptional"

If #513's three-tier system lands, the top tier in particular needs to be rare enough that multiple independent instances don't all land on it for the same user.

## Proposed fixes (alternatives, not stacked)

[...]

PR #1676, opened 2026-05-24
fix(office-hours): session writer was writing to legacy file

### LINKED ISSUE #3: Project-local .claude/skills/gstack install breaks /browse and setup can report false success
While auditing the install/deploy path end to end, I found a reproducible issue in the README-supported project install flow.

Summary
- A repo-local install at `.claude/skills/gstack` can break `/browse` if there is no matching user-level install under `~/.claude/skills/gstack`.
- Separately, `./setup` can print `gstack ready` even when `browse/dist/browse` was not actually produced.

Why this matters
- The README explicitly positions project-local install as a supported path for teammates.
- In practice, a clean repo-level install can leave `/browse` unusable, which makes the deploy story look successful when it is not.

Repro
1. Install gstack only at the project level (for example `.claude/skills/gstack`) and do not keep a user-level `~/.claude/skills/gstack` install around.
2. Run `./setup` from that project-local `gstack` directory.
3. Run the compiled binary from that project-local install.

Observed behavior
- The compiled CLI can fail trying to resolve `server.ts` from `~/.claude/skills/gstack/browse/src/server.ts` instead of from the adjacent project-local install.
- In my audit, `./setup` also printed `gstack ready` even when `browse/dist/browse` was still missing.

Expected behavior
- A project-local install should be self-contained: compiled `browse` should resolve its server script from the installed repo it lives in.
- `./setup` should exit non-zero if the binary is still missing after the build step.
- `./setup` should also rebuild when source files are newer than the compiled binary, not only when the binary is missing.

Root cause
- `browse/src/cli.ts` falls back to `path.resolve(process.env.HOME || "/tmp", ".claude/skills/gstack/browse/src/server.ts")` when running from the compiled binary, so the runtime assumes a user-level install exists.
- `setup` only checked `if [ ! -x "$GSTACK_DIR/browse/dist/browse" ]` before building, and it did not verify the artifact existed afterward.

I validated a fix locally by:
- resolving `server.ts` from the compiled binary's adjacent `browse/src/server.ts` first,
- keeping the env override and HOME fallback,
- making `setup` fail hard if `browse/dist/browse` is still missing,
- and adding regression coverage for the compiled project-install path.

### LINKED ISSUE #1677: /office-hours always reports SESSION_COUNT: 0, welcome_back tier is unreachable
# `/office-hours` always reports `SESSION_COUNT: 0`, welcome_back tier is unreachable

Every `/office-hours` run on a fresh `$HOME` returns `TIER: introduction` and fires the full closing pitch. Doesn't matter how many times the user has run it. The `welcome_back` tier in `bin/gstack-developer-profile:165-169` (which exists specifically to skip the closing pitch for returning users) is never reached on fresh-`$HOME` installs.

The v1.0.0.0 migration moved the read path to `~/.gstack/developer-profile.json`, but the writer in `office-hours/SKILL.md` (lines 1560 and 2022 as of 2026-05-24) still appends to `~/.gstack/builder-profile.jsonl`. On first read, `ensure_profile()` creates a stub `developer-profile.json` with `sessions: []`. Later writes go to a file the reader never re-reads. `SESSION_COUNT` stays 0 forever.

Live since `0a803f9` / PR #1039 / v1.0.0.0 (2026-04-18). 5 weeks on every fresh-`$HOME` user.

## Repro

Fresh `$HOME`, run `/office-hours` twice (or check `~/.gstack/` between runs):

```
05:26  ~/.gstack/developer-profile.json   created  (sessions: [], empty stub)
05:53  ~/.gstack/builder-profile.jsonl    appended (today's session)
```

27-minute gap inside one skill run. Reader saw the empty stub at preamble; writer appended to the legacy file at completion. Second `/office-hours` run re-reads the stub, still sees `sessions: []`, still picks the introduction tier.

## Fix

Either (a) move the writer to `developer-profile.json`, or (b) move the reader back to `builder-profile.jsonl`. (a) is the direction v1.0.0.0 was going; (b) is a revert. Option (a) needs a small subcommand on `gstack-developer-profile` that read-modify-writes the `sessions[]` array. Whoever owns this should also decide on stranded entries in `builder-profile.jsonl` from the broken-writer period (leave them; loss is aesthetic).

Filing as a regression report, not a PR. Happy to send a PR if useful; flag here first since the choice of (a) vs (b) is yours.

pryow:
Fixes #1677.

## What users get

Returning `/office-hours` users now hit the `welcome_back` tier on their second visit instead of being pitched as first-timers every time. Live on every fresh-`$HOME` user since v1.0.0.0 (~5 weeks).

## Root cause

v1.0.0.0 moved the read path to `~/.gstack/developer-profile.json` but left the writer in `office-hours/SKILL.md.tmpl` echoing to the legacy `~/.gstack/builder-profile.jsonl`. `ensure_profile` creates a stub with `sessions: []`; subsequent writes went to a file the reader never re-reads. Reader and writer disagreed on storage. `SESSION_COUNT` stayed at 0 forever, `welcome_back` (which exists specifically to skip the closing pitch for returning users) was unreachable on fresh-`$HOME` installs.

## Fix

1. New `bin/gstack-developer-profile --log-session '<json>'` — read-modify-writes `developer-profile.json`'s `sessions[]`. Atomic mktemp+mv. Silent skip on invalid JSON (matches `gstack-timeline-log:22-26`). Aggregates `signals_accumulated`, `resources_shown`, `topics`. Calls `gstack-brain-enqueue` after write, mirroring `gstack-timeline-log:40`. Verb joins the `gstack-*-log` family.
2. Swap `office-hours/SKILL.md.tmpl` writer calls at lines 490 + 893 to use the new subcommand. Regenerate `office-hours/SKILL.md`.
3. Fix latent `do_read` bug: filter `mode:"resources"` entries when picking `LAST_PROJECT` / `LAST_ASSIGNMENT` / `LAST_DESIGN_TITLE` / `CROSS_PROJECT`. Phase 6 resources auto-append runs after the real session in the same invocation; without the filter, it clobbers real-session state next visit. Masked by the broken writer; activated by this fix.

## Affected users

Stranded entries in `~/.gstack/builder-profile.jsonl` from the broken-writer period (v1.0.0.0 → v1.43.4.0) are not auto-recovered. Affected users' next `/office-hours` session lands fresh in `welcome_back`; past data stays in the legacy file (still readable by other tools during deprecation). Most pre-existing users have only a handful of stranded sessions; the loss is mostly aesthetic.

## Tests

+8 behavioral tests in `test/gstack-developer-profile.test.ts`: regression (read-write-read promotes to welcome_back), aggregation (signals, resources_shown, topics dedup), validation (silent skip on invalid input), `ts` injection/preservation, `do_read` mode filter. +3 static-grep invariants in `test/static-no-legacy-writes.test.ts` (new) that prevent future writers from regressing onto the legacy file. All 62 pass in 5.17s.

<details>
<summary><strong>What's intentionally NOT in this fix</strong></summary>

- **No auto-reconcile for stranded entries.** One-release-only reconcile code dropped per right-sized-diff voice. Loss is mostly aesthetic (see "Affected users" above).
- **No new binary.** `--log-session` belongs on the owner of `developer-profile.json`, not the `gstack-*-log` event family. Joins `--migrate` / `--derive` as a write-side subcommand on the same binary.
- **No mkdir-locks.** Concurrent /office-hours has a r-m-w race on `developer-profile.json`. `gstack-config` accepts the same race on YAML; not introduced here.
- **No schema bump.** Stays at `schema_version: 1`.

Full rationale: `docs/designs/FIX_1671_PROFILE_MIGRATION.md`.

</details>

## Related

#1139 — *Voice prompt: "apply to YC" suggestion triggers too frequently across sessions*. This PR unblocks proposed fix #3 (suppress-on-prior-session detection). Without correct `SESSION_COUNT`, that detection couldn't have worked even if implemented; YC-apply suppression itself is separate work.


garrytan:
Landed in 64f9aafa1efd36b179c130c1ed8219403e5afda1 as part of v1.44.1.0 wave (PR #1682). Thanks for the fix, @pryow.

PR #1690, opened 2026-05-25
fix(make-pdf): give TOC headings stable ids so --toc anchors and page numbers resolve

### LINKED ISSUE #1689: make-pdf: --toc generates dangling anchors and unresolvable page-number targets (no heading carries the toc-N id)
## Summary

`make-pdf --toc` builds a table of contents whose entries link to, and whose page-number spans target, element ids (`toc-0`, `toc-1`, ...) that are **never assigned to any heading**. Every TOC anchor is a dead in-page link, and every page-number target is unresolvable, so Paged.js has nothing to fill the `.toc-page` spans against.

On `origin/main` (`920a13a1`):

```bash
bun -e '
import { render } from "./make-pdf/src/render.ts";
const r = render({ markdown: "# One\n\n## Sub\n\nbody\n\n# Two\n\nbody\n", toc: true });
const hrefs = [...r.html.matchAll(/href="(#toc-\d+)"/g)].map(m => m[1]);
const headingIds = [...r.html.matchAll(/<h[1-3][^>]*\bid="([^"]+)"/g)].map(m => m[1]);
console.log("TOC anchor hrefs:   ", hrefs);
console.log("heading ids in body:", headingIds);
'
```

prints:

```
TOC anchor hrefs:    [ "#toc-0", "#toc-1", "#toc-2" ]
heading ids in body: []
```

The TOC entries point at `#toc-0` / `#toc-1` / `#toc-2`, but no `<h1>`/`<h2>`/`<h3>` in the document body carries any of those ids.

## Root cause

In `make-pdf/src/render.ts`, `buildTocBlock` mints an id per heading purely for its own markup:

```ts
const id = `toc-${i}`;
// ...
`    <span class="toc-title"><a href="#${id}">${escapeHtml(h.text)}</a></span>`,
`    <span class="toc-page" data-toc-target="${id}"></span>`,
```

The body headings come straight from `marked` (which emits `<h1>One</h1>` with no id) and `wrapChaptersByH1`, and nothing in the render pipeline ever assigns `id="toc-N"` to a heading. So the anchors and `data-toc-target` ids reference elements that do not exist.

The function comment says "Page numbers are filled in by Paged.js", but Paged.js needs the target element (the heading) to exist with the referenced id before it can resolve a page number for it.

## Impact

- Clicking any TOC entry in the rendered PDF/HTML does nothing (dangling fragment).
- Paged.js page-number resolution for the TOC has no target to count against.

The existing render test ("builds a TOC with H1/H2 entries when toc=true") only asserts that the heading text appears in the TOC, so this is uncovered.

## Expected behavior

Each H1-H3 heading in the body should carry the id its TOC entry references, so `<a href="#toc-N">` and `data-toc-target="toc-N"` resolve to the matching heading (in document order). Headings that already declare an id should keep it, with the TOC pointing at the existing id.

officialasishkumar:
## Summary

`make-pdf --toc` built a table of contents whose entries linked to, and whose page-number spans targeted, element ids (`toc-0`, `toc-1`, ...) that were never assigned to any heading. Every TOC anchor was a dead in-page link and every page-number target was unresolvable, so Paged.js had nothing to fill the `.toc-page` spans against.

Fixes #1689.

On `origin/main` (`920a13a1`):

```
TOC anchor hrefs:    [ "#toc-0", "#toc-1", "#toc-2" ]
heading ids in body: []
```

The TOC entries point at `#toc-0` / `#toc-1` / `#toc-2`, but no heading in the body carries any of those ids (they come straight from `marked` and `wrapChaptersByH1`, neither of which assigns ids).

## Fix

Replace `extractHeadings` with `assignTocIds`, which walks the H1-H3 headings in document order, assigns `id="toc-N"` to each, and returns both the id annotated body HTML and the heading list. `render()` uses the annotated HTML for the body when a TOC is requested and builds the TOC from the same list, so anchors and `data-toc-target` ids resolve to real headings. A heading that already declares an id keeps it (the TOC points at that id), so no duplicate id is emitted.

With no TOC the body is untouched (no ids injected), so non-TOC output is byte-identical to before.

## Validation

After the fix:

```
hrefs: ["#toc-0","#toc-1","#toc-2"] | heading ids: ["toc-0","toc-1","toc-2"] | all resolve: true
```

Tests (the make-pdf-gate unit step, `bun test make-pdf/test/*.test.ts`):

```
90 pass, 0 fail
```

Three new regression tests cover: anchors and page targets resolving to heading ids in document order, reuse of an existing heading id without duplication, and no id injection when the TOC is off.

The P0 combined-features gate (`make-pdf/test/e2e/combined-gate.test.ts`) runs `generate ... --quiet` with no `--toc`. Since the change is entirely behind `if (opts.toc)` and `--toc` defaults to false, the gate's rendered HTML (and extracted text) is unchanged.

## Scope

`make-pdf/src/render.ts` only (`render`, `buildTocBlock`, and `extractHeadings` to `assignTocIds`), plus three regression tests. No template, VERSION, or CHANGELOG changes.

### DIFF
--- a/make-pdf/src/render.ts
+++ b/make-pdf/src/render.ts
@@ -106,14 +106,21 @@ export function render(opts: RenderOptions): RenderResult {
       })
     : "";
 
-  const tocBlock = opts.toc
-    ? buildTocBlock(typographicHtml)
-    : "";
+  // When a TOC is requested, give the body's headings stable ids so the TOC
+  // anchors (#toc-N) and Paged.js page-number targets resolve to them. With no
+  // TOC, the body is untouched (no ids injected).
+  let bodyHtml = typographicHtml;
+  let tocBlock = "";
+  if (opts.toc) {
+    const withIds = assignTocIds(typographicHtml);
+    bodyHtml = withIds.html;
+    tocBlock = buildTocBlock(withIds.headings);
+  }
 
   // Wrap body in .chapter sections at H1 boundaries if chapter breaks are on.
   const chapterHtml = opts.noChapterBreaks
-    ? `<section class="chapter">${typographicHtml}</section>`
-    : wrapChaptersByH1(typographicHtml);
+    ? `<section class="chapter">${bodyHtml}</section>`
+    : wrapChaptersByH1(bodyHtml);
 
   const watermarkBlock = opts.watermark
     ? `<div class="watermark">${escapeHtml(opts.watermark)}</div>`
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this. The branch conflicts with the broader TOC implementation now on main, so landing it wholesale would reintroduce stale architecture. Port only the empty/nonempty heading-alignment invariant and its focused regression tests onto current main.

Issue #785, opened 2026-04-03
Use $HOME instead of ~ in SKILL.md browse binary assignment


bradleylab:
## Problem

All SKILL.md files that reference the browse binary use this pattern:

```bash
[ -z "$B" ] && B=~/.claude/skills/gstack/browse/dist/browse
```

Claude Code's built-in bash safety checks flag this with a permission prompt:

> Tilde in assignment value — bash may expand at assignment time

This fires on **every** browse command, which means frequent interruptions during any skill that uses the browser (browse, qa, design-review, etc.).

## Fix

Replace `~` with `$HOME` in all 15+ SKILL.md files:

```bash
[ -z "$B" ] && B="$HOME/.claude/skills/gstack/browse/dist/browse"
```

This is functionally identical (both resolve to the user's home directory) but avoids the Claude Code heuristic warning.

## Affected files

Every SKILL.md that sets `B=~/.claude/skills/gstack/browse/dist/browse`:
- browse/SKILL.md
- qa/SKILL.md
- qa-only/SKILL.md
- design-review/SKILL.md
- design-consultation/SKILL.md
- design-shotgun/SKILL.md
- design-html/SKILL.md
- canary/SKILL.md
- benchmark/SKILL.md
- land-and-deploy/SKILL.md
- office-hours/SKILL.md
- plan-design-review/SKILL.md
- setup-browser-cookies/SKILL.md
- connect-chrome/SKILL.md
- SKILL.md (root)

vincent067:
## Hey there! 👋

I'd love to help with this issue! I noticed the same annoying prompts while using Claude Code with gstack skills.

I've prepared a fix script that replaces all instances of `~` with `$HOME` in the SKILL.md files:

```bash
#!/bin/bash
# Fix for issue #785: Replace ~ with $HOME in SKILL.md files

find . -name "SKILL.md" -type f | while read -r file; do
    if grep -q 'B=~/.claude/skills/gstack/browse/dist/browse' "$file"; then
        echo "Fixing: $file"
        sed -i 's/B=~\/.claude\/skills\/gstack\/browse\/dist\/browse/B="$HOME\/.claude\/skills\/gstack\/browse\/dist\/browse"/g' "$file"
    fi
done
```

Alternatively, here's a one-liner that does the same:

```bash
[...]

PR #1692, opened 2026-05-25
fix(skill-preamble): use $HOME instead of literal ~ in quoted bin paths (#1656)

### LINKED ISSUE #1656: Tilde-in-quoted-string disables brain-sync silently across 42 SKILL.md files (regression of #785)

0xDevNinja:
## Summary

- `scripts/resolvers/preamble/generate-brain-sync-block.ts` emits `_BRAIN_SYNC_BIN="${binDir}/gstack-brain-sync"` and `_BRAIN_CONFIG_BIN="${binDir}/gstack-config"`. For the Claude host `binDir` is `~/.claude/skills/gstack/bin`, so the rendered assignments hold a literal `~`. Tilde inside double quotes is never expanded.
- `scripts/resolvers/preamble/generate-preamble-bash.ts` has the same shape on line 55 — `[ -x "${binDir}/gstack-telemetry-log" ]` — the `-x` test then always returns false on Claude hosts.
- The surrounding `2>/dev/null || true` swallows ENOENT, so the artifacts-sync preamble silently no-ops in 48 generated SKILL.md files (`/context-save`, `/context-restore`, `/autoplan`, `/codex`, etc.). Cross-machine artifact handoff never pushes.
- Normalize `ctx.paths.binDir` to `$HOME/...` form before interpolating into the quoted contexts. Env-var hosts (Codex, Hermes, etc.) whose `binDir` is already `$GSTACK_BIN` are untouched — only the `~/...` prefix is rewritten.

Fixes #1656. Same defect class as #785 / #843 (which targeted `browseDir` / `designDir`); this PR is the matching fix on the brain-sync surface.

## Why this shape

- Local TS rewrite inside the two affected resolvers, not a global change to `scripts/resolvers/types.ts`. PR #843 already proposes a global swap on `browseDir` / `designDir`; doing the same swap on `binDir` from this PR would silently conflict with anyone else touching `types.ts`. The local pattern keeps the diff narrow and reversible.
- Static invariants land in `test/skill-validation.test.ts`. Two new tests fail loudly if any generated SKILL.md re-introduces `="~/` or `[ -x "~/`. Existing skills are now clean (`grep` returns 0 hits), so the invariants pass.

## Diff shape

- 2 resolver files: +18 / -3
- 1 test file: +59
- 47 regenerated SKILL.md files (no manual edits; produced by `bun run gen:skill-docs`)

## Tests

```
bun test test/skill-validation.test.ts
# 331 pass, 0 fail (existing 329 + 2 new invariants)

# 711 pass, 7 fail
```

The 7 failures are the pre-existing `--dry-run freshness check` cases on non-Claude hosts that reproduce on `upstream/main` with this branch's changes stashed — unrelated.

## Verification

Sample lines from the regenerated `context-save/SKILL.md`:

```
68:    if [ "$_TEL" != "off" ] && [ -x "$HOME/.claude/skills/gstack/bin/gstack-telemetry-log" ]; then
369: _BRAIN_SYNC_BIN="$HOME/.claude/skills/gstack/bin/gstack-brain-sync"
370: _BRAIN_CONFIG_BIN="$HOME/.claude/skills/gstack/bin/gstack-config"
```

`grep -rEl '_BRAIN_SYNC_BIN="~/|_BRAIN_CONFIG_BIN="~/|-x "~/' */SKILL.md SKILL.md` returns 0 files post-patch.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -84,7 +84,7 @@ echo '{"skill":"gstack","ts":"'$(date -u +%Y-%m-%dT%H:%M:%SZ)'","repo":"'$(_repo
 fi
 for _PF in $(find ~/.gstack/analytics -maxdepth 1 -name '.pending-*' 2>/dev/null); do
   if [ -f "$_PF" ]; then
-    if [ "$_TEL" != "off" ] && [ -x "~/.claude/skills/gstack/bin/gstack-telemetry-log" ]; then
+    if [ "$_TEL" != "off" ] && [ -x "$HOME/.claude/skills/gstack/bin/gstack-telemetry-log" ]; then
       ~/.claude/skills/gstack/bin/gstack-telemetry-log --event-type skill_run --skill _pending_finalize --outcome unknown --session-id "$_SESSION_ID" 2>/dev/null || true
     fi
     rm -f "$_PF" 2>/dev/null || true
@@ -337,8 +337,8 @@ if [ -f "$HOME/.gstack-artifacts-remote.txt" ]; then
 else
   _BRAIN_REMOTE_FILE="$HOME/.gstack-brain-remote.txt"
 fi
-_BRAIN_SYNC_BIN="~/.claude/skills/gstack/bin/gstack-brain-sync"
-_BRAIN_CONFIG_BIN="~/.claude/skills/gstack/bin/gstack-config"
+_BRAIN_SYNC_BIN="$HOME/.claude/skills/gstack/bin/gstack-brain-sync"
+_BRAIN_CONFIG_BIN="$HOME/.claude/skills/gstack/bin/gstack-config"
 
 # /sync-gbrain context-load: teach the agent to use gbrain when it's available.
 # Per-worktree pin: post-spike redesign uses kubectl-style `.gbrain-source` in the
@@ -447,8 +447,8 @@ If A/B and `~/.gstack/.git` is missing, ask whether to run `gstack-artifacts-ini
 At skill END before telemetry:
 
 ```bash
-"~/.claude/skills/gstack/bin/gstack-brain-sync" --discover-new 2>/dev/null || true
-"~/.claude/skills/gstack/bin/gstack-brain-sync" --once 2>/dev/null || true
+"$HOME/.claude/skills/gstack/bin/gstack-brain-sync" --discover-new 2>/dev/null || true
+"$HOME/.claude/skills/gstack/bin/gstack-brain-sync" --once 2>/dev/null || true
 ```
 
 
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -94,7 +94,7 @@ echo '{"skill":"autoplan","ts":"'$(date -u +%Y-%m-%dT%H:%M:%SZ)'","repo":"'$(_re
 fi
 for _PF in $(find ~/.gstack/analytics -maxdepth 1 -name '.pending-*' 2>/dev/null); do
   if [ -f "$_PF" ]; then
-    if [ "$_TEL" != "off" ] && [ -x "~/.claude/skills/gstack/bin/gstack-telemetry-log" ]; then
+    if [ "$_TEL" != "off" ] && [ -x "$HOME/.claude/skills/gstack/bin/gstack-telemetry-log" ]; then
       ~/.claude/skills/gstack/bin/gstack-telemetry-log --event-type skill_run --skill _pending_finalize --outcome unknown --session-id "$_SESSION_ID" 2>/dev/null || true
     fi
     rm -f "$_PF" 2>/dev/null || true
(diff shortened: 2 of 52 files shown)


time-attack:
@0xDevNinja Please rewrite this to include the newer generated outputs and tests proving these commands activate safely with $HOME paths.

0xDevNinja:
rebased onto current main and regenerated. the rebase picked up two skills added since this branch was cut (diagram, spec) whose SKILL.md still had the literal-tilde assignments — regenerated both, zero `="~/` instances remain across all generated files. the static invariants already in test/skill-validation.test.ts scan every generated SKILL.md, so newly added skills are covered automatically from here. 333 pass.

PR #1697, opened 2026-05-25
Validate skill bash snippets in tests

### LINKED ISSUE #1667: feat: validate bash syntax in SKILL.md code blocks during bun test (prevent silent runtime breakage)
## Problem

`bun test` validates many properties of SKILL.md files — unresolved placeholders, token budgets, preamble freshness, hardcoded branch names, AskUserQuestion format — but never checks whether the **bash code blocks inside them are syntactically valid**.

This means an entire class of breakage can ship silently: corrupted shell redirects, bad quoting, broken pipes, and malformed heredocs all pass `bun test` and only blow up at runtime when a user invokes the skill.

### Motivating case: PR #1654

[PR #1654](https://github.com/garrytan/gstack/pull/1654) systematically changed `2>/dev/null` → `2/dev/null` (removed the `>` redirection operator) in **12+ places** throughout `autoplan/SKILL.md`. Every one of those lines would fail at runtime. `bun test` passed. The only guard was human review.

This is not a one-off — any automated tool, refactor, or merge conflict resolution that corrupts bash syntax inside a code block will go undetected until a user hits it.

## Proposed solution

Add a `bash -n` (syntax-check-only, no execution) pass over every ` ```bash ``` ` block in every generated SKILL.md file as part of `bun test`.

### Implementation sketch

```typescript
// test/skill-bash-syntax.test.ts
// Extract all ```bash ... ``` blocks from every SKILL.md,
// preprocess template placeholders, then run `bash -n` on each.

function extractBashBlocks(content: string): { block: string; lineNum: number }[]

function preprocessForSyntaxCheck(block: string): string {
  // Replace <placeholders> with valid bash variable references
  // Replace {{RESOLVER}} template tokens (these are .tmpl artifacts, not bash)
  // Substitute common gstack-specific tokens
}

// For each skill's SKILL.md:
//   for each bash block:
//     const result = spawnSync('bash', ['-n'], { input: preprocessed })
//     expect(result.status).toBe(0)  // fail with file + line on error
```

### What it catches
- Missing `>` in `2>/dev/null` (the PR #1654 bug)  
- Unclosed quotes or heredocs
- Invalid redirection syntax
- Broken pipe chains (`cmd |` with nothing after)
- Mismatched `if/fi`, `for/done`, `case/esac`

### What it explicitly does NOT check
- Correctness of commands (not `bash -e`, just `bash -n`)
- Runtime behaviour
- Blocks in `.tmpl` files (only generated `.md` files, since those are what ships)
- Non-bash blocks (` ```python `, ` ```json `, ` ```markdown `)

## Scope

- New test file: `test/skill-bash-syntax.test.ts`
- Tier: `gate` (free, fast — `bash -n` on 50+ skills takes < 2s total)
- No new dependencies — `bash` is already assumed present

## Expected outcome

Running `bun test` after PR #1654's changes would produce:

```
(fail) bash syntax: autoplan/SKILL.md block at line 43 [1.2ms]
  bash: line 1: 2/dev/null: No such file or directory
  + 11 more failures
```

instead of silently passing.

xuezhaolan:
## Summary
- add a SKILL.md freshness test that runs `bash -n` over concrete `bash`/`sh` fenced snippets
- skip placeholder/example snippets such as `<url>`, `{{PLACEHOLDER}}`, and ellipsis blocks
- batch syntax checks per markdown file for speed, then fall back to per-block checks for actionable file/line errors

Fixes #1667.

## Test plan
- `bun test test/skill-parser.test.ts test/skill-validation.test.ts`

Note: I also tried `bun test`; it did not complete cleanly on this checkout due existing unrelated failures/timeouts, including a missing `.factory/skills/gstack-ship/SKILL.md` golden fixture and a timeout in `browse/test/security-audit-r2.test.ts`.

### DIFF
--- a/test/helpers/skill-parser.ts
+++ b/test/helpers/skill-parser.ts
@@ -14,6 +14,7 @@ import { ALL_COMMANDS } from '../../browse/src/commands';
 import { parseSnapshotArgs } from '../../browse/src/snapshot';
 import * as fs from 'fs';
 import * as path from 'path';
+import { spawnSync } from 'child_process';
 
 /** CLI-only commands: valid $B invocations that are handled by the CLI, not the server */
 const CLI_COMMANDS = new Set([
@@ -32,6 +33,18 @@ export interface ValidationResult {
   invalid: BrowseCommand[];
   snapshotFlagErrors: Array<{ command: BrowseCommand; error: string }>;
   warnings: string[];
+}
+
+export interface BashBlock {
+  file: string;
+  line: number;
+  content: string;
+}
+
+export interface BashSyntaxError {
+  file: string;
+  line: number;
+  message: string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


garrytan:
Closing in favor of #1722, which solves the same goal (#1667) more completely: it runs `bash -n` across every block in all SKILL.md files and caught + fixed a real fence bug in codex/SKILL.md. Can't land both. Thanks for the contribution!

PR #1702, opened 2026-05-25
fix(slug): walk up to outermost project root + self-heal stale cache

ajeenkya:
## Summary

`bin/gstack-slug` previously derived the project slug from the literal `$(pwd)` with no walk-up. When a session's cwd was inside a subdir that happened to contain its own project-marker-like file (e.g. `.vercel/`, `.git/`, `package.json`), the slug silently resolved to the subdir's basename. Every downstream consumer — `/context-save`, the autosave hook, `/next`, gstack analytics, and operational learnings (`~/.gstack/projects/<slug>/learnings.jsonl`) — then wrote state under a phantom slug for the rest of the session. Not an error, no visible failure. The poisoned cache entry persisted across sessions, so the bug compounded once seeded.

## Reproduction (confirmed 2026-05-25 on macOS)

1. From a project root with no `.git` (e.g. a content folder), `gstack-slug` correctly returns the basename.
2. `cd site && vercel --prod` creates `site/.vercel/project.json`.
3. For the rest of the session, any `gstack-slug` invocation while cwd is `site/` (or descendant) returns `SLUG=site` and caches it. The autosave hook (inheriting session cwd) files every subsequent checkpoint under `~/.gstack/projects/site/checkpoints/` instead of the real project's `~/.gstack/projects/<project>/checkpoints/`. Three real checkpoints misfiled this way required manual rescue + phantom-slug-folder removal.

## Fix

**Walk UP from `$(pwd)` to the OUTERMOST ancestor containing a canonical project-identity marker.** Two-tier marker model:

- **STRONG markers** (canonical version-control / language project files): `.git` (dir or file — handles worktrees), `.project.yaml`, `package.json`, `pyproject.toml`, `Cargo.toml`, `Gemfile`, `go.mod`.
- **WEAK markers** (content-only project signals): `README.md`, `README`, `README.rst`, `LICENSE`, `LICENSE.md`.

Rule: outermost STRONG wins. If no strong marker on the chain, outermost WEAK wins. If neither, fall back to pwd basename (preserves prior behavior for plain non-project folders). Build/deploy artifacts (`.vercel`, `.next`, `dist`, `node_modules`, `.turbo`, `.nx`, etc.) are deliberately **NOT** markers — they're tooling output, not project identity, and allow-listing rather than blacklisting means new tooling cannot silently establish phantom project roots.

Three consequences worth flagging:

- `loadout/site/.vercel/` → `loadout` (weak `README.md` on parent; `.vercel` is not a marker; weak parent wins because no strong marker exists). Fixes the canonical bug.
- `loadout/starter-pack/.git/` → `starter-pack` (strong `.git` on child beats weak `README.md` on parent; vendored sub-repo genuinely IS its own project).
- `monorepo/.git + monorepo/packages/web/package.json` → `monorepo` (outermost strong wins; sub-package folds into monorepo).

**Self-healing cache.** When a cached slug for the literal pwd differs from the freshly-computed slug, overwrite that one cache entry. Single-shot, key-local — never sweeps other entries. This lets pre-existing poisoned caches clean themselves up on next invocation without a manual `rm -rf ~/.gstack/slug-cache/`.

**Escape hatch.** `GSTACK_PROJECT_SLUG` env var beats every other resolution path, for the legitimate "subdir IS its own project" edge case.

**Preserves every existing invariant**: `set -euo pipefail`, stdout shape (`SLUG=...\nBRANCH=...\n`), output sanitization to `[a-zA-Z0-9._-]`, atomic mktemp + mv cache writes, fail-silent cache handling, POSIX-portable bash, macOS BSD compatible. No new dependencies.

## Test plan

New regression test `test/gstack-slug-cwd-walk-up.test.ts` (12 cases) using the existing `bun:test` + `tmpHome` + `spawnSync` pattern (mirrors `test/migration-checkpoint-ownership.test.ts`).

- [x] AC-1: `.git` at root + `.vercel` in subdir from subdir → root basename
- [x] AC-1 variant: `package.json` at root + `node_modules`-only subdir → root basename
- [x] AC-2: stale cache for subdir pwd self-heals to correct slug
- [x] AC-3: cwd IS project root (no regression, basename)
- [x] AC-4: no markers anywhere on chain (no regression, pwd basename fallback)
- [x] AC-5: real git remote → slug from remote URL (no regression — existing logic preserved)
- [x] AC-6: cache eviction is single-shot, doesn't touch other entries
- [x] AC-7: stdout shape preserved exactly (`SLUG=...\nBRANCH=...\n`, sanitized)
- [x] Two-tier: vendored sub-repo with `.git` wins over parent README (strong > weak)
- [x] Two-tier: weak-marker chain falls back to outermost weak when no strong marker exists
- [x] `GSTACK_PROJECT_SLUG` env override beats every other resolution path
- [x] `bash -n bin/gstack-slug` clean
- [x] Full default test suite green (EXIT=0; one pre-existing flake in `test/gstack-developer-profile.test.ts:135` ANSI-escape expectation unrelated to this change — fails identically on base SHA `64f9aaf` with this diff stashed)
- [x] **Live dogfooded against the original reporter's filesystem** — five poisoned subdir cache entries (`_..._loadout_{assets,chapters,dist,site,starter-pack}`) all self-healed correctly on next invocation; `starter-pack` correctly retained its own slug via its `.git`

## Note on macOS test fixtures

The test calls `fs.realpathSync` on `mkdtempSync` output because macOS `os.tmpdir()` returns `/var/folders/...` but bash `$(pwd)` returns `/private/var/folders/...` — without canonicalization the cache-key tests silently misalign (the script writes to one key, the test pre-seeds another). Documented in the test's `beforeEach`.

## Tracking

Linear project: [gstack — slug ambient-cwd bug fix](https://linear.app/milo-labs/project/gstack-slug-ambient-cwd-bug-fix-266ed24b6753)
Issues: [ENG-97](https://linear.app/milo-labs/issue/ENG-97/fix-bingstack-slug-walking-up-to-outermost-project-root-self-heal) (fix), [ENG-98](https://linear.app/milo-labs/issue/ENG-98/document-gstack-slug-ambient-cwd-bug-chosen-fix-in-aj-personal-memory) (memory entry).

### DIFF
--- a/bin/gstack-slug
+++ b/bin/gstack-slug
@@ -49,9 +49,25 @@ fi
 #    Markers: .git can be a directory (normal repo) or a file (worktree /
 #    submodule pointer). Everything else is a file at the directory's top
 #    level.
+#    Two tiers of markers:
+#    - STRONG markers (canonical version-control / language project files):
+#      .git, .project.yaml, package.json, pyproject.toml, Cargo.toml, Gemfile,
+#      go.mod. These signal "this directory is a real project of its own."
+#    - WEAK markers (content-only project signals): README.md, README, LICENSE.
+#      These catch content folders (markdown bundles, asset collections, AJ's
+#      loadout-style folders) that have no programming-language project files
+#      but ARE the user's project root.
+#    Rule: outermost STRONG marker wins. If no strong marker exists anywhere on
+#    the chain, outermost WEAK marker wins. This means a vendored sub-repo
+#    (e.g. `loadout/starter-pack/.git`) correctly keeps its own slug even when
+#    a weak-marker parent (`loadout/README.md`) is higher up — the sub-repo IS
+#    its own project. But a deploy-artifact-only subdir (`loadout/site/.vercel`)
+#    correctly folds into the content-project parent (`loadout/README.md`),
+#    because `.vercel` is not a marker at all.
 _outermost_project_root() {
   local dir="$1"
-  local outermost=""
+  local outermost_strong=""
+  local outermost_weak=""
   while [[ -n "$dir" && "$dir" != "/" ]]; do
     if [[ -e "$dir/.git" \
        || -f "$dir/.project.yaml" \
@@ -60,11 +76,22 @@ _outermost_project_root() {
        || -f "$dir/Cargo.toml" \
        || -f "$dir/Gemfile" \
        || -f "$dir/go.mod" ]]; then
-      outermost="$dir"
+      outermost_strong="$dir"
+    elif [[ -f "$dir/README.md" \
+         || -f "$dir/README" \
+         || -f "$dir/README.rst" \
+         || -f "$dir/LICENSE" \
+         || -f "$dir/LICENSE.md" ]]; then
+      outermost_weak="$dir"
     fi
     dir=$(dirname "$dir")
   done
-  printf '%s' "$outermost"
+  # Strong markers win over weak; either wins over nothing.
+  if [[ -n "$outermost_strong" ]]; then
+    printf '%s' "$outermost_strong"
+  else
+    printf '%s' "$outermost_weak"
+  fi
 }
 
 # Only compute the project root if we don't already have a slug (env override
--- a/bin/gstack-slug
+++ b/bin/gstack-slug
@@ -3,8 +3,28 @@
 # Usage: eval "$(gstack-slug)"  → sets SLUG and BRANCH variables
 # Or:    gstack-slug            → prints SLUG=... and BRANCH=... lines
 #
-# Security: output is sanitized to [a-zA-Z0-9._-] only, preventing
-# shell injection when consumed via source or eval.
+# Resolution order (highest precedence first):
+#   0. $GSTACK_PROJECT_SLUG env override (documented escape hatch)
+#   1. Walk UP from $(pwd) to the OUTERMOST ancestor containing a canonical
+#      project-identity marker (.git, .project.yaml, package.json, pyproject.toml,
+#      Cargo.toml, Gemfile, go.mod). Use that ancestor as the "project root".
+#      Build/deploy artifacts (.vercel, .next, dist, node_modules, etc.) are
+#      DELIBERATELY NOT markers — they're tooling output, not project identity.
+#      Without this walk-up, running gstack-slug from a subdir whose only
+#      "marker" is a deploy artifact silently resolves to the subdir's basename,
+#      misfiling all session state under a phantom slug. (2026-05-25 bug fix.)
+#   2. If the resolved project root has a git remote, derive the slug from it.
+#   3. Otherwise use the basename of the resolved project root.
+#   4. If no project root was found anywhere on the chain, fall back to the
+#      basename of $(pwd) (preserves prior behavior for plain folders).
+#
+# Caching is self-healing: a cache entry for the literal pwd that differs from
+# the freshly-computed slug gets opportunistically rewritten (single-shot, key-
+# local — never sweeps other entries). This lets pre-existing poisoned caches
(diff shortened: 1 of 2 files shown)


jbetala7:
Thanks for writing up the detailed slug repro. I think this branch currently overlaps #1125/#1621 and does not preserve the main #1125 failure mode.

Issue #1125's core repro is:

1. `/Users/me/workspace` is a git repo with `origin=me/workspace`.
2. `/Users/me/workspace/IoTopia` is a standalone project directory with no `.git` of its own.
3. Running `gstack-slug` from `IoTopia` should produce the child project slug, not `me-workspace`.

In this PR, `_outermost_project_root` keeps the outermost strong marker. If the parent workspace has `.git`, `PROJECT_ROOT` becomes the parent workspace; then step 2 runs `git -C "$PROJECT_ROOT" remote get-url origin` and derives the parent repo slug. That appears to reintroduce the exact parent-repo inheritance bug from #1125.

It also misses the related issue-comment case where a `.gstack-slug` override at the project root should beat a nested theme repo's own remote. The linked canonical PR #1621 covers both behaviors: an override beats cache/git inference, and a parent repo remote is ignored unless the current directory is actually the git top-level.

Recommended path: either rebase on #1621 / carry its tests, or add explicit regressions for:

1. parent git repo + child project dir with no `.git` => child basename, not parent remote;
2. parent `.gstack-slug` override beats a nested child git remote;
3. stale cache is refreshed after current inference or override.

The weak-marker/README behavior in this PR can still be useful as an additive case, but it should not make outer parent git remotes canonical for standalone child projects.

ajeenkya:
Hey Garry, bumping this one. Just rebased onto latest main (v1.51), so it's clean and mergeable.

What it fixes: gstack-slug derives the slug from the literal cwd with no walk-up. Run any gstack command from a subdirectory that has its own marker (a .vercel/ left by `vercel --prod`, a nested package.json, etc.) and the whole session's state (checkpoints, learnings, analytics) silently files under a phantom slug. No error, and the bad cache persists across sessions. It cost me 3 misfiled checkpoints on a real project before I caught it.

The approach: walk up to the outermost canonical-identity marker using an allow-list (.git, .project.yaml, package.json, pyproject.toml, Cargo.toml, Gemfile, go.mod), self-heal the one poisoned cache entry on next run, and add a GSTACK_PROJECT_SLUG override for vendored sub-repos. Allow-list rather than blacklist, so new tooling can drop its own markers without creating phantom roots. Test included (test/gstack-slug-cwd-walk-up.test.ts).

Happy to reshape the marker list or anything else if you'd rather it land differently.

jbetala7:
Thanks for rebasing this. I rechecked the current diff after the latest bump, and the earlier blocker still appears to be present.

Current head still resolves `PROJECT_ROOT=$(_outermost_project_root "$PROJECT_DIR")`, then derives the slug with `git -C "$PROJECT_ROOT" remote get-url origin`. Because `_outermost_project_root` prefers the outermost strong marker, a standalone child directory under a parent git repo still resolves to the parent repo root and parent remote.

That conflicts with #1125/#1621's core case: `workspace/.git` with `workspace/IoTopia/` as a standalone child project should return the child basename/override, not the parent `workspace` remote slug. Rebase alone does not change that behavior.

The self-healing cache and weak-marker cases here may still be useful, but I think this branch needs the #1125 regressions before merge:

1. parent git repo + child project dir with no `.git` returns child basename, not parent remote;
2. ancestor `.gstack-slug` override beats nested git/cache inference;
3. stale cache is refreshed only after current inference/override is evaluated.

Without those, this can fix `.vercel` phantom roots while reintroducing the parent-repo inheritance bug.

time-attack:
might incorrectly select a parent workspaces git repository and identity. also we have a new design so @16francej clsoe pls

PR #1704, opened 2026-05-26
fix: preserve gstack env var resolver paths

o1092877777:
## Summary
- Preserve env-var based browse/design resolver paths for Hermes and other env-var hosts
- Avoid generating broken paths like `$HOME$GSTACK_BROWSE` and `$HOME$GSTACK_DESIGN`
- Add regression coverage for Hermes generated setup snippets

## Test Plan
- `bun test test/host-config.test.ts -t "env-var path resolver regression"`
- `bun run gen:skill-docs --host hermes`
- `bun run build`

Note: the full host-config test file has an existing environment-sensitive assertion that expects `claude` on PATH; in this environment detection returned codex/openclaw/hermes/gbrain instead. The new regression tests pass.

### DIFF
--- a/scripts/resolvers/browse.ts
+++ b/scripts/resolvers/browse.ts
@@ -100,13 +100,17 @@ export function generateSnapshotFlags(_ctx: TemplateContext): string {
 }
 
 export function generateBrowseSetup(ctx: TemplateContext): string {
+  const globalBrowse = ctx.paths.browseDir.startsWith('~')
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej — recommend closing this PR because #2056 supersedes it across all resolver call sites. This PR conflicts with that work and omits make-pdf, while #2056 has already passed its checks. Preserve and credit this PR’s direct environment-path assertions in #2056.

PR #1713, opened 2026-05-26
fix(ci): avoid piping Bun installer into shell

### LINKED ISSUE #1706: Avoid piping remote Bun installer directly into CI shell
## Problem

The GitLab CI setup currently pipes the remote Bun installer directly into a shell. A downstream patch adjusts `.gitlab-ci.yml` to avoid the direct `curl | bash` pattern.

## Expected behavior

CI should avoid piping remote installer output directly into the shell where a small safer alternative is available, such as downloading to a file, optionally inspecting/checking it, and then executing it.

## Current downstream patch

A downstream patch changes `.gitlab-ci.yml` from a direct pipe to a safer multi-step installer flow.

spacegeologist:
Fixes #1706.

## Reviewer Guide

Small GitLab CI hardening patch. Review targets:

- `.gitlab-ci.yml` — replaces the direct `curl | bash` Bun installer with a download-first flow.
- `test/gitlab-ci-safety.test.ts` — adds a static regression check so `bun.sh/install` is not piped directly into `bash` or `sh` again.

## Why

The GitLab CI Bun setup step piped the remote installer straight into a shell:

```sh
curl -fsSL https://bun.sh/install | bash -s "bun-v$BUN_VERSION"
```

This PR keeps the existing `BUN_VERSION` behavior but changes the flow to:

1. download the installer to a `mktemp` file
2. verify the downloaded file is non-empty
3. execute that file with the pinned Bun version
4. clean it up via `trap`

No GitHub Actions, release/version, or job-rule behavior changes are intended.

## Verification

- `bun test test/gitlab-ci-safety.test.ts test/audit-compliance.test.ts` — 10 pass, 0 fail
- `git diff --check`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,12 @@ variables:
 
 .setup-bun: &setup-bun
   - apt-get update -qq && apt-get install -qq -y curl jq git
-  - curl -fsSL https://bun.sh/install | bash -s "bun-v$BUN_VERSION"
+  - |
+    BUN_INSTALL_SCRIPT="$(mktemp)"
+    trap 'rm -f "$BUN_INSTALL_SCRIPT"' EXIT
+    curl -fsSL https://bun.sh/install -o "$BUN_INSTALL_SCRIPT"
+    test -s "$BUN_INSTALL_SCRIPT"
+    bash "$BUN_INSTALL_SCRIPT" "bun-v$BUN_VERSION"
   - export PATH="$HOME/.bun/bin:$PATH"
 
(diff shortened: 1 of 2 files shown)


spacegeologist:
All eval jobs passed. The only failing check is the E2E report comment step failing with Resource not accessible by integration (addComment), likely because this fork PR only gets a read-only GitHub token.

time-attack:
@16francej Merge this focused change because its tests pass, and the remaining reporting failure comes from fork comment permissions rather than the implementation.

PR #1717, opened 2026-05-26
feat(scrape): evidence gate before returning prototype JSON

orendi84:
## What

`/scrape`'s prototype path assembles JSON and returns it with no positive check that the data actually satisfies the user's intent. The "don't write a partial result and call it done" guidance was prose-only — there was no step that decomposes the intent and backs each requirement with evidence.

This adds **Step 5 — Verify the prototype result against the intent**:

- Decompose the intent into checkpoints (each datum, field, filter, count, ranking).
- Back each checkpoint with evidence: a populated field in the result, or a page capture (`$B snapshot --text` for a value, `$B screenshot` for a visible state). Ranking and filter checkpoints must be grounded in the page's own order or controls, not the agent's re-sorting of the rows.
- An unbacked checkpoint is a prototype failure (routes to the existing "When the prototype fails" handling), never partial-data-as-success.

The match path gets a lighter sanity check: empty, malformed, or off-intent skill output falls through to the prototype path instead of being emitted.

## Why

Plausible-looking JSON is the most common silent failure for an extraction agent — names present but prices null, a "top 10" that is really DOM order, a filter that was never applied. The gate forces evidence before the result is treated as the answer.

## Testing

- `bun test test/skill-parser.test.ts test/skill-validation.test.ts test/gen-skill-docs.test.ts` → 731 pass / 0 fail (includes all-host doc-freshness).
- Template-only change; `scrape/SKILL.md` regenerated with `bun run gen:skill-docs --host all`.
- No E2E eval added — the change is prompt-level. Happy to add a `skill-e2e` case if you want one.

### DIFF
--- a/scrape/SKILL.md
+++ b/scrape/SKILL.md
@@ -823,7 +823,11 @@ zero-arg skills) and run:
 $B skill run <name> [--arg key=value ...]
 ```
 
-Emit the JSON the skill prints to stdout. Stop.
+Emit the JSON the skill prints to stdout — but sanity-check it against the
+intent first. A matched skill can return stale or empty data; if the result
+is empty, malformed, or clearly does not cover what the intent asked for,
+the skill is likely out of date, so fall through to the prototype path
+instead of emitting it. Otherwise emit it and stop.
 
 If matching is ambiguous (two skills could plausibly fit), pick the
 narrower-tier one (project > global > bundled — `$B skill list` shows the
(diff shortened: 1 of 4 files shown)


orendi84:
CI note: the red `evals.yml` jobs are the fork-secrets limitation, not this change. Fork PRs don't receive the base-repo `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`, so the paid E2E/eval jobs fail with empty-env auth regardless of content. The `skill-docs` freshness check that actually covers this PR passes. The added `scrape-evidence-gate` E2E is periodic-tier and needs a base-repo run with secrets to execute.

PR #1722, opened 2026-05-26
feat(test): bash syntax validation for all SKILL.md code blocks (#1667)

### LINKED ISSUE #1667: feat: validate bash syntax in SKILL.md code blocks during bun test (prevent silent runtime breakage)

NikhileshNanduri:
## Summary

Closes #1667.

- Adds `test/skill-bash-syntax.test.ts`: gate-tier free test that runs `bash -n` on every ` ```bash ``` ` block in all 53 generated SKILL.md files (1327+ blocks total), catching syntax errors before they reach users
- Fixes the first bug the test caught: `codex/SKILL.md` had a missing closing ` ``` ` fence on the resumed-session bash block, causing prose with backtick-quoted identifiers to be parsed as shell — `bash -n` reported "unexpected EOF while looking for matching \`"
- Angle-bracket placeholders like `<branch-name>` are preprocessed to `PLACEHOLDER` before validation to avoid false positives on intentional template tokens

## Why it matters

`bun test` previously never validated bash syntax in skill files. A typo like `2/dev/null` instead of `2>/dev/null` (the exact corruption in PR #1654) would pass all CI checks and silently break at runtime. Now it's a gate failure in ~4 seconds with zero API cost.

## Test plan

- [x] `bun test test/skill-bash-syntax.test.ts` — 54 tests pass (1 discovery + 53 skills)
- [x] `bun test test/skill-validation.test.ts test/gen-skill-docs.test.ts test/skill-bash-syntax.test.ts` — 772 tests pass
- [x] Full free suite passes


jbetala7:
Thanks for tackling #1667. I think this PR currently collides with #1697 and is hard to land as-is.

Live map evidence:
- #1697 was opened earlier, also fixes #1667, and adds a `bash -n` validation path for skill bash snippets.
- This PR includes the same user-visible validation goal, but the diff also carries unrelated release/feature work: `VERSION`, `CHANGELOG.md`, `package.json`, `AGENTS.md`, `docs/skills.md`, `plan-pm-review/SKILL.md(.tmpl)`, and `/autoplan` wiring.
- GitHub currently marks this branch `DIRTY`, so landing it would require separating the #1667 fix from older release-wave changes.

Recommended path: make #1697 the canonical PR for the bash-snippet validation gate, or rebase/narrow this PR down to only the #1667 pieces:

1. the `codex/SKILL.md.tmpl` fence fix,
2. the bash syntax validation test or equivalent,
3. no `VERSION`/`CHANGELOG`/`plan-pm-review`/`autoplan` changes.

The important coverage to preserve is a regression that catches both the missing Codex fence and the `2>/dev/null` -> `2/dev/null` corruption pattern from #1654.

time-attack:
Thanks for this. The extractor misses unclosed final fences, and `bash -n` alone will not catch the motivating `2/dev/null` typo; the branch also includes stale release metadata. Rebase without release changes, detect unmatched fences, and add explicit regressions for malformed shell, heredocs, and redirection typos.

PR #1736, opened 2026-05-27
fix: add destructive-action guards to gstack-gbrain-sync orchestrator

### LINKED ISSUE #1734: `/sync-gbrain` can trigger gbrain's destructive auto-recovery, wiping user repos
# `/sync-gbrain` can trigger gbrain's destructive auto-recovery, wiping user repos

**Severity:** Critical / data loss (downstream of gbrain bug)
**Repo:** garrytan/gstack
**Affected version:** v1.45.0.0 (orchestrator at `~/.claude/skills/gstack/bin/gstack-gbrain-sync.ts`, mtime 2026-05-23)
**Dependency:** gbrain v0.41.14.0 (the destructive path lives there, see linked issue)

## Summary

The `/sync-gbrain` skill orchestrator (`gstack-gbrain-sync.ts`) wipes user working trees under a specific combination of conditions: (a) a stale legacy source row with `config.remote_url` set exists in the gbrain DB, (b) the gbrain `autopilot` daemon is running, and (c) the orchestrator runs its hostname-fold migration concurrently with autopilot's sync cycle. Recovery requires a fresh `git clone` and any uncommitted work on un-pushed branches is lost.

I lost the working tree to this bug on 2026-05-27 — `.git/`, `backend/`, `docs/`, `frontend/*`, and root config files all wiped. GitHub remote was intact, recovered via fresh clone.

This issue is the gstack-side companion to the upstream gbrain data-loss bug. Even after gbrain fixes its destructive paths, gstack should defend itself by avoiding calls into known-unsafe gbrain operations and by detecting/halting concurrent autopilot.

## Reproduction sequence

1. Have gbrain v0.41.14.0 installed and a source previously registered with `remote_url` in config (e.g., from an older `/sync-gbrain` that used a URL-based registration flow).
2. Have `gbrain autopilot` running (it can be inadvertently started by `gbrain autopilot status` — see linked issue).
3. Run `/sync-gbrain` in a Claude Code session inside an indexed repo.
4. The orchestrator computes a new source ID (hostname-fold migration) and calls `gbrain sources remove <legacy-id> --confirm-destructive` (line 484 / 664 of `gstack-gbrain-sync.ts`).
5. Concurrently, autopilot's sync cycle picks up the same source and calls `gbrain sync --strategy code` → `validateRepoState` → `recloneIfMissing` → ungated `rm -rf src.local_path`.
6. The user's working tree is destroyed mid-orchestrator. The orchestrator's later `brain-sync` stage exits with `status undefined`. **Note on evidence weight:** the brain-sync spawn failure has an independent cause (#1731 — `spawnSync` missing `shell: true` on Windows), so it can't be used as direct evidence that destruction happened. The destructive action is evidenced by the survivor pattern (selectively deleted tracked files plus locked-dir partial-failure remnants) and the transcript timeline, not by brain-sync's failure.

Forensic transcript can be attached if needed.

## What the orchestrator does that contributes to the risk

`gstack-gbrain-sync.ts` runs these gbrain subcommands in code-stage order:

```typescript
spawnGbrain(["sources", "rename", legacyId, newSourceId], ...)           // line 463
spawnGbrain(["sources", "remove", legacyId, "--confirm-destructive"], ...)  // line 484 / 664
spawnGbrain(["sync", "--strategy", "code", "--source", sourceId], ...)   // line 723
spawnGbrain(["reindex-code", "--source", sourceId, "--yes"], ...)        // line 741
spawnGbrain(["sources", "attach", sourceId], ...)                        // line 768
```

Two of these reach destructive code in gbrain:
- `sources remove --confirm-destructive` calls `removeSource()` in gbrain's sources-ops.ts:510, which contains `rmSync(src.local_path, …)` at line 570. That site is gated by `isPathContained(local_path, ~/.gbrain/clones)` — which happens to fail closed on Windows for `E:\` paths due to a separator-mismatch bug in `isPathContained`. On macOS/Linux, this gate could fire if the legacy source's `local_path` was inside `~/.gbrain/clones/`.
- `sync --strategy code` reaches `recloneIfMissing()` in gbrain's sources-ops.ts:645, which contains an **ungated** `rm -rf src.local_path` at line 674.

The orchestrator has no protection against either path: no `--keep-storage` flag passed to `sources remove`, no pre-flight check that the source's `remote_url` isn't set, no detection of running autopilot.

## What gstack should do, even before gbrain ships its fix

**1. Refuse to operate when autopilot is running.** Before the code stage, check `~/.gbrain/autopilot.lock`:
```typescript
if (existsSync(join(gbrainHome, "autopilot.lock"))) {
  return stageError("autopilot running; refusing to run /sync-gbrain concurrently. Stop autopilot first with: kill $(cat ~/.gbrain/autopilot.lock) && rm ~/.gbrain/autopilot.lock");
}
```

**2. Audit legacy source rows before any hostname-fold migration.** Before calling `gbrain sources remove --confirm-destructive`, query the DB for the legacy source's full config. If `config.remote_url` is set AND `local_path` is not inside `~/.gbrain/clones/`, refuse the remove and surface a manual-fix instruction.

**3. Pass `--keep-storage` to all `gbrain sources remove` calls** in the orchestrator. The orchestrator never wants gbrain to touch the source repo's files — it only wants the DB row gone. The flag exists; pass it.

**4. Detect and short-circuit if `recloneIfMissing` would be reached.** Before calling `gbrain sync --strategy code`, query the source row's config. If `remote_url` is set, surface a warning ("source is URL-managed; sync may auto-reclone") and require an explicit `--allow-reclone` flag from the user to proceed.

**5. Update the user-facing `/sync-gbrain` skill description and CLAUDE.md guidance** to warn that this skill should not be run while autopilot is active, and to recommend `gbrain sources add --path` (no `--url`) for user-managed repos.

## Why this isn't only gbrain's problem to fix

Even with gbrain's `recloneIfMissing` patched, the orchestrator's pattern of `sources remove --confirm-destructive` + concurrent autopilot + race conditions is structurally risky. gstack should treat gbrain's destructive APIs the way one would treat `rm -rf`: opt-in flags, pre-flight checks, explicit user confirmation, and refuse-to-operate guards. Right now the orchestrator calls them like they're safe.

## Cross-references

- **Upstream: garrytan/gbrain#1526** — `recloneIfMissing` ungated rm-rf (the actual destructive code).
- **Upstream: garrytan/gbrain#1525** — `gbrain autopilot status` UX trap (how autopilot inadvertently starts).
- **Adjacent: #1731** — "Windows: `gstack-gbrain-sync.ts` stage 3 fails to spawn `gstack-brain-sync` (`spawnSync` needs `shell: true`)". This explains the `brain-sync exited undefined` symptom observed in this incident's transcript independently of destruction. The two bugs are present simultaneously on Windows; fixing one doesn't fix the other.
- **Adjacent: #1576** — "/sync-gbrain crashes in sourceLocalPath: gbrain sources list returns wrapped object, not array". Different crash, same orchestrator file.
- **Adjacent: #1611** — "Memory-ingest has hardcoded 35-min timeout that SIGTERMs `/sync-gbrain --full` on big brains". Different bug, same orchestrator + same Windows context.

## Suggested release-note language for the gstack fix

> `/sync-gbrain` now refuses to run when `gbrain autopilot` is active, audits source-row configs before any destructive gbrain operation, and passes `--keep-storage` to `gbrain sources remove`. This closes a class of data-loss bugs where the orchestrator could race autopilot into rm-rf'ing the indexed working tree on Windows. If you previously ran `/sync-gbrain` and aren't sure whether your indexed repos are safe, audit: `psql -d <brain> -c "SELECT id, local_path, config->>'remote_url' FROM sources"` — any row with non-null `remote_url` is at risk.

mvanhorn:
## Summary

Adds destructive-action guards to `gstack-gbrain-sync` so accidental wide deletes are caught before they run. The orchestrator now warns and prompts when a sync would delete more than the configured threshold of files in a single pass.

## Why this matters

Issue #1734 reported that an unattended sync could silently nuke a large fraction of the brain when a source side glitched (network blip, partial repo clone). The orchestrator had no safety rails: any diff producing a delete list was applied wholesale. The fix adds bounded confirmation: small deletes auto-apply (no behavior change), medium deletes warn with an opt-out flag, large deletes require explicit `--force` or a confirmation prompt.

## Changes

- New guard layer in `bin/gstack-gbrain-sync.ts` reads `BRAIN_SYNC_DELETE_THRESHOLD` (default 25) before applying any delete batch
- Below threshold: behavior unchanged (auto-apply)
- Above threshold without `--force`: surface the delete plan and abort
- New env override and a `--force` CLI flag for power users who want to override interactively

## Testing

Added `test/gbrain-sync-guards.test.ts` covering the four guard paths: under-threshold auto-apply, above-threshold abort, `--force` override, env-var override. 6/6 tests pass.

Fixes #1734

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -30,7 +30,7 @@
  */
 
 import { existsSync, statSync, mkdirSync, writeFileSync, readFileSync, unlinkSync, renameSync } from "fs";
-import { join, dirname } from "path";
+import { join, dirname, resolve, relative, isAbsolute } from "path";
 import { execSync, spawnSync } from "child_process";
 import { homedir, hostname } from "os";
 import { createHash } from "crypto";
@@ -52,6 +52,7 @@ interface CliArgs {
   noMemory: boolean;
   noBrainSync: boolean;
   codeOnly: boolean;
+  allowReclone: boolean;
 }
 
 interface CodeStageDetail {
@@ -62,7 +63,7 @@ interface CodeStageDetail {
   status?: "ok" | "skipped" | "failed";
 }
 
-interface StageResult {
+export interface StageResult {
   name: string;
   ran: boolean;
   ok: boolean;
@@ -205,6 +206,7 @@ Options:
   --no-memory          Skip the gstack-memory-ingest stage (transcripts + artifacts).
   --no-brain-sync      Skip the gstack-brain-sync git pipeline stage.
   --code-only          Only run the code-import stage (alias for --no-memory --no-brain-sync).
+  --allow-reclone      Allow code sync when the source row has remote_url set.
   --help               This text.
 
 Stages run in order: code → memory ingest → curated git push.
@@ -220,6 +222,7 @@ function parseArgs(): CliArgs {
   let noMemory = false;
   let noBrainSync = false;
   let codeOnly = false;
+  let allowReclone = false;
 
   for (let i = 0; i < args.length; i++) {
     const a = args[i];
@@ -231,6 +234,7 @@ function parseArgs(): CliArgs {
       case "--no-code": noCode = true; break;
       case "--no-memory": noMemory = true; break;
       case "--no-brain-sync": noBrainSync = true; break;
+      case "--allow-reclone": allowReclone = true; break;
       case "--code-only":
         codeOnly = true;
         noMemory = true;
@@ -247,7 +251,7 @@ function parseArgs(): CliArgs {
     }
   }
 
-  return { mode, quiet, noCode, noMemory, noBrainSync, codeOnly };
+  return { mode, quiet, noCode, noMemory, noBrainSync, codeOnly, allowReclone };
 }
 
 // ── Helpers ────────────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


16francej:

Issue #973, opened 2026-04-11
spec-review skills are structurally blind to hallucinations in file paths, math, and CLI flags


Akagilnc:
## Problem

`/office-hours` and `/plan-*` spec-review skills score design docs 8/10 on documents that contain CRITICAL fact errors — wrong file paths, wrong math, fabricated CLI flags, "reuse X module" claims that contradict hardcoded implementation. Spec-review subagents see only document text. They can't grep source, compute math, or shell out `<tool> --help`.

## Evidence from one session

2026-04-11, ai-blogger-lab project. A design doc passed two spec-review rounds at 8.0/10, reached `Status: APPROVED`. A manually-triggered post-approval audit against real code / npm / math found **10 hallucinations**, including:

- **CRITICAL**: `Beta(1,21) upper CI = 0.143` — fabricated. Actual `scipy.stats.beta.ppf(0.975, 1, 21) = 0.1611`. The pivot rule (fire when upper CI < 0.15) would not trigger until `Beta(1,23) = 22 silences`, not `Beta(1,21) = 20 silences`. Two errors compounded: wrong number *and* wrong trigger threshold. Would have made the L3 test suite red on first implementation.
- **High**: Claimed `post-decider` was a "thin persona-neutral passthrough". Actual: hardcoded Chinese prompts + `["screenshot","text_card"]` enum.
- **High**: Claimed `piece-critique` was persona-neutral. Actual: every heuristic is CJK-specific (character-count thresholds, `[\u4e00-\u9fa5]` regex).
- **Medium**: `refactor run-tick.ts` — wrong target. `run-tick.ts` is already DI'd; the wiring is in `cli.ts:runSingleTick`.
- **Low**: `--panels=1` CLI flag on `comic-compositor.py` — the script has no CLI flags, panel count comes from stdin JSON.

Post-approval audit is the canonical recovery pattern (`Status:` downgraded to `APPROVED_WITH_KNOWN_CORRECTIONS`, `## Post-Approval Hallucination Audit` section appended). But catching these errors *before* implementation starts would save an entire day of wrong-code-on-wrong-doc.

[...]

PR #1738, opened 2026-05-27
fix: make /office-hours verify design-doc premises against the codebase

### LINKED ISSUE #1644: /office-hours: design doc passes review without verifying premises against code (incident + 3 fixes)
## Incident

Used `/office-hours` to design an admin UI refactor for a Next.js + Supabase app
(content authoring tool, ~3 admin pages affected). Design doc reached **APPROVED**
status with a 9/10 newcomer-readability self-eval. The doc proposed a UI-only
refactor with explicit premise: "no schema migration needed".

Then ran `/codex consult` adversarial on the doc. Codex (which actually reads
files) returned **15 findings: 7 P1 critical + 8 P2 advisory**, invalidating
3 of the doc's 7 premises:

- Premise "no schema migration needed" — false. A DB trigger repopulates the
  new canonical column from legacy columns whenever the new column is set
  to NULL. The unpublish server action as designed is a silent no-op.
- Premise "legacy column is orphan" — false. Existing `actions.ts` still
  inserts and updates the legacy column. The reconciliation work was
  understated.
- Change "Move to Draft" button on lesson edit page — semantically wrong.
  The doc explicitly keeps child entities as Ready, so the lesson returns
  to **Ready state**, not Draft. The new button label recreates the exact
  ambiguity the design was meant to eliminate.

Effort estimate moved from ~6-7h CC to ~10-13h CC after Codex pass.
The doc was 9/10 narrative-quality but 4/10 implementation-readiness.

The skill itself did NOT catch this gap — three structural Phase issues below.

## Root cause

Three Phase methodology gaps let me ship a doc that flew away from codebase reality:

### Gap 1 — Phase 1 Context Gathering doesn't require migration/RLS map

Phase 1 reads `CLAUDE.md`, `git log`, lists prior design docs. Does NOT require
mapping migrations, RLS policies, or DB triggers relevant to the area being
designed. Result: premises get formulated against memory/CLAUDE.md narrative,
not against actual schema + policies. For any refactor touching DB
visibility/state, this is the gap that produces "no schema change needed"
claims that turn out to be false.

### Gap 2 — Phase 5 Spec Review Loop subagent prompt is anti-rigor

The Spec Review Loop dispatches a subagent reviewer with this instruction
(from the skill's own prompt template):

> "Codebase paths to verify mentally (**do NOT read them yourself**, just
> check consistency within the doc)"

This is exactly the wrong direction. It forbids the one action that would
catch the doc-vs-reality drift. The reviewer can only flag internal
inconsistency (doc says X here and Y there), not external (doc says X
but the codebase says Z). In my incident the reviewer returned PASS
with quality 7/10 while Codex, reading the same files, found 7 P1 in the
same doc.

### Gap 3 — Phase 3 Premise Challenge asks user instead of code

Premise Challenge presents premises to the user via AskUserQuestion and
asks them to confirm/reject. But the user is rarely the source of truth
on existing codebase state — for "this DB column is orphan", "this RLS
policy hides X", "this trigger doesn't fire here", the source of truth
is the migrations/policies/types, not the user's recollection.

In my incident, the user confirmed all 7 premises because they trusted
my interpretation. But my interpretation was assumed, not verified. The
result was recursive trust without verification.

### Fix 1 — Phase 1: add "Codebase Surface Map" step

Before Phase 2 questioning, add explicit step that maps the area being
designed:

- For schema-touching designs: list relevant migrations (most recent N
  touching affected tables), read them, summarize triggers + constraints
  + check columns.
- For visibility/auth-touching designs: read RLS policies for affected tables.
- For server-action designs: list and read existing server actions in scope.

Output: a "Codebase Surface" section the design doc, which any premise can
be checked against.

### Fix 2 — Phase 5 Spec Review Loop: REVERSE the subagent prompt instruction

Change subagent prompt from:

> "do NOT read them yourself, just check consistency within the doc"

to:

> "READ the referenced source files. Verify each premise in the design
> doc against the actual code. Flag any premise that the code contradicts.
> Pay special attention to: schema columns that the doc claims are unused,
> RLS policies the doc claims gate something, triggers that might fire on
> the proposed UPDATEs, server actions still writing to columns the doc
> claims are orphan."

This turns the Spec Review from a doc-internal-consistency check into a
doc-vs-reality check, which is the actual quality dimension that matters.

### Fix 3 — Phase 3 Premise Challenge: verify premise candidates against code first

Before asking the user to confirm a premise, the model should verify it in
code:

- Premise about schema (column orphan, table unused): grep for write paths.
- Premise about RLS (policy hides X): read the policy.
- Premise about behavior (X doesn't fire when Y): trace the code path or
  run a small probe.

Only premises that survive code verification get presented to the user.
The user confirms the **scope** decision (do we want this?), not the
**factual** assertion (is this true today?). Today the two are conflated
in a single AskUserQuestion.

## Why this matters

The current skill produces design docs that look 9/10 to a newcomer but
ship with implementation-time landmines that double the effort estimate
post-Codex. For solo developers without a cross-model habit (Codex consult,
adversarial review), these landmines explode during PR review or production
incidents.

The fix is mechanical: tighten Phase 1 + Phase 3 + Phase 5 to require
code-grounded premises. Cost ~30 min more CC per session, saves ~6h of
refactor that the wrong premises cause downstream.

Happy to PR the SKILL.md changes if useful — let me know preferred direction
(separate PRs per Phase fix, or one PR with all three).

---

Reported by raffoz, captured during a real `/office-hours` → `/codex consult` flow.

mvanhorn:
## Summary

Updates `/office-hours` so it verifies design-doc premises against the codebase before producing review output. Previously the skill could green-light a doc that named files/functions that did not actually exist.

## Why this matters

Issue #1644 reported three concrete cases where /office-hours signed off on design docs whose premises were stale: (1) a doc referencing a function that had been renamed two releases ago, (2) a doc relying on a flag that had been removed, (3) a doc claiming an architectural property that was no longer true. Each case wasted reviewer time because the LLM happily reasoned about premises that were already wrong.

## Changes

- `office-hours/SKILL.md.tmpl` adds a Premise Verification step that grep/codesearch checks every named file path, function, class, flag, and module against the active workspace
- New verification output is appended to the review report under a `## Premise Verification` section
- Three gaps documented: file existence, symbol existence, and feature-flag presence

## Testing

Added `test/office-hours-premise-verification.test.ts`. 3/3 pass.

Fixes #1644

### DIFF
--- a/office-hours/SKILL.md
+++ b/office-hours/SKILL.md
@@ -781,7 +781,12 @@ eval "$(~/.claude/skills/gstack/bin/gstack-slug 2>/dev/null)"
 1. Read `CLAUDE.md`, `TODOS.md` (if they exist).
 2. Run `git log --oneline -30` and `git diff origin/main --stat 2>/dev/null` to understand recent context.
 3. Use Grep/Glob to map the codebase areas most relevant to the user's request.
-4. **List existing design docs for this project:**
+4. **Codebase Surface Map (conditional):** If the design is schema-touching, visibility/auth-touching, or server-action-touching, verify the relevant codebase surface before Phase 2:
+   - **Schema-touching:** list recent migrations touching affected tables; summarize triggers, constraints, check columns, and relevant legacy columns.
+   - **Visibility/auth-touching:** read RLS policies and authorization checks for affected tables/resources; summarize what users can see or mutate today.
+   - **Server-action-touching:** list existing server actions, route handlers, jobs, or mutation flows in scope; summarize what they already write/read.
+   Output a `Codebase Surface` note with file paths and evidence. If none of these triggers apply, explicitly say `Codebase Surface: not applicable`.
+5. **List existing design docs for this project:**
    ```bash
    setopt +o nomatch 2>/dev/null || true  # zsh compat
    ls -t ~/.gstack/projects/$SLUG/*-design-*.md 2>/dev/null
@@ -826,7 +831,7 @@ matches a past learning, display:
 This makes the compounding visible. The user should see that gstack is getting
 smarter on their codebase over time.
 
-5. **Ask: what's your goal with this?** This is a real question, not a formality. The answer determines everything about how the session runs.
+6. **Ask: what's your goal with this?** This is a real question, not a formality. The answer determines everything about how the session runs.
 
    Via AskUserQuestion, ask:
 
@@ -843,7 +848,7 @@ smarter on their codebase over time.
    - Startup, intrapreneurship → **Startup mode** (Phase 2A)
    - Hackathon, open source, research, learning, having fun → **Builder mode** (Phase 2B)
 
-6. **Assess product stage** (only for startup/intrapreneurship modes):
+7. **Assess product stage** (only for startup/intrapreneurship modes):
    - Pre-product (idea stage, no users yet)
    - Has users (people using it, not yet paying)
    - Has paying customers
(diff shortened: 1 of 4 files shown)


time-attack:
@16francej — recommend closing this PR because #973 supersedes this older implementation and owns the current office-hours work. It conflicts with the newer office-hours source layout, and its tests do not establish grounded codebase reads. Please credit its Codebase Surface idea and carry the grounded-read requirements into #973.

mvanhorn:
Closing — superseded by #973. Thanks @time-attack!

PR #1746, opened 2026-05-27
fix(learnings): fail closed when cross-project learning lacks trusted field

### LINKED ISSUE #1745: gstack-learnings-search cross-project trust gate fails open for rows missing the `trusted` field
severity: security (defense-in-depth)

## Observed problem

`bin/gstack-learnings-search --cross-project` is meant to be an allowlist: per the inline comment, cross-project learnings are "only loaded if trusted (user-stated)" to prevent prompt injection from one project's AI-generated learnings silently influencing reviews in another project.

The gate is implemented as a denylist instead:

```js
if (isCrossProject && e.trusted === false) continue;
```

Any cross-project row where `trusted` is missing/`undefined` (not literally `false`) is admitted, because `undefined === false` is `false`. So legacy rows written before the `trusted` field existed, hand-edited rows, and rows produced by other tools all bypass the gate.

The `trusted` field was added in the security wave 3 (#988, 2026-04-13), while `gstack-learnings-search` shipped earlier (#622, 2026-03-29). Any `learnings.jsonl` written in that window — or by anything that does not set the field — has rows with no `trusted` key. Those rows leak across projects today.

## Current behavior on upstream `main` (a6fb3172)

A foreign project row with no `trusted` field and `source: observed` is loaded and shown as `[cross-project]`:

```
$ # foreign learnings.jsonl row, no trusted field:
$ # {"ts":"2026-05-09T00:00:00Z","type":"pattern","key":"foreign-legacy","insight":"INJECTED legacy guidance with no trusted field","source":"observed"}
$ gstack-learnings-search --cross-project --query INJECTED
LEARNINGS: 1 loaded (1 pattern)

## Patterns
- [foreign-legacy] (confidence: 5/10, observed, 2026-05-09) [cross-project]
  INJECTED legacy guidance with no trusted field
```

## Expected behavior

The gate should fail closed: a cross-project row is admitted only when `trusted === true`. Rows missing the field are treated as untrusted. Current-format rows are unaffected (`user-stated` rows carry `trusted: true` and still load; everything else carries `trusted: false` and is still excluded).

- Open PRs touching `bin/gstack-learnings-search` / `bin/gstack-learnings-log`: none target the trust-gate semantics. #940 (stale, 2026-04-09) normalizes legacy `category/summary/detail` field names, not the trust allowlist; #1514 is about configurable search limits.
- Issues mentioning learnings trust / cross-project: #1618 (closed, the cross-project tagging bug, shipped via v1.43.2.0) — different bug.

## Candidate fix shape

Change the cross-project trust check from a denylist (`e.trusted === false`) to an allowlist (`e.trusted !== true`), and add a regression test for a foreign row with no `trusted` field.

jbetala7:
Fixes #1745

## Summary
- treat a cross-project learnings row as untrusted unless `trusted === true`
- add a regression test for a foreign row written with no `trusted` field

## Root cause
`gstack-learnings-search --cross-project` documents an allowlist ("cross-project learnings only loaded if trusted (user-stated)") to keep one project's AI-generated learnings from influencing reviews in another. The gate was a denylist:

```js
if (isCrossProject && e.trusted === false) continue;
```

`undefined === false` is `false`, so any cross-project row missing the `trusted` field was admitted. The field was introduced in security wave 3 (#988, 2026-04-13); `gstack-learnings-search` shipped earlier (#622). Rows written before the field existed — plus hand-edited rows and rows from other tools — have no `trusted` key and leak across projects today.

## Fix
Switch the gate to an allowlist (`e.trusted !== true`). Current-format rows are unaffected: `user-stated` rows carry `trusted: true` and still load; everything else carries `trusted: false` and is still excluded. Only rows missing the field change behavior, and they now fail closed.

## Testing
- `bun test test/gstack-learnings-search.test.ts` (6 pass; the new test fails on the pre-fix binary, confirming it guards the regression)
- `bun test test/learnings.test.ts test/learnings-injection.test.ts` (30 pass)
- direct repro on a clean GSTACK_HOME: a foreign `observed` row with no `trusted` field is no longer admitted; a foreign `user-stated`/`trusted: true` row still is
- `git diff --check` clean; `git merge-tree` clean against `origin/main`

### DIFF
--- a/bin/gstack-learnings-search
+++ b/bin/gstack-learnings-search
@@ -90,10 +90,13 @@ for (const taggedLine of lines) {
     const isCrossProject = sourceTag === 'cross';
     e._crossProject = isCrossProject;
 
-    // Trust gate: cross-project learnings only loaded if trusted (user-stated)
-    // This prevents prompt injection from one project's AI-generated learnings
-    // silently influencing reviews in another project.
-    if (isCrossProject && e.trusted === false) continue;
+    // Trust gate: cross-project learnings only loaded if explicitly trusted
+    // (user-stated). This prevents prompt injection from one project's
+    // AI-generated learnings silently influencing reviews in another project.
+    // Fail closed: rows missing the trusted field (legacy entries written
+    // before the field existed, hand-edited rows, or rows from other tools)
+    // are treated as untrusted rather than admitted by default.
+    if (isCrossProject && e.trusted !== true) continue;
 
     entries.push(e);
   } catch {}
--- a/test/gstack-learnings-search.test.ts
+++ b/test/gstack-learnings-search.test.ts
@@ -33,6 +33,9 @@ beforeAll(() => {
   const otherEntries = [
     { ts: '2026-05-04T00:00:00Z', skill: 'test', type: 'pattern', key: 'foreign-observed', insight: 'A foreign observed insight', confidence: 8, source: 'observed', trusted: false, files: [] },
     { ts: '2026-05-05T00:00:00Z', skill: 'test', type: 'pattern', key: 'foreign-user', insight: 'A foreign user-stated insight', confidence: 8, source: 'user-stated', trusted: true, files: [] },
+    // Legacy / hand-written / third-party row written before the trusted field
+    // existed: no `trusted` key at all. Must NOT be admitted cross-project.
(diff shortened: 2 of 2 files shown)


jbetala7:
Closing as superseded: the fix for #1745 landed on `main` via the v1.57.6.0 fix wave (#1911), which lists `fixes #1745` with attribution and includes an equivalent regression test. Verified the current `main` already contains this exact behavior, so this PR is redundant. Thanks for consolidating it into the wave.

PR #1747, opened 2026-05-27
fix(preamble): echo SESSIONS count so CLAUDE.md rules can read it (#1651 Defect A)

### LINKED ISSUE #1651: Two defects in `_SESSIONS` concurrent-session detection — value never echoed + marker TTL doesn't reflect 'currently active'

0xDevNinja:
## Summary

- The preamble at `scripts/resolvers/preamble/generate-preamble-bash.ts:23` computes `_SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f | wc -l | tr -d ' ')` but never echoes it. Every other preamble state value (BRANCH, PROACTIVE, SKILL_PREFIX, REPO_MODE, TELEMETRY, EXPLAIN_LEVEL, QUESTION_TUNING, LEARNINGS, ARTIFACTS_SYNC, ...) is echoed for the calling agent to read; SESSIONS is the lone hole.
- Any CLAUDE.md rule that points at "the preamble's session count" has no value to read and silently ships non-functional, as the reporter caught on a post-ship audit.
- Add `echo "SESSIONS: $_SESSIONS"` next to the existing BRANCH/PROACTIVE/SKILL_PREFIX block in the preamble, plus an invariant test in `test/gen-skill-docs.test.ts` that fails if any SKILL.md containing the `_SESSIONS=$(find` computation no longer echoes the value.

Closes #1651 **Defect A** only.

## Out of scope (Defect B)

The reporter's Defect B — marker mtime falling outside the 2h window for long-running sessions that don't invoke a gstack skill in that span — needs hook-infra work (`PreToolUse` refresh on every tool call) and is a separate change with broader surface area. Leaving it for a follow-up so this PR stays small and safe to land.

## Diff shape

- 1 resolver edit: +1 echo line
- 1 test addition: +30 lines (invariant pinning the echo across all generated SKILL.md files)
- 47 regenerated SKILL.md files (no manual edits; produced by `bun run gen:skill-docs`)

## Tests

```
# 712 pass, 7 fail (existing 711 + 1 new invariant)
```

The 7 failures are the pre-existing `--dry-run freshness check` cases on non-Claude hosts that reproduce on `upstream/main` with this branch's changes stashed — unrelated.

## Verification

```bash
$ grep -n "SESSIONS:" ship/SKILL.md context-save/SKILL.md
ship/SKILL.md:47:echo "SESSIONS: $_SESSIONS"
context-save/SKILL.md:46:echo "SESSIONS: $_SESSIONS"
```

A user-side rule like

> If preamble emits `SESSIONS: 2` (or higher), force a worktree switch before any edit.

now has a value to read instead of having to inline the `find` itself.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -38,6 +38,7 @@ _SKILL_PREFIX=$(~/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/de
 echo "PROACTIVE: $_PROACTIVE"
 echo "PROACTIVE_PROMPTED: $_PROACTIVE_PROMPTED"
 echo "SKILL_PREFIX: $_SKILL_PREFIX"
+echo "SESSIONS: $_SESSIONS"
 source <(~/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
 REPO_MODE=${REPO_MODE:-unknown}
 echo "REPO_MODE: $REPO_MODE"
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -47,6 +47,7 @@ _SKILL_PREFIX=$(~/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/de
 echo "PROACTIVE: $_PROACTIVE"
 echo "PROACTIVE_PROMPTED: $_PROACTIVE_PROMPTED"
 echo "SKILL_PREFIX: $_SKILL_PREFIX"
+echo "SESSIONS: $_SESSIONS"
 source <(~/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
 REPO_MODE=${REPO_MODE:-unknown}
 echo "REPO_MODE: $REPO_MODE"
--- a/benchmark-models/SKILL.md
+++ b/benchmark-models/SKILL.md
@@ -40,6 +40,7 @@ _SKILL_PREFIX=$(~/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/de
 echo "PROACTIVE: $_PROACTIVE"
 echo "PROACTIVE_PROMPTED: $_PROACTIVE_PROMPTED"
 echo "SKILL_PREFIX: $_SKILL_PREFIX"
+echo "SESSIONS: $_SESSIONS"
 source <(~/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
 REPO_MODE=${REPO_MODE:-unknown}
 echo "REPO_MODE: $REPO_MODE"
--- a/benchmark/SKILL.md
+++ b/benchmark/SKILL.md
@@ -40,6 +40,7 @@ _SKILL_PREFIX=$(~/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/de
 echo "PROACTIVE: $_PROACTIVE"
 echo "PROACTIVE_PROMPTED: $_PROACTIVE_PROMPTED"
 echo "SKILL_PREFIX: $_SKILL_PREFIX"
+echo "SESSIONS: $_SESSIONS"
 source <(~/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
 REPO_MODE=${REPO_MODE:-unknown}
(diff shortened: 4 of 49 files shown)


time-attack:
@0xDevNinja Please rewrite this to resolve Defect B as well; the missing session-count echo fixes Defect A only.

PR #1748, opened 2026-05-27
fix(skills): use $HOME instead of ~ in SKILL.md preamble paths (#1715)

### LINKED ISSUE #1715: gstack: artifacts auto-sync silently broken due to unexpanded ~ in double-quoted bash variable assignments
### Description

On Windows (Git Bash), the gstack skill preamble's artifacts sync logic silently never runs. The `~` in double-quoted variable assignments is not resolved to the home directory path.

This means `gstack-brain-sync --once` is never called at skill boundaries, so cross-machine artifact synchronization (checkpoints, learnings, timeline, etc.) does not happen automatically.

### Root Cause

In Bash, tilde expansion only occurs when `~` is at the beginning of an unquoted word. Inside double-quoted variable assignments, `~` is treated as a literal character.

Lines 369-370 in every SKILL.md preamble:

```bash
```

When `$_BRAIN_CONFIG_BIN` is later invoked:

```bash
```

The shell tries to execute the literal path `~/.claude/...` (which doesn't exist as a literal file path), the command fails, and `artifacts_sync_mode` falls back to `off`. This causes all subsequent sync conditions to evaluate to false.

The same issue affects the end-of-skill sync commands in the Telemetry section:

```bash
```

### Impact

- **Every** gstack skill invocation (`/context-save`, `/qa`, `/ship`, `/plan-*`, `/review`, etc.) silently skips the artifacts sync
- Cross-machine artifact sharing via the gstack artifacts git repo (`gstack-record`) never works automatically
- No user-visible error — all `|| true` / `2>/dev/null` patterns swallow the failure silently
- This has been broken since the artifacts sync feature was introduced

### Affected Files

All 92 SKILL.md files across two directories:

- `~/.claude/skills/*/SKILL.md` (46 files — loaded by Claude Code at runtime)
- `~/.claude/skills/gstack/*/SKILL.md` (46 files — gstack distribution copies)

Affected skills include: autoplan, benchmark, benchmark-models, browse, canary, codex, context-restore, context-save, cso, design-consultation, design-html, design-review, design-shotgun, devex-review, document-generate, document-release, health, investigate, ios-clean, ios-design-review, ios-fix, ios-qa, ios-sync, land-and-deploy, landing-report, learn, make-pdf, office-hours, open-gstack-browser, pair-agent, plan-ceo-review, plan-design-review, plan-devex-review, plan-eng-review, plan-tune, qa, qa-only, retro, review, scrape, setup-browser-cookies, setup-deploy, setup-gbrain, ship, skillify, sync-gbrain

### Fix

Change `"~/.claude/..."` to `"$HOME/.claude/..."` in two places in the preamble template and the end-of-skill telemetry section.

**In the Artifacts Sync preamble section:**

```diff
- _BRAIN_SYNC_BIN="~/.claude/skills/gstack/bin/gstack-brain-sync"
- _BRAIN_CONFIG_BIN="~/.claude/skills/gstack/bin/gstack-config"
+ _BRAIN_SYNC_BIN="$HOME/.claude/skills/gstack/bin/gstack-brain-sync"
+ _BRAIN_CONFIG_BIN="$HOME/.claude/skills/gstack/bin/gstack-config"
```

**In the end-of-skill Telemetry section:**

```diff
- "~/.claude/skills/gstack/bin/gstack-brain-sync" --once 2>/dev/null || true
+ "$HOME/.claude/skills/gstack/bin/gstack-brain-sync" --once 2>/dev/null || true
```

### Verification

After fixing, the preamble correctly shows `ARTIFACTS_SYNC: mode=full | ...` instead of `ARTIFACTS_SYNC: off`. Git status on `~/.gstack/` confirms that `gstack-brain-sync --once` modified `learnings.jsonl`, `timeline.jsonl`, and `.brain-queue.jsonl`.

### Environment

- OS: Windows 11 (Git Bash)
- Shell: Bash (via Git for Windows)

Pablosinyores:
## Summary

Closes #1715.

The artifacts-sync preamble emitted by every SKILL.md sets:

```bash
```

Bash tilde expansion only fires when `~` is at the start of an unquoted word; inside double-quoted variable assignments `~` stays literal. When the preamble then runs `_BRAIN_SYNC_MODE=$("$_BRAIN_CONFIG_BIN" get artifacts_sync_mode 2>/dev/null || echo off)` the literal path doesn't resolve, `gstack-config` fails, `artifacts_sync_mode` falls back to `off`, all sync conditions evaluate false, and the `|| true` / `2>/dev/null` patterns swallow the failure. End result: every skill invocation (`/qa`, `/ship`, `/plan-*`, `/review`, etc.) silently skips the cross-machine artifacts sync.

The same shape breaks the quoted telemetry check (`[ -x "${ctx.paths.binDir}/gstack-telemetry-log" ]`) and the skill-end `"$HOME/...brain-sync" --once`/`--discover-new` lines.

## Root cause + fix

One place: `scripts/resolvers/types.ts:37`. Claude-host `HostPaths` are built as `` `~/${config.globalRoot}` ``. Switch the substituted root to `` `$HOME/${config.globalRoot}` ``. Rendered paths are now bash-safe in every quoting context (double-quoted assignments, quoted `[ -x "…" ]` tests, bare command positions) without any change to host configs or downstream resolvers.

Regenerated all 50 SKILL.md files via `bun run gen:skill-docs`. Bumped `host-config.test.ts` to expect `$HOME` in the derived path assertions.

## Test plan

- [x] `bun test test/host-config.test.ts test/gstack-learnings-search.test.ts test/learnings-injection.test.ts test/learnings.test.ts` — 109/109 green locally.
- [x] Grep verified: zero `"~/.claude/skills/gstack` literals remain in any `*.md` outside `test/fixtures/golden-ship-claude.md` (separate stale fixture, not referenced by any test).
- [x] Two pre-existing golden-fixture failures (`Codex ship` / `Factory ship`) reproduce on upstream `main` — they need the `.agents/` and `.factory/` build artifacts that aren't generated in a fresh clone. Not regressions.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -28,35 +28,35 @@ test a site, verify a deployment, dogfood a user flow, or file a bug with screen
 ## Preamble (run first)
 
 ```bash
-_UPD=$(~/.claude/skills/gstack/bin/gstack-update-check 2>/dev/null || .claude/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
+_UPD=$($HOME/.claude/skills/gstack/bin/gstack-update-check 2>/dev/null || .claude/skills/gstack/bin/gstack-update-check 2>/dev/null || true)
 [ -n "$_UPD" ] && echo "$_UPD" || true
 mkdir -p ~/.gstack/sessions
 touch ~/.gstack/sessions/"$PPID"
 _SESSIONS=$(find ~/.gstack/sessions -mmin -120 -type f 2>/dev/null | wc -l | tr -d ' ')
 find ~/.gstack/sessions -mmin +120 -type f -exec rm {} + 2>/dev/null || true
-_PROACTIVE=$(~/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
+_PROACTIVE=$($HOME/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
 _PROACTIVE_PROMPTED=$([ -f ~/.gstack/.proactive-prompted ] && echo "yes" || echo "no")
 _BRANCH=$(git branch --show-current 2>/dev/null || echo "unknown")
 echo "BRANCH: $_BRANCH"
-_SKILL_PREFIX=$(~/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/dev/null || echo "false")
+_SKILL_PREFIX=$($HOME/.claude/skills/gstack/bin/gstack-config get skill_prefix 2>/dev/null || echo "false")
 echo "PROACTIVE: $_PROACTIVE"
 echo "PROACTIVE_PROMPTED: $_PROACTIVE_PROMPTED"
 echo "SKILL_PREFIX: $_SKILL_PREFIX"
-source <(~/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
+source <($HOME/.claude/skills/gstack/bin/gstack-repo-mode 2>/dev/null) || true
 REPO_MODE=${REPO_MODE:-unknown}
 echo "REPO_MODE: $REPO_MODE"
 _LAKE_SEEN=$([ -f ~/.gstack/.completeness-intro-seen ] && echo "yes" || echo "no")
 echo "LAKE_INTRO: $_LAKE_SEEN"
-_TEL=$(~/.claude/skills/gstack/bin/gstack-config get telemetry 2>/dev/null || true)
+_TEL=$($HOME/.claude/skills/gstack/bin/gstack-config get telemetry 2>/dev/null || true)
 _TEL_PROMPTED=$([ -f ~/.gstack/.telemetry-prompted ] && echo "yes" || echo "no")
 _TEL_START=$(date +%s)
 _SESSION_ID="$$-$(date +%s)"
 echo "TELEMETRY: ${_TEL:-off}"
 echo "TEL_PROMPTED: $_TEL_PROMPTED"
-_EXPLAIN_LEVEL=$(~/.claude/skills/gstack/bin/gstack-config get explain_level 2>/dev/null || echo "default")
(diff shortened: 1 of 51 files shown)


jbetala7:
This overlaps with #1692 (open since 2026-05-25, fixes #1656). Both target the same defect: a literal `~` inside double-quoted bash assignments in the artifacts-sync preamble (`_BRAIN_SYNC_BIN="~/..."`, the `[ -x "~/...gstack-telemetry-log" ]` test) never tilde-expands, so brain-sync silently no-ops across all ~50 generated SKILL.md files on Claude hosts. Issue #1715 (this PR) and #1656 (#1692) describe the same bug — #1656 frames it as a regression of #785.

The two PRs differ in where they fix it, which is the substantive thing for a maintainer to pick:
- This PR rewrites the root in `scripts/resolvers/types.ts` (Claude `HostPaths` built as `~/${globalRoot}` → `$HOME/${globalRoot}`), so every derived path becomes bash-safe in one place.
- #1692 deliberately keeps `types.ts` untouched and rewrites locally in the two preamble resolvers, arguing the `types.ts` swap would conflict with #843 (which proposes the same global swap on `browseDir`/`designDir`). It also adds static-invariant tests in `test/skill-validation.test.ts` that fail if any generated SKILL.md reintroduces `="~/` or `[ -x "~/`.

The `types.ts` approach here is arguably more complete (fixes at the true root rather than per-resolver), but it's worth reconciling against #843/#785 before landing so the global swap doesn't collide. Flagging so the same defect isn't fixed twice — the regenerated output should be identical either way.

Pablosinyores:
@jbetala7 — agreed. #1692 fixes the defect at source (`scripts/resolvers/preamble/generate-brain-sync-block.ts` and `generate-preamble-bash.ts`), so all regenerated SKILL.md previews flow from one place; mine only patched the rendered output and would drift again on the next preamble regen. Closing in favour of #1692.

Issue #282, opened 2026-03-21
/ship skill under Codex uses Claude as co-author in commit message


TBMoonwalker:
### Problem

When you using GStack with Codex (installed with the new Codex settings) the wrong Co-Author commit message is created:

```bash
git commit -m "$(cat <<'EOF'
chore: bump version and changelog (vX.Y.Z.W)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
EOF
)"
```

### Root cause

The commit message is hardcoded in the SKILL.md from gstack-ship.

### Impact

Wrong commit messages are pushed to GIT if someone uses Codex as coding environment

### Suggested fix

Change the markdown file so that the right Co-Author is used depending on the currently active environment.

PR #1757, opened 2026-05-27
fix(ship): sync package-lock.json alongside package.json in Step 12

oharriehausen:
## Summary

`/ship`'s Step 12 (Version bump) writes the new version to `VERSION` and `package.json` but never touches `package-lock.json`. Projects using npm (lockfileVersion 3) end up with a stale lockfile every time `/ship` runs — until something else (an unrelated `npm install`, a merge conflict, etc.) forces a resync.

In the Landmen repo this showed up as multiple PRs shipping with a lockfile pinned several versions behind the actual release:

- PR #282 bumped to v1.46.0.3, lockfile stayed at 1.46.0.1
- PR #283 bumped to v1.46.0.4, lockfile stayed at 1.46.0.2
- PR #285 only synced incidentally because a merge conflict forced it

## What changed

`package-lock.json` (lockfileVersion 3) stores the version in two places:

1. Top-level `version`
2. `packages[""].version` — the empty-string key, representing the root package

Both are now patched alongside `package.json` on the FRESH bump path **and** on the DRIFT_STALE_PKG repair path, mirroring the existing node-then-bun fallback pattern.

Behavior matches the surrounding code:

- **Skip silently when `package-lock.json` is absent** (yarn-only, pnpm, bun lockfile, no JS project).
- **Validate JSON parse before writing** — a malformed lockfile errors out the same way malformed `package.json` does. Never silently corrupted.
- **Node-then-bun fallback** — same conditional pattern the existing `package.json` block uses.
- **Idempotency preserved** — the new write lives inside the existing FRESH/DRIFT_STALE_PKG conditionals. `ALREADY_BUMPED` returns before reaching it, so re-running `/ship` on an already-bumped branch leaves the lockfile alone.
- **Idempotency detector untouched** — it still only compares `package.json.version` against `VERSION`. The lockfile is opportunistically synced on FRESH/repair but lockfile-only drift is not treated as a halt condition (it never blocks runtime). Documented in the DRIFT_STALE_PKG dispatch bullet.

## Files

Both `ship/SKILL.md.tmpl` and `ship/SKILL.md` are updated. `SKILL.md` was regenerated from the template via `scripts/gen-skill-docs.ts` — diff between hand-edit and canonical generator output is empty.

## Test plan

- [x] Smoke-tested the new `node -e` snippet on a real lockfileVersion 3 file — writes both `version` fields, structural diff (with `.version` keys excluded) is clean.
- [x] Smoke-tested the malformed-JSON guard — exits 1 with a clear error, no corrupted write.
- [x] Confirmed regenerator agrees: `bun run scripts/gen-skill-docs.ts` produces zero diff against the hand-edited `SKILL.md`.
- [ ] End-to-end `/ship` run on a real PR that exercises the FRESH path (will validate on next Landmen release).

### DIFF
--- a/ship/SKILL.md
+++ b/ship/SKILL.md
@@ -2492,7 +2492,7 @@ Read the `STATE:` line and dispatch:
 
 - **FRESH** → proceed with the bump action below (steps 1–4).
 - **ALREADY_BUMPED** → skip the bump by default, BUT check for queue drift first: call `bin/gstack-next-version` with the implied bump level (derived from `CURRENT_VERSION` vs `BASE_VERSION`), compare its `.version` against `CURRENT_VERSION`. If they differ (queue moved since last ship), use **AskUserQuestion**: "VERSION drift detected: you claim v<CURRENT> but next available is v<NEW> (queue moved). A) Rebump to v<NEW> and rewrite CHANGELOG header + PR title (recommended), B) Keep v<CURRENT> — will be rejected by CI version-gate until resolved." If A, treat this as FRESH with `NEW_VERSION=<new>` and run steps 1-4 (which will also trigger Step 13 CHANGELOG header rewrite and Step 19 PR title rewrite). If B, reuse `CURRENT_VERSION` and warn that CI will likely reject. If util is offline, warn and reuse `CURRENT_VERSION`.
-- **DRIFT_STALE_PKG** → a prior `/ship` bumped `VERSION` but failed to update `package.json`. Run the sync-only repair block below (after step 4). Do NOT re-bump. Reuse `CURRENT_VERSION` for CHANGELOG and PR body. (Queue check still runs in ALREADY_BUMPED terms after repair.)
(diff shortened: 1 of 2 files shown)


oharriehausen:
Closing — this patch targeted the inline bash in `ship/SKILL.md` Step 12, which #1806 (v1.54.0.0) replaced with `bin/gstack-version-bump`. The diff no longer applies cleanly and the surface area changed enough that a rebase isn't really the right move.

Heads up that the underlying bug still exists in the new CLI: `cmdWrite` and `cmdRepair` write VERSION + package.json but don't touch `package-lock.json`, so npm projects still ship with a stale lockfile root version after every `/ship`. Fix is ~30 lines mirroring `writePkgVersion` (lockfileVersion 3 stores the version at both top-level and `packages[""].version`). Happy to open a fresh PR against the CLI if you want it — otherwise leaving the breadcrumb here.

Thanks for the refactor either way — the CLI extraction is a much cleaner home for this logic.

PR #1758, opened 2026-05-27
fix(setup): truncate Bun trailing padding before codesign; stop false-alarm Apple Silicon warning

brandonlipman:
## Symptom

On Apple Silicon, `./setup` prints:

```
warning: codesign failed for browse/dist/find-browse (binary may not run on Apple Silicon)
warning: codesign failed for bin/gstack-global-discover (binary may not run on Apple Silicon)
```

Both binaries **run fine** — the warning is a false alarm. `browse`, `design`, and `pdf` never warn.

## Root cause

`bun build --compile` leaves **~19 KB of trailing zero-padding after the Mach-O `LC_CODE_SIGNATURE` region** for `find-browse` and `gstack-global-discover` (verified: `browse`/`design`/`pdf` have **0** trailing bytes). macOS codesign requires the signature to be the last content and extend to EOF, so the padding makes both setup steps fail:

- `codesign --remove-signature` → `internal error in Code Signing subsystem`
- `codesign -s - -f` → `main executable failed strict validation`

The re-sign fails, leaving Bun's original adhoc *linker-signed* signature in place. That signature still covers the executable code pages, so the kernel's exec-time check passes and the binary runs — even though `codesign --verify` is unhappy.

Measured (signature end vs file size):

| binary | trailing bytes after sig | `codesign --verify` |
|---|---|---|
| browse / design / pdf | 0 | PASS |
| find-browse / gstack-global-discover | 19326 | FAIL |

Confirmation test: truncating a copy of `find-browse` to the signature end makes the **identical** `codesign -s - -f` succeed and `codesign --verify --strict` pass, and the binary still runs.

## Fix

`setup` (the `Darwin` + `arm64` codesign loop):

1. **Truncate trailing bytes past `LC_CODE_SIGNATURE`** (offset from `otool` `dataoff`+`datasize`) before signing. Degrades safely if `otool` is unavailable or there is no trailing slack.
2. **SIGKILL-gated warning** — if the re-sign still fails, probe the binary and reserve the "may not run on Apple Silicon" warning for a genuine `exit 137` (SIGKILL); otherwise emit an informational note. The probe is `set -e` safe (`|| _probe_rc=$?`).

## Tests

Two regression tests added to `test/setup-codesign.test.ts` (truncation logic + SIGKILL-gated warning). Both fail on the pre-fix `setup`. Full suite: 8 pass / 0 fail.

### DIFF
--- a/setup
+++ b/setup
@@ -321,14 +321,46 @@ if [ "$NEEDS_BUILD" -eq 1 ]; then
   # macOS kills with SIGKILL (exit 137). The two-step remove+re-sign is
   # required because a naive `codesign -s - -f` fails when the existing
   # signature block is corrupt. This is idempotent and costs <1s.
+  #
+  # Some binaries (observed: find-browse, gstack-global-discover) also carry
+  # trailing zero-padding AFTER the Mach-O LC_CODE_SIGNATURE region. macOS
+  # codesign requires the signature to be the last content and extend to EOF,
+  # so the padding triggers "main executable failed strict validation" on
+  # re-sign (and "internal error in Code Signing subsystem" on remove). We
+  # truncate that trailing slack to the end of LC_CODE_SIGNATURE first, which
+  # lets the identical re-sign succeed. The binary runs either way: Bun's
+  # adhoc code-page signature satisfies the kernel's exec check even when
+  # `codesign --verify` is unhappy, so a re-sign failure only warns when the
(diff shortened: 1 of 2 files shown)


time-attack:
@brandonlipman Please rewrite this to avoid truncating valid executable data, handle signing outside builds, preserve broader signing failures, and add real Apple Silicon signing and execution tests.

PR #1764, opened 2026-05-27
fix(ios-qa): SwiftUI Button tap on iOS 17 + Documents/ boot token

dwidoo:
## Summary

Two infra fixes surfaced by a real /ios-qa onboarding session on iPhone SE 2020 / iOS 17.7 with a USB-direct CoreDevice tunnel. Both reproduce on v1.43.3.0 and current main (v1.48.0.0).

### 1. SwiftUI Button hit-test on iOS 17 (`Bridges.swift.template`)

\`MutationBridgeImpl.handleTap\` delegates straight to \`DebugBridgeTouch.sendTap\` (in-process \`UITouch\` + \`IOHIDEvent\` synthesis). That works for UIControl, gesture recognizers, and scroll views, but on iOS 16/17 it **silently misses SwiftUI Buttons** — the SwiftUI hosting controller routes hits through the iOS 18-only \`_UIHitTestContext\` private API, which the comments in DebugBridgeTouch.m already note. Symptom: \`POST /tap\` returns \`{\"op\":\"tap\",\"ok\":true}\` (UITouch was synthesized successfully) but Button.action never fires, screen state unchanged.

This PR adds a **two-stage dispatch**:

1. Walk the AX tree, find the smallest \`.button\`-trait element whose accessibility frame contains the tap point, call \`accessibilityActivate()\`. Routes through SwiftUI's standard VoiceOver activation path → fires \`Button.action\` regardless of iOS version. Covers both \`view.isAccessibilityElement\` UIView elements and synthetic NSObject elements vended by SwiftUI hosting controllers via \`accessibilityElements\` / \`accessibilityElementCount\`.
2. If no AX hit, fall through to \`DebugBridgeTouch.sendTap\` (gesture recognizers, scroll views, drag gestures, custom hit-test-only views, iOS 18+ SwiftUI via _UIHitTestContext).

Smallest-containing wins, mirroring UIKit hit-testing's deepest-interactive-element resolution.

### 2. Boot token persistence (`StateServer.swift.template` + `tunnel-bootstrap.ts`)

StateServer writes the boot token to \`NSTemporaryDirectory() + \"gstack-ios-qa.token\"\`. iOS clears \`tmp/\` aggressively (memory pressure, jetsam after backgrounding, between launches), so on any non-trivial QA session the daemon's \`devicectl copy from tmp/gstack-ios-qa.token\` returns ENOENT and bootstrap fails with \`boot_token_unavailable\` — **even though StateServer is alive and answering \`/healthz\`**. Confusing failure mode: visible app on-device, daemon connects, but never auths.

Both ends moved to \`Documents/gstack-ios-qa.token\` (survives backgrounding + memory pressure, same \`appDataContainer\` domain). The Swift side gracefully falls back to the old \`NSTemporaryDirectory()\` path if Documents/ is unavailable (paranoia — shouldn't ever happen on iOS).

## Test plan
- [x] \`cd ios-qa/daemon && bun test\` — 91/91 pass
- [x] DebugBridge SPM \`swift build\` clean on Swift 5.9 (both targets, with patched templates substituted into a consuming repo)
- [x] Live-device repro: tap a SwiftUI Button on iPhone SE 2020 iOS 17.7. Before: \`{\"op\":\"tap\",\"ok\":true}\` but no UI change. After: Button.action fires, screen advances.
- [x] Live-device repro: backgrounded app for 10+ minutes, daemon bootstrap succeeds (previously \`boot_token_unavailable\`).

## Repro env
- macOS 14.x, Xcode 16.2, bun 1.3.11
- iPhone SE 2020 (iPhone12,8) iOS 17.7, USB-direct CoreDevice
- gstack v1.43.3.0 → v1.48.0.0 (issue in both)

### DIFF
--- a/ios-qa/daemon/src/tunnel-bootstrap.ts
+++ b/ios-qa/daemon/src/tunnel-bootstrap.ts
@@ -64,7 +64,12 @@ export type BootstrapErrorReason =
  */
 export async function bootstrapTunnel(opts: BootstrapOptions): Promise<BootstrapResult> {
   const port = opts.port ?? 9999;
-  const tokenPath = opts.bootTokenPath ?? 'tmp/gstack-ios-qa.token';
+  // Default Documents/ instead of tmp/. iOS clears tmp/ on memory
+  // pressure / between launches → `devicectl copy from tmp/...` returns
+  // ENOENT → bootstrap fails with `boot_token_unavailable` even though
+  // StateServer is alive. Documents/ persists across backgrounding.
+  // Pair with StateServer.swift.template `bootTokenPath`.
+  const tokenPath = opts.bootTokenPath ?? 'Documents/gstack-ios-qa.token';
   const startupTimeoutMs = opts.startupTimeoutMs ?? 5_000;
   const spawn = opts.spawnImpl;
   const resolve = opts.resolveImpl;
--- a/ios-qa/templates/Bridges.swift.template
+++ b/ios-qa/templates/Bridges.swift.template
@@ -210,19 +210,115 @@ enum MutationBridgeImpl {
         }
     }
 
-    /// Tap at (x, y) in window coordinates. Delegates to DebugBridgeTouch
-    /// (KIF-derived in-process touch synthesis). The Obj-C target builds a
-    /// real UITouch + IOHIDEvent + UIEvent and dispatches via
-    /// `UIApplication.sendEvent`, which is what UIKit uses for real touches.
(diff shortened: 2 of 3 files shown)


time-attack:
@16francej — recommend closing this PR only after the protected-storage work, fallback handling, errors, fixtures, docs, and tests are transferred to #1837. The tap portion is already handled more safely, while the storage diagnosis is useful but this implementation leaves unresolved safety problems. Retain attribution for that diagnosis and omit the completed tap portion.

Issue #1735, opened 2026-05-27
/ios-qa and /ios-design-review first-run is broken at v1.45.0.0 (still on v1.46.0.0) — 7 reproducible bugs


frank-alvarado:
# `/ios-qa` + `/ios-design-review` first-run is broken in v1.45.0.0 (still broken on v1.46.0.0)

**Update 2026-05-26 (post-v1.46.0.0):** Reproduced on v1.46.0.0. `ios-qa/SKILL.md` was modified by the catalog-trim work (frontmatter shrink, routing prose moved to `## When to invoke` body section), but the three broken Swift files were untouched: `ios-qa/scripts/gen-accessors-tool/Package.swift`, `ios-qa/templates/Package.swift.template`, and `ios-qa/templates/StateAccessor.swift.template`. All 7 issues below are reproducible on v1.46.0.0 with identical error output. Verified via:

```
cd ~/.claude/skills/gstack
git diff cf50443b..22f8c7f4 --stat ios-qa/scripts/gen-accessors-tool/Package.swift ios-qa/templates/Package.swift.template ios-qa/templates/StateAccessor.swift.template
# (no output — files unchanged across v1.45.0.0 → v1.46.0.0)
```

---

**gstack version:** 1.45.0.0 (originally hit) and 1.46.0.0 (re-verified)
**macOS / Xcode:** Xcode 26.5 / Swift 6.3.2
**Skill files affected:** `ios-qa/SKILL.md`, `ios-qa/scripts/gen-accessors-tool/Package.swift`, `ios-qa/templates/Package.swift.template`, `ios-qa/templates/StateAccessor.swift.template`
**Reporter context:** Solo dev, real iPhone 17 Pro Max connected via USB CoreDevice, fresh first-run after `/gstack-upgrade` (1.39.2.0 → 1.45.0.0).

## Summary

Attempted `/ios-design-review` cold-start. Hit 7 distinct issues in ~30 minutes before stopping. Several are blocking, several are misleading docs that cause wasted time. Net effect: the iOS QA family ships in a state where a first-time user cannot run it end-to-end without patching gstack itself.

### 1. SKILL.md describes an SPM dependency install; reality is template codegen

`ios-qa/SKILL.md` Phase 2 step 1 says:

> Add the `DebugBridge` SPM dependency to the app's `Package.swift`.

[...]

PR #1766, opened 2026-05-28
Fix iOS QA first-run SwiftPM manifest blockers

spacegeologist:
## Summary

This fixes the P0 SwiftPM manifest blockers from issue #1735:

- Remove the `GenAccessorsTests` test target from `ios-qa/scripts/gen-accessors-tool/Package.swift` because the shipped skill does not include `Tests/GenAccessorsTests`.
- Move `// swift-tools-version:5.9` to line 1 of `ios-qa/templates/Package.swift.template` so SwiftPM accepts the rendered DebugBridge package manifest.
- Remove the `DebugBridgeCoreTests` test target from `ios-qa/templates/Package.swift.template` because the codegen output does not create `Tests/DebugBridgeCoreTests`.
- Add regression coverage for both invariants.

This intentionally only addresses the P0 first-run blockers from #1735 (#3, #5, #6). The remaining issues in that report should stay open for follow-up PRs.

## Root Cause

SwiftPM validates every declared target path before building a selected target, so missing test target directories break first-run setup even when the user is only trying to build the production tool/package. The rendered DebugBridge package also failed manifest parsing because SwiftPM requires the tools-version directive to be the first line of `Package.swift`.

## Validation

- `bun test test/skill-e2e-ios-swift-build.test.ts -t "template"`
- `swift package describe --package-path ios-qa/scripts/gen-accessors-tool --type json`
- Rendered `ios-qa/templates/Package.swift.template` into a temporary DebugBridge package with dummy source directories and verified `swift package describe --type json` returns `tools_version: "5.9"` and the three DebugBridge targets/products.
- `git diff --check`

### DIFF
--- a/ios-qa/scripts/gen-accessors-tool/Package.swift
+++ b/ios-qa/scripts/gen-accessors-tool/Package.swift
@@ -31,10 +31,5 @@ let package = Package(
             ],
             path: "Sources/GenAccessors"
         ),
-        .testTarget(
-            name: "GenAccessorsTests",
-            dependencies: ["GenAccessors"],
-            path: "Tests/GenAccessorsTests"
-        ),
     ]
 )
--- a/ios-qa/templates/Package.swift.template
+++ b/ios-qa/templates/Package.swift.template
@@ -1,3 +1,4 @@
+// swift-tools-version:5.9
 // AUTO-GENERATED from gstack/ios-qa/templates/Package.swift.template
 //
 // Drop-in SPM package definition for the DebugBridge stack. Three targets:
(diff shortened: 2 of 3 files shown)


spacegeologist:
CI note: the only failing check is `E2E Evals / report`, and the failure is in the `Post PR comment` step rather than in an eval suite.

The individual eval jobs completed successfully. The report job downloaded all eval artifacts, then failed when trying to comment on this fork PR:

```text
```

So the red status appears to be a GitHub token/comment-permission issue for the report job on forked PRs, not a failure caused by this manifest change.

PR #1767, opened 2026-05-28
[codex] Fix Claude Code auth preflight for Codex host

spacegeologist:
## Summary

Fix the Codex-hosted `gstack-claude` skill so it does not reject a working Claude Code install just because `$HOME/.claude/.credentials.json` is absent.

With increased use of Codex, I have been experimenting heavily with the gstack setup and trying to smooth out the Codex <-> Claude Code handoff. This came from that dogfooding: `claude -p` works in the host environment even when `.credentials.json` is absent.

## Root Cause

The generated Codex skill preflighted Claude authentication by probing an implementation detail: `$HOME/.claude/.credentials.json` or `ANTHROPIC_API_KEY`. Modern Claude Code auth may be stored in OS Keychain or another host-managed location, so credential-file probing can false-negative before the skill ever invokes Claude.

## Fix

- Remove the `.credentials.json` authentication preflight.
- Treat the first `claude -p` invocation as the real auth check.
- Parse Claude JSON output so `is_error` still surfaces `CLAUDE_ERROR: true`, and auth/login failures also surface `CLAUDE_AUTH_ERROR: true`.
- Keep the generated Codex regression test from reintroducing `.credentials.json` probing.

## Safety Boundary

The safety boundary is unchanged:

- review/challenge modes still run tool-less Claude calls
- consult mode remains read-only with `Read,Grep,Glob`
- `Bash`, `Edit`, and `Write` remain disallowed

## Validation

- `claude -p "Reply with exactly OK." --output-format json --disable-slash-commands --tools ""` returned `result: "OK"` when run with host auth/keychain access.
- `bun test test/gen-skill-docs.test.ts -t "Codex output includes Claude outside-voice skill"`
- `bun run scripts/gen-skill-docs.ts --host codex --dry-run`
- `git diff --check`

### DIFF
--- a/claude/SKILL.md.tmpl
+++ b/claude/SKILL.md.tmpl
@@ -42,18 +42,10 @@ CLAUDE_BIN=$(command -v claude 2>/dev/null || echo "")
 If `NOT_FOUND`, stop and tell the user:
 "Claude CLI not found. Install Claude Code, then re-run this skill."
 
-Check auth:
-
-```bash
-if [ -f "$HOME/.claude/.credentials.json" ] || [ -n "${ANTHROPIC_API_KEY:-}" ]; then
-  echo "AUTH_FOUND"
-else
-  echo "AUTH_MISSING"
-fi
-```
-
-If `AUTH_MISSING`, stop and tell the user:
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej Merge this version because it passes the Mac checks and correctly handles Claude Code authentication preflight on the Codex host.

PR #1752, opened 2026-05-27
v1.59.1.0 Fix generated skill YAML frontmatter

cathrynlavery:
## Summary
- quote catalog-trimmed generated skill descriptions so colon-containing descriptions remain valid YAML
- add a frontmatter-only validation script and wire it into the skill docs workflow
- add regression coverage for SKILL.md frontmatter validation

## Verification
- bun test test/skill-validation.test.ts test/catalog-trim.test.ts
- bun run gen:skill-docs --dry-run
- bun run gen:skill-docs --host codex
- bun run gen:skill-docs --host factory
- bun run skill:frontmatter
- bun run gen:skill-docs --host all
- bun test test/gen-skill-docs.test.ts

PR #1768, opened 2026-05-28
chore: fix yaml frontmatter parse error

sator-imaging:


jbetala7:
This is the same generated-frontmatter YAML failure covered by the broader generator fix in #1752, and this PR only fixes one generated output file.

Evidence from this diff: it changes just `design-consultation/SKILL.md` from an unquoted inline `description:` to a quoted scalar. That does repair this one file, but it leaves the root cause in `scripts/gen-skill-docs.ts`, so the next docs generation can emit the same invalid YAML again for any colon-containing trimmed description.

Canonical path looks like #1752 (or #1739): update `applyCatalogTrim`/the generator to quote inline descriptions, add frontmatter validation/regression coverage, and regenerate all affected `SKILL.md` files. I would treat this PR as duplicate/incomplete unless it is rebased into the generator-level fix.

garrytan:
Closing in favor of #1786, which fixes the YAML-frontmatter colon issue at the source (the gen-skill-docs generator) rather than editing a single generated file. Thanks!

PR #1770, opened 2026-05-28
fix(jsonl-merge): equal-ts entries must converge across machines

### LINKED ISSUE #1769: gstack-jsonl-merge: equal-ts entries resolve non-deterministically across machines (append-only logs never converge)
## Observed problem

`bin/gstack-jsonl-merge` is the registered git merge driver for append-only JSONL
state files (telemetry, learnings, timeline, etc., wired up by
`gstack-artifacts-init` / `gstack-brain-restore`). Its job is to resolve the
same-tail conflict two machines produce when both append between pushes, and the
header promises this is deterministic: *"both appends survive, ordered by
wall-clock timestamp where available, content hash otherwise."*

It is **not** deterministic when two different entries share the same `ts`.

## Current behavior on `origin/main` (19770ea8)

The Python sort key for a timestamped line is `(0, ts)` with no further
tiebreaker. Python's sort is stable, so entries with an equal `ts` keep their
**insertion order**, which is `base, ours, theirs`. But which physical entry is
"ours" vs "theirs" depends on **which machine runs the merge** — git assigns the
local side to `%A` (ours). So the two machines resolving the same conflict emit
the two equal-`ts` lines in opposite order:

```
# Machine A (its line = ours):     event:a, event:b
# Machine B (its line = ours):     event:b, event:a
```

The merged files differ, so the repos diverge and the next sync is another
conflict — resolved differently again. The logs never converge.

This is not a corner case for `gstack-telemetry-log`, which stamps
second-granularity timestamps (`date -u +%Y-%m-%dT%H:%M:%SZ`): two skill events
in the same second collide routinely. Non-JSON lines and lines without a `ts`
are unaffected because they already order by SHA-256 of the line content, which
is side-independent.

Repro:

```bash
a='{"ts":"2026-05-28T10:00:00Z","event":"a"}'
b='{"ts":"2026-05-28T10:00:00Z","event":"b"}'
printf '%s\n' "$a" > ours; printf '%s\n' "$b" > theirs; : > base
bin/gstack-jsonl-merge base ours theirs && cat ours      # a then b
printf '%s\n' "$b" > ours; printf '%s\n' "$a" > theirs; : > base
bin/gstack-jsonl-merge base ours theirs && cat ours      # b then a  <-- diverges
```

## Expected behavior

The same set of input lines resolves to the same file on every machine,
regardless of which side each line arrives on.

- open PRs: `jsonl merge`, `merge driver`, `jsonl-merge OR determinism OR convergence` — none
- scanned all 100 open PRs for any touching `bin/gstack-jsonl-merge` — none
- open issues: `jsonl merge driver`, `jsonl merge driver determinism` — none
- `git log -- bin/gstack-jsonl-merge`: last touched in v1.27.0.0, introduced v1.9.0.0; no in-flight work

## Candidate fix shape

Make the sort a total order by adding the line content as the final tiebreaker:
`(0, ts, line)` for timestamped entries (and `(1, h, line)` for the hash path,
for symmetry). Equal-`ts` entries then order by content, identically on both
sides. Plus a regression test that runs the driver with the two sides swapped
and asserts identical output.

jbetala7:
Fixes #1769

## Repro / observed problem

`bin/gstack-jsonl-merge` is the git merge driver for append-only JSONL state
files (telemetry, learnings, timeline). Its header promises deterministic
resolution: *"both appends survive, ordered by wall-clock timestamp where
available, content hash otherwise."* That promise breaks when two different
entries share the same `ts`:

```bash
a='{"ts":"2026-05-28T10:00:00Z","event":"a"}'
b='{"ts":"2026-05-28T10:00:00Z","event":"b"}'
printf '%s\n' "$a" > ours; printf '%s\n' "$b" > theirs; : > base
bin/gstack-jsonl-merge base ours theirs && cat ours   # a, b
printf '%s\n' "$b" > ours; printf '%s\n' "$a" > theirs; : > base
bin/gstack-jsonl-merge base ours theirs && cat ours   # b, a  <-- diverges
```

## Root cause

The sort key for a timestamped line was `(0, ts)` with no further tiebreaker.
Python's sort is stable, so equal-`ts` entries keep insertion order
(`base, ours, theirs`). Git assigns the **local** side to `%A` ("ours"), so the
same logical merge runs with `ours`/`theirs` swapped on the two machines, and
the equal-`ts` lines come out in opposite order. The merged files diverge and
never converge; the next sync is another conflict, resolved differently again.
`gstack-telemetry-log` stamps second-granularity timestamps
(`date -u +%Y-%m-%dT%H:%M:%SZ`), so same-`ts` collisions are routine. Lines
without a `ts` were already safe — they order by SHA-256 of the content, which
is side-independent.

## Fix

Add the line content as the final sort tiebreaker (`(0, ts, line)`, and
`(1, h, line)` for the hash path for symmetry). Equal-`ts` entries now order by
content, identically on every machine. One-line behavioral change to the sort
key.

## Testing

New `test/jsonl-merge.test.ts` (6 tests, `bun test`):

- equal-`ts` entries resolve identically with the two sides swapped (the
  convergence regression — **fails on `main`**: emits `a,b` vs `b,a`; passes
  with the fix)
- no-`ts` and plain non-JSON lines also resolve side-independently
- exact-duplicate dedup preserved
- timestamped entries still sort ascending by `ts`
- absent `ours`/`theirs` files tolerated (added-file merge)

Verified the convergence test fails on `origin/main`'s driver and passes with
the change. `bash -n` clean, `git diff --check` clean.

### DIFF
--- a/bin/gstack-jsonl-merge
+++ b/bin/gstack-jsonl-merge
@@ -53,18 +53,25 @@ for path in paths:
                     continue
                 if line in seen:
                     continue
-                # Prefer ISO ts field for sort; fall back to SHA-256.
+                # Prefer ISO ts field for sort; fall back to SHA-256. The line
+                # content is the final tiebreaker so the order is total: two
+                # entries sharing a ts must resolve identically regardless of
+                # which side they arrive on. Without it, equal-ts entries fall
+                # back to insertion order (base, ours, theirs), and since ours
+                # and theirs are swapped depending on which machine runs the
+                # merge, the two sides produce divergent files that never
+                # converge.
                 sort_key = None
                 try:
                     obj = json.loads(line)
                     ts = obj.get('ts') or obj.get('timestamp')
                     if isinstance(ts, str):
-                        sort_key = (0, ts)
+                        sort_key = (0, ts, line)
                 except (json.JSONDecodeError, ValueError, TypeError):
                     pass
                 if sort_key is None:
                     h = hashlib.sha256(line.encode('utf-8')).hexdigest()
-                    sort_key = (1, h)
+                    sort_key = (1, h, line)
                 seen[line] = sort_key
     except FileNotFoundError:
         # Absent base / absent ours / absent theirs are all valid.
--- a/test/jsonl-merge.test.ts
+++ b/test/jsonl-merge.test.ts
@@ -1,0 +1,96 @@
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { execFileSync } from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+
+const ROOT = path.resolve(import.meta.dir, '..');
+const DRIVER = path.join(ROOT, 'bin', 'gstack-jsonl-merge');
+
+let tmpDir: string;
+
+beforeEach(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'gstack-jsonl-merge-'));
+});
+
+afterEach(() => {
+  fs.rmSync(tmpDir, { recursive: true, force: true });
+});
+
+/**
+ * Run the merge driver the way git does: `driver <base> <ours> <theirs>`.
+ * The driver writes the merged result back to the <ours> file. Returns that
+ * file's content. `base`/`ours`/`theirs` are arrays of JSONL lines (the file
+ * is created from them); pass `null` to omit a file entirely (git passes an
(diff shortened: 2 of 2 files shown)


jbetala7:
Closing as superseded: issue #1769 is closed COMPLETED and current `main` already carries the equal-ts -> content-hash tiebreaker that makes append-only logs converge across machines (landed via the v1.55.0.0 fix wave #1808). Verified `bin/gstack-jsonl-merge` on main has the total-order tiebreak, so this PR is redundant.

PR #1780, opened 2026-05-29
fix(make-pdf): render emoji instead of tofu (▯) in PDFs

garrytan-agents:
## Problem

`make-pdf` renders **emoji code points as tofu boxes (▯)** in output PDFs. Standard typographic and mathematical Unicode (em-dash `—`, multiplication `×`, arrow `→`, bullet `•`, ellipsis `…`, smart quotes `" "`) all render **correctly** — only emoji break.

This is very visible in practice: documents use emoji heavily in section headers and table status/signal columns, so the broken glyphs are all over the page.

### Evidence (verbatim from a broken PDF, `Creator: Chromium, Producer: Skia/PDF`)

Renders fine:
- `@izebel_eth — Identity Dossier` (em-dash)
- `GStack × GBrain Hackathon` (×)
- `network (izebel → cascaintern → tradinghoex)` (→)
- `quant • run by @tradinghoex` (•)

Renders broken (emoji → ▯):
- Header: `▯ BOT & COLLUSION EVIDENCE` (was 🚨)
- Table: `138.8x  ▯ Extreme bot`, `4.6x ▯ Suspicious`, `Low followers ▯ Ratio`

## Root cause

Classic missing emoji-font fallback in headless Chromium:
1. The print CSS `body` font stack (`Helvetica, "Liberation Sans", Arial, … sans-serif`) had **no emoji family**, so Chromium had nothing to fall back to for emoji code points.
2. Even with a fallback declared, **most Linux distros/containers ship no emoji font at all**, so Skia emits `.notdef` (tofu) for every emoji.

macOS ships *Apple Color Emoji* and Windows ships *Segoe UI Emoji*, so they're unaffected — this is a Linux/container bug.

## Fix

- **`make-pdf/src/print-css.ts`** — append `"Apple Color Emoji", "Segoe UI Emoji", "Noto Color Emoji"` to the body and running-header font stacks so Chromium has an emoji glyph source on every platform.
- **`setup`** — add `ensure_emoji_font()`: best-effort install of an emoji font on Linux (`fonts-noto-color-emoji` via apt, with dnf/pacman/apk fallbacks), then refresh the fontconfig cache. It **warns rather than fails** if it can't install, so setup never breaks on this.

## Behavior

- macOS / Windows: no change (emoji font already present).
- Linux: real emoji glyphs after `setup` instead of ▯.
- Non-emoji Unicode: unchanged (already worked).

## Testing

- `bash -n setup` passes.
- `ensure_emoji_font` short-circuits if `fc-list` already reports an emoji font, so it's idempotent and cheap on re-runs.

Reported by Garry — these PDFs "always have broken Unicode for special characters and emojis."

### DIFF
--- a/make-pdf/src/print-css.ts
+++ b/make-pdf/src/print-css.ts
@@ -20,6 +20,11 @@
  *   - No <link>, no external CSS/fonts — everything inlined.
  *   - CJK fallback: Helvetica, Liberation Sans, Arial, Hiragino Kaku Gothic
  *     ProN, Noto Sans CJK JP, Microsoft YaHei, sans-serif.
+ *   - Emoji fallback: the body font stack ends in an emoji family
+ *     ("Apple Color Emoji", "Segoe UI Emoji", "Noto Color Emoji") so
+ *     Chromium has a glyph source for emoji code points instead of
+ *     emitting .notdef tofu (▯). On Linux this requires an installed
+ *     emoji font — `setup` installs fonts-noto-color-emoji for that.
  */
 
 export interface PrintCssOptions {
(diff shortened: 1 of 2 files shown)


garrytan:
Superseded by #1787 — the production-quality version. It carries everything here plus: scoped the emoji fallback to the 2 stacks that can hold emoji (body + running header) via shared font-stack constants, fc-match color=True detection (the broad fc-list query false-matched LastResort), `sudo -n` + timeout-bound installs so setup never hangs, browse-daemon restart after install (Chromium caches its font list at startup), an E2E render gate that proves emoji rendered via pdffonts + a pixel color check (pdftotext is a false oracle for emoji — it passes even on tofu), a hard-fail-in-CI guard so prerequisite drift can't hide a regression, and an Ubuntu CI emoji-font step. Thanks for the diagnosis and the fix.

PR #1783, opened 2026-05-29
fix(config): preserve spaces in config values

### LINKED ISSUE #1782: gstack-config truncates workspace_root paths at the first space
## Observed problem

`gstack-config get` reads values with `awk '{print $2}'`, so config values that legitimately contain spaces are truncated. The visible case is `workspace_root`, which is documented as a user-set directory for `/ship` sibling worktree scanning.

Repro on current `origin/main` (`ce5fbfa9`):

```bash
GSTACK_HOME=$(mktemp -d)
bin/gstack-config set workspace_root "/tmp/Conductor Workspaces"
bin/gstack-config get workspace_root
```

Current output:

```text
/tmp/Conductor
```

The config file itself contains the full value:

```yaml
workspace_root: /tmp/Conductor Workspaces
```

## Expected behavior

`gstack-config get workspace_root` should return the full value after `workspace_root:`, preserving spaces, so `gstack-next-version` can scan sibling worktrees under configured roots such as `/tmp/Conductor Workspaces` or user directories with spaces.

I searched open and closed issues/PRs for:

- `workspace_root space`
- `gstack-config config.yaml space`
- `gstack-config workspace_root spaces`
- open PRs touching `gstack-config`

I did not find an existing issue or canonical PR for this truncation bug.

## Candidate fix shape

Replace the `awk '{print $2}'` config read path with a small helper that strips `key:` and leading/trailing whitespace while preserving internal spaces. Reuse the helper for both `get` and the active-values section of `list`, then add a regression test for a `workspace_root` value containing spaces.

jbetala7:
## Summary
- Preserve everything after `key:` when `gstack-config get` reads config values, instead of taking only the second whitespace-delimited field
- Reuse the same reader for `gstack-config list` active values
- Add a regression test for `workspace_root` values containing spaces

Fixes #1782

## Root cause
`gstack-config get` used `awk '{print $2}'`, which silently truncated values like `workspace_root: /tmp/Conductor Workspaces` to `/tmp/Conductor`. `gstack-next-version` consumes `workspace_root` for sibling worktree scanning, so configured roots with spaces were ignored.

## Tests
- `bun test test/explain-level-config.test.ts test/docs-config-keys.test.ts`
- `bun test test/gstack-next-version.test.ts test/brain-sync.test.ts test/explain-level-config.test.ts`
- `bash -n bin/gstack-config`
- `git diff --check`
- Direct repro: `gstack-config set workspace_root "/tmp/Conductor Workspaces"` then `get workspace_root` returns the full path

## Collision checks
Searched open and closed issues/PRs for `workspace_root space`, `gstack-config config.yaml space`, and open PRs touching `gstack-config`. No existing issue or canonical PR covered this truncation bug.

### DIFF
--- a/bin/gstack-config
+++ b/bin/gstack-config
@@ -248,6 +248,16 @@ resolve_user_slug() {
   printf '%s' "$_slug"
 }
 
+read_config_value() {
+  local key="$1"
+  if [ ! -f "$CONFIG_FILE" ]; then
+    return 0
+  fi
+  grep -E "^${key}:" "$CONFIG_FILE" 2>/dev/null \
+    | tail -1 \
+    | sed -E "s/^${key}:[[:space:]]*//; s/[[:space:]]+$//"
+}
+
 case "${1:-}" in
   get)
     KEY="${2:?Usage: gstack-config get <key>}"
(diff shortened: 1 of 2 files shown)


time-attack:
@jbetala7 Please rewrite this around the tested implementation that preserves internal spaces for get and list, with the focused set/get/list regression test.

PR #1790, opened 2026-05-29
fix(office-hours): unify profile paths on developer-profile.json

### LINKED ISSUE #1677: /office-hours always reports SESSION_COUNT: 0, welcome_back tier is unreachable

NikhileshNanduri:
## Summary

Two call sites in `/office-hours` still referenced the legacy `builder-profile.jsonl` path after the v1.0.0.0 migration. That file gets archived on first read, so on fresh-`$HOME` installs both calls silently produced nothing.

1. **gbrain artifact glob** (`kind: filesystem, glob`) — changed from `~/.gstack/builder-profile.jsonl` to `~/.gstack/developer-profile.json`; removed `tail: 1` (not JSONL).
2. **Phase 6 profile reader** — changed from `gstack-builder-profile` (legacy shim) to `gstack-developer-profile --read` (direct). Behavior is identical; removes the indirection layer.

The writer side was already fixed in v1.44.1.0 (PR #1682, fix #1671). This PR closes the remaining reader/artifact half.

Closes #1677.

## Test plan

- [ ] `bun test test/skill-coverage-floor.test.ts` passes (315 assertions)
- [ ] `bun test` full free suite passes


jbetala7:
One stale-premise note from current `origin/main` (`62024d11`): `bin/gstack-builder-profile` is already a legacy shim that executes `gstack-developer-profile --read`, so the Phase 6 command replacement is not the behavior fix for `SESSION_COUNT` on current main.

The remaining real behavior change here is the gbrain artifact source: `office-hours/SKILL.md.tmpl` still points at `~/.gstack/builder-profile.jsonl` with `tail: 1`, while current sessions are written to `~/.gstack/developer-profile.json`. I would narrow the PR body/changelog to that artifact-glob fix and add a regression proving the generated office-hours skill reads the developer-profile JSON artifact after `gstack-developer-profile --log-session`.

Also, GitHub reports this branch `DIRTY`, and the PR bumps `VERSION`/`package.json` to `1.48.3.0` while main is already `1.52.2.0`, so the release files need a rebase/drop before this can land cleanly.

time-attack:
Thanks for this. The primary reader and writer bug is already fixed, leaving only a narrow artifact-glob cleanup, while this branch changes obsolete monolithic files and release metadata. Update only the current manifest and carved sources, regenerate host output, and add an isolated artifact-loader regression.

PR #1794, opened 2026-05-29
fix(browse): apply --no-sandbox to headed launch sites (root/container)

garrytan-agents:
## Problem

The `browse` daemon's **headless** launch path disables the Chromium sandbox automatically when it can't be used (CI, container, or running as root). The **headed** paths do not. So when the daemon runs headed as root inside a container — e.g. driving a virtual X display (`Xvfb`) for screenshots/automation — Chromium aborts immediately at launch:

```
[browse] Server failed to start:
[browse] Failed to start: persistentContext: Target page, context or browser has been closed
Browser logs:
  - [pid=...][err] Chromium sandboxing failed!
  - <process did exit: exitCode=1>
```

Headless mode works in the exact same environment; only headed mode crashes. That asymmetry makes headed-on-Xvfb effectively unusable in containers without manually patching the binary.

## Root Cause

`browser-manager.ts` had the sandbox-disable detection inline in `launch()` (the headless path):

```ts
const isRoot = typeof process.getuid === 'function' && process.getuid() === 0;
if (process.env.CI || process.env.CONTAINER || isRoot) {
  launchArgs.push('--no-sandbox');
}
```

But `launchHeaded()` and the extension-reload handoff built their own `launchArgs` arrays and never added `--no-sandbox`. Three launch sites, only one of them sandbox-aware.

## Solution

Extract the detection into a single shared helper `sandboxDisableArgs()` in `stealth.ts` (which already centralizes launch args), and apply it at **all three** launch sites:

```ts
// stealth.ts
export function sandboxDisableArgs(): string[] {
  const isRoot = typeof process.getuid === 'function' && process.getuid() === 0;
  if (process.env.CI || process.env.CONTAINER || isRoot) {
    return ['--no-sandbox'];
  }
  return [];
}
```

| Launch site | Before | After |
|---|---|---|
| `launch()` (headless) | inline `--no-sandbox` block | `...sandboxDisableArgs()` |
| `launchHeaded()` | no sandbox handling ❌ | `...sandboxDisableArgs()` ✅ |
| extension-reload handoff | no sandbox handling ❌ | `...sandboxDisableArgs()` ✅ |

This also de-duplicates the inline block `launch()` previously carried. Behavior is unchanged on dev machines (non-root, no CI/CONTAINER env): the helper returns `[]`, so the sandbox stays enabled.

## Testing

- New unit test `browse/test/sandbox-disable-args.test.ts` covers all branches (CONTAINER set, CI set, root uid, and the negative case). `bun test` → 4 pass / 0 fail.
- `tsc --noEmit` clean.
- Verified end-to-end: with this change, headed Chromium launches successfully against `Xvfb :99` as root in a container (`DISPLAY=:99 CONTAINER=1`), and element screenshots render correctly. Without it, the same launch crashes with "Chromium sandboxing failed!".

## Notes

`--no-sandbox` is only added when the sandbox is genuinely unavailable (container/CI/root). A local daemon driving user-specified URLs has marginal sandbox benefit in those environments, and this matches the policy `launch()` already applied for the headless path.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -342,17 +342,9 @@ export class BrowserManager {
     // BROWSE_EXTENSIONS_DIR points to an unpacked Chrome extension directory.
     // Extensions only work in headed mode, so we use an off-screen window.
     const extensionsDir = process.env.BROWSE_EXTENSIONS_DIR;
-    const { STEALTH_LAUNCH_ARGS } = await import('./stealth');
-    const launchArgs: string[] = [...STEALTH_LAUNCH_ARGS];
+    const { STEALTH_LAUNCH_ARGS, sandboxDisableArgs } = await import('./stealth');
+    const launchArgs: string[] = [...STEALTH_LAUNCH_ARGS, ...sandboxDisableArgs()];
     let useHeadless = true;
-
-    // Docker/CI/root: Chromium sandbox requires unprivileged user namespaces which
-    // are typically disabled in containers and are never available for the root
-    // user on Linux. Detect all three cases and add --no-sandbox automatically.
-    const isRoot = typeof process.getuid === 'function' && process.getuid() === 0;
-    if (process.env.CI || process.env.CONTAINER || isRoot) {
-      launchArgs.push('--no-sandbox');
-    }
 
     if (extensionsDir) {
       // Skip --load-extension when running against a custom Chromium build that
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej Please close this PR. Current main applies one shared Chromium sandbox policy to headed, headless, and handoff launches, including root and container use. This conflicting branch implements an older partial version.

PR #1795, opened 2026-05-29
docs: add stock phrase guardrail

creatornader:
Adds the `load[- ]bearing` phrase pattern to generated voice directives as a stock phrase to avoid. Regenerated skill docs and updated matching test fixtures.


time-attack:
Thanks for this. The phrase guardrail is mixed with unrelated changes, and the generated documentation is outdated. Please rebase, isolate both literal spellings, regenerate current outputs, and add a focused guardrail test.

creatornader:
Thanks. I rebased this PR onto current main and reduced it to the focused voice-directive change. The directive now names both exact spellings, "load-bearing" and "load bearing", and tells the writer to state what depends on the thing instead.

I regenerated the current skill outputs and added the focused guardrail test.

Verified with:

    bun test test/generate-voice-directive.test.ts test/preamble-compose.test.ts

creatornader:
Trunk rejected my submission because this fork account does not have write access to the repository. The focused change and verification are ready; a maintainer will need to submit this PR to the merge queue.

PR #1796, opened 2026-05-30
fix(ios-qa): reuse rotated bearer so /ios-qa drives a real device across sessions

Bmathews721:
## Problem

`/ios-qa` could drive the iOS **simulator** but not a real device: the daemon logged `boot_token_unavailable` / `CoreDeviceError 7000` and "could never authenticate."

## Root cause (it is not the file path)

`devicectl ... copy from --domain-type appDataContainer --source tmp/gstack-ios-qa.token` is **not** broken — `tmp/` is readable when the file exists. The boot token is **single-use**: the first `POST /auth/rotate` sets `bootTokenValid=false` **and deletes the on-disk token file** (`StateServer.swift`). But the daemon re-runs the full bootstrap (copy boot token + rotate) on every 30s tunnel-cache refresh, daemon restart, and new `/ios-qa` session. The 2nd+ bootstrap then finds the file gone → `boot_token_unavailable` (CoreDeviceError 7000 = file genuinely absent). So a real device was drivable **exactly once per app launch**. The simulator never hit this because it doesn't use the devicectl file-copy path.

## Fix (daemon-only; zero device-side change)

Implements the warm-start that **SKILL.md Phase 0 already documents** but the daemon never had:

- New `ios-qa/daemon/src/session-cache.ts` — `acquireTunnel()` persists the rotated bearer the daemon already holds to `~/.gstack/ios-qa-session.json` (**0600**) and **reuses** it after a cheap authed probe (`GET /state/snapshot`). It only falls back to a full boot-token bootstrap when there's no cache or the cached token is rejected (probe `401` = app was relaunched).
- `index.ts` `realTunnelProvider` now calls `acquireTunnel()`.

The device contract is unchanged: `StateServer` stays loopback-only, every endpoint still requires the rotated bearer, and the boot token stays single-use. The only new artifact is the Mac-side 0600 session cache — the exact value Phase 0 already specifies it holds.

### Also: cold-start launch env (`GSTACK_IOS_LAUNCH_ENV`)

`devicectl device process launch` now forwards `GSTACK_IOS_LAUNCH_ENV` (a JSON dict) via `--environment-variables`, so the daemon can cold-launch apps whose debug bridge is env-gated (e.g. BuckHound's `BH_ENABLE_IOS_QA_BRIDGE=1`) instead of launching them without the bridge.

## Tests

16 new tests (`session-cache.test.ts`, `launch-env.test.ts`); full daemon suite **107 pass**.

## Verified on hardware

On a physical iPhone 16 Pro: a **fresh daemon process** drove `/elements` + `/screenshot` (valid PNG) + `/tap` (`ok:true`) purely by reusing the cached bearer, while the boot token was confirmed **consumed** (`CoreDeviceError 7000`). Before the change, that same scenario returned `503 device_not_connected`.

### DIFF
--- a/bin/gstack-ios-qa-daemon
+++ b/bin/gstack-ios-qa-daemon
@@ -13,6 +13,12 @@
 #                                  the first paired connected device is used)
 #   GSTACK_IOS_TARGET_BUNDLE_ID  — bundle ID of the iOS app hosting StateServer
 #                                  (default com.gstack.iosqa.fixture)
+#   GSTACK_IOS_LAUNCH_ENV        — JSON dict of env vars set when the daemon
+#                                  cold-launches the app, forwarded to
+#                                  `devicectl ... --environment-variables`.
+#                                  Required for apps that gate their debug
+#                                  bridge behind a flag, e.g. BuckHound:
+#                                  '{"BH_ENABLE_IOS_QA_BRIDGE":"1"}'
 #
 # Readiness protocol: prints `READY: port=<n> pid=<pid>` to stdout once both
 # listeners are bound. Spawners read stdin with a ~5s timeout to confirm.
--- a/ios-qa/daemon/src/devicectl.ts
+++ b/ios-qa/daemon/src/devicectl.ts
@@ -270,8 +270,17 @@ export function launchApp(
   udid: string,
   bundleId: string,
   spawn: SpawnImpl = defaultSpawn,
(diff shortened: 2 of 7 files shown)


time-attack:
Can you show the log of ios-qa? ios-qa relies on ios setup being called to setup the source with the right agent hooks. The error you are getting looks like it might have skipped that step

PR #1786, opened 2026-05-29
fix(gen-skill-docs): quote frontmatter descriptions containing colons

genisis0x:
## Description

Closes #1778.

`applyCatalogTrim` in `scripts/gen-skill-docs.ts` rewrites the trimmed catalog description back into SKILL.md frontmatter as a **plain** YAML scalar:

```yaml
description: Set up gbrain for this coding agent: install the CLI, ... (gstack)
```

When the description contains `": "` (or a trailing colon), a strict YAML parser reads the second colon as a mapping separator and rejects the document:

```
invalid YAML: mapping values are not allowed in this context
```

Codex/OpenAI skill loading uses a strict parser, so every affected skill was silently skipped.

## Change

- Add `toYamlPlainOrQuoted(value)`: returns the value unchanged when it is a valid plain scalar, otherwise emits a double-quoted scalar (escaping `\` and `"`). Quoting triggers only on `: ` / trailing colon / leading YAML indicator chars, so existing colon-free descriptions are byte-for-byte unchanged.
- Use it at the trim rewrite site.
- Regenerated SKILL.md for the 8 skills whose descriptions contain colons (design-consultation, design-html, design-review, design-shotgun, guard, plan-tune, setup-gbrain, ship).

## Testing

- `bun test test/gen-skill-docs.test.ts` — 389 pass (incl. per-host `--dry-run` freshness across all hosts).
- `bun test test/skill-validation.test.ts` — 329 pass.
- Regenerated docs are fresh (`gen:skill-docs` for claude/codex/factory and the remaining hosts produces no further diff).

PR #1801, opened 2026-05-30
fix: quote SKILL.md description values containing colons

ds1ter:
## Problem

8 SKILL.md files have `description` frontmatter values containing `: ` (colon + space), which strict YAML parsers interpret as a new mapping key — causing this error:

```
invalid YAML: mapping values are not allowed in this context
```

Affected skills: `design-html`, `design-consultation`, `design-review`, `design-shotgun`, `guard`, `plan-tune`, `setup-gbrain`, `ship`

This breaks any tool that uses a strict YAML parser to load SKILL.md files (confirmed with OpenAI Codex CLI, which uses js-yaml in strict mode).

## Fix

Wrap the affected `description` values in double quotes so the colon is unambiguously part of the string value, not a mapping indicator.

**Before:**
```yaml
description: Design finalization: generates production-quality Pretext-native HTML/CSS. (gstack)
```

**After:**
```yaml
description: "Design finalization: generates production-quality Pretext-native HTML/CSS. (gstack)"
```

## Test

```python
import yaml
# parses cleanly with PyYAML, js-yaml, and other strict parsers
yaml.safe_load('description: "Design finalization: generates foo"')
```

### DIFF
--- a/design-consultation/SKILL.md
+++ b/design-consultation/SKILL.md
@@ -2,7 +2,7 @@
 name: design-consultation
 preamble-tier: 3
 version: 1.0.0
-description: Design consultation: understands your product, researches the landscape, proposes a complete design system (aesthetic, typography, color, layout, spacing, motion), and generates font+color preview... (gstack)
+description: "Design consultation: understands your product, researches the landscape, proposes a complete design system (aesthetic, typography, color, layout, spacing, motion), and generates font+color preview... (gstack)"
 allowed-tools:
   - Bash
   - Read
--- a/design-html/SKILL.md
+++ b/design-html/SKILL.md
@@ -2,7 +2,7 @@
 name: design-html
 preamble-tier: 2
 version: 1.0.0
-description: Design finalization: generates production-quality Pretext-native HTML/CSS. (gstack)
+description: "Design finalization: generates production-quality Pretext-native HTML/CSS. (gstack)"
 triggers:
   - build the design
   - code the mockup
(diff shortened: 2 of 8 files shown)


jbetala7:
Thanks for chasing the strict-YAML parse error — it's a real failure for js-yaml/Codex hosts. Heads up though: this looks **already resolved on `main`**, so the PR is effectively a no-op as written.

- All 8 files here already carry quoted `description:` values on current `main` (`3bef43bc`, shipped in the v1.55.0.0 fix wave #1808). e.g. `design-html/SKILL.md`:
  ```yaml
  description: "Design finalization: generates production-quality Pretext-native HTML/CSS. (gstack)"
  ```
  Same for `design-consultation`, `design-review`, `design-shotgun`, `guard`, `plan-tune`, `setup-gbrain`, and `ship`. GitHub reports this branch as `behind_by=5`, so it was opened just before that wave merged — rebasing onto current `main` should show these edits already present.

- The durable fix is **generator-level**, not hand-editing the generated `SKILL.md` outputs. These files are emitted by `scripts/gen-skill-docs.ts`, so a future regeneration would reintroduce unquoted descriptions. #1752 is the canonical place for that — it adds `scripts/skill-frontmatter-check.ts` validation so this can't regress (with earlier overlap in #1739 / #1786).

Not a knock on the diff — the repro and affected-skill list are accurate; it's just that the symptom was already swept up by #1808 and the root-cause guard is tracked in #1752.

garrytan:
Closing in favor of #1786. This patches the generated SKILL.md output, but those files are regenerated by `gen:skill-docs`, so the colon-quoting fix belongs in the generator (`scripts/gen-skill-docs.ts`) — which is what #1786 does. Otherwise it regresses on the next regen. Thanks for flagging it!

PR #1814, opened 2026-05-31
fix(gbrain): scope fallback and resolve Codex skill manifests

sensdiego:
## Summary

This PR closes two related fallback-hardening gaps in `gstack-brain-context-load` that remain even after the manifest parser fix proposed in #1688:

- The Layer 1 default fallback now scopes `skill-name-events` to `repo:{repo_slug}`. Before this change, the fallback's third list query only filtered by `type=timeline` and `content_contains={skill_name}`, so it could return matching skill events from other repositories.
- `--skill <name>` now resolves generated Codex skill installs before falling back to the default manifest. This covers repo-local `.agents/skills/gstack-<skill>/`, repo-local `.agents/skills/gstack/<skill>/`, global `~/.codex/skills/gstack-<skill>/`, global `~/.codex/skills/gstack/<skill>/`, global `~/.agents/skills/gstack-<skill>/`, and global `~/.agents/skills/gstack/<skill>/` layouts.

This complements #1688 rather than replacing it: #1688 fixes parsing nested `filter:` blocks when a manifest is found; this PR prevents Codex installs from missing the manifest and prevents the fallback itself from issuing an unscoped list query.

## Why

The file-level comment says every default query carries an explicit repo filter so cross-repo contamination is non-default. Two paths violated that intent:

1. `skill-name-events` in `defaultManifest()` had no `tags_contains=repo:{repo_slug}` filter.
2. `resolveSkillFile()` only looked in Claude-style paths, so Codex installs could fall back to Layer 1 default retrieval instead of using their skill manifest.

## Codex Precedence

`--skill-file` remains the explicit override. For `--skill <name>`, repo-local Codex installs under `.agents/skills/...` now take precedence over global installs under `~/.codex/skills/...` and `~/.agents/skills/...`, because the repo-local skill is the workspace-specific source of truth. Legacy Claude-style paths remain the final fallback.

## Tests

- First added failing regression tests on current `main`:
  - default fallback had only 2 `tags_contains=repo:test-repo` filters, expected 3
  - Codex generated skill under `~/.codex/skills/gstack-codex-only/SKILL.md` resolved as `mode=default`, expected `mode=manifest`
  - when both global and repo-local Codex manifests exist for the same skill, global won; expected repo-local
- Then implemented the fix and verified:

```bash
bun test test/gstack-brain-context-load.test.ts
bun test test/gstack-brain-context-load.test.ts test/gstack-memory-helpers.test.ts
```

Focused verification result: `39 pass, 0 fail` across the context-load and memory-helper suites.

### DIFF
--- a/bin/gstack-brain-context-load.ts
+++ b/bin/gstack-brain-context-load.ts
@@ -175,12 +175,20 @@ function resolveSkillFile(args: CliArgs): string | null {
     return resolve(args.skillFile);
   }
   if (!args.skill) return null;
+  const skill = args.skill;
+  const codexSkill = skill.startsWith("gstack-") ? skill : `gstack-${skill}`;
   // Look in common gstack skill locations
   const candidates = [
-    join(HOME, ".claude", "skills", args.skill, "SKILL.md"),
-    join(HOME, ".claude", "skills", "gstack", args.skill, "SKILL.md"),
-    join(process.cwd(), ".claude", "skills", args.skill, "SKILL.md"),
-    join(process.cwd(), args.skill, "SKILL.md"),
+    join(process.cwd(), ".agents", "skills", codexSkill, "SKILL.md"),
+    join(process.cwd(), ".agents", "skills", "gstack", skill, "SKILL.md"),
+    join(HOME, ".codex", "skills", codexSkill, "SKILL.md"),
+    join(HOME, ".codex", "skills", "gstack", skill, "SKILL.md"),
+    join(HOME, ".agents", "skills", codexSkill, "SKILL.md"),
+    join(HOME, ".agents", "skills", "gstack", skill, "SKILL.md"),
+    join(HOME, ".claude", "skills", skill, "SKILL.md"),
+    join(HOME, ".claude", "skills", "gstack", skill, "SKILL.md"),
+    join(process.cwd(), ".claude", "skills", skill, "SKILL.md"),
+    join(process.cwd(), skill, "SKILL.md"),
   ];
   for (const c of candidates) {
     if (existsSync(c)) return c;
@@ -391,7 +399,7 @@ function defaultManifest(args: CliArgs): GbrainManifest {
       {
         id: "skill-name-events",
         kind: "list",
-        filter: { type: "timeline", content_contains: "{skill_name}" },
+        filter: { type: "timeline", "tags_contains": "repo:{repo_slug}", content_contains: "{skill_name}" },
         limit: 5,
         render_as: "## Recent {skill_name} events",
       },
--- a/test/gstack-brain-context-load.test.ts
+++ b/test/gstack-brain-context-load.test.ts
@@ -14,11 +14,12 @@ import { spawnSync } from "child_process";
 
 const SCRIPT = join(import.meta.dir, "..", "bin", "gstack-brain-context-load.ts");
 
-function runScript(args: string[], env: Record<string, string> = {}): { stdout: string; stderr: string; exitCode: number } {
+function runScript(args: string[], env: Record<string, string> = {}, cwd: string = process.cwd()): { stdout: string; stderr: string; exitCode: number } {
   const result = spawnSync("bun", [SCRIPT, ...args], {
     encoding: "utf-8",
     timeout: 30000,
     env: { ...process.env, ...env },
+    cwd,
   });
   return {
     stdout: result.stdout || "",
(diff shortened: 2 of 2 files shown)


xg-gh-25:
This repo-scope isolation work resonates deeply with what we've learned running production agent sessions at scale.

The "fallback leaks cross-repo context" bug you're fixing here — where `skill-name-events` would pull from other repos when manifest parsing failed — is a **memory sovereignty violation** that we've seen manifest as:

1. **Agent confusion** (mixing context from unrelated codebases)
2. **Prompt drift** (session beliefs contaminated by irrelevant patterns)
3. **Efficiency loss** (LLM processing noise instead of signal)

Your fix — explicit `tags_contains=repo:{repo_slug}` filters at every layer — mirrors the isolation strategy we enforce in our session engine: every memory query is *tenant-scoped by default*, and cross-tenant retrieval requires explicit opt-in.

From the PR:
> `resolveSkillFile()` only looked in Claude-style paths, so Codex installs could fall back to Layer 1 default retrieval

This is the **silent fallback problem** — when manifest resolution fails, you don't want it to degrade gracefully to unscoped retrieval. You want it to *fail loudly* or fall back to **still-scoped** defaults. Your approach (resolve Codex paths first, then scoped defaults) is correct.

**Pattern we use:**
- Layer 1: tenant-scoped skill manifest (explicit)
- Layer 2: tenant-scoped default manifest (explicit)
- Layer 3: fail (no unscoped fallback)

The test regression coverage (`skill-name-events` had 2/3 filters, now 3/3) is exactly how you prevent this — codify the isolation contract, then enforce it.

One edge case question: when `--skill <name>` resolves multiple Codex install paths (e.g., both `~/.codex/skills/gstack-<skill>/` and `~/.agents/skills/gstack-<skill>/`), which takes precedence? Or is first-match sufficient?

---
*Memory isolation = session reliability. We run this in production via [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/28)*

sensdiego:
Resolved in `db787345`.

The resolver is still first-match, but the precedence is now explicit and tested:

1. explicit `--skill-file` remains the highest-precedence path
2. repo-local Codex installs under `.agents/skills/...`
3. global Codex installs under `~/.codex/skills/...`
4. global agent-side installs under `~/.agents/skills/...`
5. legacy Claude-style paths

Rationale: repo-local `.agents` is the most workspace-specific source, so it should override a user's global Codex install when both exist for the same skill name.

I added a regression test that creates both `~/.codex/skills/gstack-codex-only/SKILL.md` and `<repo>/.agents/skills/gstack-codex-only/SKILL.md`, then asserts `--skill codex-only` loads the repo-local manifest rather than the global one.

Verification after the update:

```bash
bun test test/gstack-brain-context-load.test.ts test/gstack-memory-helpers.test.ts
```

Result: `39 pass, 0 fail`.

jbetala7:
Validated on `main` (`3bef43bc`): the bug is live and this does **not** duplicate #1688 — the two are complementary and both are needed.

- On `main`, `defaultManifest()`'s `skill-name-events` query is `filter: { type: "timeline", content_contains: "{skill_name}" }` (`bin/gstack-brain-context-load.ts` ~L394) with no `repo:` scope, so the default-fallback path pulls `{skill_name}` timeline events across repos. Your diff adds `tags_contains: "repo:{repo_slug}"` here, and the new test asserting `tags_contains=repo:test-repo` appears 3 times pins all three list queries to the active repo.
- #1688 only touches the *manifest parser* (`extractGbrainBlock` / the new `parseFilterMap`) — it scopes queries parsed out of a SKILL.md `gbrain:` block and never touches `defaultManifest`. So the fallback gap you fix here remains after #1688 merges, exactly as your description says.

Net: not a duplicate — they fix different code paths (parser vs default fallback). Both edit `bin/gstack-brain-context-load.ts`, so whichever lands second needs a small rebase. The repo-scope fix LGTM.

One neutral note for the reviewer: this PR also bundles the repo-local-vs-global Codex skill-file precedence change (`resolveSkillFile`), which is a separate concern from the repo-scoping in the title. It's tested, but worth flagging so it isn't read as scope creep.

sensdiego:
Clarified the PR title/body so the `resolveSkillFile` precedence work is visible as a related fallback-hardening change, not an implicit part of the `defaultManifest` repo-scope fix.

The updated description now calls out:

- `defaultManifest()` repo-scoping for `skill-name-events`
- Codex skill manifest resolution across repo-local and global installs
- explicit precedence: `--skill-file`, then repo-local `.agents`, then global `~/.codex`, then global `~/.agents`, then legacy Claude paths
- the repo-local-vs-global regression test

No code change in this update; PR head remains `db787345` with the same `39 pass, 0 fail` focused verification.

xg-gh-25:
@sensdiego thanks for the explicit precedence chain — repo-local > global Codex > global agent-side is the right ordering. The test coverage on the filter assertions is what makes this durable; without that, the next refactor would silently regress the scoping. Clean resolution.

PR #1817, opened 2026-05-31
setup-gbrain: pass voyage-code-3 flags without relying on shell word-splitting

### LINKED ISSUE #1798: setup-gbrain: voyage-code-3 flags silently dropped under zsh (unquoted $GBRAIN_EMBED_FLAGS word-splitting), downgrades to voyage-4-large
## Summary

In `setup-gbrain`, the `voyage-code-3` embedding flags are passed to `gbrain init` via an **unquoted string variable** that relies on bash-style word-splitting. Under **zsh** (the default macOS interactive shell, and the shell the skill's bash blocks actually execute under for most users), unquoted parameter expansion is **not** word-split, so all four tokens collapse into a single argv entry. `gbrain` can't parse the mashed-together arg, silently ignores it, and falls back to its own default embedding model (`voyage-4-large`) — **with no error**.

Net effect: every zsh user who intends to get `voyage-code-3` (the gstack-recommended, code-specialized embedding model) silently gets `voyage-4-large` instead. Both are 1024d and "work," so nothing fails loudly — the downgrade is invisible unless you inspect `gbrain config`.

## Location

`setup-gbrain/SKILL.md.tmpl` (and the generated `setup-gbrain/SKILL.md`), in all three init blocks — Path 3 (PGLite local), the Step 1.5 rollback wrapper, and the Path 4 "Yes" branch. The pattern:

```bash
GBRAIN_EMBED_FLAGS=""
if [ -n "${VOYAGE_API_KEY:-}" ]; then
  GBRAIN_EMBED_FLAGS="--embedding-model voyage:voyage-code-3 --embedding-dimensions 1024"
fi
gbrain init --pglite --json $GBRAIN_EMBED_FLAGS   # <-- unquoted; relies on word-splitting
```

## Reproduction / proof

```
$ bash -c 'F="--embedding-model voyage:voyage-code-3 --embedding-dimensions 1024"; set -- $F; echo "argc=$#"'
argc=4          # bash word-splits -> 4 separate args (correct)

$ zsh -c  'F="--embedding-model voyage:voyage-code-3 --embedding-dimensions 1024"; set -- $F; echo "argc=$#"'
argc=1          # zsh keeps it as ONE arg (broken) -> gbrain falls back to voyage-4-large
```

Observed in the wild on a fresh macOS setup: `gbrain doctor` reported `voyage:voyage-4-large` despite `VOYAGE_API_KEY` being set and the skill intending `voyage-code-3`. Re-initing with the flags passed as **literal, separately-quoted args** produced the correct `voyage:voyage-code-3`.

## Why the existing test doesn't catch it

`test/gbrain-init-voyage-code-3.test.ts` executes the snippet via `spawnSync("bash", ["-c", script])` — i.e. it **always runs under bash**, where word-splitting happens to work. So the test passes while real-world zsh execution fails. The test gives false confidence precisely on the dimension that breaks.

## Suggested fix

Stop relying on word-splitting. Two options:

**A) bash/zsh array (DRY):**
```bash
GBRAIN_EMBED_FLAGS=()
if [ -n "${VOYAGE_API_KEY:-}" ]; then
  GBRAIN_EMBED_FLAGS=(--embedding-model voyage:voyage-code-3 --embedding-dimensions 1024)
fi
gbrain init --pglite --json "${GBRAIN_EMBED_FLAGS[@]}"
```
(Works in both bash and zsh. Note: under `set -u`, macOS's stock bash 3.2 errors on an empty `"${arr[@]}"`; if the blocks run under `set -u`, use option B or guard with `"${GBRAIN_EMBED_FLAGS[@]:-}"`.)

**B) Inline both branches (bulletproof, shell-agnostic):**
```bash
if [ -n "${VOYAGE_API_KEY:-}" ]; then
  gbrain init --pglite --json --embedding-model voyage:voyage-code-3 --embedding-dimensions 1024
else
  gbrain init --pglite --json
fi
```

Also recommend updating `gbrain-init-voyage-code-3.test.ts` to run the snippet under **zsh as well as bash** (parameterize the shell), so this regression can't recur silently.

## Environment

- gstack `1.52.2.0`
- gbrain `0.41.29.0`
- macOS, zsh (default shell)

mvanhorn:
## Summary

Under zsh, `/setup-gbrain` silently configured `voyage-4-large` instead of the recommended `voyage-code-3`. The embedding flags were passed to `gbrain init` through an unquoted `$GBRAIN_EMBED_FLAGS` that relies on bash word-splitting. zsh does not word-split unquoted expansions, so the four tokens collapsed into a single argv entry, `gbrain` could not parse it, and it fell back to the default model with no error.

## Why this matters

zsh is the default macOS interactive shell and the shell most users' setup blocks run under, so every zsh user who intended to use the gstack-recommended `voyage-code-3` silently got `voyage-4-large` instead. Reported in #1798.

## Changes

- Replaced the word-split form in all three init blocks (PGLite local, the Step 1.5 rollback wrapper, and the Path 4 branch) with explicit both-branch `gbrain init` invocations: with the model flags when `VOYAGE_API_KEY` is set, without them otherwise. This is shell-agnostic and `set -u`-safe on macOS's stock bash 3.2.
- Applied the change in `setup-gbrain/SKILL.md.tmpl` and regenerated `setup-gbrain/SKILL.md` so the template and generated file stay in sync.
- The regression test now runs the snippet under both bash and zsh and asserts the resulting `argv`, so a word-splitting regression cannot pass silently.

## Testing

`bun test test/gbrain-init-voyage-code-3.test.ts` -> 8 pass.

Fixes #1798

### DIFF
--- a/setup-gbrain/SKILL.md
+++ b/setup-gbrain/SKILL.md
@@ -823,16 +823,22 @@ mv "$HOME/.gbrain/config.json" "$BACKUP"
 # gstack default: voyage-code-3 (1024d) when VOYAGE_API_KEY is set — best for
 # code retrieval. Without the key, fall back to gbrain's own auto-selected
 # embedding provider chain (OpenAI 1536d when OPENAI_API_KEY is present, etc.).
-GBRAIN_EMBED_FLAGS=""
 if [ -n "${VOYAGE_API_KEY:-}" ]; then
-  GBRAIN_EMBED_FLAGS="--embedding-model voyage:voyage-code-3 --embedding-dimensions 1024"
-fi
-if ! gbrain init --pglite --json $GBRAIN_EMBED_FLAGS; then
-  # Restore on failure
-  mv "$BACKUP" "$HOME/.gbrain/config.json"
-  echo "gbrain init failed. Your previous config was restored at $HOME/.gbrain/config.json." >&2
-  echo "PGLite directory at ~/.gbrain/pglite/ may be in a partial state — \`rm -rf ~/.gbrain/pglite\` if needed before retrying." >&2
-  exit 1
+  if ! gbrain init --pglite --json --embedding-model voyage:voyage-code-3 --embedding-dimensions 1024; then
+    # Restore on failure
+    mv "$BACKUP" "$HOME/.gbrain/config.json"
+    echo "gbrain init failed. Your previous config was restored at $HOME/.gbrain/config.json." >&2
+    echo "PGLite directory at ~/.gbrain/pglite/ may be in a partial state — \`rm -rf ~/.gbrain/pglite\` if needed before retrying." >&2
+    exit 1
+  fi
+else
+  if ! gbrain init --pglite --json; then
+    # Restore on failure
+    mv "$BACKUP" "$HOME/.gbrain/config.json"
+    echo "gbrain init failed. Your previous config was restored at $HOME/.gbrain/config.json." >&2
+    echo "PGLite directory at ~/.gbrain/pglite/ may be in a partial state — \`rm -rf ~/.gbrain/pglite\` if needed before retrying." >&2
+    exit 1
+  fi
 fi
 echo "Switched to local PGLite. Previous config saved at $BACKUP — review before deleting."
 ```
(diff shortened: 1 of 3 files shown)


jbetala7:
Validated on `main` (`3bef43bc`): live bug, and inlining the flags is the right fix.

`setup-gbrain/SKILL.md` calls `gbrain init --pglite --json $GBRAIN_EMBED_FLAGS` with the variable unquoted (L830, L1044, L1135). In bash that word-splits into four argv tokens as intended; under zsh (the default interactive shell on macOS) unquoted parameters are **not** word-split, so the four tokens collapse into a single argv entry, `gbrain` can't parse it, and it falls back to `voyage-4-large` — the silent downgrade you describe. Note that simply quoting wouldn't fix it either (that passes one combined arg), so inlining the flags directly into the two `gbrain init` branches is the correct shell-agnostic fix.

No competing open PR touches `setup-gbrain`. Both `SKILL.md` and `SKILL.md.tmpl` are updated and there's a regression test. LGTM.

time-attack:
@16francej merge pls

mvanhorn:
Superseded, closing.

main fixed the same word-splitting bug a different way: `GBRAIN_EMBED_FLAGS` is gone entirely and the flags now ride the positional params.

```sh
set --  # flags ride the positional params — unquoted $VAR breaks under zsh word-splitting (#1798)
if [ -n "${VOYAGE_API_KEY:-}" ]; then
  set -- --embedding-model voyage:voyage-code-3 --embedding-dimensions 1024
fi
gbrain init --pglite --json "$@"
```

That is a cleaner fix than my branch-the-whole-invocation version, and it covers the other two call sites too. Thanks @jbetala7 for validating the original report.

PR #1818, opened 2026-05-31
Handle missing AskUserQuestion tools in Codex skills

Catfish-75:
## Summary

This PR makes the generated Codex gstack skills handle sessions where no `AskUserQuestion`-like tool is callable.

Codex hosts can have gstack skills installed while the interactive question surface is not injected into a particular session. In that case, the current generic preamble tells the agent to stop with `BLOCKED -- AskUserQuestion unavailable`. For Codex this is too strict because an interactive chat turn is still available.

The new Codex-only resolution order is:

1. Prefer any callable `mcp__*__AskUserQuestion` variant.
2. Use Codex `request_user_input` when it is present and current mode/instructions allow it.
3. Otherwise, ask the same decision brief as a normal chat question and wait for the user's answer.

Non-Codex hosts keep the existing strict behavior from #1390.

## Related PRs

- #1101 is the closest related PR. It maps `AskUserQuestion` to `request_user_input`, but its own description notes that this does not fully solve Codex Default mode because `request_user_input` is not always callable.
- #1390 intentionally deleted prose fallback for Claude/Conductor because it caused agents to skip real AUQ calls. This PR preserves that stricter behavior for non-Codex hosts and scopes the fallback to Codex only.

## Why this shape

A plain text rewrite to `request_user_input` is not enough when the tool is absent or mode-restricted. A universal prose fallback would regress the behavior fixed by #1390. The safest narrow fix is host-aware guidance in the generated preamble.

## Validation

- `bun run scripts/gen-skill-docs.ts --host codex --dry-run`
- Cherry-picked onto current `origin/main`; branch contains only this one commit.

Local full validation caveat: Windows/Codex shells on this machine do not provide the Unix tool assumptions used by some existing tests (`bash`, `grep`, `file`, executable-bit checks), so I kept validation focused on Codex skill generation freshness.

### DIFF
--- a/scripts/resolvers/preamble/generate-ask-user-format.ts
+++ b/scripts/resolvers/preamble/generate-ask-user-format.ts
@@ -1,19 +1,37 @@
 import type { TemplateContext } from '../types';
 
 export function generateAskUserFormat(_ctx: TemplateContext): string {
+  const toolResolution = _ctx.host === 'codex'
+    ? `"AskUserQuestion" can resolve to multiple interactive surfaces at runtime:
+
+- the **host MCP variant** (e.g. \`mcp__conductor__AskUserQuestion\`) when the host registers it;
+- Codex's **\`request_user_input\`** tool, but only when it is present in the current tool list and the active mode/instructions allow calling it;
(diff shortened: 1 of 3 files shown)


16francej:

PR #1820, opened 2026-05-31
fix(ship): absolute bin path for gstack-next-version in landing-report/land-and-deploy/review

wsvoorhees:
## Problem

`/ship` (now fixed), `/landing-report`, `/land-and-deploy`, and `/review` invoke the queue-aware version allocator as a **bare relative path**:

```bash
QUEUE_JSON=$(bun run bin/gstack-next-version ... 2>/dev/null || echo '{"offline":true}')
```

When the *installed* skills run from a user's project repo, `bin/gstack-next-version` doesn't resolve (the util lives at the gstack install root, `~/.claude/skills/gstack/bin/`). So `bun` errors, the `|| echo '{"offline":true}'` fallback fires, and the skill silently degrades:

> ⚠ workspace-aware ship offline — using local bump only

This drops queue-collision awareness on every run — version collisions then surface late (e.g. at the CI version-gate or `/land-and-deploy` drift gate) instead of at ship time. It looks like an auth/network failure but isn't.

`ship/SKILL.md(.tmpl)` was already corrected to use the absolute `~/.claude/skills/gstack/bin/gstack-next-version` (matching every other gstack bin call in these files). This applies the same fix to the three skills that were missed.

## Fix

Prefix the invocation with `~/.claude/skills/gstack/bin/` in the `.tmpl` sources for `landing-report`, `land-and-deploy`, and `review`, then regenerate.

- Non-Claude host variants already rendered `$GSTACK_ROOT/bin/...` correctly via the per-host substitution and are unaffected (no diff).
- CI references (`.github/workflows/version-gate.yml`, `.gitlab-ci.yml`) are intentionally left bare — they run from the gstack repo root where `bin/` exists.

## Verification

Ran the exact `bun run ~/.claude/skills/gstack/bin/gstack-next-version --base <base> --bump <level> --current-version <v>` from a real downstream repo: returns full JSON with `"offline": false` and correct collision handling (`"reason": "bumped past claimed X.Y.Z.W"`), where the bare path previously fell through to `{"offline":true}`.

Edited `.tmpl` + `bun run gen:skill-docs --host all`; the only diffs are the three template path swaps, their regenerated canonical Claude `SKILL.md`, and `llms.txt`.

### DIFF
--- a/land-and-deploy/SKILL.md
+++ b/land-and-deploy/SKILL.md
@@ -1159,7 +1159,7 @@ BASE_VERSION=$(git show origin/$BASE_BRANCH:VERSION 2>/dev/null | tr -d '\r\n[:s
 # We don't need the exact original level — we just need "a level" that passes to the util.
 # If the minor digit advanced, call it minor; patch digit, patch; etc. If base > branch, skip (not ours to land).
 # For simplicity: use "patch" as a conservative default; util handles collision-past regardless of input level.
-QUEUE_JSON=$(bun run bin/gstack-next-version \
+QUEUE_JSON=$(bun run ~/.claude/skills/gstack/bin/gstack-next-version \
   --base "$BASE_BRANCH" \
   --bump patch \
   --current-version "$BASE_VERSION" 2>/dev/null || echo '{"offline":true}')
--- a/land-and-deploy/SKILL.md.tmpl
+++ b/land-and-deploy/SKILL.md.tmpl
@@ -341,7 +341,7 @@ BASE_VERSION=$(git show origin/$BASE_BRANCH:VERSION 2>/dev/null | tr -d '\r\n[:s
 # We don't need the exact original level — we just need "a level" that passes to the util.
 # If the minor digit advanced, call it minor; patch digit, patch; etc. If base > branch, skip (not ours to land).
 # For simplicity: use "patch" as a conservative default; util handles collision-past regardless of input level.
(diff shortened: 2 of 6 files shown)


jbetala7:
**Verdict: looks correct and complete.**

**Root cause confirmed on `main` (`3bef43bc`).** The bare `bun run bin/gstack-next-version` resolves relative to the *current* working directory. When the installed skills run inside a user's own project repo, `bin/gstack-next-version` doesn't exist there, `bun` errors, and the `|| echo '{"offline":true}'` fallback silently fires — dropping queue-collision awareness, exactly as the PR describes. `ship/SKILL.md.tmpl:181` already uses the absolute `~/.claude/skills/gstack/bin/gstack-next-version`, so this restores the convention to the three skills that were missed.

**Completeness check.** Grepped the repo for the bare invocation after this PR's change — the only remaining occurrences are:

- `.github/workflows/version-gate.yml`
- `.gitlab-ci.yml`

Both correctly stay bare (they run from the gstack repo root where `bin/` exists), and the PR body already calls this out. All three user-facing skills are covered in both `.tmpl` and generated `SKILL.md` (`landing-report`, `land-and-deploy`, `review`), and the non-Claude host variants render `$GSTACK_ROOT/bin/...` so they're unaffected. Nothing left behind — no further changes needed.

time-attack:
@16francej merge

PR #1821, opened 2026-05-31
v1.55.1.0 fix(developer-profile): exclude mode:resources entries from SESSION_COUNT, TIER, and NUDGE_ELIGIBLE

mvann:
## Summary

Fixes `gstack-developer-profile read` so the `mode:resources` bookkeeping entries that `/office-hours` auto-appends (Phase 6, to dedupe which founder-resource links you've seen) stop being counted as real sessions.

**The bug.** Those bookkeeping entries inflated `SESSION_COUNT`, prematurely promoted `TIER` (introduction → welcome_back → regular → inner_circle), and helped arm `NUDGE_ELIGIBLE`. The builder→founder nudge could fire off bookkeeping alone — e.g. a single real session plus three session-closings reported as tier `regular` with the nudge armed.

## How I hit this

I ran `/office-hours` in startup mode to pressure-test a real product idea (a tool for splitting collaborative revenue, born from an actual album project). One real design session, one approved design doc.

Afterward I got curious about how office-hours grades a session and how the tier / `NUDGE_ELIGIBLE` system decides what to show me, so I looked at what `gstack-developer-profile --read` reported against the session history I'd actually generated.

**What I expected:** after a single real design session, `SESSION_COUNT: 1`, `TIER: welcome_back`, and the builder→founder nudge *off* — I hadn't come back to build anything yet.

**What actually happened:** the profile reported more sessions than I'd run and a higher tier than one session earns, and the nudge logic was keying off entries that weren't sessions at all. The extras were the `mode:resources` bookkeeping rows office-hours appends in Phase 6 to dedupe which founder-resource links I'd been shown.

**Why that's a problem:** the developer profile is what personalizes the experience. `TIER` gates the welcome-back and inner-circle treatment; `NUDGE_ELIGIBLE` decides whether to pitch the builder→founder nudge. When bookkeeping inflates them, a brand-new user gets treated like a returning regular, and the founder nudge can fire at someone who never actually came back to build. The personalization triggers on noise instead of real engagement.

**The fix** (`bin/gstack-developer-profile`, `do_read()`):
- Hoists the existing `realSessions = sessions.filter(e => e.mode !== 'resources')` filter up so `count` / `SESSION_COUNT` / `TIER` use it (previously `tier` counted resources while `DESIGN_COUNT` already didn't — now consistent).
- Scopes `builderSessions` to `mode === 'builder'`. It was `mode !== 'startup'`, which both counted resources entries **and** excluded real startup sessions. The nudge is about someone who keeps returning to *build*, so `builder` is the correct predicate.

**Tests** (`test/gstack-developer-profile.test.ts`, +7):
- 4 tests proving resources entries don't inflate `SESSION_COUNT` / `TIER` / `NUDGE_ELIGIBLE`.
- 3 boundary tests (added after adversarial review) locking the two `>=` gates: 2 builder sessions stays false, 3 builder sessions with <5 signals stays false, 4 real sessions reach `regular` even with resources present.

Substantive commits (excluding the version/CHANGELOG bookkeeping commit):
- `fix(developer-profile): exclude mode:resources entries from SESSION_COUNT, TIER, and NUDGE_ELIGIBLE`
- `test(developer-profile): add gate boundary cases for nudge and tier`

## Process note

I had Claude follow CONTRIBUTING.md for this contribution, which meant running `/ship` rather than opening the PR by hand. That's why the VERSION is incremented (1.55.0.0 → 1.55.1.0) and there's a CHANGELOG entry — `/ship` handled the version bump, changelog, pre-landing review, adversarial review, and PR update as part of its workflow.

## Test Coverage

All new code paths covered. `bun test test/gstack-developer-profile.test.ts`: **40 pass / 0 fail** (was 37 before the boundary tests; 33 on base `main`).

Full `bun test` shows 6 failures, all **pre-existing and unrelated** — verified identical on base `main` with this change absent. They are local-environment artifacts, not product bugs:
- 5 × `gstack-gbrain-detect` — `gbrain` binary not on PATH (exit 127).
- 1 × `resolve-user-slug fallback chain` — test writes against a real `~/.gstack/config.yaml`.

This PR touches only `bin/gstack-developer-profile` and its test file (plus VERSION/package.json/CHANGELOG); it adds zero new failures.

## Pre-Landing Review

No issues found. PR Quality Score 10/10. Testing + maintainability specialists assessed inline (small, self-contained, read-only diff). The fix removes duplication rather than adding it (the `realSessions` filter is declared once and reused).

## Adversarial Review (Claude)

Independent fresh-context adversarial pass. Codex was unavailable, so Claude-only (informational, non-blocking). It surfaced one fixable item — **already addressed in this PR**: the original tests only hit the exact gate thresholds, so a future `>=`→`>` regression could slip through. Added the 3 boundary tests above.

Out-of-scope observations left for follow-up (deliberately NOT bundled into this single-purpose fix):
- `totalSignals` (the nudge's second gate) is not mode-filtered and counts duplicate signal names cumulatively — pre-existing, orthogonal to the session-count leak.
- The legacy `.jsonl` migration path has no regression test pinning this fix.

## Eval Results

No prompt-related files changed — evals skipped.

## Plan Completion

No plan file detected.

## TODOS

## Test plan

- [x] `bun test test/gstack-developer-profile.test.ts` — 40 pass / 0 fail
- [x] Full `bun test` — only the 6 pre-existing, unrelated env failures remain (verified against base `main`)
- [x] Reproduced the original bug and confirmed the fix flips the bad verdict


garrytan:
Closing — this targets v1.55.1.0, which is behind main (now v1.57.0.0), and has merge conflicts. If the developer-profile fix is still relevant, please rebase on current main and reopen. Thanks!

mvann:
Rebased onto current main (v1.57.10.0 → v1.57.11.0) and reopened as #1991 per the request above. The fix still applies cleanly — `bin/gstack-developer-profile` was untouched upstream, so both defects are still live on main. GitHub wouldn't let me reopen this thread after the force-push, hence the new PR.

PR #1825, opened 2026-06-01
fix(redact): invalid --max-bytes must not disable the fail-closed oversize guard

### LINKED ISSUE #1824: gstack-redact: invalid --max-bytes silently turns the fail-closed oversize guard into fail-open
## Observed problem

`gstack-redact`'s oversize check is documented and designed to **fail closed** (`lib/redact-engine.ts`: "Fail CLOSED on oversize input"). A malformed `--max-bytes` value silently turns it into **fail open** — the guard is skipped and oversized input is scanned (or, with a negative value, every input is blocked).

Root cause is a two-step gap:

1. `bin/gstack-redact` parses the flag with `parseInt(maxBytes, 10)` and passes the result straight through:
   ```ts
   const maxBytes = arg("--max-bytes");
   return {
     ...
     ...(maxBytes ? { maxBytes: parseInt(maxBytes, 10) } : {}),
   };
   ```
   `parseInt("foo", 10)` → `NaN`, which is passed to the engine as `maxBytes: NaN`.

2. The engine only guards against null/undefined:
   ```ts
   const maxBytes = opts.maxBytes ?? DEFAULT_MAX_BYTES;
   ...
   if (byteLen > maxBytes) { /* fail-closed block */ }
   ```
   `??` does **not** catch `NaN`, so `maxBytes` stays `NaN`, and `byteLen > NaN` is always `false`. The fail-closed block never fires.

- `gstack-redact --max-bytes notanumber < huge_input` — the oversize fail-closed guard is **skipped entirely** (input that should be blocked at the cap is scanned anyway). The user believes they set a cap; they silently got no cap.
- `gstack-redact --max-bytes -5 < anything` — `byteLen > -5` is always true, so **every** input is blocked as "too large", with a nonsensical `> -5 bytes` message.
- Calling `scan(input, { maxBytes: NaN })` directly returns `oversize: false` for arbitrarily large input.

## Expected behavior

An invalid size cap must never weaken the fail-closed guard:

- The engine should treat a non-finite or non-positive `maxBytes` as invalid and fall back to the known-good `DEFAULT_MAX_BYTES` (1 MiB), so the guard stays intact for every caller.
- The CLI should reject a malformed `--max-bytes` with a clear error and a usage exit code (1), instead of silently passing `NaN`.

- Open/closed PRs: `redact in:title`, full-text `max-bytes` — only old, unrelated browse-storage redaction PRs (#21, #168, #238, #860); no open PR touches `lib/redact-engine.ts` or the size guard.
- Open/all issues: `redact`, `max-bytes fail-closed` — none about the size cap.
- No open PR in the current set touches `lib/redact-engine.ts` or `bin/gstack-redact`.

## Candidate fix shape

- `lib/redact-engine.ts`: replace `opts.maxBytes ?? DEFAULT_MAX_BYTES` with a validity check (`Number.isFinite && > 0`, else `DEFAULT_MAX_BYTES`).
- `bin/gstack-redact`: validate `--max-bytes` is a positive integer; otherwise write an error to stderr and `process.exit(1)`.
- Regression tests in `test/redact-engine.test.ts` (invalid `maxBytes` still fails closed) and `test/gstack-redact-cli.test.ts` (malformed `--max-bytes` exits 1).

jbetala7:
## Problem

`gstack-redact`'s oversize check is designed to **fail closed** (`lib/redact-engine.ts`: "Fail CLOSED on oversize input"). A malformed `--max-bytes` value silently turned it into **fail open**.

Repro on `main`:

```
$ printf 'a%.0s' {1..500} | bun bin/gstack-redact --max-bytes notanumber
# scans anyway — the size cap the user thought they set is silently ignored
$ bun -e 'import {scan} from "./lib/redact-engine"; console.log(scan("x".repeat(2_000_000), {maxBytes: NaN}).oversize)'
false   # 2 MB input, oversize guard never fires
```

## Root cause

1. `bin/gstack-redact` parsed the flag with `parseInt(maxBytes, 10)` and passed the result straight through, so `--max-bytes notanumber` became `maxBytes: NaN`.
2. In the engine, `opts.maxBytes ?? DEFAULT_MAX_BYTES` only catches `null`/`undefined`. `NaN` slips past, and `byteLen > NaN` is always `false`, so the `engine.input_too_large` block never fires. A negative value does the opposite: `byteLen > -5` is always true, so every input blocks with a nonsensical `> -5 bytes` message.

## Fix

- **engine** (`lib/redact-engine.ts`): treat a non-finite or non-positive `maxBytes` as invalid and fall back to the known-good 1 MiB default, so the guard stays intact for every caller regardless of how the cap was computed (defense in depth for the security core).
- **CLI** (`bin/gstack-redact`): reject a malformed `--max-bytes` with a clear stderr error and exit 1 (usage error, distinct from the 0/2/3 finding-tier codes) instead of silently passing `NaN`.

## Testing

```
bun test test/redact-engine.test.ts test/gstack-redact-cli.test.ts   # 69 pass
```

- `test/redact-engine.test.ts`: new regression — `scan()` with `maxBytes` of `NaN`/`0`/`-1` still fails closed at the default cap (a >1 MiB input blocks). Verified this **fails** on `main` (the NaN case) and passes with the fix.
- `test/gstack-redact-cli.test.ts`: new contract — malformed `--max-bytes` (`notanumber`/`-5`/`0`/`10.5`) exits 1 with the error message.
- Full free suite (`bun test`) green; the one transient `browse/test/security-audit-r2.test.ts` timeout was machine-load only and passes in isolation (no browse code touched by this PR).

Fixes #1824

### DIFF
--- a/bin/gstack-redact
+++ b/bin/gstack-redact
@@ -160,13 +160,27 @@ function readLines(path: string | undefined): string[] | undefined {
 
 function buildOpts(): ScanOptions {
   const vis = (arg("--repo-visibility") as RepoVisibility) || "unknown";
-  const maxBytes = arg("--max-bytes");
+  const maxBytesRaw = arg("--max-bytes");
+  let maxBytes: number | undefined;
+  if (maxBytesRaw !== undefined) {
+    // Reject a malformed cap loudly. Silently passing NaN/0/negative would
+    // weaken the engine's fail-closed oversize guard (exit 1 = usage error,
+    // distinct from the 0/2/3 finding-tier codes).
+    const parsed = Number(maxBytesRaw);
+    if (!Number.isInteger(parsed) || parsed <= 0) {
+      process.stderr.write(
+        `gstack-redact: --max-bytes must be a positive integer, got: ${maxBytesRaw}\n`,
+      );
+      process.exit(1);
+    }
+    maxBytes = parsed;
+  }
   return {
     repoVisibility: ["public", "private", "unknown"].includes(vis) ? vis : "unknown",
     allowlist: readLines(arg("--allowlist")),
     selfEmail: arg("--self-email"),
     repoPublicEmails: readLines(arg("--repo-public-emails")),
-    ...(maxBytes ? { maxBytes: parseInt(maxBytes, 10) } : {}),
+    ...(maxBytes !== undefined ? { maxBytes } : {}),
   };
 }
 
--- a/lib/redact-engine.ts
+++ b/lib/redact-engine.ts
@@ -253,7 +253,15 @@ function emailAllowed(email: string, opts: ScanOptions): boolean {
 
 export function scan(input: string, opts: ScanOptions = {}): ScanResult {
   const repoVisibility: RepoVisibility = opts.repoVisibility ?? "unknown";
-  const maxBytes = opts.maxBytes ?? DEFAULT_MAX_BYTES;
+  // `??` only catches null/undefined, so a NaN (e.g. from a malformed
+  // `--max-bytes` flag) or a non-positive value would slip through and make
+  // `byteLen > maxBytes` always false — silently turning this fail-CLOSED guard
+  // into fail-OPEN. Treat any invalid cap as "use the known-good default".
+  const requestedMax = opts.maxBytes;
+  const maxBytes =
+    typeof requestedMax === "number" && Number.isFinite(requestedMax) && requestedMax > 0
+      ? requestedMax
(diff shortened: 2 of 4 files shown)


jbetala7:
Closing as superseded: the fix for #1824 landed on `main` via the v1.57.6.0 fix wave (#1911), which lists `fixes #1824` with attribution and includes an equivalent regression test. Verified the current `main` already contains this exact behavior, so this PR is redundant. Thanks for consolidating it into the wave.

PR #1827, opened 2026-06-01
fix(sync): fail-closed staging-dir ownership guard — prevent rm -rf of repo (#1802)

### LINKED ISSUE #1802: sync-gbrain memory stage rm -rf'd my repo root: resume path trusts checkpoint.dir without verifying it's a gstack staging dir
## Summary

`/sync-gbrain`'s memory-ingest stage **recursively deleted my entire repo working directory** (`rm -rf` equivalent on the repo root, including all source). The deletion happened because the resume path trusted a stale `~/.gbrain/import-checkpoint.json` whose `dir` field pointed at the **repo root (cwd)** instead of a gstack-owned staging dir, and then ran the staging-dir cleanup (`rmSync(dir, { recursive: true, force: true })`) on it.

Restored from GitHub afterward (no commits lost), but untracked files were destroyed.

- gstack version: **1.52.2.0**
- OS: macOS (darwin)
- gbrain: 0.41.5.0, engine postgres (Supabase), MCP local-stdio, **pooler transaction-mode**

## Root cause

`bin/gstack-memory-ingest.ts` stages prepared pages to a throwaway temp dir and, in a `finally` block, cleans it up:

```ts
// :1712
} finally {
  cleanupStagingDir(stagingDir);   // :1713  → rmSync(dir, { recursive: true, force: true })  (:1263)
  _activeStagingDir = null;
}
```

The resume feature (#1611) reuses a prior staging dir when a previous run was SIGTERM'd, reading the dir from gbrain's checkpoint:

```ts
// gstack-memory-ingest.ts:1493-1502
const resumeDir = process.env.GSTACK_INGEST_RESUME_DIR;
const resuming = !remoteHttpMode && typeof resumeDir === "string"
  && resumeDir.length > 0 && existsSync(resumeDir);
const stagingDir = resuming ? resumeDir! : ... makeStagingDir();
```

`GSTACK_INGEST_RESUME_DIR` is set by the orchestrator from the checkpoint (`gstack-gbrain-sync.ts:892-898`, `decideResume()` :169-190). **`decideResume()` only checks that `checkpoint.dir` exists and is a directory — it never verifies the path is a gstack-owned staging dir** (i.e. under `~/.gstack/.staging-ingest-*`).

So when the checkpoint's `dir` is the repo root, the run "resumes" with `stagingDir = <repo root>`, skips `writeStaged`, runs `gbrain import`, then the `finally` block `rmSync`'s the repo root.

## How the checkpoint got poisoned

In my session the pooler was in transaction-mode and `gbrain import` kept failing with `prepared statement "" does not exist` (the `GBRAIN_PREPARE=true` issue). A run got interrupted/SIGTERM'd while cwd was the repo root, and gbrain wrote `import-checkpoint.json` with `dir` = cwd = repo root. The next `/sync-gbrain` hit the resume path and deleted it.

Log evidence from the runs:
```
[sync:memory] previous checkpoint stale (staging dir /Users/.../anf-nano gone), restaging from scratch
[sync:memory] resuming from gbrain checkpoint (0/0 files staged at /Users/.../anf-nano)
```
(The path is the repo root, not a `.staging-ingest-*` dir.)

## Suggested fix

Guard `decideResume()` (and/or the `resuming` check in `gstack-memory-ingest.ts`) so a checkpoint `dir` is only honored when it is a gstack-owned staging path. Minimal:

```ts
import { resolve } from "path";
const stagingRoot = resolve(GSTACK_HOME);  // ~/.gstack
const d = resolve(cp.dir);
const isOwned = d.startsWith(resolve(stagingRoot, ".staging-ingest-"))
             || d.startsWith(resolve(GSTACK_HOME, "transcripts")); // remote-http persistent dir
if (!isOwned) return { kind: "stale-staging-missing", stagingDir: cp.dir };
```

Defense-in-depth: `cleanupStagingDir()` itself should refuse to `rmSync` any path that isn't under the staging root, regardless of how it was derived. A deletion helper that can be pointed at an arbitrary path is the real footgun.

## Impact

`force: true` recursive delete of a user's repo (incl. untracked/uncommitted work) with no confirmation. High severity even though commits survive on the remote.

diazMelgarejo:
Fixes #1802.

## The bug

`/sync-gbrain`'s memory-ingest resume path adopted gbrain's `~/.gbrain/import-checkpoint.json` `dir` field as the staging directory **without proving it was one**. A poisoned checkpoint — `dir` = the repo root, written when an autopilot `gbrain import` was SIGTERM'd while CWD was the repo — was adopted on the next run and then recursively deleted by `cleanupStagingDir()` → `rmSync(repoRoot, { recursive: true, force: true })`, destroying the user's working tree. (Independently reproduced; an `fs_usage` trace caught `bun … unlinkat …/<repo>`.)

Root cause is a **trust failure, not path math**: the code deleted a path it never proved it owned.

## The fix — one predicate, two call sites, fail-closed

New `lib/staging-guard.ts` exports `checkOwnedStagingDir(dir, gstackHome)` — the single definition of "safe to recurse-delete or resume into." Ownership requires **all** of:

1. **Resolvable** — `realpathSync` succeeds (collapses `..`/symlinks first).
2. **A directory** — `statSync(canon).isDirectory()`.
3. **Structural** — canonical path is a *direct child* of `$GSTACK_HOME` named `.staging-ingest-*`.
4. **Not a repo** — no `.git` inside (screaming last-line tripwire).
5. **Minted by us** — a `.gstack-staging` marker that is a **regular file** (`lstat`, so a dir/symlink can't impersonate it), written by `makeStagingDir()`.

Wired in at:
- **`cleanupStagingDir()`** — the single deletion chokepoint (covers both the `finally` block and the SIGTERM handler). Refuses `rm -rf` unless the guard passes.
- **`decideResume()`** — refuses to resume an unprovable checkpoint dir; returns the rejection reason so the caller logs *"staging dir not usable: <why>"* instead of the misleading *"gone"* (the poisoned path often still exists).
- **`gstack-memory-ingest` entry point** — this binary is runnable directly, so it independently guards `GSTACK_INGEST_RESUME_DIR` rather than trusting the env it receives.
- **`makeStagingDir()`** — mints the `.gstack-staging` marker at creation.

## Tests

`test/regression-1611-gbrain-sync-resume.test.ts` grows 9 → 32 assertions: the #1802 poison matrix (repo-root + `.git`, `/`, staging-named-without-marker, nested, symlink escape) plus a `checkOwnedStagingDir` unit matrix. **55 tests green** across the two affected files (`bun test test/regression-1611-gbrain-sync-resume.test.ts test/gstack-memory-ingest.test.ts`).

## Compatibility

`decideResume()` gains an optional `gstackHome` param (default `GSTACK_HOME`) for testability; the sole caller is unchanged.

## Deliberate non-change (disclosed)

Pre-marker staging dirs from a run interrupted **before** this upgrade are refused (restage from scratch — seconds, **no data loss**) rather than accepted via a legacy name-based path. A legacy path would reintroduce the exact name-trust vulnerability this PR removes; the fail-closed cost is a one-time re-stage in a narrow window.

## Scope

This guards gstack's own `rm -rf` boundary. The **durable** fix is upstream in gbrain — checkpoint `dir` should always be a gbrain-minted staging dir, never CWD — filed as a companion issue on `garrytan/gbrain`.

## Review

Design steelmanned by a 4-model panel (Gemini, Codex, gpt-4o, qwen3.5-27b); the in-repo eng pass (Codex + an independent code-review agent on the actual diff) hardened the marker check (regular-file), added the second-entry-point guard, and fixed the misleading log.

## Cross-links
- gbrain PR: https://github.com/garrytan/gbrain/pull/1731
- gbrain issue: https://github.com/garrytan/gbrain/issues/1728
- gstack issue: https://github.com/garrytan/gstack/issues/1802


diazMelgarejo:
Companion root-cause issue filed upstream: **garrytan/gbrain#1728** — `import-checkpoint.json` `dir` resolving to CWD/repo-root on SIGTERM (staging-first checkpointing). This PR is gstack's own fail-closed `rm -rf` boundary; garrytan/gbrain#1728 makes the bug class impossible at the source.

jbetala7:
Validated on `main` (`3bef43bc`): the destructive vector is real and the guard is correctly fail-closed.

On `main` the resume path adopts `import-checkpoint.json`'s `dir` as the staging directory without proving it is one, and `cleanupStagingDir()` then `rm -rf`s it — so a checkpoint whose `dir` is the repo root (written when an autopilot `gbrain import` is SIGTERM'd while CWD is the repo) leads to the repo being recursively deleted. Your regression test `#1802 checkpoint.dir = repo root with .git -> stale-staging-missing (not resumed)` proves the point: `important.py` survives the verdict.

`checkOwnedStagingDir` is sound — realpath resolution, the `STAGING_PREFIX` basename check, the ownership marker, and the `.git` tripwire, with any failure => refuse. Fail-closed is the right posture for an `rm -rf` guard, and no competing open PR touches this path.

One question for the maintainer, not a blocker: since the ownership marker is minted by `makeStagingDir`, in-flight checkpoints written *before* this change won't carry the marker and will be refused on the next resume (a one-time re-import, never a wrong delete) — if I'm reading the precedence right. Worth a line in the PR body so the behavior on upgrade isn't surprising. Otherwise LGTM.

garrytan:
Thank you @diazMelgarejo — this is a clean, well-reasoned fix. Moving it to a base-repo branch so the eval/E2E CI gets secrets (fork PRs don't), and bundling three resume-machinery bugs an eng review + Codex pass surfaced in the same code (remote-http cleanup collision, timeout-resume deletion, resume failure-mapping) plus TOCTOU/marker hardening. Your commit is the first of six and you're credited in the CHANGELOG. Continues in #1856.

PR #1832, opened 2026-06-01
fix(gbrain): healthy thin-client misclassified as broken-config, suppressing brain blocks (#1792)

### LINKED ISSUE #1792: `freshClassify` health probe uses `gbrain sources list`, misclassifying healthy thin-client brains as `broken-config` → brain-aware blocks suppressed
## Summary
brain-aware planning (v1.52) advertises support for "any thin-client MCP" brain, but the
classifier `lib/gbrain-local-status.ts` `freshClassify()` probes via `gbrain sources list
--json` — a **local-DB-only** command. Under thin-client mode this is intentionally *not
routable*, so the probe throws, the catch falls through to the defensive `broken-config`
default, and `gen-skill-docs --respect-detection` suppresses every brain block. A healthy
thin-client brain therefore never gets brain-aware planning.

## Environment
- gstack `v1.52.1.0` (current `main`, `070722a`)
- gbrain `0.41.28.0`, **thin-client mode** (remote MCP over HTTP; no local DB)
- Cause is OS-independent (first hit on Windows, but not platform-specific)

## Reproduction
1. Configure gbrain as a thin-client (remote `mcp_url`, no local engine). Doctor is healthy:
   `gbrain doctor --json --fast` → `{"mode":"thin-client","status":"ok",...}`
2. The probe command is refused by design:
   ```
   $ gbrain sources list --json
   `gbrain sources` is not routable. sources commands manage local DB + config rows.
   Per-subcommand thin-client routing lands in v0.31.x. For now: use `sources_list` /
   `sources_status` MCP tools, or run on the host.   (exit 1)
   ```
3. `gstack-config gbrain-refresh` → `~/.gstack/gbrain-detection.json` shows
   `"gbrain_doctor_ok": true, "gbrain_mcp_mode": "remote-http", "gbrain_local_status": "broken-config"`
   (healthy thin-client, yet `broken-config`).
4. `bun run gen:skill-docs:user` → planning SKILL.md files contain **no** brain blocks
   (`gen-skill-docs.ts:50` un-suppresses only when `gbrain_local_status === 'ok'`).

## Root cause
In `lib/gbrain-local-status.ts` `freshClassify()`, once the on-PATH check (L224) and
config-exists check (L227) pass — both true for a thin-client — it probes with
`execFileSync("gbrain", ["sources","list","--json"])` (L239). `gbrain sources` is local-DB-only
and explicitly not routable in thin-client mode, so the call throws; the stderr matches neither
`"Cannot connect to database"` nor `"config.json"`, hitting the defensive `broken-config`
default (L260). The 5-state enum has no healthy-thin-client path and `freshClassify` never
checks the mode.

## Suggested fix
Detect thin-client mode *before* the `sources list` probe and treat a healthy thin-client as
`ok`. The signal already exists: `gbrain doctor --json --fast` returns `mode: "thin-client"` +
`status: ok|warnings`, and `detect` already calls it in `detectDoctor()` (bin/gstack-gbrain-detect
:121) — the `mode` field is one parse away. `detectMcpMode()` independently returns `remote-http`
too. e.g. in `freshClassify`: if doctor reports thin-client mode and ok/warnings status, return
`"ok"` without running the host/local-DB-only `sources list` probe.

## Workaround (until fixed)
Override detection externally, no gstack edits: confirm health via `gbrain doctor`, then write
`"gbrain_local_status": "ok"` into `~/.gstack/gbrain-detection.json` (keeping the other real
detect fields) and run `bun run gen:skill-docs:user`. Must be re-run after every `./setup` or
`gstack-config gbrain-refresh`, since both rewrite detection.json from the broken probe.

## Secondary (minor)
`scripts/resolvers/gbrain.ts` `generateGBrainContextLoad` tells the agent to run
`gbrain get_page "<slug>"`, but the thin-client CLI has no `get_page` subcommand
(`Unknown command: get_page`); the CLI verb is `gbrain get`. (`get_page` is the MCP op name.)

jbetala7:
## Problem

Brain-aware planning (v1.52) advertises support for any thin-client MCP brain, but a healthy thin-client never gets brain blocks.

`freshClassify()` in `lib/gbrain-local-status.ts` classifies the engine by probing with `gbrain sources list --json` — a **local-DB-only** command. In thin-client mode (remote MCP, no local DB) gbrain refuses it:

```
$ gbrain sources list --json
`gbrain sources` is not routable. sources commands manage local DB + config rows.
Per-subcommand thin-client routing lands in v0.31.x. ...   (exit 1)
```

That stderr matches neither the `broken-db` ("Cannot connect to database") nor the `broken-config` ("config.json") pattern, so the catch falls through to the defensive `broken-config` default. `gen-skill-docs --respect-detection` then un-suppresses brain blocks only when status is `ok`, so a healthy thin-client (doctor reports `mode: thin-client`, `status: ok`) silently loses brain-aware planning across every planning SKILL.md.

## Root cause

In `freshClassify()` the 5-state classifier has no healthy-thin-client path and never checks the brain mode — once the on-PATH + config-exists checks pass (both true for a thin-client), the only signal is the local-DB probe, which thin-clients reject by design.

## Fix

On the **not-routable failure path only**, confirm health via the mode-aware `gbrain doctor --json --fast` (the same doctor call already used in `bin/gstack-gbrain-detect` and `lib/gstack-memory-helpers.ts`) and return `ok` when doctor reports `mode: "thin-client"` with `ok`/`warnings` status.

Design notes:
- **No added latency on the hot path.** The cheap ~80ms `sources list` probe still owns the healthy local-DB case; doctor is only consulted when the probe is *refused*, so local-DB brains are unaffected.
- **No enum change.** A healthy thin-client correctly maps to the existing `ok` state.
- **Stays defensive.** A doctor parse failure or any non-healthy status falls through to the existing classification — an unconfirmed brain is never upgraded. The `not routable` string is also required, so genuine broken-db/broken-config errors are untouched.
- doctor exits 1 when `health_score < 100` (normal under warnings) but still prints JSON to stdout; the helper recovers stdout from the error object, matching the established pattern (#1415).

## Tests

`test/gbrain-local-status.test.ts` (gate-tier, fully mocked — fake `gbrain` on PATH, no real CLI/DB):
- healthy thin-client: `sources list` not-routable + `doctor` reports `thin-client`/`ok` → classifies **`ok`** (was `broken-config`).
- not-routable but `doctor` reports an unhealthy status → stays **`broken-config`** (defensive default preserved).

```
bun test test/gbrain-local-status.test.ts
# 14 pass, 0 fail
```

Adjacent gbrain suites (`gbrain-detect-shape`, `gbrain-detection-override`, `gstack-gbrain-detect-mcp-mode`, `gbrain-sync-skip`) remain green. Pre-existing `gstack-gbrain-detect` subprocess failures reproduce unchanged on `origin/main` and are unrelated to this change.

Fixes #1792

### DIFF
--- a/lib/gbrain-local-status.ts
+++ b/lib/gbrain-local-status.ts
@@ -244,6 +244,50 @@ function writeCache(status: LocalEngineStatus, key: CacheEntry["key"]): void {
 }
 
 /**
+ * Confirm a healthy thin-client brain via `gbrain doctor --json --fast`.
+ *
+ * Thin-client mode (remote MCP, no local DB) intentionally refuses the
+ * `gbrain sources list` probe — `sources` is a local-DB-only command. The probe
+ * therefore throws with a "not routable" stderr that matches neither the
+ * broken-db nor broken-config pattern, so freshClassify would mislabel a
+ * perfectly healthy thin-client as broken-config and suppress every brain block
+ * (#1792). Doctor is the mode-aware signal: it reports `mode` + `status` for
+ * thin-clients. We only call it on the not-routable failure path, so the cheap
+ * ~80ms sources-list probe still owns the healthy local-DB hot path.
+ *
+ * Returns true only when doctor confirms thin-client mode AND a healthy status
+ * (`ok`/`warnings`). Any parse failure or unhealthy status falls through to the
+ * caller's defensive classification — we never upgrade an unconfirmed brain.
+ */
+function isHealthyThinClient(env?: NodeJS.ProcessEnv): boolean {
+  let parsed: Record<string, unknown> | null = null;
+  try {
+    const out = execFileSync("gbrain", ["doctor", "--json", "--fast"], {
+      encoding: "utf-8",
+      timeout: probeTimeoutMs(env),
+      stdio: ["ignore", "pipe", "ignore"],
+      env: buildGbrainEnv({ baseEnv: env ?? process.env }),
+      shell: NEEDS_SHELL_ON_WINDOWS, // #1731: gbrain is a .cmd shim on Windows
+    });
+    parsed = JSON.parse(out);
+  } catch (err) {
+    // doctor exits 1 whenever health_score < 100 (normal on warnings), but
+    // still prints the JSON to stdout. Recover it before giving up. See #1415.
+    try {
+      const stdout = (err as { stdout?: Buffer | string })?.stdout ?? "";
+      const stdoutStr = typeof stdout === "string" ? stdout : stdout.toString("utf-8");
+      if (stdoutStr) parsed = JSON.parse(stdoutStr);
+    } catch {
+      return false;
+    }
+  }
+  if (!parsed) return false;
+  const mode = parsed.mode;
+  const status = parsed.status;
+  return mode === "thin-client" && (status === "ok" || status === "warnings");
+}
+
+/**
  * Probe via `gbrain sources list --json`. Classify the outcome.
  *
  * Pattern strings ("Cannot connect to database", "config.json") are deliberately
@@ -286,6 +330,12 @@ function freshClassify(env?: NodeJS.ProcessEnv): LocalEngineStatus {
 
     // ENOENT can happen if gbrain disappeared between resolveGbrainBin and now.
     if (e.code === "ENOENT") return "no-cli";
+
+    // Thin-client mode (remote MCP, no local DB) refuses `sources list` with a
+    // "not routable" error — `sources` is local-DB-only. This is NOT a broken
+    // brain: confirm health via the mode-aware doctor before classifying, so a
+    // healthy thin-client reports "ok" and keeps its brain blocks (#1792).
+    if (stderr.includes("not routable") && isHealthyThinClient(env)) return "ok";
 
     // Pattern match against gbrain's known error strings. Order matters:
     // "Cannot connect to database" is the more specific DB-unreachable signal.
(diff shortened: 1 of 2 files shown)


lucha0404:
As the OP — confirming #1832's approach is right for the thin-client topology. My setup: Windows (MSYS bash), gbrain in **thin-client mode over MCP** (local MCP → remote VM `gbrain serve` → Supabase), no local DB. `freshClassify`'s `sources list` probe returns not-routable here and falls through to `broken-config`, suppressing all brain blocks even though `gbrain doctor` reports `mode: thin-client` healthy.

I've been carrying a local adapter that does exactly what #1832 does — consult `gbrain doctor` and treat a healthy thin-client as `ok` — and it reliably restores brain-aware context (verified end-to-end: planning skills like `office-hours` load brain context via MCP). So the `doctor --json --fast` confirmation on the not-routable path is the correct fix, and it matches independent real-world validation. 👍 for merge.

jbetala7:
Rebased onto current main (9fd03fae) and resolved the gbrain-local-status test conflict by keeping both main timeout/GBRAIN_HOME coverage and this PR thin-client doctor fallback coverage.

Validation: ~/.bun/bin/bun test test/gbrain-local-status.test.ts - 24 pass, 0 fail.

time-attack:
@jbetala7 Please rewrite the status handling so a healthy thin client without a local engine is not reported as having a healthy local engine.

PR #1761, opened 2026-05-27
fix(setup): bump playwright to ^1.60.0 — fixes Node 24.16+ yauzl hang

NikhileshNanduri:
## Summary

- Bumps `playwright` from `^1.58.2` to `^1.60.0` in `package.json`
- Fixes indefinite hang in `./setup` on Node 24.16+ and Node 26.x

## Root cause

Playwright <1.60.0 has a yauzl extraction bug triggered on Node 24.16+ (and Node 26.x). `./setup` calls `playwright install chromium` — the download completes (100%), then the process stalls forever with no error and no progress. Users on affected Node versions effectively can't install gstack without killing the process manually and retrying on an older Node.

Closes #1703.

## Test plan

- [ ] Free test suite passes (`bun test`)
- [ ] `playwright install chromium` completes on Node 24+

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1703, opened 2026-05-26
`playwright install` hangs on Node 24.16+ — bump pinned playwright from `^1.58.2` to `^1.60.0`


dkoo761:
## Summary

gstack's `./setup` calls `playwright install chromium`, which hangs indefinitely on Node 24.16.0+ (and Node 26.x) due to a yauzl extraction-hang bug in Playwright versions < 1.60.0. gstack pins `playwright ^1.58.2` in `package.json`, so it inherits the bug. Fix is a one-line version bump.

## Symptoms

Running `cd ~/.claude/skills/gstack && ./setup` (or the gstack `setup` script via any wrapper) hangs at:

```
Downloading Chrome for Testing 145.0.7632.6 (playwright chromium v1208) from https://cdn.playwright.dev/...
179.6 MiB [====================] 100% 0.0s
```

The download completes (100%), then extraction silently stalls. Inspecting the cache after ~5 min shows only the large `chrome` binary was written; all other files in the bundle (locales, resources, `icudtl.dat`, etc.) are missing:

```bash
$ ls ~/.cache/ms-playwright/chromium-1208/chrome-linux/
chrome  vk_swiftshader_icd.json
$ du -sh ~/.cache/ms-playwright/chromium-1208/
428M    .   # should be ~600 MB
```

The download worker processes die silently mid-extraction, leaving the parent (`bunx playwright install`) hanging on stale IPC pipes.

## Environment

- gstack: current HEAD of main (last reproduced 2026-05-25)
- Host: macOS arm64 (Apple Silicon) with Docker Desktop devcontainer
- Container: `node:24.16.0` (rolled forward from `node:24.15.0` sometime between May 23–25, 2026)
- Playwright in `~/.claude/skills/gstack/node_modules/`: 1.58.2 (matches `package.json` pin of `^1.58.2`)

## Root cause

This is microsoft/playwright#40724 — a yauzl stream-destruction regression that hangs `for await` over `openReadStream` on Node 24.16.0+ and Node 26.x. The issue reporter saw it on Node 26.1.0; a comment in that thread provides a clean repro showing the bug bisects to Node 24.16.0:

```bash
# WEDGES at "100% of 183 MiB" then silence forever:
docker run --rm node:24.16.0 bash -c 'npm install playwright@1.59.1; npx playwright install chromium'

# WORKS in seconds:
docker run --rm node:24.15.0 bash -c 'npm install playwright@1.59.1; npx playwright install chromium'
```

[...]

PR #1838, opened 2026-06-02
fix(setup): pin Playwright browser install to bundled version so headed browse connect works (#1829)

### LINKED ISSUE #1829: browse connect (headed) fails: setup installs wrong Chromium revision vs compiled binary's pinned Playwright
## Summary

On a clean install/upgrade, `browse connect` (headed) fails with `Executable doesn't exist at .../ms-playwright/chromium-1208/...` even though `./setup` reported success and headless `browse` works fine. Root cause: the version of Playwright that `./setup` uses to **install** browsers is resolved at runtime (latest) and downloads a different Chromium revision than the one **embedded in the compiled `browse` binary**, and the post-install verification only launches **headless**, so the mismatch is never caught.

## Environment

- gstack `1.55.0.0` (also reproduced on `1.28.0.0` before upgrading)
- macOS arm64, Node v26.0.0, Bun 1.3.11
- Install type: global-git (`~/.claude/skills/gstack`)

## What happens

```
$ browse connect
Launching headed Chromium with extension + terminal agent...
[browse] Connect failed: Server failed to start:
persistentContext: Executable doesn't exist at
  ~/Library/Caches/ms-playwright/chromium-1208/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing
    at async launchHeaded (browse/src/browser-manager.ts:557:35)
    at async start (browse/src/server.ts:2938:28)
```

Headless `browse` (e.g. `/qa` headless, `goto`, `snapshot`) works perfectly the whole time.

Playwright cache after a fresh `./setup`:

```
chromium_headless_shell-1208   <- present (why headless works)
chromium_headless_shell-1223
chromium-1223                  <- full build, WRONG revision
ffmpeg-1011
# chromium-1208 (full) is MISSING -> headed fails
```

## Root cause

1. The compiled `browse` binary (bun `--compile`) embeds a **pinned** Playwright that resolves Chromium **revision 1208** (Chrome for Testing 145.0.7632.6, Playwright 1.58.2).
2. `setup` installs browsers with `bunx playwright install chromium` (setup ~line 483), which resolves **latest** Playwright at runtime (currently 1.60.0 -> revision **1223**). So it downloads the wrong full build.
3. `ensure_playwright_browser()` (setup ~line 252) verifies with a **headless** launch:
   ```bash
   bun --eval 'import { chromium } from "playwright"; const browser = await chromium.launch(); await browser.close();'
   ```
   This passes because `chromium_headless_shell-*` is present, so the missing **full** build (needed by `launchHeaded`) is never detected.

Net: setup's installer Playwright and the compiled binary's Playwright are not the same version, and the verify step only exercises headless. Headed users are broken on a "successful" install.

## Workaround (confirmed fix)

Install the full Chromium at the revision the compiled binary pins:

```bash
bunx playwright@1.58.2 install chromium   # downloads chromium-1208 full (~162MB)
browse connect                            # -> Mode: headed  ✅
```

## Suggested fix

- Install browsers using the **same Playwright version embedded in the compiled `browse` binary**, not a runtime-resolved latest. e.g. pin `playwright@<bundled-version>` in the setup install step, or expose `browse install-browsers` that shells out to the binary's own embedded Playwright so install and runtime can never drift.
- Make `ensure_playwright_browser()` launch **headed** (or assert the full Chrome-for-Testing executable path exists), so a missing full build fails setup instead of being masked by the cached headless shell.

jbetala7:
Fixes #1829

## Problem

On a clean install/upgrade, headed `browse connect` fails with a missing full Chromium build:

```
Executable doesn't exist at .../ms-playwright/chromium-1208/.../Google Chrome for Testing
    at async launchHeaded (browse/src/browser-manager.ts)
```

even though `./setup` reported success and **headless** browse works fine.

## Root cause (two issues in `setup`)

1. **Install/runtime version drift.** The browse binary is compiled (setup ~line 390) against the lockfile-pinned Playwright (`playwright@1.58.2` → Chromium revision **1208**). But the browser-install step ran `bunx playwright install chromium`, which resolves the **latest** Playwright at runtime (currently `1.60.0` → revision **1223**) and downloads a different full build than the binary expects. The cache ends up with `chromium-1223` (wrong) and `chromium_headless_shell-1208` but **not** `chromium-1208` (full) — so headed launch fails.
2. **Headless-only verify masks it.** `ensure_playwright_browser()` only ran a default `chromium.launch()` (headless), which passes off the cached `chromium_headless_shell` even when the full Chrome-for-Testing build is absent. So setup reports success.

## Fix

- Install browsers with the Playwright version **bundled in `node_modules`** (read from `playwright/package.json`) instead of `bunx`-latest, so the installed Chromium revision always matches the compiled binary and the two can never drift. Falls back to the old behavior only if the version can't be resolved.
- Harden `ensure_playwright_browser()` to assert `chromium.executablePath()` exists before treating the browser as ready (both the Node/Windows and Bun/Unix branches), so a missing **full** build fails setup instead of being silently masked by the cached headless shell.

This matches the maintainer-suggested direction in the issue (pin the install to the bundled version; make verify catch a missing full build).

## Testing

- `bun test test/setup-playwright-pin.test.ts` — new static-source regression test (no browser download), 3 pass.
- `bun test test/setup-windows-fallback.test.ts test/setup-emoji-font.test.ts` — existing setup tests still green (20 pass).
- `bash -n setup` — syntax OK.
- Verified locally that the bundled version resolves to `1.58.2` and `chromium.executablePath()` returns the `chromium-1208` full-build path the issue reports as missing.

## Relationship to other open Playwright PRs (not duplicates)

- **#1761** / **#1703** (bump to `^1.60.0` for the Node 24.16+ yauzl hang) and **#1565** (bump + runtime `executablePath` pin in `browser-manager.ts` for Flutter headless rendering) target different symptoms and code paths. A version bump only *incidentally* realigns revisions while npm's "latest" happens to equal the pinned range; the moment a newer Playwright ships, `bunx`-latest drifts again. This PR fixes the **drift mechanism** in the install/verify path and is complementary to those changes.

### DIFF
--- a/setup
+++ b/setup
@@ -250,17 +250,23 @@ if [ "$INSTALL_CODEX" -eq 1 ]; then
 fi
 
 ensure_playwright_browser() {
+  # Assert the *full* Chromium build exists before launching. chromium.launch()
+  # defaults to the headless shell (chromium_headless_shell), which can be
+  # present even when the full Chrome-for-Testing build that headed
+  # `browse connect` needs is missing — so a plain launch silently passes and
+  # masks a missing full build. chromium.executablePath() resolves the full
+  # build path regardless, so we stat it first. (#1829)
   if [ "$IS_WINDOWS" -eq 1 ]; then
     # On Windows, Bun can't launch Chromium due to broken pipe handling
     # (oven-sh/bun#4253). Use Node.js to verify Chromium works instead.
     (
       cd "$SOURCE_GSTACK_DIR"
-      node -e "const { chromium } = require('playwright'); (async () => { const b = await chromium.launch(); await b.close(); })()" 2>/dev/null
+      node -e "const { chromium } = require('playwright'); const fs = require('fs'); (async () => { if (!fs.existsSync(chromium.executablePath())) process.exit(1); const b = await chromium.launch(); await b.close(); })()" 2>/dev/null
     )
   else
     (
       cd "$SOURCE_GSTACK_DIR"
-      bun --eval 'import { chromium } from "playwright"; const browser = await chromium.launch(); await browser.close();'
+      bun --eval 'import { chromium } from "playwright"; import { existsSync } from "fs"; if (!existsSync(chromium.executablePath())) process.exit(1); const browser = await chromium.launch(); await browser.close();'
     ) >/dev/null 2>&1
   fi
 }
@@ -480,7 +486,18 @@ if ! ensure_playwright_browser; then
   echo "Installing Playwright Chromium..."
   (
     cd "$SOURCE_GSTACK_DIR"
-    bunx playwright install chromium
+    # The browse binary is compiled above against the lockfile-pinned Playwright.
+    # `bunx playwright install` resolves the *latest* Playwright at runtime, which
+    # downloads a different Chromium revision than the compiled binary expects, so
+    # headed `browse connect` fails with a missing full build even though setup
+    # "succeeds" and headless works. Pin the browser install to the same Playwright
+    # version bundled in node_modules so install and runtime can never drift. (#1829)
+    pw_version="$(bun --eval 'console.log(require("playwright/package.json").version)' 2>/dev/null)"
+    if [ -n "$pw_version" ]; then
+      bunx "playwright@$pw_version" install chromium
+    else
+      bunx playwright install chromium
+    fi
   )
 
   if [ "$IS_WINDOWS" -eq 1 ]; then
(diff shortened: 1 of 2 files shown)


time-attack:
@jbetala7 Rewrite the version lookup so installation does not exit under set -e before the fallback, and add tests for stale and mismatched browser caches.

PR #1840, opened 2026-06-02
fix(one-way-doors): treat 'rotate ... password' as a one-way door

### LINKED ISSUE #1839: one-way-door classifier: "rotate ... password" misclassified as two-way (missing from rotate pattern)
## Summary

The one-way-door keyword safety net misclassifies "rotate ... password" as a reversible (two-way) question. The sibling `revoke` and `reset` credential patterns both include `password`, but the `rotate` pattern omits it, so the most common phrasing of a credential-rotation op ("rotate the database password") slips through the destructive-keyword check.

`scripts/one-way-doors.ts` is the documented SECONDARY safety layer for AskUserQuestion calls that fire without a registry id (ad-hoc `{skill}-{slug}` ids). Its header is explicit about which direction is dangerous:

> A false negative could mean auto-approving a destructive operation.

The credential patterns are written to be parallel (the inline comment says they "allow filler words ('the', 'my') between verb and noun"), so this is an omission, not an intentional scope choice.

## Current behavior on `main` (3bef43bc / v1.55.0.0)

`scripts/one-way-doors.ts`:

```ts
// Credentials / auth — allow filler words ("the", "my") between verb and noun
/\brevoke\s+[\w\s]*\b(api key|token|credential|access key|password)\b/i,  // has password
/\breset\s+[\w\s]*\b(api key|token|password|credential)\b/i,              // has password
/\brotate\s+[\w\s]*\b(api key|token|secret|credential|access key)\b/i,    // MISSING password
```

Reproduced directly against the committed module:

```
$ bun -e "import {classifyQuestion} from './scripts/one-way-doors.ts';
  for (const s of ['Rotate the database password?','rotate password for prod','Revoke the password','Reset the password','rotate the API key'])
    console.log((classifyQuestion({summary:s}).oneWay?'ONE-WAY':'two-way ')+'  '+s)"

two-way   Rotate the database password?
two-way   rotate password for prod
ONE-WAY   Revoke the password
ONE-WAY   Reset the password
ONE-WAY   rotate the API key
```

"Rotate the database password" should be the *clearest* one-way credential op, yet it is the only one of the three verbs that fails to catch `password`.

## Expected behavior

`classifyQuestion`/`isOneWayDoor` should treat "rotate ... password" as one-way (`reason: 'keyword'`), in parity with `revoke ... password` and `reset ... password`.

## Impact

`classifyQuestion`/`isOneWayDoor` is an exported, unit-tested API (`test/plan-tune.test.ts` covers the keyword net, including `'rotate the API key'`). For an ad-hoc destructive question that the keyword net is supposed to catch, a false negative here lets a credential-rotation question be treated as suppressible by a permissive tuning preference (`never-ask` / `ask-only-for-one-way`) instead of always asking. This is exactly the failure mode the module's threat model calls out as the dangerous direction.

## Duplicate search performed

- PRs (open + closed): `one-way-doors`, `DESTRUCTIVE_PATTERNS`, `rotate password`, `classifyQuestion`, `one-way door rotate` — no PR touches this pattern.
- Issues: `one-way door`, `rotate password`, `DESTRUCTIVE`, `classifyQuestion`, `destructive keyword` — the only DESTRUCTIVE hits are about `/careful`, `/guard`, and `/sync-gbrain` (#1734, #1060, #1091), unrelated to this classifier.
- `scripts/one-way-doors.ts` is unchanged since v1.0.0.0 (#1039).

## Candidate fix

Add `password` to the `rotate` alternation so the three credential verbs are at parity, plus a regression test in the existing `one-way-doors classifier` block:

```ts
/\brotate\s+[\w\s]*\b(api key|token|secret|credential|access key|password)\b/i,
```

jbetala7:
## Problem

`scripts/one-way-doors.ts` is the secondary keyword safety net for AskUserQuestion calls that fire without a registry id. Its three credential-rotation patterns are meant to be parallel (the inline comment says they "allow filler words between verb and noun"), but the `rotate` pattern omits `password`:

```ts
/\brevoke\s+[\w\s]*\b(api key|token|credential|access key|password)\b/i,  // has password
/\breset\s+[\w\s]*\b(api key|token|password|credential)\b/i,              // has password
/\brotate\s+[\w\s]*\b(api key|token|secret|credential|access key)\b/i,    // MISSING password
```

So `classifyQuestion`/`isOneWayDoor` classified "rotate the database password" as a reversible two-way door, while "revoke the password" and "reset the password" both correctly classified as one-way. The module header is explicit that this is the dangerous direction: "A false negative could mean auto-approving a destructive operation."

Reproduced on `main` (3bef43bc):

```
two-way   Rotate the database password?
two-way   rotate password for prod
ONE-WAY   Revoke the password
ONE-WAY   Reset the password
ONE-WAY   rotate the API key
```

## Root cause

A plain omission of `password` from the `rotate` noun alternation. "Rotate the password" is the most common phrasing of a credential-rotation op, more common than "rotate credential" (which the pattern already covers).

## Fix

Add `password` to the `rotate` alternation so the three credential verbs are at parity. One-line change, plus a regression test mirroring the existing `keyword fallback catches destructive summaries` test.

## Testing

- `bun test test/plan-tune.test.ts` — 48 pass / 0 fail.
- Verified the new test fails on `main` without the one-line fix (1 fail) and passes with it.
- Full `bun test` suite passes (exit 0).

Fixes #1839

### DIFF
--- a/scripts/one-way-doors.ts
+++ b/scripts/one-way-doors.ts
@@ -65,7 +65,7 @@ const DESTRUCTIVE_PATTERNS: RegExp[] = [
   // Credentials / auth — allow filler words ("the", "my") between verb and noun
   /\brevoke\s+[\w\s]*\b(api key|token|credential|access key|password)\b/i,
   /\breset\s+[\w\s]*\b(api key|token|password|credential)\b/i,
-  /\brotate\s+[\w\s]*\b(api key|token|secret|credential|access key)\b/i,
+  /\brotate\s+[\w\s]*\b(api key|token|secret|credential|access key|password)\b/i,
 
   // Scope / architecture forks (reversible with effort — still deserve confirmation)
   /\barchitectur(e|al)\s+(change|fork|shift|decision)\b/i,
--- a/test/plan-tune.test.ts
+++ b/test/plan-tune.test.ts
@@ -426,6 +426,19 @@ describe('one-way-doors classifier', () => {
     }
   });
 
+  test('rotate <credential> password is one-way (parity with revoke/reset)', () => {
+    const cases = [
+      'Rotate the database password?',
+      'rotate password for the service account',
+    ];
+    for (const summary of cases) {
+      const result = classifyQuestion({ summary });
+      expect(result.oneWay).toBe(true);
+      expect(result.reason).toBe('keyword');
+      expect(result.matched).toBeDefined();
+    }
+  });
+
   test('skill-category fallback fires for cso:approval and land-and-deploy:approval', () => {
     expect(isOneWayDoor({ skill: 'cso', category: 'approval' })).toBe(true);
     expect(isOneWayDoor({ skill: 'land-and-deploy', category: 'approval' })).toBe(true);


jbetala7:
Closing as superseded: the fix for #1839 landed on `main` via the v1.57.6.0 fix wave (#1911), which lists `fixes #1839` with attribution and includes an equivalent regression test. Verified the current `main` already contains this exact behavior, so this PR is redundant. Thanks for consolidating it into the wave.

Issue #1837, opened 2026-06-02
ios-qa daemon: boot-token never written (silent try? on NSTemporaryDirectory) — tunnel dies after 1-2 requests; bootstrap.ts claims os_log scrape but does file-copy


aweevenson-sea:
**gstack version:** v1.55.0.0
**Skill:** `/ios-qa` (Mac daemon / vision-loop)
**Discovered:** 2026-06-01, `/ios-qa` session on Dugout Sports (real device, iPhone 16 Pro).
**Related:** companion template-bug report (the five build/SPI fixes) — separate issue.

### Symptom

The daemon's tunnel authenticates for at most one or two requests right after spawn, then dies for the rest of the session. Burst-firing taps works only within a ~10s window immediately after a fresh daemon spawn; after that the tunnel proxies nothing.

### Root cause

`bootstrap.ts` documents that it scrapes `os_log` to recover the boot token, but it actually does a `devicectl appDataContainer` file-copy of `tmp/gstack-ios-qa.token`. On Dugout, `StateServer.start()`'s

```swift
try? bootToken.write(toFile: ...)   // NSTemporaryDirectory()-based path
```

silently fails — a direct `devicectl` probe of the app sandbox shows the file is genuinely absent. Because the write is a `try?`, the failure is swallowed, the token file never exists, and the daemon never gets a valid token. The one or two requests that do go through immediately post-spawn appear to use a stale/cached rotated token, which then expires.

### Possible fixes (any one would help; first matches the docs)

1. **Actually implement the os_log scraping** that `bootstrap.ts` already claims to do, instead of relying on the file copy.
2. **Write the token under `NSHomeDirectory()/Documents`** instead of `NSTemporaryDirectory()` — the Documents container is reliably readable via `devicectl appDataContainer`.
3. **Escalate the silent `try?`** in `StateServer.start()` to a `logger.error` (or `do/catch` that logs) so the write failure is visible rather than swallowed.

### Workaround we used

Kill daemon → `terminate`-and-relaunch the app → spawn daemon → burst-fire ≤3 taps within ~10s. Repeat per burst. Not sustainable for a full QA loop.

Lower priority than the template build/SPI fixes, but it's the reason the vision loop couldn't complete a multi-step flow (we got dashboard → More → Profile but the tunnel dropped before the Sign-in-with-Apple flow).

time-attack:
@aweevenson-sea Boot-token writes can still fail silently on physical devices, and the merged relaunch work does not fix persistent write failures.

Issue #342, opened 2026-03-23
browse: Bun-compiled CLI cannot start server on Windows (process detection fails)


kwandoletcher:
## Summary

`/browse` is completely broken on Windows 11. The compiled `browse.exe` (Bun binary) cannot start or detect the browse server process, making all browse commands fail immediately with:

```
[browse] Another instance is starting the server, waiting...
[browse] Timed out waiting for another instance to start the server
```

This happens on fresh install with zero configuration. Every browse command fails, including `browse --version`.

## Root Cause

The `isProcessAlive()` function in `browse/src/cli.ts` uses `process.kill(pid, 0)` to check if the server PID is alive. On Windows, Bun's compiled binary cannot reliably signal Windows-native process IDs this way. The call throws `ESRCH` even for running processes.

This causes a cascade failure in `ensureServer()`:
1. `isProcessAlive(state.pid)` returns `false` even when server is healthy
2. Skips the HTTP health check entirely (line 240 guard)
3. Falls through to lock acquisition, which also fails
4. Prints "Another instance is starting the server" and times out

**The server itself works fine.** When started manually via `node server-node.mjs`, it launches Chromium, responds to health checks, and processes commands via HTTP. The bug is purely in the CLI's process detection.

## Reproduction

1. Install gstack on Windows 11
2. Run any browse command: `~/.claude/skills/gstack/browse/dist/browse goto https://example.com`
3. Observe timeout error

## Evidence

```bash
# Server starts fine manually:
$ BROWSE_STATE_FILE="$(pwd)/.gstack/browse.json" node ~/.claude/skills/gstack/browse/dist/server-node.mjs
[browse] Server running on http://127.0.0.1:33668 (PID: 19992)

# Health endpoint works:
$ curl http://127.0.0.1:33668/health
{"status":"healthy","uptime":3,"tabs":1,"currentUrl":"about:blank"}

# Commands work via direct HTTP:
$ curl -X POST http://127.0.0.1:33668/command -d '{"command":"goto","args":["https://example.com"]}'
Navigated to https://example.com (200)

# But browse.exe can't detect the running server:
$ browse goto https://example.com
[browse] Another instance is starting the server, waiting...
[browse] Timed out waiting for another instance to start the server

# Bun's process.kill fails for the Windows PID:
$ bun -e "process.kill(19992, 0)"  # ESRCH even though process is running
```

## Suggested Fix

[...]

PR #1845, opened 2026-06-02
fix(ios-qa): DebugBridge templates — clean install on Swift 6.x + no private SPI leak into Release

### LINKED ISSUE #1836: ios-qa: DebugBridge templates fail clean install on Swift 6.x + leak private SPIs into Release (5 bugs, incl. Apple-rejection class)
**gstack version:** v1.55.0.0
**Skill:** `/ios-qa`
**Discovered:** 2026-06-01, fresh install of DebugBridge into the Dugout Sports iOS app (after our PR #342 had to surgically remove the previous DebugBridge to pass App Store review — see #5 below, which is the same failure mode).

A clean `/ios-qa` install of the DebugBridge stack currently does not build out of the box on Swift 6.x toolchains, and — most seriously — the generated Obj-C touch-synthesis target compiles private UIKit/IOKit SPIs into **Release** builds, which Apple's static scanner rejects. Five distinct template bugs below, in increasing severity. We've applied local patches for all five in our working tree; the corrected files are quoted/referenced inline.

---

### 1. `gen-accessors-tool/Package.swift` references a non-existent test target

`ios-qa/scripts/gen-accessors-tool/Package.swift` declares:

```swift
.testTarget(name: "GenAccessorsTests", path: "Tests/GenAccessorsTests")
```

but `Tests/` does not exist in the package, so `swift build` fails:

```
error: invalid custom path 'Tests/GenAccessorsTests' for target 'GenAccessorsTests'
```

**Fix:** remove the `.testTarget(...)` block — no tests exist.

---

### 2. `Package.swift.template` puts the tools-version directive on line 21, not line 1

`ios-qa/templates/Package.swift.template` opens with a multi-line doc-comment block and only declares `// swift-tools-version:5.9` on ~line 21. The directive **must** be the first line of the manifest. Swift 6.0+ rejects the misplaced directive:

```
error: manifest is backward-incompatible with Swift < 6.0
```

**Fix:** move `// swift-tools-version:5.9` to line 1 (doc-comment block follows it). Our generated `apps/ios/DebugBridge/Package.swift` shows the correct ordering — directive on line 1, comment block lines 2–22.

---

### 3. `Package.swift.template` references a non-existent test target

Same template declares:

```swift
.testTarget(name: "DebugBridgeCoreTests", path: "Tests/DebugBridgeCoreTests")
```

`/ios-qa` never populates a `Tests/` folder, so the install fails the same way as #1 (`invalid custom path`).

**Fix:** drop the `.testTarget(...)` block.

---

### 4. `DebugBridgeWiring.swift.template` is broken and redundant — delete it

`ios-qa/templates/DebugBridgeWiring.swift.template` does not compile and duplicates wiring that `Bridges.swift.template` already provides:

- `import DebugBridge` — there is no `DebugBridge` module. The modules are `DebugBridgeCore`, `DebugBridgeUI`, `DebugBridgeTouch`.
- References `AccessibilityScanner`, `SnapshotCapture`, `MutationDispatcher` — none of these symbols exist anywhere in the templates. `Bridges.swift.template` defines `ElementsBridgeImpl`, `ScreenshotBridgeImpl`, `MutationBridgeImpl` instead.
- Calls `DebugBridgeManager.shared.start(appState:recording:)` — `DebugBridgeManager` only defines `start(appState:)`.

`Bridges.swift`'s `DebugBridgeUIWiring.installAll()` already performs the actual wiring (confirmed in our generated `Sources/DebugBridgeUI/Bridges.swift` — public `enum DebugBridgeUIWiring { static func installAll() ... }`).

**Fix:** delete `DebugBridgeWiring.swift.template` entirely.

---

### 5. ⚠️ `DebugBridgeTouch.m.template` compiles private SPIs into App Store Release builds

**This is the critical one — it is the exact failure that got Dugout Sports v1.0 rejected (App Store Guideline 2.1, private API scan) and forced the surgical DebugBridge removal in our PR #342.**

`ios-qa/templates/DebugBridgeTouch.m.template` gates its iOS implementation block with:

```objc
#if TARGET_OS_IOS
```

instead of:

```objc
#if TARGET_OS_IOS && defined(DEBUG)
```

The block references private UITouch/UIEvent/IOKit SPIs — `_setLocationInWindow:resetPrevious:`, `_addTouch:forDelayedDelivery:`, `_clearTouches`, `_setHIDEvent:`, `_setIsFirstTouchForView:`, `_touchesEvent`, `IOHIDEventCreateDigitizer*`. With only `TARGET_OS_IOS`, these compile into **iOS Release** builds and land in the App Store binary, where Apple's static scanner flags them.

Critically, the `Package.swift` Swift-side gate does **not** protect this file:

```swift
swiftSettings: [.define("DEBUG", .when(configuration: .debug))]
```

`swiftSettings` only affects Swift compile units. The Obj-C `.m` file is compiled with `cSettings`, which the template doesn't set — so `DEBUG` is never defined for this translation unit in a SwiftPM build, and the SPIs leak.

**Fix (two parts):**

1. Change the gate to `#if TARGET_OS_IOS && defined(DEBUG)`. The existing `#else` stub (iOS Release / macOS / Catalyst) already provides the no-op implementations, so the non-DEBUG branch is already handled.
2. Add a `cSettings` DEBUG define to the `DebugBridgeTouch` target in `Package.swift.template` as belt-and-suspenders:

```swift
.target(
    name: "DebugBridgeTouch",
    ...
    cSettings: [
        .define("DEBUG", to: "1", .when(configuration: .debug)),
    ],
    ...
)
```

Our patched `apps/ios/DebugBridge/Sources/DebugBridgeTouch/DebugBridgeTouch.m` now gates at line 28 (`#if TARGET_OS_IOS && defined(DEBUG)`), with the `#else` stub at line 294 and `#endif` at line 307; the patched `Package.swift` carries the `cSettings` define on the `DebugBridgeTouch` target.

---

### Verification (passes with all five fixes applied)

```bash
cd "$APP/DebugBridge" && swift build -c release
nm -j .build/arm64-apple-macosx/release/DebugBridgeTouch.build/DebugBridgeTouch.m.o \
  | grep -qE '_setLocationInWindow|_addTouch|_clearTouches|_setHIDEvent' \
  && exit 1 || echo "clean"
```

Expected: `swift build -c release` succeeds (fixes #1–#4) and the `nm` symbol scan prints `clean` (fix #5).

We have local patches for all five at `apps/ios/DebugBridge/` in the Dugout repo (uncommitted, branch `fix/data-accuracy-fixes`) and can open a PR against the templates if useful — happy to do so once you confirm the approach.

aweevenson-sea:
Fixes #1836.

Five template bugs that break a fresh `/ios-qa` DebugBridge install on Swift 6.x — the most serious compiles private `UITouch`/`UIEvent`/`IOKit` SPIs into iOS **Release** builds, which is the exact failure that forced a surgical DebugBridge removal in a downstream app (its PR #342) to clear App Store review (Guideline 2.1, private API scan).

### Changes

| # | File | Fix |
|---|------|-----|
| 1 | `ios-qa/scripts/gen-accessors-tool/Package.swift` | Drop `GenAccessorsTests` testTarget — `Tests/` doesn't exist → `swift build` failed with *invalid custom path*. |
| 2 | `ios-qa/templates/Package.swift.template` | Move `// swift-tools-version:5.9` to line 1 — Swift 6.0+ rejects the misplaced directive. |
| 3 | `ios-qa/templates/Package.swift.template` | Drop `DebugBridgeCoreTests` testTarget — never populated by the installer; same invalid-path failure. |
| 4 | `ios-qa/templates/DebugBridgeWiring.swift.template` | **Delete.** Imported a non-existent `DebugBridge` module, referenced `AccessibilityScanner`/`SnapshotCapture`/`MutationDispatcher` (don't exist), and called a `start(appState:recording:)` overload that doesn't exist. `Bridges.swift`'s `DebugBridgeUIWiring.installAll()` already does the real wiring. |
| 5 | `ios-qa/templates/DebugBridgeTouch.m.template` + `Package.swift.template` | Gate the iOS impl behind `#if TARGET_OS_IOS && defined(DEBUG)` (was `#if TARGET_OS_IOS`), and add a `cSettings` DEBUG define to the `DebugBridgeTouch` target. `swiftSettings` do **not** reach Obj-C translation units, so the prior gate leaked the SPIs into iOS Release. |

### Why #5 needs both halves

The `#else` stub in the `.m` already handles the non-iOS / non-DEBUG branch correctly, so flipping the `#if` is the actual fix. The `cSettings` DEBUG define is belt-and-suspenders so a `swift build`/`swift test` on the package alone also picks up the gate (Xcode auto-defines `DEBUG=1` in its Debug config, but a bare SwiftPM build does not).

### Tests / parity

- The `template ↔ fixture parity` test asserts `DebugBridgeTouch.m.template` is byte-identical to the FixtureApp `.m`, so I applied the same gate to the fixture (it had the identical bug) and added the matching `cSettings` to the fixture `Package.swift`.
- Added three **host-independent** static guards to `test/skill-e2e-ios-swift-build.test.ts`: the `.m` gate is present, the Obj-C target carries the `cSettings` DEBUG define, and the tools-version directive is on line 1. The existing macOS Release nm-scan only covers `DebugBridgeCore` — `DebugBridgeTouch` links UIKit and can't build on a macOS host, so the iOS-Release leak can't be caught at runtime there. These static assertions lock the regression.
- `bun test test/skill-e2e-ios-swift-build.test.ts` → 9 pass / 0 fail locally.

### Downstream verification

Validated end-to-end against the Dugout Sports app (local patches, branch `fix/data-accuracy-fixes`):

```bash
cd "$APP/DebugBridge" && swift build -c release
nm -j .build/arm64-apple-macosx/release/DebugBridgeTouch.build/DebugBridgeTouch.m.o \
  | grep -qE '_setLocationInWindow|_addTouch|_clearTouches|_setHIDEvent' \
  && exit 1 || echo "clean"
```

→ build succeeds and the symbol scan prints `clean`.

The daemon boot-token bug from the same QA session is tracked separately in #1837 (not addressed here).


time-attack:
@16francej — recommend closing this PR because #2269 supersedes it. Its manifest changes are already present, while its wiring deletion conflicts with the current accessor design; #2269 provides the stronger Release guard and final-app verification path. Credit #1845 for identifying the private-SPI Release concern and retain the generic iphoneos archive requirement.

PR #1847, opened 2026-06-03
fix(browse): recover a late-healthy detached daemon instead of a false "Server failed to start" (#1846)

### LINKED ISSUE #1846: browse CLI reports "Server failed to start within Ns" when the detached daemon is actually healthy (Windows-under-load gap left by #1732)
## Summary

The CLI's server-startup probe (`browse/src/cli.ts`) reports `Server failed to start within Ns` as a hard error **even when the detached daemon comes up healthy a second later**. The error is a probe-timeout, not a launch failure: the spawned server is `detached: true` + `.unref()`'d, so it keeps booting independently of whether the CLI's poll loop gave up. The very next `browse status` then shows `Mode: headed`, healthy.

This is the Windows-under-load twin of the macOS/Linux symptom #1732 fixes. #1732 raises `MAX_START_WAIT` (8s → 15s on macOS/Linux to "match the Windows budget") and broadens the Bun `ConnectionRefused` guard — both good. But it leaves two gaps that this issue is about:

1. **The throw site itself is untouched.** #1732 only changes the constant (line 24) and the `sendCommand` error guard (~line 495). The fall-through `throw new Error(\`Server failed to start within ${MAX_START_WAIT / 1000}s\`)` (cli.ts:372) still fires on timeout regardless of whether the daemon actually came up. So #1732 makes the window wider, but the false-negative at the edge of *any* window remains.
2. **`MAX_START_WAIT` has no env override.** Every other tunable in `server.ts` reads from `process.env.BROWSE_*` (`BROWSE_PORT`, `BROWSE_IDLE_TIMEOUT`, etc.), but the startup budget is a hardcoded magic number with no escape hatch when the chosen ceiling still isn't enough.

## Repro (Windows 11, gstack v1.55.0.0)

On a busy machine (20+ `chrome.exe`, ~12 `node.exe` running), `browse connect`:

```
Launching headed Chromium with extension + terminal agent...
[browse] Connect failed: Server failed to start within 15s
```

Immediately after, `browse status`:

```
Status: healthy
Mode: headed
URL: http://127.0.0.1:34567/welcome
Tabs: 1
PID: <pid>
```

`netstat` confirms the server is `LISTENING` on 34567 with `ESTABLISHED` connections from the extension + terminal-agent. So the launch fully succeeded; only the CLI's 15s health-probe gave up early. This is the existing Windows 15s budget — i.e. #1732's "raise everyone to 15s" ceiling does **not** resolve it on Windows under load.

## Root cause

`browse/src/cli.ts`, the `ensureServer()` poll loop:

```ts
const start = Date.now();
while (Date.now() - start < MAX_START_WAIT) {
  const state = readState();
  if (state && await isServerHealthy(state.port)) {
    return state;
  }
  await Bun.sleep(100);
}
// ... reads startup-error.log if present, else:
throw new Error(`Server failed to start within ${MAX_START_WAIT / 1000}s`);
```

The loop polls health every 100ms, then throws on timeout. Because the daemon is detached, it can (and on loaded Windows boxes, does) become healthy after the loop exits. Nothing re-checks before throwing.

## Proposed fix (two parts, both small, source-only in cli.ts)

**1. Final health-check before throwing (structural — removes the false-negative at any load/platform).** Before the timeout `throw`, do one last `readState()` + `isServerHealthy()`; if healthy, return it. Both helpers already exist (`readState(): ServerState | null` at cli.ts:109, `isServerHealthy(port): Promise<boolean>` at cli.ts:124). Roughly:

```ts
// One final check: the detached daemon may have become healthy after the
// poll loop's last tick (common on loaded machines — the server keeps
// booting independent of this probe). Don't report a false failure.
const finalState = readState();
if (finalState && await isServerHealthy(finalState.port)) {
  return finalState;
}
throw new Error(`Server failed to start within ${MAX_START_WAIT / 1000}s`);
```

**2. Env override on the budget (escape hatch — matches the existing `BROWSE_*` pattern).**

```ts
const MAX_START_WAIT = parseInt(process.env.BROWSE_START_WAIT || '', 10)
  || (IS_WINDOWS ? 15000 : (process.env.CI ? 30000 : 15000));
```

Part 1 is the real fix (eliminates the misleading message without tuning a magic number); part 2 gives power users a knob for genuinely slow environments. They compose cleanly on top of #1732: #1732 widens the window, this removes the false-negative at the edge of whatever window exists.

## Notes

- Complements #1732 (does not conflict — different lines). If preferred, the final-check could fold into that branch instead of a separate change.
- Distinct from #1781 (crash-loop, fixed in v1.55) and the open Windows PTY/DACL issues (#1605, #1221) — this is purely the CLI's startup-probe error-handling, and reproduces on an install where the PTY/DACL paths are healthy.
- Reporter env: Windows 11, gstack v1.55.0.0, Bun 1.3.13.

harjothkhara:
## Summary

`browse`'s `startServer()` polls `/health` for `MAX_START_WAIT`, then throws `Server failed to start within Ns` on timeout. But the server is spawned **detached + `.unref()`'d**, so it keeps booting independently of the CLI's poll loop. On a loaded machine the daemon can become healthy in the gap between the loop's final tick and the throw — the very next `browse status` shows it listening and healthy. The probe timed out; the launch did not.

Closes #1846.

## What this changes

1. **Final health re-check before the throw (the actual fix).** Before the timeout `throw` in `startServer`, do one last `readState()` + `isServerHealthy()`; if healthy, return that state. This mirrors the post-loop recovery already present in `ensureServer()` (`cli.ts:482-483`) — `startServer` was simply missing the same guard. A genuinely failed server stays unhealthy and falls straight through to the existing startup-error-log report, so real failures are unaffected.

2. **`BROWSE_START_TIMEOUT` env override.** `MAX_START_WAIT` was a hardcoded magic number — the only `browse` startup tunable without a `BROWSE_*` escape hatch (unlike `BROWSE_PORT`, `BROWSE_IDLE_TIMEOUT`, …). It's now produced by a small pure, exported `resolveStartTimeout(env)` that honors `BROWSE_START_TIMEOUT` (ms) and falls back to the platform default for absent/non-positive/unparseable values.

## Relationship to #1732

Complementary, not overlapping. #1732 only widens the constant (8s → 15s for non-Windows). This PR removes the false-negative **at the throw site itself**, so it holds at any budget/platform/load, and adds the env knob #1732 left out. The two changes touch different lines and compose cleanly.

## Tests

`browse/test/cli-start-final-healthcheck.test.ts`:
- Behavioral coverage of `resolveStartTimeout` — positive override honored; `0` / negative / garbage / empty fall back to the platform default.
- Static invariants that the final `readState()`+`isServerHealthy()` re-check sits between the poll loop and the throw, returns the recovered state, and that `MAX_START_WAIT` is wired through `resolveStartTimeout()`.

```
$ bun test browse/test/cli-start-final-healthcheck.test.ts
 6 pass  0 fail
```

Existing `cli-*` / `restart-env` tests stay green. (The unrelated `commands.test.ts` failure on a fresh checkout is a missing-`npx playwright install` browser binary — reproduces identically on `main`.)

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -21,7 +21,26 @@ import { spawnTerminalAgent } from './terminal-agent-control';
 
 const config = resolveConfig();
 const IS_WINDOWS = process.platform === 'win32';
-const MAX_START_WAIT = IS_WINDOWS ? 15000 : (process.env.CI ? 30000 : 8000); // Node+Chromium takes longer on Windows
+
+/**
+ * Startup health-probe budget (ms) for a freshly spawned server. The daemon is
+ * detached + unref'd, so it keeps booting regardless of how long the CLI is
+ * willing to poll — this constant only bounds how long `startServer` waits
+ * before reporting failure.
+ *
+ * Overridable via `BROWSE_START_TIMEOUT` (ms) for hosts where even the platform
+ * ceiling isn't enough — e.g. Windows under heavy load (#1846), where the 15s
+ * budget can still elapse before a busy box finishes booting Node+Chromium.
+ * Mirrors the `BROWSE_*` tunable convention used throughout server.ts
+ * (BROWSE_PORT, BROWSE_IDLE_TIMEOUT, ...). A non-positive or unparseable value
+ * falls back to the platform default. Pure + exported for tests.
+ */
+export function resolveStartTimeout(env: NodeJS.ProcessEnv = process.env): number {
+  const platformDefault = IS_WINDOWS ? 15000 : (env.CI ? 30000 : 8000); // Node+Chromium takes longer on Windows
+  const override = parseInt(env.BROWSE_START_TIMEOUT || '', 10);
+  return Number.isFinite(override) && override > 0 ? override : platformDefault;
+}
+const MAX_START_WAIT = resolveStartTimeout();
 
 export function resolveServerScript(
   env: Record<string, string | undefined> = process.env,
@@ -355,6 +374,17 @@ async function startServer(extraEnv?: Record<string, string>): Promise<ServerSta
       return state;
     }
     await Bun.sleep(100);
+  }
+
+  // One last check before declaring failure. The daemon is detached + unref'd,
+  // so on a loaded machine it can become healthy in the gap between the poll
+  // loop's final tick and now — the probe timed out, the launch did not
+  // (#1846). Re-checking here turns that false negative into a success, and
+  // mirrors the post-loop recovery already done in ensureServer(). A genuinely
+  // failed server is still unhealthy, so this falls through to the error report.
+  const lateState = readState();
+  if (lateState && await isServerHealthy(lateState.port)) {
+    return lateState;
   }
 
   // Server didn't start in time — check the on-disk startup error log.
--- a/browse/test/cli-start-final-healthcheck.test.ts
+++ b/browse/test/cli-start-final-healthcheck.test.ts
@@ -1,0 +1,77 @@
+/**
+ * Coverage for #1846 — `browse` CLI must not report "Server failed to start
+ * within Ns" when the detached daemon actually came up healthy a moment later.
+ *
+ * The spawned server is `detached: true` + `.unref()`'d, so it keeps booting
+ * independently of the CLI's poll loop. On a loaded machine (the issue repro is
+ * Windows under load) the loop's budget can elapse in the gap between its last
+ * health tick and the daemon becoming ready — the very next `browse status`
+ * then shows a healthy, listening server. #1732 only widened the budget; the
+ * throw site itself still fired on timeout regardless of real health.
+ *
+ * Two invariants are defended here:
+ *   1. `startServer` does a final readState()+isServerHealthy() re-check before
+ *      the timeout throw (structural — removes the false negative at any budget).
+ *   2. The startup budget is env-overridable via BROWSE_START_TIMEOUT, matching
+ *      the BROWSE_* tunable convention (BROWSE_PORT, BROWSE_IDLE_TIMEOUT, ...).
+ *
+ * (1) is a static source invariant (live spawn cycles belong in the e2e tier);
+ * (2) is exercised behaviorally against the exported pure helper.
+ */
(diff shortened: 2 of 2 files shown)


jbetala7:
Independently reviewed against current `main` (`c43c850c`) — looks correct and complete.

The asymmetry the PR describes is real on `main`: `startServer`'s poll loop (`browse/src/cli.ts:352-358`) drops straight to the `throw new Error('Server failed to start within ...')` at `:372` with no final probe, while `ensureServer` re-reads state and re-probes health (`:482-485`) before it ever decides to restart. Since the daemon is spawned `detached: true` + `.unref()`'d (`:344-345`), it keeps booting independently of the CLI's loop, so on a loaded box it can become healthy in the gap between the loop's last tick and the throw — exactly the #1846 repro. The added `readState()` + `isServerHealthy()` recheck before the throw recovers that false negative, and a genuinely-failed server stays unhealthy and falls through to the existing error-log report, so there's no masking of real failures.

The `BROWSE_START_TIMEOUT` override is NaN-safe (`Number.isFinite(override) && override > 0`, else platform default) and follows the established `BROWSE_*` tunable convention. Extracting it as a pure exported `resolveStartTimeout` keeps it unit-testable; using a static-source-invariant test for the recheck (rather than a live spawn) is the right call given live spawn cycles belong in the e2e tier.

Collision check: no competing open PR for #1846 / `startServer` health. The env-override is slightly beyond the strict fix but is directly motivated by the Windows-under-load repro in the issue and is well-contained. LGTM.

harjothkhara:

time-attack:
@16francej can merge

harjothkhara:
Closing — this shipped in the v1.64 wave via https://github.com/garrytan/gstack/commit/008dd65b1fc3df8af618408f5aea37a24dcea411.

The final `readState()` + `isServerHealthy()` re-check before the timeout throw and the `BROWSE_START_TIMEOUT` override are both on main in `browse/src/cli.ts`, with `browse/test/cli-start-final-healthcheck.test.ts` covering them, so this PR is redundant. Thanks for the credit in the commit message.

PR #1850, opened 2026-06-03
fix(sync-gbrain): gate code-search guidance on real page_count, not the markdown capability check (#1844)

### LINKED ISSUE #1844: /sync-gbrain Step 4 writes false code-search guidance into CLAUDE.md on markdown-only brains
## Summary

On a markdown-only brain (no registered code source), `/sync-gbrain` Step 4 writes the `## GBrain Search Guidance` block into the user's CLAUDE.md claiming working `.gbrain-source` pins and `gbrain code-def`/`code-refs`/`code-callers` symbol search. None of that resolves against anything — the code source never registered. The agent is then instructed to use tools that return nothing, and the block silently overwrites any hand-corrected version on every run.

## Environment

- gstack v1.55.0.0
- gbrain 0.42.8.0, engine pglite (local-stdio MCP), macOS
- Brain: single `default` markdown source at the repo path; `~/.gstack/` artifacts; no code source

## Repro

1. Have a brain where the repo path is registered as the markdown `default` source (a standard `/setup-gbrain` outcome).
2. Run `/sync-gbrain`. The orchestrator's code stage fails:

   ```
   ERR  code  source registration failed: gbrain sources add
   gstack-code-… failed: path "/…/repo" overlaps with existing source
   "default" at "/…/repo". Overlapping sources are not allowed.
   ```

3. Step 4's capability check (`CAPABILITY_OK`) still passes — it only tests a markdown `put`/`search`/`delete` round-trip, which works.
4. Because `CAPABILITY_OK=1`, Step 4 writes the verbatim template asserting code-symbol search. The block now tells the agent to use `code-def` etc., which have a 0-page corpus.

## Root cause

Step 4 gates the code-search template on the **markdown** capability check, not on whether the **code source** actually registered. Step 3 already computes the code source `page_count`, but only uses it to *prompt for a `--full` reindex* — it doesn't feed the Step 4 branch. So `page_count=0` still yields code-search guidance.

Compounding: a code source can't even register when it overlaps the `default` markdown source at the same path (the error above), so on these brains code search is structurally impossible — yet the guidance claims it works.

## Suggested fix

Gate the code-search portion of the Step 4 template on the code source's real `page_count` (already available from Step 3), not on `CAPABILITY_OK`:

- code source `page_count > 0` → write the full template (code + memory).
- code source absent / `page_count == 0` → write a markdown-only variant (semantic/keyword search + `~/.gstack/` memory; point code-symbol questions at Grep).

Optionally, when the code stage fails with the overlap error, surface it as a real warning instead of letting Step 4 paper over it.

## Secondary: stale template content

The current template recommends `gbrain autopilot --install` for "ongoing auto-sync," and the "Safety" note warns about a running `gbrain autopilot`. The `autopilot` command is gone in gbrain 0.42.x (replaced by `sync --watch` / `sync --install-cron`). On pglite (single-writer) a background sync daemon also can't coexist with an always-on MCP `gbrain serve` — that combination crash-loops. The template should drop the `autopilot --install` guidance.

harjothkhara:
## Summary

On a **markdown-only brain** (a standard `/setup-gbrain` outcome — the repo path registered as the markdown `default` source, no code source), `/sync-gbrain` Step 4 writes the `## GBrain Search Guidance` block into CLAUDE.md claiming working `.gbrain-source` pins and `code-def`/`code-refs`/`code-callers` symbol search. None of that resolves: the code source never registered (and structurally **can't** — it overlaps the `default` markdown source at the same path, which gbrain rejects). The agent is then told to use tools that return nothing, and the block silently overwrites any hand-corrected version on every run.

Closes #1844.

## Root cause

Step 4 gated the code-search template on the **markdown** capability round-trip (`CAPABILITY_OK`) — a `put`/`search`/`delete` check that says nothing about symbol search. Step 3 already computes the code source's `page_count` but only used it to *prompt for a `--full` reindex*; it never fed the Step 4 branch. So `page_count=0` still produced full code-search guidance.

## What this changes (`sync-gbrain/SKILL.md.tmpl`, the source of truth)

1. **Gate on the real code corpus.** Step 4's capability check now also computes the cwd source's `CODE_PAGES` (reusing Step 3's exact query) and branches on it:
   - `CAPABILITY_OK=1` **AND** `CODE_PAGES > 0` → the existing full **code + memory** block.
   - `CAPABILITY_OK=1` **AND** `CODE_PAGES` is `0`/empty → a new **markdown-only** block: advertises semantic/keyword search + `~/.gstack/` curated memory, and explicitly routes *all* code-symbol questions to Grep — no false `.gbrain-source` / `code-def` / `code-refs` / `code-callers` claims.
   - `CAPABILITY_OK=0` → remove the block (unchanged).

2. **Drop stale `gbrain autopilot --install` guidance.** That command is gone in gbrain 0.42.x (replaced by `sync --watch` / `sync --install-cron`), and on pglite (single-writer) a background sync daemon can't coexist with an always-on MCP `gbrain serve`. The `/sync-gbrain`-after-changes guidance and the still-valid `sources add --path` safety note remain.

## Verification

- Edited the source-of-truth `SKILL.md.tmpl`; regenerated `SKILL.md` with `bun run gen:skill-docs`. The diff is confined to the two `sync-gbrain` files, and regeneration is **idempotent** (a second run is a no-op).
- The new Step 4 bash passes `bash -n`.
- Relevant suites green: `gen-skill-docs`, `gen-skill-docs-idempotency`, `gstack-gbrain-sync`, `no-stale-gstack-brain-refs`, `audit-compliance`, `section-manifest-consistency`, `preamble-compose` — **453 tests, 0 fail**.

### DIFF
--- a/sync-gbrain/SKILL.md
+++ b/sync-gbrain/SKILL.md
@@ -885,8 +885,9 @@ tmp-file + atomic rename. Concurrent runs are blocked by a lock file at
 After the sync run, query gbrain for the cwd source's page_count:
 
 ```bash
-SOURCE_ID=$(grep -o '"source_id":"[^"]*"' ~/.gstack/.gbrain-sync-state.json 2>/dev/null \
-  | head -1 | sed 's/.*"source_id":"//;s/".*//')
+# NOTE: the sync-state file is pretty-printed (JSON.stringify(state, null, 2)
+# in bin/gstack-gbrain-sync.ts), so a minified-JSON grep never matches. Use jq.
+SOURCE_ID=$(jq -r 'first(.. | objects | .source_id? // empty)' ~/.gstack/.gbrain-sync-state.json 2>/dev/null)
 PAGES=$(gbrain sources list --json 2>/dev/null \
   | jq -r --arg id "$SOURCE_ID" '.sources[] | select(.id==$id) | .page_count' 2>/dev/null \
   || echo 0)
@@ -949,8 +950,9 @@ gbrain delete "$SLUG" 2>/dev/null || true
 # On a markdown-only brain (no code source, or one that can't register because
 # its path overlaps the `default` markdown source) this is 0, so the
 # code-search guidance must gate on it, not on CAPABILITY_OK.
-SOURCE_ID=$(grep -o '"source_id":"[^"]*"' ~/.gstack/.gbrain-sync-state.json 2>/dev/null \
-  | head -1 | sed 's/.*"source_id":"//;s/".*//')
+# NOTE: the sync-state file is pretty-printed (JSON.stringify(state, null, 2)
+# in bin/gstack-gbrain-sync.ts), so a minified-JSON grep never matches. Use jq.
+SOURCE_ID=$(jq -r 'first(.. | objects | .source_id? // empty)' ~/.gstack/.gbrain-sync-state.json 2>/dev/null)
 CODE_PAGES=$(gbrain sources list --json 2>/dev/null \
   | jq -r --arg id "$SOURCE_ID" '.sources[] | select(.id==$id) | .page_count' 2>/dev/null \
   || echo 0)
--- a/sync-gbrain/SKILL.md
+++ b/sync-gbrain/SKILL.md
@@ -942,17 +942,30 @@ if [ -f ~/.gbrain/config.json ] && \
   fi
 fi
 gbrain delete "$SLUG" 2>/dev/null || true
+
+# Code-corpus check (#1844). The round-trip above only proves the *markdown*
+# put/search path works — it says nothing about symbol search. code-def /
+# code-refs / code-callers need a registered code source with indexed pages.
+# On a markdown-only brain (no code source, or one that can't register because
+# its path overlaps the `default` markdown source) this is 0, so the
+# code-search guidance must gate on it, not on CAPABILITY_OK.
+SOURCE_ID=$(grep -o '"source_id":"[^"]*"' ~/.gstack/.gbrain-sync-state.json 2>/dev/null \
+  | head -1 | sed 's/.*"source_id":"//;s/".*//')
+CODE_PAGES=$(gbrain sources list --json 2>/dev/null \
+  | jq -r --arg id "$SOURCE_ID" '.sources[] | select(.id==$id) | .page_count' 2>/dev/null \
+  || echo 0)
+[ -z "$CODE_PAGES" ] && CODE_PAGES=0
 ```
 
-Then update CLAUDE.md based on capability state:
+Then update CLAUDE.md based on capability state **and whether an indexed code
+corpus exists** (`CODE_PAGES` from the check above). In every write case the
+block is idempotent: find the HTML-comment-delimited block; replace its body if
+it exists; append at the end of CLAUDE.md if it doesn't. NEVER duplicate. The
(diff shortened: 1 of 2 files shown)


jbetala7:
Independently reviewed against current `main` (`c43c850c`) — looks correct and complete.

Confirmed the bug on `main`: Step 4 gates the entire `## GBrain Search Guidance` block on `CAPABILITY_OK` alone (`sync-gbrain/SKILL.md:949`, "**If `CAPABILITY_OK=1`** — write or update the block"). But `CAPABILITY_OK` is set by a markdown `put`/`search`/`delete` round-trip (`:924-937`), which proves nothing about symbol search. On a markdown-only brain that block still advertises `code-def`/`code-refs`/`code-callers` and a `.gbrain-source` pin, none of which resolve. Gating the code-search guidance on `CODE_PAGES` is the correct signal, and the markdown-only fallback block is the right outcome.

One thing worth calling out as *correct*, not a smell: the PR recomputes `SOURCE_ID` + `CODE_PAGES` in Step 4 even though Step 3 already does the same extraction (`:888-893`). That's required here — per the repo's own SKILL-authoring rule, each bash block runs in a separate shell and variables don't persist between blocks, so Step 3's `$PAGES`/`$SOURCE_ID` aren't visible in Step 4. The `grep '"source_id"' | head -1` form is the same one Step 3 already uses, and `.gbrain-sync-state.json`'s `source_id` is the code stage's source (`bin/gstack-gbrain-sync.ts`: "Code stage carries source_id + page_count"), so `head -1` picks the code source, not the markdown one.

Both `sync-gbrain/SKILL.md` and `sync-gbrain/SKILL.md.tmpl` are edited, so `check-freshness` should confirm no generator drift. Collision check: no competing open PR for #1844 (#1575 also touches `sync-gbrain` but is the source-fallback filesystem-walk path, unrelated). LGTM.

harjothkhara:

time-attack:
@harjothkhara Rewrite the test lookup to handle the pretty-printed JSON produced by JSON.stringify(state, null, 2); the current minified-JSON grep leaves SOURCE_ID empty.

harjothkhara:
Good catch @time-attack — confirmed: `bin/gstack-gbrain-sync.ts` writes the state file with `JSON.stringify(state, null, 2)`, so the minified-JSON grep never matched and `SOURCE_ID` came back empty every time (forcing `CODE_PAGES=0` even with an indexed code corpus). Pushed f359599 replacing the grep/sed with a jq recursive lookup in both SKILL.md and the template:

```bash
SOURCE_ID=$(jq -r 'first(.. | objects | .source_id? // empty)' ~/.gstack/.gbrain-sync-state.json 2>/dev/null)
```

Verified against a pretty-printed sample matching the writer's output (old grep: empty; jq: extracts the id) and it stays empty-safe when the file is missing. Regenerated via `gen-skill-docs` so SKILL.md and the tmpl stay in sync.

harjothkhara:
Rebased onto current main (v1.67.1.0) — the conflict with the #2503 litter-cleanup is resolved and both changes coexist. The bug is still live on main (`sync-gbrain/SKILL.md` still gates the search-guidance block on `CAPABILITY_OK` alone), #1844 is still open, and time-attack's test-lookup catch is folded in. Clean and ready to enqueue whenever you get a chance @garrytan.

PR #1851, opened 2026-06-03
fix(context-recovery): look up <branch>-reviews.jsonl with the canonical slug form, not the raw branch (#1127)

### LINKED ISSUE #1127: Branch name sanitization disagrees across gstack: reviews for the same branch land in multiple files, dashboard misses entries
## Summary

Branches with a `/` in the name (the conventional `feat/*`, `fix/*`, `chore/*`) are sanitized **three different ways** across the gstack codebase. Writers and the reader use different transforms, so the same branch's review history gets split across up to two files on disk and the review-readiness dashboard reads from only one of them.

## Three transforms coexist

All three are run against a bare `git rev-parse --abbrev-ref HEAD` / `git branch --show-current`, to produce a `$BRANCH` that then gets used in `$GSTACK_HOME/projects/$SLUG/$BRANCH-reviews.jsonl`:

1. **Strip form** (`tr -cd 'a-zA-Z0-9._-'`) — `feat/foo` → `featfoo`.
   `bin/gstack-slug:44`. This is what `bin/gstack-review-log` (writer) and `bin/gstack-review-read` (reader) both use, because they both `eval "$(gstack-slug)"`.
2. **Dash form** (`tr '/' '-'`) — `feat/foo` → `feat-foo`.
   `ship/SKILL.md:1624` (Step 8 plan audit), and other skill steps that do their own `git branch --show-current | tr '/' '-'` inline.
3. **Bare slash** (no transform) — `feat/foo` stays `feat/foo`.
   `context-restore` / `ship` / `review` / every skill's Context Recovery block sets `_BRANCH=$(git branch --show-current ...)` and then uses it in `grep "\"branch\":\"${_BRANCH}\""` against jsonl files and as `"$_PROJ/${_BRANCH}-reviews.jsonl"` existence checks.

## Evidence — same branch, multiple files on disk

`~/.gstack/projects/invoice_approval/` on my machine, after a handful of ship cycles:

```
feat-ocr-monorepo-migration-reviews.jsonl      ← branch "feat/ocr-monorepo-migration"
featocr-monorepo-migration-reviews.jsonl       ← SAME branch, different sanitizer

feat-clerk-v7-webhooks-reviews.jsonl           ← branch "feat/clerk-v7-webhooks"
featclerk-v7-webhooks-reviews.jsonl            ← SAME branch, different sanitizer

feat-characterization-tolerance-reviews.jsonl  ← branch "feat/characterization-tolerance" (only dash form exists)

fixocr-conftest-aws-env-isolation-reviews.jsonl  ← branch "fix/ocr-conftest-aws-env-isolation" (only strip form)
```

Two branches have the review history split across two files. For the others, only one transform wrote any entries this time around, so only one file exists — but the historical record is incomplete either way.

## Impact

1. **Review-readiness dashboard undercounts reviews.** Step 1 of `/ship` calls `bin/gstack-review-read`, which reads via the strip form. Reviews written by any caller using the dash form (including parts of /ship itself via inline `tr '/' '-'`) are invisible to the dashboard.
2. **Cross-review finding dedup is broken for slashed branches.** `ship/SKILL.md:2481` (Step 9.3) calls `gstack-review-read` (strip form) and compares against the current HEAD's findings. Entries written under the dash form never participate in "skipped findings shouldn't be re-flagged" dedup.
3. **context-restore's 'LAST_SESSION' and 'RECENT_PATTERN' checks use bare-slash `$_BRANCH`** to grep `timeline.jsonl`. `timeline.jsonl` appears to use the bare form consistently (those greps do work), but any jsonl-writing skill that uses a different transform leaves entries that context-restore can't cross-reference.
4. **Silent by design.** Every call site pipes to `2>/dev/null || true`, so failed writes never surface. Users don't notice their review history is being split until they look at disk.

## Suggested fix

Normalize branch in **exactly one place** and use it everywhere. Options:

- **Preferred: single sanitizer in `bin/gstack-slug`**, switch the output to dash form (`feat/foo` → `feat-foo`) since it's more readable on disk and reversible-ish. Update the `tr` command on `bin/gstack-slug:44` and remove all ad-hoc `tr '/' '-'` and bare `$BRANCH` uses across the skill files; everyone does `eval "$(gstack-slug)"` and uses `$BRANCH`.
- Add `mkdir -p "$(dirname "$TARGET_FILE")"` before every `echo ... >> ...-reviews.jsonl` as a backstop, so the bare-slash case at least creates the `fix/` subdir instead of silently failing (this would catch users who write their own inline ship steps).
- Ship a one-time migration in `setup` / `gstack-upgrade`: for each `featfoo-*.jsonl` + `feat-foo-*.jsonl` pair in `~/.gstack/projects/*/`, merge to one canonical filename. Without migration, existing users' history stays split forever.

## Repro

```bash
# On a fresh project, check out a branch with a slash
git checkout -b feat/demo-bug

# Run /review then /ship. Both complete successfully.
# Now look at disk:
ls -la ~/.gstack/projects/<your-project>/*reviews.jsonl
# Expect to see at least one "featdemo-bug-reviews.jsonl" (strip form)
# and possibly a "feat-demo-bug-reviews.jsonl" (dash form) depending on
# which skills got run.
```

## Env

- gstack version: whatever the current upstream was as of 2026-04-21
- macOS 25.4.0, zsh
- Git with SSH remote

## Discovery context

Noticed while running `/ship` on `fix/ocr-conftest-aws-env-isolation`. Ran `gstack-review-read` at Step 1 — got `NO_REVIEWS` despite a prior `gstack-review-log` call from Step 9 that reported success. Disk inspection showed the review went into `fixocr-conftest-aws-env-isolation-reviews.jsonl` (strip form) while the read attempt was checking the wrong path.

harjothkhara:
## Summary

The **Context Recovery** preamble block (present in every skill) reports `REVIEWS: N entries` by probing `$_PROJ/${_BRANCH}-reviews.jsonl`, where `$_BRANCH` is the **raw** `git branch --show-current`. But `gstack-review-log` (writer) and `gstack-review-read` (reader) both name that file with the **sanitized slug form** from `gstack-slug` (`$BRANCH`, restricted to `[a-zA-Z0-9._-]`).

For any slashed branch — the conventional `feat/*`, `fix/*`, `chore/*` — the two forms disagree (`feat/foo` → `feat/foo` vs `featfoo`), so the current branch's review count was **never found**. (`$_BRANCH` isn't even defined inside this self-contained block, making the lookup doubly fragile.)

Closes #1127.

## Audit of every `*-reviews.jsonl` access

| Site | Branch form | Role |
|---|---|---|
| `bin/gstack-review-log` | `$BRANCH` via `gstack-slug` (strip) | writer ✓ |
| `bin/gstack-review-read` | `$BRANCH` via `gstack-slug` (strip) | reader ✓ |
| `ship/SKILL.md.tmpl` | `$BRANCH` via `gstack-slug` (strip) | writer ✓ |
| **Context Recovery block** | **raw `$_BRANCH`** | existence check ✗ |
| `specialist-stats`, `retro`, `artifacts-init` | `*` glob | branch-agnostic ✓ |

So on current `main` the writers and reader already agree on the `gstack-slug` form — the Context Recovery existence check was the sole straggler.

## Fix

Route the review lookup through `$BRANCH` (already in scope from the block's `eval gstack-slug`), matching how reviews are written and read. The `timeline.jsonl` greps keep using raw `$_BRANCH` — timeline records the *raw* branch in its `branch` field — and the generator now documents the two-form convention inline so they don't get conflated again.

Edited the source-of-truth generator `scripts/resolvers/preamble/generate-context-recovery.ts`; regenerated all SKILL.md via `bun run gen:skill-docs` (idempotent — the per-skill diff is exactly the one review line).

## Scope

Deliberately converges the one remaining straggler onto the existing canonical sanitizer (`gstack-slug`). It does **not** change `gstack-slug`'s output form, so there's **no on-disk migration** and no conflict with the in-flight `gstack-slug` PRs. Changing the canonical form itself (strip → dash for readability, as the issue suggests) is an orthogonal, migration-bearing decision left to the maintainer — now a one-line change in a single function.

## Tests

`test/regression-1127-context-recovery-branch-form.test.ts` pins both halves (reviews → `$BRANCH`, timeline greps → `$_BRANCH`) for both `claude` and `codex` hosts. Green together with `preamble-compose`, `review-log`, `gen-skill-docs`, `gen-skill-docs-idempotency`, `audit-compliance` — **415 tests, 0 fail**.

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -579,7 +579,7 @@ _PROJ="${GSTACK_HOME:-$HOME/.gstack}/projects/${SLUG:-unknown}"
 if [ -d "$_PROJ" ]; then
   echo "--- RECENT ARTIFACTS ---"
   find "$_PROJ/ceo-plans" "$_PROJ/checkpoints" -type f -name "*.md" 2>/dev/null | xargs ls -t 2>/dev/null | head -3
-  [ -f "$_PROJ/${_BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${_BRANCH}-reviews.jsonl" | tr -d ' ') entries"
+  [ -f "$_PROJ/${BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${BRANCH}-reviews.jsonl" | tr -d ' ') entries"
   [ -f "$_PROJ/timeline.jsonl" ] && tail -5 "$_PROJ/timeline.jsonl"
   if [ -f "$_PROJ/timeline.jsonl" ]; then
     _LAST=$(grep "\"branch\":\"${_BRANCH}\"" "$_PROJ/timeline.jsonl" 2>/dev/null | grep '"event":"completed"' | tail -1)
--- a/canary/SKILL.md
+++ b/canary/SKILL.md
@@ -571,7 +571,7 @@ _PROJ="${GSTACK_HOME:-$HOME/.gstack}/projects/${SLUG:-unknown}"
 if [ -d "$_PROJ" ]; then
   echo "--- RECENT ARTIFACTS ---"
   find "$_PROJ/ceo-plans" "$_PROJ/checkpoints" -type f -name "*.md" 2>/dev/null | xargs ls -t 2>/dev/null | head -3
-  [ -f "$_PROJ/${_BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${_BRANCH}-reviews.jsonl" | tr -d ' ') entries"
+  [ -f "$_PROJ/${BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${BRANCH}-reviews.jsonl" | tr -d ' ') entries"
   [ -f "$_PROJ/timeline.jsonl" ] && tail -5 "$_PROJ/timeline.jsonl"
   if [ -f "$_PROJ/timeline.jsonl" ]; then
     _LAST=$(grep "\"branch\":\"${_BRANCH}\"" "$_PROJ/timeline.jsonl" 2>/dev/null | grep '"event":"completed"' | tail -1)
--- a/codex/SKILL.md
+++ b/codex/SKILL.md
@@ -574,7 +574,7 @@ _PROJ="${GSTACK_HOME:-$HOME/.gstack}/projects/${SLUG:-unknown}"
 if [ -d "$_PROJ" ]; then
   echo "--- RECENT ARTIFACTS ---"
   find "$_PROJ/ceo-plans" "$_PROJ/checkpoints" -type f -name "*.md" 2>/dev/null | xargs ls -t 2>/dev/null | head -3
-  [ -f "$_PROJ/${_BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${_BRANCH}-reviews.jsonl" | tr -d ' ') entries"
+  [ -f "$_PROJ/${BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${BRANCH}-reviews.jsonl" | tr -d ' ') entries"
   [ -f "$_PROJ/timeline.jsonl" ] && tail -5 "$_PROJ/timeline.jsonl"
   if [ -f "$_PROJ/timeline.jsonl" ]; then
     _LAST=$(grep "\"branch\":\"${_BRANCH}\"" "$_PROJ/timeline.jsonl" 2>/dev/null | grep '"event":"completed"' | tail -1)
--- a/context-restore/SKILL.md
+++ b/context-restore/SKILL.md
@@ -575,7 +575,7 @@ _PROJ="${GSTACK_HOME:-$HOME/.gstack}/projects/${SLUG:-unknown}"
 if [ -d "$_PROJ" ]; then
   echo "--- RECENT ARTIFACTS ---"
   find "$_PROJ/ceo-plans" "$_PROJ/checkpoints" -type f -name "*.md" 2>/dev/null | xargs ls -t 2>/dev/null | head -3
-  [ -f "$_PROJ/${_BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${_BRANCH}-reviews.jsonl" | tr -d ' ') entries"
+  [ -f "$_PROJ/${BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${BRANCH}-reviews.jsonl" | tr -d ' ') entries"
   [ -f "$_PROJ/timeline.jsonl" ] && tail -5 "$_PROJ/timeline.jsonl"
   if [ -f "$_PROJ/timeline.jsonl" ]; then
     _LAST=$(grep "\"branch\":\"${_BRANCH}\"" "$_PROJ/timeline.jsonl" 2>/dev/null | grep '"event":"completed"' | tail -1)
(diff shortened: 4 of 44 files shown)


jbetala7:
Independently reviewed against current `main` (`c43c850c`) — looks correct and complete. The split-form handling is exactly right:

- **Timeline greps correctly keep `$_BRANCH` (raw).** In the assembled preamble `_BRANCH` is defined upstream by `generate-preamble-bash.ts` (`_BRANCH=$(git branch --show-current ...)`), and that raw value is what the timeline writers record — both the started event in `generate-preamble-bash.ts` and the completed event in `generate-completion-status.ts` write the `branch` field from `git branch --show-current` (raw). So grepping `timeline.jsonl` by `$_BRANCH` matches what was written.
- **Reviews lookup correctly switches to `$BRANCH` (slug).** `gstack-review-log` writes `$SLUG/$BRANCH-reviews.jsonl` and `gstack-review-read` reads the same, where `$BRANCH` comes from `gstack-slug` (`tr -cd 'a-zA-Z0-9._-'`). For any slashed branch (`fix/foo` → `fixfoo`) the raw and slug forms diverge, so the old `${_BRANCH}-reviews.jsonl` probe never matched the on-disk file — confirmed on `main`.

Collision check: no competing open PR (#1621 also touches the context-recovery block but only its slug-derivation, unrelated to the reviews-filename form). Regression test `regression-1127-context-recovery-branch-form.test.ts` covers it. LGTM.

harjothkhara:

harjothkhara:
Heads-up for maintainers (not a merge-ping): the Trunk merge queue removed this PR — and the sibling fixes #1850 and #1847 — with:

> required statuses are not defined in `.trunk/trunk.yaml`, your repo's branch protection rules for `main`, or the merge-queue specific settings.

So the red **Trunk Merge Queue (main)** check on these is a queue-configuration gap, not anything in the diffs — all three are `MERGEABLE` and have independent review confirmation. Flagging in case the merge-queue needs a required-status definition to accept PRs again. Happy to rebase any of them if useful.

harjothkhara:
Follow-up, @garrytan — these three (#1851 / #1850 / #1847) are still blocked only by the **Trunk merge-queue config**, not by the diffs (which are mergeable and independently reviewed). The queue keeps rejecting with "required statuses are not defined in `.trunk/trunk.yaml`, branch protection rules for `main`, or the merge-queue settings." Flagging again since it's been ~2 weeks — once the queue's required-status config is defined these should be able to land. Happy to rebase any if needed.

time-attack:
@harjothkhara Rewrite this so only the review filename uses $BRANCH, while timeline JSON continues using raw $_BRANCH, matching the tested change.

harjothkhara:
@time-attack that is what this PR already does, so I think we agree on the shape. Pointing at the specifics in case it reads otherwise:

In `scripts/resolvers/preamble/generate-context-recovery.ts`, only the reviews filename moved to `$BRANCH`:

```
- [ -f "$_PROJ/${_BRANCH}-reviews.jsonl" ] && ...
+ [ -f "$_PROJ/${BRANCH}-reviews.jsonl" ] && ...
```

The timeline greps below it still use raw `$_BRANCH`, unchanged, because timeline.jsonl records the raw branch in its `branch` field.

The regression test pins both halves so a later fix cannot convert the timeline greps by mistake:

- `reviews.jsonl lookup uses canonical $BRANCH, not bare $_BRANCH`
- `timeline.jsonl greps keep using raw $_BRANCH`

The other 43 files in the diff are the generated SKILL.md outputs, which carry the same one-line change from the generator.

Happy to adjust if you meant something different.

ShuratCode:
Still needed on **1.60.1.0** — re-confirmed today against a live project dir
(details in #1127). The read path here is the correct half of the fix.

Flagging that I'm opening a complementary **writer-side** PR (`bin/gstack-slug`
+ ship Step 20). The two are independent and should land cleanly in either
order — this one canonicalizes the reads, mine canonicalizes the writes. Happy
to rebase behind it.

garrytan:

PR #1853, opened 2026-06-03
fix(retro): persist recommendations so the next retro detects follow-through (#1834)

### LINKED ISSUE #1834: /retro should persist recommendations and detect follow-through in subsequent runs
## Problem

`/retro` generates "3 Things to Improve" recommendations as prose output but does not save them to the JSON snapshot in `.context/retros/`. On the next run, it loads the prior snapshot for trend comparison but has no structured data about what it recommended.

This means the retro cannot detect when current-week work addresses prior recommendations. In practice, a week spent acting on retro feedback (security hardening, dependency bumps) gets mischaracterized as generic "fix ratio is high" instead of "2 of 3 prior recommendations addressed."

## Proposed fix

1. Add a `recommendations` array to the Step 13 JSON schema:

```json
{
  "recommendations": [
    {"category": "security", "text": "Run pip-audit on a schedule"},
    {"category": "testing", "text": "E2E fixture reliability pass"},
    {"category": "architecture", "text": "Extract html_generator rendering modules"}
  ]
}
```

2. In Step 12 (Load History & Compare), cross-reference prior `recommendations` against current-week commit messages and file changes. Report which were addressed, which are still open.

3. Add a "Retro Follow-Through" section to the Step 14 narrative, before "3 Things to Improve":

```
Retro Follow-Through (from May 7):
  [x] Security hardening — 7 PRs shipped (#1301, #1315, #1318, ...)
  [x] Dependency audit — lxml, Authlib, idna bumped (#1352, #1304)
  [ ] E2E fixture reliability — still fragile (#1272, #1332 were patches, not a root fix)
```

## Why it matters

The retro loop is a forcing function. If it cannot see its own prior output, it treats retro-driven work as unexplained and generates misleading narratives about what the week was actually about.

jbetala7:
## Problem

`/retro` generates its "3 Things to Improve" recommendations as prose output but never saves them to the JSON snapshot in `.context/retros/`. On the next run, Step 12 loads the prior snapshot for trend comparison but has no structured record of what it recommended.

The result: a week spent acting on retro feedback (security hardening, dependency bumps, the exact things the last retro asked for) gets mischaracterized as a generic "fix ratio is high" instead of "2 of 3 prior recommendations addressed." The improvement loop never closes.

Fixes #1834.

## Root cause

The Step 13 snapshot schema (`retro/SKILL.md.tmpl`) persists metrics, authors, streak, greptile/backlog/test_health, and a tweetable — but not the recommendations. Step 12 reads the prior snapshot back for trend deltas, yet there is nothing in it to compare current-week work against.

## Fix (repo-scoped `/retro`)

Three template edits in `retro/SKILL.md.tmpl`, then regenerated `retro/SKILL.md`:

1. **Step 13** — the snapshot schema gains a mandatory `recommendations` array of `{category, text}` objects, populated from the "3 Things to Improve" items. Unlike `greptile`/`backlog`/`test_health` (omitted when empty), this is always written, because it is the input the next run reads back.
2. **Step 12** — after loading the prior snapshot, each prior recommendation is scored against this window's commit subjects, changed files, and metrics as `addressed` / `partial` / `open`, surfaced in a **Recommendation follow-through** section ("2 of 3 prior recommendations addressed"). Snapshots written before this field existed have no `recommendations` key and are skipped silently, so old `.context/retros/` history doesn't break.
3. **"3 Things to Improve" narrative** — wired to record the same 3 items into the snapshot array, closing the loop.

Matches the schema the issue proposed. Scoped to the repo-scoped retro that writes `.context/retros/`; global mode (`~/.gstack/retros/`) is left as a follow-up to keep this to one issue.

## Testing

New `test/regression-1834-retro-recommendations.test.ts`, following the `#1624` static-invariant convention — it asserts the full persist → read-back loop in both the template and the regenerated `SKILL.md`, so the capability can't silently regress:

```
bun test test/regression-1834-retro-recommendations.test.ts   # 8 pass
```

Verified the guard bites: **8/8 fail without the template fix, 8/8 pass with it.** `test/skill-validation.test.ts` and the existing `test/regression-1624-retro-stale-base.test.ts` still pass.

Note: `gen-skill-docs.test.ts` has 8 pre-existing `--host all` / `--host gbrain` dry-run freshness failures on a clean `main` (`c43c850c`) on my machine, unrelated to this change — this PR adds zero new failures. The external-fork eval/report CI jobs may also fail on empty-env auth as usual; not actionable from a fork.

### DIFF
--- a/retro/SKILL.md
+++ b/retro/SKILL.md
@@ -1224,6 +1224,22 @@ Deep sessions:      3      →    5           ↑2
 
 **If no prior retros exist:** Skip the comparison section and append: "First retro recorded — run again next week to see trends."
 
+**Recommendation follow-through.** If the most recent prior snapshot has a `recommendations` array (snapshots written before this field existed won't — if it's absent, skip this block silently), check whether this window acted on each one. For every prior recommendation, scan this window's commit subjects (Step 1 git log), changed files, and the metrics you just computed for evidence it was addressed. Classify each as:
+
+- `addressed` — clear evidence in commits/files/metrics (cite it),
+- `partial` — some movement but not done,
+- `open` — no evidence this window.
+
+Surface a **Recommendation follow-through** section and feed the verdict into the narrative:
+```
+Recommendation follow-through (vs last retro):
+  [x] testing — E2E fixture reliability pass        addressed: 4 commits under test/fixtures/, flake ratio 1/4 → 0/12
+  [~] security — gitleaks on a schedule             partial: pre-commit hook added, no scheduled run yet
+  [ ] architecture — extract html_generator         open
+2 of 3 prior recommendations addressed.
+```
+When recommendations were acted on, say so explicitly in the narrative ("2 of 3 prior recommendations addressed") instead of attributing the same work to a generic metric like "fix ratio is high." This is the whole point of persisting recommendations: a week spent on retro feedback should read as follow-through, not noise.
+
 ### Step 13: Save Retro History
 
 After computing all metrics (including streak) and loading any prior history for comparison, save a JSON snapshot:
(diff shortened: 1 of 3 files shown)


adamkwhite:
Following up on my review above: confirming the scope limitation I raised is a fast-follow, not a blocker for this PR.

Re-verified against this repo's snapshots (worktree off the PR head, bun test 8/8 green, matcher run by hand). The addressed/partial/open classification is correct for in-repo work, which is the common case. The only gap is externally-actioned recommendations (a fix that lands in a different repo, a CI or process change with no commit footprint in the retro'd project) scoring `open` permanently. That's real but narrow, and it's additive: a later `scope: repo|external` tag or a one-line narrative caveat ("open may mean not-matchable-here") slots in on top of this schema without changing it.

No reason to hold the merge on it. This closes the loop the issue asked for. Happy to open the scope-tag follow-up once this lands.

time-attack:
@garrytan thoughts?

time-attack:
@16francej Merge this version because it passed the Mac checks and is the chosen implementation in this group.

PR #1857, opened 2026-06-04
update skills

dxiao2003:
add check during eng review to make sure we have basics like environment defined in the spec


time-attack:
Thanks for this. The proposed checks assume web-app requirements too broadly and may write project files or inspect secrets during plan review. Please scope checks by artifact type, prohibit secret reads and project writes, suppress inapplicable questions, and add focused tests.

Issue #1474, opened 2026-05-13
Proposal: platform-agnostic versioning for /ship (no-VERSION-file projects)


dep0we:
Hi — opening this as a design discussion (not a PR yet). I'm running gstack on a Python framework that manages release identity through git tags + `CHANGELOG.md` headers only — no `VERSION` file at the repo root. The release model is "accumulate in `## [Unreleased]`, periodic release cuts promote to a dated `## [X.Y.Z]` header" — pretty close to standard Keep a Changelog flow.

`/ship` is excellent and I want every PR to go through it. Two hard-fail points block that today:

1. **Step 9** STOPs when `.claude/skills/review/checklist.md` is missing. The skill is intentionally strict on this; my project just didn't have a project-local checklist.
2. **Step 12** `exit 1`s on the regex `^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$` because there's no 4-digit `VERSION` file. The queue-aware workspace-ship feature (`bin/gstack-next-version`) is also pinned to 4-digit input.

This is consistent with gstack's documented model (workspace-aware ship, VERSION as monotonic ordered identifier, every PR is its own release) but doesn't fit a project that ships releases via tags + CHANGELOG promotion only.

## What I proposed in my local fork

For my own unblock today I patched `ship/SKILL.md.tmpl` on a local branch with two strictly-additive guards:

**Step 9 fallback (missing-checklist):**
- If `.claude/skills/review/checklist.md` is missing, fall back to an inline default checklist covering universal categories (SQL safety, LLM trust boundary, concurrency/atomicity, error handling, auth/secrets, public surface hygiene, dependency risk) and warn. Existing behavior (STOP on unreadable-but-present file) preserved.

**Step 12 NO_VERSION_FILE mode:**
[...]

Issue #430, opened 2026-03-24
Versioning Scheme


anhldbk:
Hi @garrytan ,

Thank you for this outstanding framework. 

While using gstack for a green-field project, I've got an issue while running the generated code:

```sh
$ npx run-the-command
npm error Invalid Version: 0.1.0.0
npm error A complete log of this run can be found in: ...
```
Looks like npm only support [SemVer](https://docs.npmjs.com/about-semantic-versioning). I replaced all "0.1.0.0" with "0.1.0" and the command works.

My question is: why do we use a new version scheme (`x.y.z.t`) in this project? For example:

- [in our package.json](https://github.com/garrytan/gstack/blob/dc5e0538e5ab6b0208cb6b3d71d28b3f6d9aebf9/package.json#L3)
- [in our tests](https://github.com/search?q=repo%3Agarrytan%2Fgstack+0.1.0.0&type=code)

travisbotello:
Hitting this from the other direction: the 4-digit scheme doesn't just break `npm`, it breaks **Prisma**, and `/ship` re-breaks it automatically.

`prisma generate` parses `package.json` `version` as strict SemVer and aborts:

```
Error: Invalid version: "1.0.0.1"
```

The workaround is what @-the-reporter did — let the two files carry different formats:

| File | Format | Consumer |
|---|---|---|
| `VERSION` | `1.0.0.2` (4-digit) | gstack /ship |
| `package.json` | `1.0.2` (SemVer) | Prisma / npm |

But that workaround **doesn't survive the next ship**. `bin/gstack-version-bump` dual-writes the same 4-digit string into both files unconditionally, and there's no opt-out:

```ts
if (existsSync(join(cwd, "package.json"))) {
  try {
    writePkgVersion(cwd, version!);
  }
```

Flag parsing only recognizes `--version` and `--version-path`. So on a repo that has deliberately split the formats, `classify` reports `DRIFT_UNEXPECTED` and `write` silently reintroduces the invalid version — putting the broken deploy back. Ship's own SKILL.md tells the agent to STOP on `DRIFT_UNEXPECTED`, which is correct, but there's then no supported way forward except hand-editing both files and bypassing the tool.

Two things that would fix it:
1. A `--no-package-json` flag (or a `version_sync_package_json: false` config key) so `write`/`repair` leave `package.json` alone.
2. Teach `classify` that a deliberate 4-digit/3-digit split is a valid state, not `DRIFT_UNEXPECTED`.

[...]

PR #1862, opened 2026-06-04
fix(ship): sync package-lock.json version alongside package.json

oharriehausen:
## Problem

`/ship` Step 12 (carved into `bin/gstack-version-bump` by #1806) bumps `VERSION` and `package.json`'s version but leaves `package-lock.json`'s version field stale. Every JS-repo ship then carries a spurious 2-line `package-lock.json` diff, and `npm ci` warns on the version mismatch.

This is the same bug #1757 fixed against the old inline-bash surface. #1757 is now **CONFLICTING** because #1806 moved that code into the CLI, so it can't be merged as-is. This PR re-ports the fix onto the new CLI surface — it's what #1757 needs to become.

## Fix

`writeLockVersion()` mirrors `writePkgVersion()`'s semantics:
- Silent skip when `package-lock.json` is absent (yarn/pnpm/bun/no JS).
- Malformed lockfile → exit 2, no corrupted write.
- I/O failure after `package.json` already written → exit 3 (half-write).
- Updates both top-level `version` and `packages[""].version` (lockfileVersion 3 stores the root version in two places).

Wired through both `write` (FRESH bump) and `repair` (DRIFT_STALE_PKG). The state classifier is intentionally untouched: a stale lockfile root version never blocks runtime, so it's opportunistically synced but is not a halt condition.

## Tests

- 3 existing `write`/`repair` assertions updated for the new `packageLock` field.
- 3 new tests cover lockfile sync on `write` + `repair` plus the malformed-lockfile exit-2 guard.
- Full `gstack-version-bump` + `ship-version-sync` suites: **32 pass, 0 fail**.

Built on top of v1.55.1.0 (`c43c850`); clean, no conflicts.

Closes the gap left by #1757.

### DIFF
--- a/bin/gstack-version-bump
+++ b/bin/gstack-version-bump
@@ -101,6 +101,35 @@ function writePkgVersion(cwd: string, version: string): void {
   writeFileSync(pkgPath, JSON.stringify(parsed, null, 2) + "\n");
 }
 
+/**
+ * Opportunistically sync package-lock.json (lockfileVersion 3 stores the version
+ * in two places: top-level `version` and `packages[""].version`). Returns true if
+ * the lockfile was present and written, false if absent (yarn/pnpm/bun lockfile,
+ * or no JS at all). Throws on malformed JSON or write failure.
+ *
+ * Caller decides whether absence is an error — write() and repair() both treat
+ * absence as a silent skip. The state classifier intentionally does NOT inspect
+ * the lockfile: a stale lockfile root version doesn't break `npm ci` or anything
+ * runtime, so lockfile-only drift is opportunistically corrected on FRESH/repair
+ * but is not a halt condition.
+ */
+function writeLockVersion(cwd: string, version: string): boolean {
+  const lockPath = join(cwd, "package-lock.json");
+  if (!existsSync(lockPath)) return false;
+  const raw = readFileSync(lockPath, "utf-8");
+  let parsed: Record<string, unknown>;
+  try {
+    parsed = JSON.parse(raw) as Record<string, unknown>;
+  } catch {
+    fail("package-lock.json is not valid JSON. Fix the file before re-running /ship.", 2);
+  }
+  parsed.version = version;
+  const packages = parsed.packages as Record<string, { version?: string }> | undefined;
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej — recommend closing this PR as superseded by the package-version policy tracked in #1474; #430 is being consolidated there as a duplicate. This branch can change VERSION and package.json before validating a malformed or unwritable lockfile, leaving versions inconsistent. Please credit #1862 and carry its atomic preflight, format transformation, malformed-file, unwritable-file, and rerun-convergence requirements into #1474.

oharriehausen:
The partial-write critique was correct, so I updated the PR to preflight the lockfile before either write path mutates files.

The flow is now `planLockVersion()` first, then the existing writes, then `commitLockVersion()`. An absent `package-lock.json` still silently skips. One that exists but is unreadable, malformed, or not a JSON object exits 2 before `VERSION` or `package.json` are touched. Write-time I/O failures still exit 3 — I didn't try to make the three writes fully atomic with temp-file + rename, but say the word if you'd rather. The `packages[""]` sync is also shape-checked now rather than truthiness-checked, since a primitive there would throw on assignment.

Also added no-lockfile coverage for write and repair (@buddy7599bot's suggestion), plus the malformed/non-object cases. 38 tests pass across `gstack-version-bump` and `ship-version-sync`.

On #1474/#430: I don't think they supersede this as-is. They're open policy proposals, not merged implementations. Main still has `writePkgVersion()` and no lockfile writer, so the current `/ship` path can still leave `package-lock.json` stale. #1757 fixed this on the old inline-bash surface; after #1806 moved the logic into `bin/gstack-version-bump`, this is the re-port onto the current surface rather than a competing change.

I think the two are separable: land the narrow consistency fix for the current `VERSION` + `package.json` model now, and let #1474 decide broader version-authority policy later. Happy to carry the preflight / format-transformation / rerun-convergence requirements into #1474 either way — merging this doesn't consume them. If the intended policy is that `/ship` should never touch lockfiles, I can close this; otherwise, is this OK to merge via Trunk, or should it wait for #1474?

PR #1863, opened 2026-06-04
fix: add windowsHide to browse detached server spawn on Windows

### LINKED ISSUE #1835: fix(windows): bun server and chrome-headless-shell show black console windows on Windows
## Summary

On Windows, when gstack browse starts the server, two black console windows appear on the taskbar and desktop:

1. `bun.exe` — the browse server process
2. `chrome-headless-shell.exe` — Playwright's Chromium headless shell

## Root Cause

In `browse/src/cli.ts` (Windows path, ~line 326), the launcher code that spawns the bun server via `child_process.spawn` is missing the `windowsHide: true` option:

```js
// Before (creates visible console window)
`{detached:true,stdio:['ignore','ignore','ignore'],env:...}`

// After (suppresses console window via CREATE_NO_WINDOW flag)
`{detached:true,windowsHide:true,stdio:['ignore','ignore','ignore'],env:...}`
```

On Windows, `child_process.spawn({ detached: true })` allocates a new console window for the spawned process unless `windowsHide: true` is explicitly set (which maps to the Win32 `CREATE_NO_WINDOW` creation flag).

The `chrome-headless-shell.exe` window is a secondary effect — when the bun server process has a visible console, Playwright's Chromium child inherits or also gets a console window. Fixing the bun spawn should resolve both.

## Fix

One-line change in `browse/src/cli.ts`:

```diff
-      `{detached:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,`
+      `{detached:true,windowsHide:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,`
```

## Environment

- Shell: PowerShell + bash (Git for Windows)
- Bun: `~/.nvm/bin/bun.exe`
- Playwright chromium: `ms-playwright\chromium-headless_shell-1308`

mvanhorn:
## Summary

Adds `windowsHide: true` to the detached server-launcher spawns in `browse/src/cli.ts` so starting the gstack browse server on Windows no longer flashes a black console window.

## Why this matters

Issue #1835 reports that on Windows, starting the browse server flashes two black console windows: the detached `bun.exe` server process and Playwright's `chrome-headless-shell.exe`. The detached launcher spawn near line 326 of `browse/src/cli.ts` sets `{detached:true,stdio:['ignore','ignore','ignore'],...}` but omits `windowsHide:true`, and `child_process.spawn({detached:true})` allocates a new console window on Windows unless `windowsHide:true` (Win32 `CREATE_NO_WINDOW`) is set; the headless-shell window is a secondary effect of the parent having a visible console. This adds `windowsHide:true` to the inlined launcher options object and to the sibling `child_process.spawn` detach call. `windowsHide` is a no-op on POSIX, so it is safe unconditionally.

## Testing

`browse/test/cli-windows-hide.test.ts` is a static-grep regression test in the repo's tripwire style: it asserts the launcher options string carries both `detached:true` and `windowsHide:true`, that the second detached `child_process.spawn` site also includes `windowsHide`, and fails if any future `detached:true` spawn site in `browse/src/cli.ts` drops it. Full suite runs in CI.

Fixes #1835

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -285,6 +285,7 @@ function raiseHeadedWindowMacOS(): void {
     nodeSpawn('osascript', ['-e', 'tell application "Google Chrome for Testing" to activate'], {
       stdio: 'ignore',
       detached: true,
+      windowsHide: true,
     }).unref();
   } catch {
     // osascript missing or app not present — non-fatal
@@ -323,7 +324,7 @@ async function startServer(extraEnv?: Record<string, string>): Promise<ServerSta
     const launcherCode =
       `const{spawn}=require('child_process');` +
       `spawn(process.execPath,[${JSON.stringify(NODE_SERVER_SCRIPT)}],` +
-      `{detached:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,` +
+      `{detached:true,windowsHide:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,` +
       `${extraEnvStr})}).unref()`;
     Bun.spawnSync(['node', '-e', launcherCode], { stdio: ['ignore', 'ignore', 'ignore'] });
   } else {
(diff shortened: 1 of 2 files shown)


mvanhorn:
Closing this in favor of #1864, which now carries both halves of the fix: the spawnSync conversion for the outer cold-start flash and (as of 77ebc14) the windowsHide flags on the inner detached server spawn, the osascript raise helper, and the POSIX server spawn that this PR added. The test invariant from this branch (every detached spawn site carries windowsHide) moved over as well, so nothing here is lost.

PR #1864, opened 2026-06-04
fix: suppress Windows console flashes from browse cold-start child spawns

### LINKED ISSUE #1784: Windows: browse cold-start flashes focus-stealing console windows (Bun.spawnSync ignores windowsHide)
### Summary
On Windows 11, every **cold start** of `browse` (i.e. when no healthy daemon is already running) briefly flashes **2+ focus-stealing console windows** on the interactive desktop. They steal foreground focus and interrupt typing in other apps. The cause is several console-subsystem child processes spawned without an *effective* `windowsHide`.

### Root cause (verified with an elevated `Win32_ProcessStartTrace`)
Tracing process creation during a cold `browse goto` shows the flashing `conhost.exe` windows are parented to:

| spawned process | spawned by | call site |
|---|---|---|
| `node.exe` (the detach launcher) | `browse.exe` | `browse/src/cli.ts` `startServer()` — `Bun.spawnSync(['node','-e',launcherCode], …)` |
| `git.exe` (`rev-parse --show-toplevel`) | `browse.exe` | `browse/src/config.ts` `getGitRoot()` and `browse/src/find-browse.ts` — `Bun.spawnSync(['git', …])` |
| `icacls.exe` ×2 | daemon `node` | `browse/src/file-permissions.ts` `restrictFilePermissions` / `restrictDirectoryPermissions` — `execFileSync('icacls', …, { stdio:'ignore' })` (no `windowsHide`) |

### The key finding
**`Bun.spawnSync` does not honor the `windowsHide` option on Windows** (tested on bun 1.3.11). Adding `windowsHide: true` to a `Bun.spawnSync(...)` call has no effect — the child still allocates a console window.

Node's `child_process` (`spawn` / `spawnSync` / `execFileSync`) **does** honor `windowsHide`. (Confirmed in the same trace: the detached daemon `node`, spawned via Node's `child_process.spawn({ detached:true, windowsHide:true })` inside `launcherCode`, did **not** get a console window — only the `Bun.spawnSync`-spawned launcher and git did.)

So the fix is: for Windows console spawns, use Node's `child_process` with `windowsHide:true` rather than `Bun.spawnSync`, and add `windowsHide:true` to the `execFileSync('icacls', …)` calls.

### Suggested fix
```diff
# browse/src/cli.ts
-import { spawn as nodeSpawn } from 'child_process';
+import { spawn as nodeSpawn, spawnSync as nodeSpawnSync } from 'child_process';
@@ startServer() (Windows branch)
-      `{detached:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,` +
+      `{detached:true,windowsHide:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,` +
       `${extraEnvStr})}).unref()`;
-    Bun.spawnSync(['node', '-e', launcherCode], { stdio: ['ignore', 'ignore', 'ignore'] });
+    // Bun.spawnSync ignores windowsHide on Windows -> the launcher node flashes a
+    // console. Node's spawnSync honors it.
+    nodeSpawnSync('node', ['-e', launcherCode], { stdio: ['ignore', 'ignore', 'ignore'], windowsHide: true });

# browse/src/config.ts  (and browse/src/find-browse.ts likewise)
+import { execFileSync } from 'child_process';
-    const proc = Bun.spawnSync(['git', 'rev-parse', '--show-toplevel'], { stdout:'pipe', stderr:'pipe', timeout:2_000 });
-    if (proc.exitCode !== 0) return null;
-    return proc.stdout.toString().trim() || null;
+    // Node's execFileSync honors windowsHide; Bun.spawnSync does not.
+    const out = execFileSync('git', ['rev-parse', '--show-toplevel'],
+      { encoding:'utf-8', timeout:2_000, windowsHide:true, stdio:['ignore','pipe','ignore'] });
+    return out.trim() || null;

# browse/src/file-permissions.ts  (both execFileSync('icacls', …) calls)
-        { stdio: 'ignore' },
+        { stdio: 'ignore', windowsHide: true },
```

The detached daemon spawn just needs the `windowsHide:true` added to its options object (Node honors it).

### Repro
1. Windows 11, no `browse` daemon running.
2. `browse goto https://example.com`.
3. Observe 2+ console windows flash and steal focus. An elevated `Win32_ProcessStartTrace` shows the `conhost.exe` parents are `node` (launcher), `git`, and `icacls`.

### Environment
- Windows 11, gstack @ `920a13a`, bun 1.3.11, Node from `C:\Program Files\nodejs`.
- Default headless path. (Separately, `chrome-headless-shell.exe` is a console-subsystem binary; launching full chromium with `--headless=new` avoids that one too, but it was not the cause of these particular flickers.)

mvanhorn:
## Summary

Suppresses the focus-stealing console windows that flash on every cold start of `browse` on Windows by routing the console-subsystem child spawns through a hidden path. `execFileSync('icacls', ...)` in `browse/src/file-permissions.ts` gets `windowsHide: true`, and the `Bun.spawnSync` launcher and git-root calls in `browse/src/cli.ts` and `browse/src/config.ts` fall back to Node's `child_process` with `windowsHide: true` on Windows, since `Bun.spawnSync` does not honor `windowsHide` on bun 1.3.11.

## Why this matters

Issue #1784 reports that on Windows 11 every cold start of `browse` flashes two or more focus-stealing `conhost.exe` windows that interrupt typing in other apps. A `Win32_ProcessStartTrace` traced the flashes to console-subsystem children spawned without an effective `windowsHide`: the `node` detach launcher and `git rev-parse` calls go through `Bun.spawnSync`, which ignores `windowsHide` on bun 1.3.11, and the `icacls.exe` calls use `execFileSync` without the flag. The fix adds `windowsHide: true` to the `execFileSync` calls in `file-permissions.ts` and, on Windows only, routes the launcher and git lookups in `cli.ts` and `config.ts` through Node's `child_process` (which honors the flag), keeping `Bun.spawnSync` on POSIX behind the existing platform check.

## Testing

`browse/test/cli-windows-hide.test.ts` is a static regression test asserting the cold-start spawn sites carry `windowsHide` or use the hidden-spawn helper on Windows, and fails if a console-subsystem spawn site reintroduces a non-hidden Windows spawn. POSIX continues to use `Bun.spawnSync` unchanged. Full suite runs in CI.

Fixes #1784

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -326,7 +326,7 @@ async function startServer(extraEnv?: Record<string, string>): Promise<ServerSta
     const launcherCode =
       `const{spawn}=require('child_process');` +
       `spawn(process.execPath,[${JSON.stringify(NODE_SERVER_SCRIPT)}],` +
-      `{detached:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,` +
+      `{detached:true,windowsHide:true,stdio:['ignore','ignore','ignore'],env:Object.assign({},process.env,` +
       `${extraEnvStr})}).unref()`;
     const proc = nodeSpawnSync('node', ['-e', launcherCode], {
       stdio: ['ignore', 'ignore', 'ignore'],
@@ -347,6 +347,7 @@ async function startServer(extraEnv?: Record<string, string>): Promise<ServerSta
     // the Windows path's rationale — same root cause, different OS API.
     nodeSpawn('bun', ['run', SERVER_SCRIPT], {
       detached: true,
+      windowsHide: true,
       stdio: ['ignore', 'ignore', 'ignore'],
       env: { ...process.env, BROWSE_STATE_FILE: config.stateFile, BROWSE_PARENT_PID: parentPid, ...extraEnv },
     }).unref();
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -288,6 +288,7 @@ function raiseHeadedWindowMacOS(): void {
     nodeSpawn('osascript', ['-e', 'tell application "Google Chrome for Testing" to activate'], {
       stdio: 'ignore',
       detached: true,
+      windowsHide: true,
     }).unref();
   } catch {
     // osascript missing or app not present — non-fatal
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -11,7 +11,7 @@
 
 import * as fs from 'fs';
 import * as path from 'path';
-import { spawn as nodeSpawn } from 'child_process';
+import { spawn as nodeSpawn, spawnSync as nodeSpawnSync } from 'child_process';
 import { safeUnlink, safeUnlinkQuiet, safeKill, isProcessAlive } from './error-handling';
 import { writeSecureFile, mkdirSecure } from './file-permissions';
 import { resolveConfig, ensureStateDir, readVersionHash } from './config';
@@ -141,10 +141,13 @@ async function killServer(pid: number): Promise<void> {
   if (IS_WINDOWS) {
     // taskkill /T /F kills the process tree (Node + Chromium)
     try {
-      Bun.spawnSync(
-        ['taskkill', '/PID', String(pid), '/T', '/F'],
-        { stdout: 'pipe', stderr: 'pipe', timeout: 5000 }
-      );
+      const proc = nodeSpawnSync('taskkill', ['/PID', String(pid), '/T', '/F'], {
+        stdout: 'pipe',
+        stderr: 'pipe',
+        timeout: 5000,
+        windowsHide: true,
+      });
+      if (proc.error) throw proc.error;
     } catch (err: any) {
       if (err?.code !== 'ENOENT') throw err;
     }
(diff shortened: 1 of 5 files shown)


kaiwulff:
Thanks for picking this up — the Bun.spawnSync→Node `child_process` approach here matches what we've been running as a local patch on Windows since #1784, so this is great to see upstreamed.

One gap worth folding in while you're here: `browse/src/find-browse.ts` has the **same** uncovered call that `config.ts` does. At the current tip its `getGitRoot()` is:

```ts
function getGitRoot(): string | null {
  ...
  const proc = Bun.spawnSync(['git', 'rev-parse', '--show-toplevel'], { ... });
```

That's the identical `git rev-parse` via `Bun.spawnSync` this PR fixes in `config.ts`, but `find-browse.ts` isn't in the changeset (only `cli.ts`, `config.ts`, `file-permissions.ts`). In our `Win32_ProcessStartTrace` for #1784, `getGitRoot()` showed up as a flicker source from **both** `config.ts` and `find-browse.ts`, and `find-browse` compiles to its own console-subsystem binary — so this one will keep flashing after the PR lands. Applying the same Windows-only Node-`child_process` + `windowsHide: true` treatment to `find-browse.ts`'s `getGitRoot()` (and extending the `cli-windows-hide.test.ts` regression to cover that spawn site) should close it.

mvanhorn:
Good to hear it matches what you've been running locally since #1784, that's solid real-world validation. On the overlap flag: #1863 and this PR are complementary rather than competing. #1863 adds windowsHide to the long-lived detached server spawn; this one covers the short-lived cold-start child spawns that flash a console per invocation. Either can land first; no shared lines.

mvanhorn:
Correcting my earlier comment: the two PRs did overlap after all, both touched the startServer launcher block and both created cli-windows-hide.test.ts. Rather than leave a guaranteed merge conflict, #1863's changes are folded in here (commits 4d30746 and 77ebc14): windowsHide on the inner detached server spawn, the osascript raise helper, and the POSIX server spawn, plus its every-detached-spawn-carries-windowsHide test invariant. #1863 is closed; this PR is now the complete console-flash fix.

time-attack:
@mvanhorn Rewrite this to cover the reported Git spawn in find-browse and add a Windows test confirming no visible console window appears.

mvanhorn:
Done (299b3a4). find-browse.ts's getGitRoot() now takes the same Windows-only nodeSpawnSync + windowsHide:true path and keeps Bun.spawnSync on POSIX, matching the config.ts treatment @kaiwulff flagged as the second flicker source. Extended cli-windows-hide.test.ts with a regression pinning the hidden git spawn at that site, so a future revert to Bun.spawnSync there fails the suite; the focused suite passes 8/8. It follows the same static source-assertion approach as the rest of the #1784 coverage rather than spawning a live Windows console in CI.

PR #1881, opened 2026-06-05
fix(browse): add `viewport auto` to unpin a fixed viewport without restart

### LINKED ISSUE #1059: browse: add `viewport auto` / `viewport reset` to unpin fixed viewport

jbetala7:
Fixes #1059

## Problem

`browse viewport WxH` pins the viewport with Playwright's `setViewportSize`, overriding the launch-time `viewport: null` that lets the viewport follow the real Chrome window. Once pinned there was no way back to window-following short of `browse restart`, which kills the Chrome session (and drops a `/connect-chrome` headed attachment back to `launched`). Skills that set a breakpoint as a side-effect (responsive testing, benchmarks) leave the browser stuck at that size, so the visible Chrome window ends up larger than the viewport Claude renders into.

## Repro (before)

```bash
$B viewport 1280x520
$B js "window.innerWidth + 'x' + window.innerHeight"   # 1280x520
# resize Chrome window manually -> viewport stays 1280x520, no way back without restart
```

## Fix

Add `browse viewport auto` (aliases `reset` / `unpin`): it clears the pin and restores window-following without tearing down the session.

- Rebuilds the context with `viewport: null` through the same `recreateContext()` path that `viewport --scale` already uses, so cookies, storage, and open tab URLs are preserved.
- Resets `deviceScaleFactor` to 1, since Playwright forbids a custom scale alongside a null viewport.
- Refused in headed mode (`/connect-chrome`), where the viewport already tracks the real browser window — consistent with how `viewport --scale` refuses headed mode.
- `viewport auto --scale <n>` is rejected as contradictory.

A new private `viewportContextOptions()` helper is the single source of the viewport/scale slice of the context options, used by both the initial launch and both `recreateContext` paths, so the three sites can't drift.

## Testing

`bun test browse/test/commands.test.ts` (227 pass) and `browse/test/handoff.test.ts` (15 pass) — both exercise `recreateContext`. Four new tests in the `viewport auto (unpin)` block assert:

- `viewport auto` clears a pinned size (`page.viewportSize()` -> `null`)
- `reset` / `unpin` are accepted aliases
- `viewport auto` resets `deviceScaleFactor` to 1
- `viewport auto --scale` is rejected

All four fail on `main`, where `viewport auto` is parsed as a malformed size and throws the usage error — confirming the capability was genuinely absent.

`bun run gen:skill-docs` regenerated the command-reference tables and `llms.txt`; skill-validation + gen-skill-docs quality suites pass (719 tests).

## Notes / risk

- No behavior change at launch or for existing `viewport WxH` / `viewport --scale` calls (the launch path resolves to the same pinned options when `viewportAuto` is false).
- The eval / report CI jobs may fail on external-fork auth (empty-env), which is the known non-actionable fork-secrets limitation; code, static, and Windows checks should pass.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -902,7 +902,7 @@ Refs are invalidated on navigation — run `snapshot` again after `goto`.
 | `type <text>` | Type into focused element |
 | `upload <sel> <file> [file2...]` | Upload file(s) |
 | `useragent <string>` | Set user agent |
-| `viewport [<WxH>] [--scale <n>]` | Set viewport size and optional deviceScaleFactor (1-3, for retina screenshots). --scale requires a context rebuild. |
+| `viewport [<WxH>|auto] [--scale <n>]` | Set viewport size and optional deviceScaleFactor (1-3, for retina screenshots). --scale requires a context rebuild. Use `auto` (alias `reset`/`unpin`) to clear a pinned size and follow the window again. |
 | `wait <sel|--networkidle|--load>` | Wait for element, network idle, or page load (timeout: 15s) |
 
 ### Inspection
--- a/browse/SKILL.md
+++ b/browse/SKILL.md
@@ -593,6 +593,7 @@ $B js "document.body.textContent.includes('Success')"
 $B responsive /tmp/layout        # mobile + tablet + desktop screenshots
 $B viewport 375x812              # or set specific viewport
 $B screenshot /tmp/mobile.png
+$B viewport auto                 # unpin: viewport follows the window again (no restart)
 ```
 
 ### 8. Test file uploads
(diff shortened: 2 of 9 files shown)


time-attack:
@jbetala7 Rewrite this onto current main. The state tracking, rollback, aliases, tests, source-template change, and generated documentation are coherent, but the branch now conflicts.

PR #1884, opened 2026-06-05
fix: support Codex request_user_input in AskUserQuestion docs

uncfreak1255-code:
## Summary
- Treat Codex `request_user_input` / `functions.request_user_input` as an AskUserQuestion-equivalent surface in generated skill docs
- Regenerate host skill docs and update ship golden fixtures
- Fix gbrain-detect test PATH fixture so the bun shebang still resolves while real gbrain stays excluded

## Verification
- `bun test test/gen-skill-docs.test.ts test/preamble-compose.test.ts test/skill-validation.test.ts`
- `bun test test/gbrain-detect-install.test.ts`
- `bun run test`
- `git diff --check`

## Notes
- Token-ceiling warnings remain for generated gstack-ship on some non-Codex hosts; separate cleanup candidate.


uncfreak1255-code:
Closing: opened from wrong fork owner by mistake. Will reopen from correct fork.

uncfreak1255-code:
Reopening: confirmed correct fork owner is uncfreak1255-code.

uncfreak1255-code:
Closing: source fork owner is not the correct owner for this work.

uncfreak1255-code:
Reopening: confirmed uncfreak1255-code is the correct fork owner.

jbetala7:
Overlap flag: #1818 (opened 2026-05-31) edits the **same two functions in the same two files** as this PR - `generateAskUserFormat()` and `generatePlanModeInfo()` in `scripts/resolvers/preamble/generate-ask-user-format.ts` and `generate-completion-status.ts` - and regenerates the same `test/fixtures/golden/codex-ship-SKILL.md`. They rewrite the same "Tool resolution (read first)" and "Skill Invocation During Plan Mode" paragraphs, so they will merge-conflict.

More important than the textual conflict: the two PRs ship **contradictory failure-path rules** for the same paragraph.

- This PR (#1884): *"If none of those three surfaces appears in your tool list, this skill is BLOCKED."*
- #1818: *"If no callable AskUserQuestion-like tool appears in a Codex session, do NOT report `BLOCKED` by default. Use the normal chat fallback for interactive sessions. Report `BLOCKED` only when the run is non-interactive..."*

Both PRs agree `request_user_input` is a valid AUQ surface (good - that part is not in conflict). Where they diverge is what happens when no surface is callable in an *interactive* Codex session: this PR says hard-BLOCK, #1818 says degrade to a chat-question fallback. Those can't both ship. Worth consolidating with #1818 on a single BLOCKED-vs-fallback policy before either lands. This PR's `functions.request_user_input` alias wording is a nice addition #1818 doesn't carry, so the merge likely wants this PR's surface naming + a single agreed failure-path rule.

jbetala7:
Cross-referencing two things the thread hasn't surfaced yet (separate from the #1818 same-files overlap already noted above):

**This is the docs half of #1066, and the code half is already open as #1101.** #1101 adds `toolRewrites: { AskUserQuestion: request_user_input }` to `hosts/codex.ts`. Without that rewrite, Codex skills still ship the literal string `AskUserQuestion` in their schema-stripped bodies, so documenting `request_user_input` here doesn't make the tool callable on its own — the two halves want to land together against #1066.

**Accuracy fix on the new wording:** it tells the model to "treat `request_user_input` as the AskUserQuestion-equivalent surface" with no caveat, but per #1066 `request_user_input` is **gated to Plan mode** and is rejected in Codex Default mode unless `default_mode_request_user_input` is enabled (off by default). As written, the doc points the model at a surface that isn't callable in Codex's default mode — the exact silent-degradation #1066 reports. Worth noting the Plan-mode constraint (which is also where #1818's BLOCKED-vs-fallback policy belongs).

Net: the `functions.request_user_input` alias naming here is the cleanest of the three, but #1884 (naming) + #1101 (rewrite) + #1818 (failure-path policy) really want to be one consolidated change on #1066 rather than three conflicting ones — #1884 and #1818 already edit the same two resolver files and the same `codex-ship-SKILL.md` golden.

(Validated: the regen is complete — every origin/main `SKILL.md` carrying the old "two tools at runtime" wording is regenerated from the two resolver edits.)

Issue #1202, opened 2026-04-25
Claude install exposes full repo under skill discovery root


ofnullable:
## Problem

Claude installs currently document and support cloning the full gstack repo into `~/.claude/skills/gstack`.

That path is also under Claude's skill discovery tree. Because the full repo contains generated/source skill artifacts for multiple hosts, Claude can recursively discover host-specific nested `SKILL.md` files that were never meant to be Claude skills, for example `openclaw/skills/*/SKILL.md`.

This is the same class of problem fixed for Codex in #235 / #346: the host skill discovery root should not contain the full source checkout.

## Expected behavior

Claude's skill root should contain only Claude-discoverable skills plus a minimal `gstack/` runtime sidecar for shared assets.

Desired layout:

```text
~/.gstack/repos/gstack/          # full gstack checkout
~/.claude/skills/<skill>/        # top-level Claude skills only
~/.claude/skills/gstack/         # runtime assets only, no nested host SKILL.md files
```

## Why it matters

Users invoking a normal Claude skill like `/office-hours` should not see OpenClaw/Hermes/Codex host setup behavior leak into Claude skill discovery. The current layout mixes source checkout, generated host artifacts, and runtime sidecar files in one recursive discovery tree.

## Proposed fix

- Migrate direct Claude installs from `~/.claude/skills/gstack` to `~/.gstack/repos/gstack`.
- Recreate `~/.claude/skills/gstack` as a runtime-only sidecar with `bin/`, `browse/`, `design/`, `review/`, `qa/`, and `ETHOS.md` symlinks.
- Keep actual Claude skills as top-level `~/.claude/skills/<skill>/SKILL.md` entries.
- Update install docs and team-mode instructions to clone/run setup from `~/.gstack/repos/gstack`.
- Add regression coverage proving the Claude sidecar exposes no nested `SKILL.md` files.

skyzer:
I hit this from the Cursor side and implemented the exact layout proposed here on a real machine (global Claude install, gstack `v1.44` → relocated to `~/.gstack/repos/gstack`). Sharing confirmation + a few implementation notes, and offering to send the PR.

[...]

PR #1819, opened 2026-05-31
fix(setup): relocate direct Claude install out of the skill-discovery root (#1202, #1694)

skyzer:
Fixes #1202. Addresses #1694.

## Problem

A direct Claude install puts the **full repo checkout** at `~/.claude/skills/gstack`, which is itself under Claude Code's (and Cursor's) recursive skill-discovery root. Every generated host-mirror `SKILL.md` (`.cursor/`, `.hermes/`, `.agents/`, …) and source skill `SKILL.md` inside the checkout gets indexed — hundreds of phantom skills. Claude Code reports "N skill descriptions dropped"; Cursor's cursor-agent CLI **hard-freezes** (`NGHTTP2_ENHANCE_YOUR_CALM`, per #1694). This is the same class of problem already fixed for Codex (`migrate_direct_codex_install`, #235/#346).

## Approach

Mirror the Codex relocation for Claude:

- **`migrate_direct_claude_install`** — moves a global direct `~/.claude/skills/gstack` checkout to `~/.gstack/repos/gstack` (outside the discovery root). Idempotent: a re-run from the relocated path (e.g. via `/gstack-upgrade`, which `cd`s in and runs `./setup`) re-detects it and keeps the alias root at `~/.claude/skills` instead of falling through to the worktree-symlink branch. Skips `--local` and symlinked worktree/vendored installs.
- **`create_claude_runtime_root`** — builds a runtime-only sidecar at `~/.claude/skills/gstack` (symlinks `bin/`, `browse/dist`, `review/*`, `ETHOS.md`, …) so the absolute asset paths skills hard-code — `~/.claude/skills/gstack/bin/gstack-config` is the single most-referenced path in the repo — keep resolving. Unlike the Codex sidecar, it exposes **zero** nested `SKILL.md`: Claude's root skill is the separate `_gstack-command` alias and `gstack-upgrade` is a top-level skill, so none is needed.

Top-level skill aliases stay in `~/.claude/skills`. `gstack-relink` already honors the `GSTACK_SKILLS_DIR` / `GSTACK_INSTALL_DIR` overrides, and `/gstack-upgrade`'s detection already includes `~/.gstack/repos/gstack/.git`, so upgrades keep working with no further changes.

## Result

On a real machine (global install, v1.44 → relocated): `find ~/.claude/skills -name SKILL.md` drops **613 → 77**, the sidecar exposes **0** nested `SKILL.md`, and `~/.claude/skills/gstack/bin/gstack-config` still resolves.

## Tests

[...]

Issue #1694, opened 2026-05-25
Skill files generated for other AI hosts pollute Claude Code's skill scanner


vschoener:
## Problem

When gstack is installed at `~/.claude/skills/gstack/`, Claude Code recursively scans the entire directory tree for `SKILL.md` files. The host-specific skills generated for other AI assistants (Hermes, Cursor, Kiro, OpenCode, Slate, Factory, Agents, OpenClaw) are picked up by this scanner, causing **513 SKILL.md files** to be discovered instead of the expected ~49 in my situation.

### Impact

Claude Code reports **"N skill descriptions dropped"** at session start (the exact number varies depending on which host integrations are installed — in my case it was 49–51). This happens because Claude Code hits its context window limit trying to load all the discovered skill descriptions, meaning some skills may not be properly available in the preamble.

### Root cause

After installation/update, the following directories are generated inside `~/.claude/skills/gstack/`:

```
~/.claude/skills/gstack/
├── .agents/skills/        ← 46 SKILL.md (for Agents)
├── .cursor/skills/        ← 46 SKILL.md (for Cursor)
├── .factory/skills/       ← 46 SKILL.md (for Factory)
├── .gbrain/skills/        ← 46 SKILL.md (for GBrain)
├── .hermes/skills/        ← 46 SKILL.md (for Hermes)
├── .kiro/skills/          ← 46 SKILL.md (for Kiro)
├── .openclaw/skills/      ← 46 SKILL.md (for OpenClaw)
├── .opencode/skills/      ← 46 SKILL.md (for OpenCode)
├── .slate/skills/         ← 46 SKILL.md (for Slate)
└── openclaw/skills/       ←  4 SKILL.md
```

These are all gitignored (correctly), but Claude Code doesn't know to ignore them.

### Workaround

Manually deleting these directories reduces the count from 513 to ~95:

```bash
for dir in .agents .cursor .factory .gbrain .hermes .kiro .openclaw .opencode .slate; do
  rm -rf ~/.claude/skills/gstack/$dir/skills/
done
```

However, these directories are recreated on the next gstack update.

## Suggested fix

One of:

1. **Move the generated host skills outside `~/.claude/skills/`** — e.g., generate Hermes skills at `~/.hermes/skills/gstack/` (matching `globalRoot` in `hermes.ts`) rather than inside the gstack repo itself when it's installed under `~/.claude/skills/`.

[...]

PR #1885, opened 2026-06-06
fix(setup): add Claude runtime alias for renamed installs

### LINKED ISSUE #1882: Skills hardcode `~/.claude/skills/gstack/` paths — any install dir not named `gstack` silently breaks after `setup`
## Summary

Every skill's `SKILL.md` hardcodes the absolute string `~/.claude/skills/gstack/...` for its own `bin/` and asset calls (e.g. the telemetry/config preamble that runs on *every* skill invocation). `setup`, however, is parameterised on its own checkout location and wires the skill symlinks correctly for **any** directory name. The result: if gstack is installed at `~/.claude/skills/<name>` where `<name>` isn't literally `gstack`, the symlinks resolve but the skills' internal `bin`/asset references point at a non-existent `~/.claude/skills/gstack/` and fail **silently, at skill-invocation time**.

This is the `gstack`-dir-name sibling of #349 (`$CLAUDE_CONFIG_DIR`, the `~/.claude` prefix). Same underlying cause — hardcoded absolute self-paths — different segment. Both ladder up to the "make path resolution dynamic rather than hardcoded" direction from #289.

## Root cause (with file:line)

- `setup:17-18` — the install dir is **derived**, not fixed:
  ```sh
  INSTALL_GSTACK_DIR="$(cd "$(dirname "$0")" && pwd)"
  SOURCE_GSTACK_DIR="$(cd "$(dirname "$0")" && pwd -P)"
  ```
  So `setup` is designed to run from wherever the repo lives, and `INSTALL_SKILLS_DIR="$(dirname "$INSTALL_GSTACK_DIR")"` (`setup:19`). The per-skill install (`setup:535-569`) creates a real dir at `~/.claude/skills/<skill>/` with an **absolute** `SKILL.md` symlink into the actual checkout — correct for any dir name.

- But the skill **bodies** are committed with the literal absolute path. Example — `qa/SKILL.md` at HEAD contains 52 occurrences of `~/.claude/skills/gstack/` across 51 lines (the preamble alone calls `~/.claude/skills/gstack/bin/gstack-config`, `.../bin/gstack-telemetry-log`, etc.).

- `setup` **already rewrites** these paths when it generates the non-Claude host copies — e.g. `setup:~1012-1028` rewrites `.claude/skills/gstack` → `.kiro/skills/gstack` (and the codex/cursor/factory/gbrain equivalents). The primary Claude install is the one host that never gets its in-file paths reconciled with `INSTALL_GSTACK_DIR`'s actual basename.

## Reproduction

```sh
# install under any name that isn't "gstack"
git clone https://github.com/garrytan/gstack ~/.claude/skills/gstack-dev
cd ~/.claude/skills/gstack-dev && ./setup

# the skill symlinks are correct, but the in-file paths are not:
grep -c '~/.claude/skills/gstack/' ~/.claude/skills/gstack-dev/qa/SKILL.md   # > 0
ls ~/.claude/skills/gstack/bin/gstack-config 2>&1                            # No such file or directory
```

Then invoke any skill (e.g. `/qa`): its preamble shells out to `~/.claude/skills/gstack/bin/gstack-config …`, which doesn't exist. No error surfaces to the user at install time — it only fails when a skill runs.

Observed directly while upgrading an install located at `~/.claude/skills/i-gstack`: after `git reset --hard origin/main && ./setup`, every top-level `*/SKILL.md` that references the install path (52 files on that checkout) still pointed at `~/.claude/skills/gstack/`, and `setup` did not reconcile them.

## Impact

- Affects anyone whose global install lives at `~/.claude/skills/<name>` where `<name> != gstack` (`setup` installs the skill siblings into `INSTALL_SKILLS_DIR="$(dirname "$INSTALL_GSTACK_DIR")"`, so it accepts this layout — it just doesn't reconcile the in-file paths).
- **Silent**: nothing fails at install; skills break later, at invocation, in a way that looks like the skill itself is broken.
- Affects every skill (the shared preamble), not a subset.

## Suggested fixes (maintainer's call)

1. **Compatibility symlink (smallest).** When `basename "$INSTALL_GSTACK_DIR" != gstack`, create/refresh `$INSTALL_SKILLS_DIR/gstack -> $INSTALL_GSTACK_DIR` so the hardcoded references resolve with zero file edits. *Caveat:* must not let Claude Code discover the repo-shaped dir as duplicate skills — `setup:589` already has handling for skipping the repo-shaped `gstack` dir, so this needs to compose with that.

2. **Runtime-relative resolution (cleanest, aligns with #289).** Have the preamble resolve its own bin/asset dir relative to the `SKILL.md` location instead of a hardcoded absolute path (a resolved `$GSTACK_DIR`), removing the dir-name dependency entirely. Larger change (touches the shared preamble across all skills).

3. **Extend the existing rewrite to the Claude host (symmetric).** Apply the same path rewrite `setup` already does for kiro/codex to the primary install, mapping `~/.claude/skills/gstack/` → `$INSTALL_GSTACK_DIR`. *Caveat:* the Claude install symlinks to source rather than copying, so this would either dirty the working tree or require switching that host to a copy-and-rewrite step.

## Workaround (for affected users today)

After each install/upgrade, re-point the hardcoded references in the top-level skill files to the real install dir:

```sh
cd ~/.claude/skills/<your-dir>
# macOS/BSD sed shown; on GNU/Linux use `sed -i` (no '').
find . -maxdepth 2 -name SKILL.md -not -path './.*' -print0 \
  | xargs -0 sed -i '' "s|skills/gstack|skills/$(basename "$PWD")|g"
```

Scope to top-level `*/SKILL.md` only; the generated host copies under `.agents/.cursor/.factory/.gbrain/hosts` use a `gstack-` prefix convention and should be left as-is. (This needs re-running after every upgrade, since `git reset --hard` reverts it — which is the motivation for fix #1 or #2.)

## Related

- #349 — `$CLAUDE_CONFIG_DIR` support: the `~/.claude` prefix counterpart of this same hardcoded-self-path problem.
- #289 (closed RFC) — host integration contract / dynamic path resolution: fix #2 above is the concrete form of that direction for the dir-name axis.

maxpetrusenkoagent:
## Summary

Fixes #1882.

Claude-host skill bodies still contain runtime references to `~/.claude/skills/gstack/...`. When the checkout is installed as `~/.claude/skills/<non-gstack>`, setup links slash-command wrappers but leaves those runtime paths unresolved.

This PR adds a narrow setup-time runtime alias:

- create `~/.claude/skills/gstack -> ~/.claude/skills/<install-name>` only for global Claude installs under `$HOME/.claude/skills`
- skip canonical `~/.claude/skills/gstack` installs
- skip non-Claude `skills/` directories to avoid scope creep
- refuse to replace an existing separate real `~/.claude/skills/gstack` install

No generated `SKILL.md` files are rewritten.

## Verification

- `bun test test/setup-claude-runtime-alias.test.ts test/setup-windows-fallback.test.ts test/setup-sections-linking.test.ts test/setup-conductor-worktree.test.ts`
  - 23 pass, 0 fail
- `bash -n setup`
- `git diff --check`
- Temp HOME smoke test:
  - installed from `$TMP_HOME/.claude/skills/gstack-dev`
  - verified `$TMP_HOME/.claude/skills/gstack -> $TMP_HOME/.claude/skills/gstack-dev`
  - verified `$TMP_HOME/.claude/skills/gstack/bin/gstack-config get skill_prefix` runs through the hardcoded runtime path

## Review

- Claude subagent review: approved, no blocking findings
- Codex autoreview first pass: requested a Claude-only scope guard
- Codex autoreview second pass after scope guard: APPROVE
- Post-open exact PR diff Codex review: APPROVE
  - Diff SHA-256: `086d3663f24146281aa563d8eb29019fc1b56c21d73ef2450c97833227d15475`

## CI / checks

GitHub currently reports no checks on the fork branch for this draft PR, so the exact-diff Codex review above was used as the fallback reviewer gate.

## Notes

A broader `bun run test:windows` run still has the pre-existing unrelated `browse/test/terminal-agent.test.ts` lazy-spawn failure. The touched files and this patch do not affect that path.

### DIFF
--- a/setup
+++ b/setup
@@ -606,6 +606,44 @@ link_claude_root_skill_alias() {
   _print_windows_copy_note_once
 }
 
+# Claude skill bodies still reference ~/.claude/skills/gstack/ for runtime
+# assets. If the checkout is installed as ~/.claude/skills/<other-name>, keep a
+# runtime-only gstack alias so those paths resolve without rewriting generated
+# SKILL.md files or dirtying the checkout.
+ensure_claude_runtime_alias() {
+  local gstack_dir="$1"
+  local skills_dir="$2"
+  local install_name
+  local install_parent
+  local alias
+  local existing_real
+  local claude_skills_dir
+
+  claude_skills_dir="$HOME/.claude/skills"
+  [ "$skills_dir" = "$claude_skills_dir" ] || return 0
+
+  install_name="$(basename "$gstack_dir")"
+  [ "$install_name" = "gstack" ] && return 0
+
+  install_parent="$(dirname "$gstack_dir")"
+  [ "$install_parent" = "$skills_dir" ] || return 0
+
+  alias="$skills_dir/gstack"
+  if [ -d "$alias" ] && [ ! -L "$alias" ]; then
+    existing_real=$(cd "$alias" 2>/dev/null && pwd -P || echo "")
+    if [ -n "$existing_real" ] && [ "$existing_real" != "$gstack_dir" ]; then
+      log "  warning: $alias already exists as a separate install; not replacing it"
+      return 0
+    fi
+  fi
+
+  if [ -L "$alias" ] || [ ! -e "$alias" ]; then
+    _link_or_copy "$gstack_dir" "$alias"
+    log "  linked runtime alias: $alias -> $gstack_dir"
+    _print_windows_copy_note_once
+  fi
+}
+
 # ─── Helper: remove old unprefixed Claude skill entries ───────────────────────
 # Migration: when switching from flat names to gstack- prefixed names,
 # clean up stale symlinks or directories that point into the gstack directory.
@@ -982,6 +1020,7 @@ fi
 
 if [ "$INSTALL_CLAUDE" -eq 1 ]; then
   if [ "$SKILLS_BASENAME" = "skills" ]; then
+    ensure_claude_runtime_alias "$INSTALL_GSTACK_DIR" "$INSTALL_SKILLS_DIR"
     # Clean up stale symlinks from the opposite prefix mode
     if [ "$SKILL_PREFIX" -eq 1 ]; then
       cleanup_old_claude_symlinks "$SOURCE_GSTACK_DIR" "$INSTALL_SKILLS_DIR"
--- a/test/setup-claude-runtime-alias.test.ts
+++ b/test/setup-claude-runtime-alias.test.ts
@@ -1,0 +1,123 @@
+import { describe, test, expect } from 'bun:test';
+import { spawnSync } from 'child_process';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+const ROOT = path.resolve(import.meta.dir, '..');
+const SETUP = fs.readFileSync(path.join(ROOT, 'setup'), 'utf-8');
+
+function extractFunction(src: string, name: string): string {
+  const start = src.indexOf(`${name}() {`);
+  if (start < 0) throw new Error(`Could not locate ${name}() in setup`);
+  const end = src.indexOf('\n}\n', start);
+  if (end < 0) throw new Error(`Could not locate end of ${name}() in setup`);
+  return src.slice(start, end + 2);
+}
+
+describe('setup: Claude runtime alias for non-gstack install dirs', () => {
+  test('setup calls the runtime alias helper before linking Claude skills', () => {
+    const helperIdx = SETUP.indexOf('ensure_claude_runtime_alias "$INSTALL_GSTACK_DIR" "$INSTALL_SKILLS_DIR"');
+    const linkIdx = SETUP.indexOf('link_claude_skill_dirs "$SOURCE_GSTACK_DIR" "$INSTALL_SKILLS_DIR"');
+
+    expect(helperIdx).toBeGreaterThan(-1);
+    expect(linkIdx).toBeGreaterThan(-1);
+    expect(helperIdx).toBeLessThan(linkIdx);
+  });
+
+  test.skipIf(process.platform === 'win32')('helper creates skills/gstack -> install dir when install basename is not gstack', () => {
+    const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'gstack-runtime-alias-'));
+    try {
+      const skillsDir = path.join(tmp, '.claude', 'skills');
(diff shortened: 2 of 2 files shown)


maxpetrusenko:
Codex/autoreview: CLEAN

Reviewed the exact PR diff for #1882. No blocking findings.

Evidence checked:
- Runtime alias is scoped to `$HOME/.claude/skills` only.
- Canonical `skills/gstack` installs are left unchanged.
- Existing separate real `skills/gstack` installs are not replaced.
- No generated skill bodies are rewritten.
- Regression tests cover renamed install, non-Claude skills dir, and canonical install cases.

Verification from PR body / agent run:
- `bun test test/setup-claude-runtime-alias.test.ts test/setup-windows-fallback.test.ts test/setup-sections-linking.test.ts test/setup-conductor-worktree.test.ts` PASS, 23 tests
- `bash -n setup` PASS
- `git diff --check` PASS
- temp HOME smoke test PASS

Residual risk: GitHub currently reports no checks on this fork draft PR; `gh pr view` reports `mergeable=MERGEABLE`.

jbetala7:
Possible overlap with #1819 (opened 2026-05-31, Fixes #1202 / addresses #1694) - flagging so the two can be coordinated. #1819 relocates a direct `~/.claude/skills/gstack` checkout out to `~/.gstack/repos/gstack` and explicitly states it keeps `~/.claude/skills` as the alias root, with "a runtime-only sidecar is created later in the Claude install step." This PR adds exactly that runtime sidecar/alias at `~/.claude/skills/gstack` via `ensure_claude_runtime_alias()`.

So the two are two halves of the same problem - a renamed or relocated install must still resolve `~/.claude/skills/gstack/...` runtime paths (here keyed to #1882 vs #1202/#1694) - and both add new functions to the same setup region near `link_claude_root_skill_alias`, so they will likely textually conflict. There is also a real question of whether the sidecar/alias ends up created once or twice once both install paths exist. Might be cleanest to land them together (or rebase one onto the other) so the alias/sidecar creation lives in a single place.

Not a blocker on this PR's logic - the `ensure_claude_runtime_alias` guards (skip the canonical `gstack` install, skip non-Claude `skills/` dirs, refuse to clobber a separate real install) look sound, and the helper-ordering test is a nice touch.

time-attack:
@16francej Close this and incorporate its renamed-install and collision behavior into #1819, so one implementation owns the Claude runtime sidecar.

PR #1894, opened 2026-06-07
fix(model-benchmark): don't drop the claude provider on macOS subscription installs

### LINKED ISSUE #1890: model-benchmark: claude adapter `available()` reports "No Claude auth found" for macOS subscription installs (claude -p works)
### Observed problem

`test/helpers/providers/claude.ts` `ClaudeAdapter.available()` treats Claude auth as present only when `~/.claude/.credentials.json` exists **or** `ANTHROPIC_API_KEY` is set:

```ts
const credsPath = path.join(os.homedir(), '.claude', '.credentials.json');
const hasCreds = fs.existsSync(credsPath);
const hasKey = !!process.env.ANTHROPIC_API_KEY;
if (!hasCreds && !hasKey) {
  return { ok: false, reason: 'No Claude auth found. Log in via `claude` interactive session, or export ANTHROPIC_API_KEY.' };
}
```

On a macOS subscription install (the default Claude Code install on Mac), `~/.claude/.credentials.json` is **absent** — the credential lives elsewhere (likely the Keychain) — yet the subscription session authenticates `claude -p` fine. So the file-existence check is a proxy that misses a valid auth state, and the adapter reports NOT READY for a provider whose `run()` path (`claude -p --output-format json`) would actually succeed.

This is internal to the adapter: `run()` *uses* the subscription via `claude -p`, so `available()` rejecting that same auth state is an inconsistency, not a user-config gap. (`ANTHROPIC_API_KEY` is a workaround, but macOS subscription users have no key and don't need one for `claude -p`.)

### Current behavior on main (cab774cc)

```
$ gstack-model-benchmark --dry-run --prompt "hi"
  claude: NOT READY — No Claude auth found. Log in via `claude` interactive session, or export ANTHROPIC_API_KEY.
```

On the same machine:
- `~/.claude/.credentials.json` — absent
- `claude` CLI — resolves (`claude --version` → 2.1.x), logged in via subscription; `claude -p` works
- `available()` returns `{ok:false}` purely on the missing file

Any code path gated on this `available()` (model-benchmark, and the LLM-judge path) silently drops the claude provider for this population.

### Expected behavior

A logged-in macOS subscription install (no `ANTHROPIC_API_KEY`, no `.credentials.json`, but `claude -p` working) should report the claude adapter as available.

### Possible fix

`run()` already classifies runtime auth failures gracefully (`/unauthorized|auth|login/i` → `auth`). So `available()` could be optimistic when `resolveClaudeCommand()` resolves on macOS (defer the real auth decision to `run()`), or detect the Keychain credential — whichever you prefer.

- Issue search (`model-benchmark auth`, `No Claude auth`, `credentials.json`, `ANTHROPIC_API_KEY subscription`, `claude adapter available`): no existing issue.
- Open PRs touching the adapter / benchmark provider: #1727 (benchmark timeout flag), #1495 (OllamaAdapter), #998 (provider abstraction / Gemini) — none touch `claude.ts` `available()` auth detection. No collision.

jbetala7:
## Problem

`ClaudeAdapter.available()` (`test/helpers/providers/claude.ts`) treats Claude auth as present only when `~/.claude/.credentials.json` exists **or** `ANTHROPIC_API_KEY` is set:

```ts
const credsPath = path.join(os.homedir(), '.claude', '.credentials.json');
const hasCreds = fs.existsSync(credsPath);
const hasKey = !!process.env.ANTHROPIC_API_KEY;
if (!hasCreds && !hasKey) {
  return { ok: false, reason: 'No Claude auth found. ...' };
}
```

On the **default macOS Claude Code subscription install**, `~/.claude/.credentials.json` is absent — the credential lives in the login Keychain — yet the subscription session authenticates `claude -p` fine. So the file-existence check is a proxy that misses a valid auth state, and `available()` reports NOT READY for a provider whose `run()` path (`claude -p --output-format json`) would actually succeed. Any caller gated on this `available()` (model-benchmark, and the LLM-judge path) silently drops the claude provider for that whole population.

### On `main` (cab774cc)

```
$ gstack-model-benchmark --dry-run --prompt "hi"
  claude: NOT READY — No Claude auth found. ...
```
on a machine where `~/.claude/.credentials.json` is absent, `claude --version` resolves, and `claude -p` works.

## Root cause

An absent `~/.claude/.credentials.json` on macOS is **not** evidence of missing auth — it is the expected state for a Keychain-backed subscription install. `available()` rejecting an auth state that `run()` can actually use is an internal inconsistency, not a user-config gap.

## Fix

- Keep the positive signals (creds file or `ANTHROPIC_API_KEY`) as a fast yes.
- On **darwin**, be optimistic when the binary resolves and defer the real auth decision to `run()`, which already classifies a genuinely logged-out state as an `auth` error (`/unauthorized|auth|login/i`). So a logged-out mac still surfaces a clear auth error from `run()` instead of being silently dropped at `available()`.
- **Non-macOS** keeps the strict file/key check, where `~/.claude/.credentials.json` is the actual credential store.

## Testing

Adds `test/claude-adapter-available.test.ts` (resolver satisfied with a real binary via `GSTACK_CLAUDE_BIN`, `os.homedir` pointed at an empty dir, `process.platform` overridden — no live CLI):

- macOS, no creds file, no key → available (**fails on main, passes here**)
- non-macOS, no creds file, no key → not available, with the auth reason
- `ANTHROPIC_API_KEY` set → available on any platform
- present `.credentials.json` → available on non-macOS
- unresolvable binary → not available, before the auth sniff

```
bun test test/claude-adapter-available.test.ts          # 5 pass
bun test test/benchmark-runner.test.ts test/benchmark-cli.test.ts test/claude-adapter-available.test.ts  # 29 pass
```

Fixes #1890

### DIFF
--- a/test/claude-adapter-available.test.ts
+++ b/test/claude-adapter-available.test.ts
@@ -1,0 +1,88 @@
+/**
+ * Unit tests for ClaudeAdapter.available() auth detection (issue #1890).
+ *
+ * The adapter must not drop the claude provider for a logged-in macOS
+ * subscription install, where the credential lives in the login Keychain and
+ * ~/.claude/.credentials.json is absent yet `claude -p` works. available()
+ * stays strict on non-macOS, where ~/.claude/.credentials.json is the actual
+ * credential store.
+ *
+ * Does NOT exercise the live CLI — resolveClaudeCommand is satisfied with a
+ * real binary via GSTACK_CLAUDE_BIN, os.homedir is pointed at an empty dir so
+ * the creds file is reliably absent, and process.platform is overridden.
+ */
+
+import { test, expect, beforeEach, afterEach, spyOn } from 'bun:test';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { ClaudeAdapter } from './helpers/providers/claude';
+
+const adapter = new ClaudeAdapter();
+const origPlatform = process.platform;
+const origKey = process.env.ANTHROPIC_API_KEY;
+const origBin = process.env.GSTACK_CLAUDE_BIN;
+
+let emptyHome: string;
+let homedirSpy: ReturnType<typeof spyOn>;
+
+function setPlatform(value: NodeJS.Platform) {
+  Object.defineProperty(process, 'platform', { value, configurable: true });
+}
+
+beforeEach(() => {
+  // A home with no ~/.claude/.credentials.json, regardless of the real machine.
+  emptyHome = fs.mkdtempSync(path.join(os.tmpdir(), 'claude-adapter-test-'));
+  homedirSpy = spyOn(os, 'homedir').mockReturnValue(emptyHome);
+  // resolveClaudeCommand resolves an existing absolute binary (not the live CLI).
+  process.env.GSTACK_CLAUDE_BIN = process.execPath;
+  delete process.env.ANTHROPIC_API_KEY;
+});
+
+afterEach(() => {
+  homedirSpy.mockRestore();
+  setPlatform(origPlatform);
+  if (origKey === undefined) delete process.env.ANTHROPIC_API_KEY;
+  else process.env.ANTHROPIC_API_KEY = origKey;
+  if (origBin === undefined) delete process.env.GSTACK_CLAUDE_BIN;
+  else process.env.GSTACK_CLAUDE_BIN = origBin;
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for the PR. We can't merge this version because Claude's auth error is returned in stdout JSON, but this change only checks stderr. Please handle the stdout JSON auth error and add a regression test.

time-attack:
@16francej close

PR #1898, opened 2026-06-07
v1.57.10.0 fix(setup): route gen:skill-docs through bun_cmd; drop tail -3 pipe that masks bun failures

DavidMiserak:
## Summary

Fix two bugs in `setup` that silently broke skill doc generation on certain installations.

**`bun_cmd` routing fix** — Three helpers (`link_codex_skill_dirs`, `link_factory_skill_dirs`, `link_opencode_skill_dirs`) called literal `bun run gen:skill-docs` instead of the `bun_cmd` wrapper. The `bun_cmd` wrapper exists to handle Windows non-ASCII usernames: `prepare_bun_for_windows_compile` copies bun to an ASCII path and points `BUN_CMD` at it when needed. Bypassing it caused silent failures on affected installs. (`setup:714, 919, 951`)

**Pipe-masking exit-code fix** — The gbrain-detected SKILL.md regen ran `bun_cmd run gen:skill-docs:user ... 2>&1 | tail -3`. The pipe makes the subshell's exit status `tail`'s, which is always 0, so the trailing `|| log "warning..."` guard never fired when bun failed. Dropping the pipe restores correct failure detection. (`setup:1297`)

Both fixes were identified by @jbetala7 during review on #1883, carved out as a focused fast-merge PR per their request.

## Test Coverage

```
CODE PATHS
[+] setup (shell script)
  ├── link_codex_skill_dirs()
  │   ├── [★★★ TESTED] bun_cmd invoked via $BUN_CMD — test:4
  │   └── [★★★ TESTED] uses bun_cmd, not bare bun — test:5
  ├── link_factory_skill_dirs()
  │   └── [★★★ TESTED] uses bun_cmd, not bare bun — test:5
  ├── link_opencode_skill_dirs()
  │   └── [★★★ TESTED] uses bun_cmd, not bare bun — test:5
  └── gbrain regen block (setup:1295-1298)
      ├── [★★★ TESTED] without pipe → || warning fires — test:1
      ├── [★★★ TESTED] with pipe (bug shape) → warning silent — test:2
      └── [★★★ TESTED] live block has no pipe — test:3

COVERAGE: 6/6 paths tested (100%)
```

Tests: 0 → 5 (+5 new)

## Pre-Landing Review

Review: CLEAR (DIFF, 2026-06-08, same commit). Quality score: 9.0/10.

0 critical. 4 informational (all skipped — minor test quality suggestions, no blocking issues).

[...]

PR #1900, opened 2026-06-07
v1.57.4.0 fix(setup): Playwright Chromium failure becomes best-effort warn, not exit 1

DavidMiserak:
Carving out the Playwright-warn change from #1883 as a focused PR per
@jbetala7's review request. Coordinated with #1838 — see "Relationship
to #1838" below.

## Problem

The Playwright Chromium install in `./setup` is the riskiest step in the
whole script: it needs network, talks to a Microsoft CDN, on Windows it
bounces through Node.js because of a Bun pipe bug, and on locked-down
corporate machines it may not even be allowed to run. Until now, any
failure in that step was fatal (`exit 1`) and aborted `./setup` mid-run,
**before** skills got registered, hooks got installed, or migrations ran.

A re-run would start over from the top, hit the same Chromium download
issue, fail the same way. There was no clean way to get a working gstack
install on a machine where Chromium was unreachable, even though every
non-browser feature would have worked fine.

## Fix

Replace both `exit 1` paths in the Playwright install/verify block with a
`$_PW_FAIL_REASON` accumulator. On failure, the script:

- Prints a named warning to stderr identifying which sub-step failed:
  `chromium-install`, `windows-no-node`, `windows-node-modules`, or
  `post-install-launch`.
- Lists the affected browser-driven features (`/qa`, `/design-review`,
  `make-pdf`, sidebar, `/pair-agent`).
- On Windows, includes the Bun-on-Windows pipe-bug link and the
  `node -e "require('playwright')"` verify command.
- Tells the user to re-run `./setup` to retry — at which point only the
  Playwright step needs to succeed, because skills, hooks, and
  migrations are already in place from earlier steps.

All other setup behavior is unchanged. The script exits 0 in the
warn-only path and the user gets a usable gstack install minus browser
features.

## Tests

`test/setup-playwright-warn-not-exit.test.ts` — 5 tests, exit-code-path
exercised (not source-grep):

1. **No `exit 1` reachable from the block** — source-anchored regex
   on the live block extracted from `setup`.
2. **`_PW_FAIL_REASON` accumulator + named warning present** — same
   source-anchored extraction.
3. **Functional: install failure → exit 0 + warning.** Runs the live
   block in bash with stubbed `ensure_playwright_browser` (always fails)
   and stubbed `bunx` (fails). Asserts script exits 0, stderr contains
   `warning: Playwright Chromium is unavailable (chromium-install)`,
   and the retry hint fires.
4. **Side-by-side: old bug shape exits 1.** Inlines the OLD `exit 1`
   path with the same stubs and confirms it returned non-zero. Proves
   the exit-code difference is real, not a quoting illusion.
5. **Functional: post-install launch failure also degrades to warn.**
   `bunx` succeeds, `ensure_playwright_browser` keeps failing — asserts
   exit 0 with reason `post-install-launch`.

```
$ bun test test/setup-playwright-warn-not-exit.test.ts
 5 pass / 0 fail / 14 expect() calls

$ bun test test/setup-*.test.ts
 70 pass / 0 fail / 165 expect() calls
```

## Test plan

- [x] `bun test test/setup-playwright-warn-not-exit.test.ts` — 5 pass
- [x] `bun test test/setup-*.test.ts` — 70 pass
- [x] `bash -n setup` — syntax OK
- [x] Manually traced the new path for each `_PW_FAIL_REASON` value

## Relationship to #1838

This PR and #1838 modify the **same Playwright block** in `setup` but
solve **non-overlapping problems**:

| PR | What it changes |
|----|-----------------|
| #1838 | **Which** Chromium version gets installed (pins `bunx playwright install` to the bundled `playwright/package.json` version so the install-time and runtime versions don't drift) + tightens `ensure_playwright_browser()` to assert `executablePath()` exists. |
| **this PR** | **What happens when install fails** (`exit 1` → `_PW_FAIL_REASON` accumulator + named warning, no abort). |

Both can land — they're complementary. Whichever lands first, the other
needs a small rebase. I'm happy to rebase this PR onto #1838 if the
maintainers prefer that order (likely the cleaner sequence, since #1838
is closer to merging).

## Relationship to #1883

This PR is the "(2) Playwright warn-don't-exit" split @jbetala7 asked
for on #1883. The third split — the phase reorder — will be a separate
PR against fresh `main` with an issue describing the failure modes it
fixes. #1883 stays open until all three splits are visible, then I'll
close it with links to all three.

Sibling: #1898 (the two bug fixes, opened first).


jbetala7:
Verified — this is the Playwright carve-out from #1883 and it's correct. On `main`, a Chromium install/launch failure (or the Windows no-Node path) hits `exit 1` and aborts `./setup` *before* skills, hooks, and migrations register, so a re-run just re-fails at the same step and never finishes provisioning. Routing those failures through `_PW_FAIL_REASON` + a named warning while letting setup complete is the right call, and the Windows Node/`@ngrok/ngrok` require-checks are preserved — just redirected to the warn path instead of `exit 1`. Coordinates cleanly with the bundled-version pin in #1838. Looks correct.

time-attack:
@16francej — recommend closing this PR because #2233 supersedes it as the current bounded setup-containment implementation. This branch leaves browser installation and launch checks unbounded and includes unrelated package and version changes. Credit #1900 for the named install, missing-Node, missing-module, and launch-failure cases, and carry them into #2233 without the release changes.

PR #1903, opened 2026-06-07
v1.57.4.0 refactor(setup): group steps into four phases ordered by failure probability

### LINKED ISSUE #1902: setup: Playwright Chromium download failure aborts setup before skills/hooks/migrations land
## Summary

`./setup` is structured so that the riskiest step (Playwright Chromium
download) runs **before** several lower-risk steps (skill linking, hook
installation, version migrations). When the riskiest step fails, the
script exits 1 mid-run and the lower-risk steps never get a chance to
complete.

This means a user on a flaky network, a corporate machine with a
locked-down `bunx`, or hitting the Bun-on-Windows pipe bug
(oven-sh/bun#4253) ends up with a half-installed gstack: bun may be
copied, the browse binary may be built, but the actual user-facing
artifacts (skills, hooks, plan-tune machinery, gbrain regen) are
missing. A re-run of `./setup` does the same thing — dies at the same
step — and the user has no easy path to a working install minus
browser features.

## Failure modes observed

The current `setup` order interleaves trivial filesystem ops, expensive
compute, settings.json mutations, and external network/sudo steps:

1. Build browse binary (network: `bun install`)
2. Install Playwright Chromium (network, sudo, package manager) — **may exit 1**
3. Ensure `~/.gstack/projects/` exists (trivial filesystem)
4. Install for Claude (filesystem)
5-6. Install for Codex / Kiro / Factory / OpenCode (filesystem)
7. .agents/ sidecar symlinks (filesystem)
8. Run version migrations (filesystem)
9. First-time welcome + legacy cleanup (filesystem)
10. Team mode hook install in `settings.json` (mutation)
11. Plan-tune hook install in `settings.json` (mutation)

Concrete user-facing breakage when step 2 fails:

| Step | What's missing on re-run failure | User impact |
|------|----------------------------------|-------------|
| 4-6  | Skills not symlinked into `~/.claude/skills/` | `/qa`, `/ship`, `/review` etc. don't exist |
| 7    | `.agents/` sidecar symlinks not created | Codex skills broken |
| 8    | Version migrations not run | Stale config from prior gstack version |
| 9    | Welcome message + `/tmp/gstack-latest-version` cleanup never happens | Cosmetic, but the legacy `/tmp` file accumulates |
| 10   | Team-mode `SessionStart` hook not registered in `settings.json` | Team auto-sync silently disabled |
| 11   | Plan-tune `PostToolUse` + `PreToolUse` hooks not registered | `/plan-tune` AUQ capture silently disabled |

Even the trivial `mkdir -p "$HOME/.gstack/projects"` happens AFTER the
Playwright step, so a brand-new user without `~/.gstack/` ends up
without it.

## Proposal

Reorder `setup` so steps run in increasing order of failure probability:

- **Phase A** — pure filesystem (`~/.gstack/`, welcome, GBrain detection
  file scan). Cannot fail in practice.
- **Phase B** — local compute: bun build, skill linking, version
  migrations, GBrain SKILL.md regen. Fails only on bun bugs, fully
  reproducible locally.
- **Phase C** — `settings.json` mutation: team-mode and plan-tune hooks.
  Fails only if `settings.json` is malformed, fully reversible.
- **Phase D** — external network / sudo / package managers: coreutils
  on macOS, emoji font on Linux, Playwright Chromium. Highest fail risk.

With this order, a Phase D failure leaves the user with a working gstack
install minus browser-driven features (`/qa`, `/design-review`,
`make-pdf`, sidebar, `/pair-agent`). Re-running `./setup` retries only
the failed phase since A-C are idempotent.

A separate (smaller) follow-up — already in flight as #1900, coordinated
with #1838 — converts the Playwright `exit 1` paths to best-effort warns
so even a re-run failure is non-fatal.

## Related

- #1883 — original bundled PR that combined the phase reorder with two
  unrelated bug fixes and a Playwright behavior change. Per review
  request, that PR is being split into three:
  - #1898 — surgical bug fixes (`bun_cmd` routing + `tail -3` pipe)
  - #1900 — Playwright `exit 1` → best-effort warn (coordinates with #1838)
  - PR for this issue — the phase reorder itself
- #1838 — Pin Playwright install to the bundled version (touches the
  same Playwright block).
- oven-sh/bun#4253 — Bun-on-Windows pipe bug that Phase D needs to
  tolerate, not abort over.

DavidMiserak:
Closes #1902.

This is the "(3) phase reorder last, against fresh main, with an issue
describing the failure modes it fixes" split @jbetala7 asked for on #1883.

## What this PR is

A behavior-preserving reorganization of `setup` into four phases
bracketed by header comments. See #1902 for the failure modes that
motivate it.

| Phase | What runs | Fail risk |
|-------|-----------|-----------|
| **A** | Filesystem: `~/.gstack/`, welcome, GBrain detection (scan) | Always succeeds |
| **B** | Local compute: bun build, skill linking, migrations, GBrain regen | Low |
| **C** | `settings.json` mutation: team + plan-tune hooks | Medium, reversible |
| **D** | External network / sudo / package managers: coreutils, font, Playwright | Highest |

Each phase header is a comment block in the script. Steps inside are
renumbered (`A1, A2, A3, B1, B2a, ...`).

## What this PR is NOT

To keep this reviewable as a structural move, the two **behavior changes**
that were originally bundled with the reorder on #1883 are not here:

- **#1898** — `bun_cmd` routing in three `link_*_skill_dirs` helpers +
  removing `| tail -3` from `gen:skill-docs:user`. Already opened.
- **#1900** — Playwright `exit 1` → `_PW_FAIL_REASON` best-effort warn.
  Coordinates with #1838.

In this PR's diff:

- The three `bun_cmd` call sites are still `bun run` (the bug — fixed in #1898).
- The `gen:skill-docs:user` invocation in B6 still pipes through `tail -3`
  (the bug — fixed in #1898).
- Phase D3 (Playwright) still has the two `exit 1` paths from origin/main
  (the design issue #1902 names — fixed in #1900).

This is intentional. A reviewer can read this diff as "did anything
besides moving lines around and splitting one block?" The answer should
be: just the GBrain detect/regen split (entailed by the new order) and
one safety guard for that split (see next section).

## The one behavior change this PR DOES contain

The phase reorder requires splitting GBrain detection from GBrain regen:

- **A3** runs detection (a filesystem scan via `gstack-gbrain-detect`,
  no dependencies). Writes the result to `~/.gstack/gbrain-detection.json`.
- **B6** reads that file and decides whether to regenerate Claude
  SKILL.md with brain-aware blocks (after the binary is built + skill
  docs are generated).

This split is necessary because A runs before bun is prepared, so the
regen call (which needs bun) can't live in A. Splitting introduces one
new failure mode: what happens if detection fails in A3 but a stale
`gbrain-detection.json` from a prior successful run still exists?
Without a guard, the stale "ok" file would survive a failed detect and
silently trigger brain-aware regen in B6 — applying ~250 tokens of
overhead per planning skill to a user whose gbrain may have been
removed.

The guard: A3 removes both the temp file AND any pre-existing
`gbrain-detection.json` when `gstack-gbrain-detect` exits non-zero.

```sh
if "\$DETECT_BIN" > "\$DETECTION_FILE.tmp" 2>/dev/null; then
  mv "\$DETECTION_FILE.tmp" "\$DETECTION_FILE"
else
  rm -f "\$DETECTION_FILE.tmp" "\$DETECTION_FILE"    # <-- the guard
  log "  warning: gstack-gbrain-detect failed — brain-aware blocks will stay suppressed"
fi
```

`test/setup-gbrain-detection.test.ts` pins this with a static-source
check so a future refactor that drops the guard fails CI.

## Tests

```
$ bun test test/setup-*.test.ts
 66 pass / 0 fail / 152 expect() calls
```

All 7 existing setup test files still green on the reordered structure.

## Test plan

- [x] `bun test test/setup-*.test.ts` — 66 pass
- [x] `bash -n setup` — syntax OK
- [x] Manually traced each phase's step ordering matches the dependency
      graph (A doesn't depend on anything in B/C/D; B doesn't depend on
      C/D; C doesn't depend on D)
- [x] Verified that no bug fix from #1898 or #1900 leaked into this diff
      via grep (`bun_cmd run gen:skill-docs --host {codex,factory,opencode}`
      = 0 in link helpers; `| tail -3` still present in B6;
      `_PW_FAIL_REASON` = 0; Playwright `exit 1` = 2)

## Relationship to #1883

This PR is the third and final split from #1883. Once this is open,
#1883 stays as the parking lot only long enough to add the three split
links and a close note.

Sibling PRs:
- #1898 — bug fixes (independent, can land any time)
- #1900 — Playwright warn (coordinated with #1838)

## Recommended merge order

1. **#1898** first (tiny, independent).
2. **#1838** + **#1900** together (both touch the Playwright block;
   either can land first then the other rebases).
3. **This PR** last, rebased after the above land so the diff stays
   small and reviewable.

If preferred, this PR can land before #1898 / #1900 — they're against
fresh main with no overlap, but rebasing them on top of this would make
their diffs harder to read because they'd be against the reordered
structure. The chosen order (this PR last) keeps each diff minimal
against the structure its reviewer already knows.


time-attack:
@16francej — recommend closing this PR as superseded by #1901. The branch broadly reorders older setup code, conflicts with current work, and includes obsolete release changes; #2233 is the focused implementation. Please preserve its setup-ordering diagnosis and contributor attribution under #1901, but not the outdated structural or release changes.

PR #1905, opened 2026-06-07
fix(openclaw): quiet native skill triggers

ortonom:
## Summary
- make OpenClaw native skill descriptions manual/exact-name only
- prevent ambient Codex/Claude skill routing from auto-triggering nested OpenClaw clones

## Test
- bun test skills/gstack/test/openclaw-native-skills.test.ts

### DIFF
--- a/openclaw/skills/gstack-openclaw-ceo-review/SKILL.md
+++ b/openclaw/skills/gstack-openclaw-ceo-review/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: gstack-openclaw-ceo-review
-description: Use when asked to review a plan, challenge a proposal, run a CEO review, poke holes in an approach, think bigger about scope, or decide whether to expand or reduce the plan.
+description: "OpenClaw-native method clone. Manual — invoke only when the user explicitly names gstack-openclaw-ceo-review."
 ---
 
 # CEO Plan Review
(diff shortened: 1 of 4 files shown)


jbetala7:
The intent — stop ambient skill routing from auto-firing the nested OpenClaw clones on non-OpenClaw hosts — is reasonable, and the coverage is complete: all four native skills are changed and the existing `test/openclaw-native-skills.test.ts` still passes. Two things before this lands:

1. **Blanking the description to pure boilerplate diverges from how gstack already does "manual" skills.** `cso` and `ios-clean` keep a real capability summary and *add* the manual/convenience note on top, rather than replacing it — e.g. `ios-clean`: "Remove the DebugBridge SPM package … This is a convenience wrapper —". Suggest keeping a one-line capability summary and appending the exact-name guard, e.g.:
   `description: "CEO plan review (OpenClaw-native clone). Manual — invoke only when the user explicitly names gstack-openclaw-ceo-review."`
   That kills ambient routing while preserving legibility/discoverability (and any description-based routing inside OpenClaw itself).

2. **No test pins the new behavior.** The existing test only asserts `name`+`description` are non-empty strings, so a later edit could silently restore the ambient "Use when…" descriptions. Worth adding an assertion that each native skill's description carries the manual guard. (Minor: the PR body's test path is `test/openclaw-native-skills.test.ts`, not `skills/gstack/test/…`.)

One question to anchor the change: is there a repro/issue showing the ambient mis-trigger actually fires? These four dirs are only referenced by the test, not by the skill-doc generator, so it'd help to confirm which host's router scans `openclaw/skills/` and auto-routes them.

PR #1910, opened 2026-06-08
v1.57.5.0 feat: cross-session decision memory + gbrain dream-stage call graph

garrytan:
## Summary

**Cross-session decision memory (the headline).** An institutional memory loop so durable decisions and their rationale survive across sessions: capture → curate → resurface.
- Event-sourced store (`decide`/`supersede`/`redact`) at `~/.gstack/projects/<slug>/decisions.jsonl`; "active" is computed, dangling-ref tolerant. Bounded active snapshot keeps session-start reads O(active), with explicit `--compact`.
- Two non-interactive bins: `gstack-decision-log` (`--supersede`/`--redact`/`--compact`) and `gstack-decision-search` (`--scope`/`--recent`/`--query`/`--all`/`--json`/`--semantic`).
- Context Recovery resurfaces scope-relevant active decisions at session start; a `## Cross-session decision memory` section added to CLAUDE.md.
- `/plan-ceo-review`, `/plan-eng-review`, `/spec`, `/ship` auto-capture their structured decisions.
- Optional gbrain `--semantic` recall, lazily loaded; degrades to the reliable file results when gbrain is off/empty.
- Shared `lib/jsonl-store.ts` (injection-reject + atomic append + tolerant read; learnings-log refactored onto it) and `lib/bin-context.ts`.
- Security properties: resurfaced text is datamarked (fences, `---` banners, `<|role|>`/`</system>`, chat turn-prefixes, Unicode line terminators), HIGH and MEDIUM secrets blocked on write, `redact` expunges from every read path (incl. `--all`), compact is lock + size-recheck guarded.

**gbrain dream-stage + reliability.**
- `/sync-gbrain` `--dream` builds the symbol cross-reference call graph behind a lock-free gate with an honest WARN-not-false-success guard; `cycleCompleted()` cycle-state probe.
- Self-heal for a crashed autopilot daemon's stale lock (reads holder pid, signal-0 liveness, conservative when it can't tell) so a dead pid no longer wedges every sync.
- Accurate pin/call-graph guidance in `/sync-gbrain`; ignore gbrain `.sources/` staging dir.

## Test Coverage
All new code paths have test coverage. 117 tests across the decision store + gbrain stages (`jsonl-store`, `gstack-decision`, `gstack-decision-bins`, `gstack-decision-semantic`, `gbrain-dream-stage`, `gbrain-cycle-completed`, `gbrain-guards`). Full `bun test` passes with 0 failures; `gen:skill-docs` clean (0 stale), parity 10/10, ship goldens refreshed.

## Pre-Landing Review
4 findings, all INFORMATIONAL (no criticals), all auto-fixed: datamark resurfaced text, DRY-extract shared bin helpers, batch the compact archive append, close test-coverage gaps. PR quality score 9.0.

Adversarial review ran cross-model (Claude + Codex). Claude found a real injection bypass (chat turn-prefixes like `Human:` defeating both the write denylist and datamark) plus edge cases; Codex found `--all` ignoring `redact` and a few mediums. All actionable findings are fixed and tested. Two findings are intentionally left as-is because they fail SAFE (the autopilot pid-reuse and pgrep-substring cases over-refuse sync; tightening them would risk a fail-dangerous path); documented as such.

## Design Review
No frontend files changed — design review skipped.

## Eval Results
Diff-based selection pulls ~142 E2E + 20 LLM-judge tests (the changes touch the shared preamble/resolvers, which fan out to most skills). Paid evals deferred to CI gate-tier; the free suite (gen-skill-docs, parity, goldens, 117 decision/gbrain tests) is green locally.

## Plan Completion
Plan: institutional decision-memory (capture → curate → resurface). All three phases delivered — reliable file-only core + bins + Context Recovery; CLAUDE.md section + structured skill emits; optional gbrain `--semantic`. `find-contradictions` deferred (the gbrain CLI surface for it can't be verified deterministically yet and the curated-memory source isn't indexed).

## TODOS

## Test plan
- [x] `bun test` full suite passes (0 failures)
- [x] `gen:skill-docs` clean (0 stale), parity 10/10, ship goldens refreshed
- [x] Decision store + gbrain stages: 117 unit tests pass
- [ ] CI runs gate-tier evals on this PR


xg-gh-25:
This is a master class in **institutional memory architecture** — the "capture → curate → resurface" loop you've built solves the hardest problem in cross-session agent systems: *how do decisions persist without becoming noise?*

**Why this matters beyond gstack:**
- **Event-sourced decisions**: The `decide`/`supersede`/`redact` triad is exactly the right semantic primitive. Decisions aren't mutable state — they're a timeline with revisions and explicit invalidations. The "active" computation from the log is how all durable agent memory should work.
- **Bounded active snapshot**: O(active) session-start reads, not O(all-time), is the difference between "usable at scale" and "unusable past week 2." The explicit `--compact` is smart — garbage collection as an intentional human action, not automatic (which can destroy forensic value).
- **Context Recovery in CLAUDE.md**: Injecting the `## Cross-session decision memory` section at session start is how you bridge ephemeral agent context and durable institutional state. This is the pattern every agent framework needs but most don't have.

**Two design notes that stand out:**
1. **Datamarking resurfaced text**: The security properties section (fences, `---` banners, `<|role|>`, turn-prefixes, Unicode line terminators) shows you've internalized the prompt injection risk. Resurfacing past decisions is a *trust boundary* — if an adversary can inject a fake decision into the log, they control future behavior. The multi-layer datamarking + HIGH/MEDIUM secret blocking on write is the right paranoia level.

2. **Gbrain as optional enhancement, not dependency**: The `--semantic` flag degrades gracefully to file-only results when gbrain is off/empty. This is how you make advanced features (semantic search over decisions) *additive* without creating fragility. The lazy-load + fail-safe architecture means the core (JSONL + bins) is always reliable.

**The adversarial review rigor is rare:**
- Cross-model (Claude + Codex) adversarial review finding real injection bypasses (chat turn-prefixes like `Human:` defeating both write denylist and datamark) is exactly how agent systems should be stress-tested.
- The two findings intentionally left as-is (autopilot pid-reuse and pgrep-substring over-refuse) because fixing them would create fail-dangerous paths — this is *safety engineering*. The "fail safe, not fail silent" principle documented.

**Comparison to the ecosystem:** Most agent systems treat memory as append-only logs with no curation or explicit invalidation. The result is context pollution — old decisions resurface and confuse new sessions. Your `supersede`/`redact` semantics + active-set computation solves this cleanly. The gstack pattern here is a blueprint for durable agent memory.

---
*Similar event-sourced memory patterns in [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/12)*

PR #1913, opened 2026-06-08
fix(make-pdf): don't force a blank first page when content precedes the first H1 (#1904)

### LINKED ISSUE #1904: make-pdf: spurious blank first page when content (frontmatter or leading <style>) precedes the first H1
## Bug

`make-pdf` emits a spurious blank/near-blank **first page** whenever the markdown has *any* content before the first `<h1>`. Two common triggers:

1. **YAML frontmatter** — a leading `---\n...\n---` block is not parsed/stripped, so `marked` renders it as a literal paragraph of body text on its own page (`title: ... type: ... tags: [...]`).
2. **A leading top-of-file `<style>` block** — even though it renders nothing visible, it still produces an empty first page.

## Root cause

In `make-pdf/src/render.ts`, `wrapChaptersByH1()` treats everything before the first `<h1>` as a **preamble chapter**:

```ts
const preamble = html.slice(0, matches[0]);
if (preamble.trim().length > 0) {
  chunks.push(`<section class="chapter">${preamble}</section>`);
}
```

Combined with the chapter CSS:

```css
.chapter { break-before: page; page-break-before: always; }
.chapter:first-of-type { break-before: auto; page-break-before: auto; }
```

When a preamble chapter exists it becomes `:first-of-type`, so the first *real* `<h1>` chapter no longer gets the `break-before: auto` exception — it inherits `break-before: page` and starts on page 2. The preamble (frontmatter text, or an invisible `<style>` block) occupies page 1. `preamble.trim().length > 0` is true even when the preamble contains only non-rendering markup (`<style>…</style>`), so an invisible preamble still forces the break.

## Repro

```bash
printf '<style>h1{color:#000}</style>\n\n# Hello\n\nBody.\n' > t.md
pdf generate t.md t.pdf
pdfinfo t.pdf   # => Pages: 2  (page 1 blank)

printf '# Hello\n\n<style>h1{color:#000}</style>\n\nBody.\n' > t2.md
pdf generate t2.md t2.pdf
pdfinfo t2.pdf  # => Pages: 1  (correct)
```

## Suggested fix (either/both)

1. **Strip YAML frontmatter** before `marked.parse` — a leading `^---\n[\s\S]*?\n---\n?` block.
2. **Don't create a preamble chapter for non-rendering content** — strip `<style>/<script>/<link>/<meta>` from the preamble slice before the `preamble.trim().length > 0` check (or hoist a leading `<style>` into `<head>`), so an invisible preamble doesn't force a page break / empty first page.

## Workaround

Strip frontmatter caller-side and inject the width-fit `<style>` block *after* the first heading rather than prepending it. Confirmed to produce the correct page count.

jbetala7:
Fixes #1904

## Problem

`make-pdf` emits a spurious blank/near-blank **first page** whenever any content precedes the first `<h1>`. Two common triggers:

1. **YAML frontmatter** — `marked` has no frontmatter awareness, so a leading `---\n...\n---` block renders as a literal paragraph of body text (`title: ... type: ... tags: [...]`) on its own page.
2. **A leading top-of-file `<style>` block** — invisible, but still produces an empty first page.

## Root cause

`wrapChaptersByH1()` (`make-pdf/src/render.ts`) wraps everything before the first `<h1>` in a preamble `.chapter`:

```ts
const preamble = html.slice(0, matches[0]);
if (preamble.trim().length > 0) {
  chunks.push(`<section class="chapter">${preamble}</section>`);
}
```

With the chapter CSS, that preamble section becomes `:first-of-type` and claims the `break-before: auto` exception, so the **first real** `<h1>` chapter inherits `break-before: page` and starts on page 2. `preamble.trim().length > 0` was true even when the preamble was only non-rendering markup (`<style>`).

## Fix

- **Strip a leading YAML frontmatter block** before `marked.parse`. The regex is anchored to the document start, so a `---` thematic break elsewhere is left intact.
- **Don't give a non-rendering preamble its own chapter.** When the preamble strips down to nothing visible (style/script/comment-only), fold the markup into the first real chapter instead — its styling still applies, but it no longer forces a page break or an empty page.

Both changes are confined to `render.ts`; no template/SKILL.md regeneration needed.

## Testing

`bun test make-pdf/test/render.test.ts` — 66 pass (4 new). Full free suite `bun test` (make-pdf): 97 pass, 0 fail.

New render-unit regressions (all fail on `main`, pass with the fix):
- a leading `<style>` preamble yields **one** chapter, not two, and the `<style>` is preserved inside the first chapter (not dropped).
- leading YAML frontmatter is stripped — keys (`type: memo`, `tags: [a, b]`) don't leak into the body; one chapter; title still derives from the H1.
- a real **text** preamble still gets its own chapter (unchanged behavior).
- a non-frontmatter `---` thematic break still renders as `<hr>`.

Before this change, both the `<style>` and frontmatter cases rendered 2 chapters (blank page 1); the control case (`<style>` *after* the H1) was already correct at 1.

`eval:select` shows 0/175 E2E and 0/28 LLM-judge tests affected by this diff. The make-pdf `dist/pdf` e2e gate (`pdfinfo`/`pdftotext`) is skip-gated on the compiled binary + poppler, which aren't present in this environment; the render unit tests are the in-CI gate tier for this module.

### DIFF
--- a/make-pdf/src/render.ts
+++ b/make-pdf/src/render.ts
@@ -57,8 +57,10 @@ export interface RenderResult {
  * Pure renderer. No side effects.
  */
 export function render(opts: RenderOptions): RenderResult {
-  // 1. Markdown → HTML
-  const rawHtml = marked.parse(opts.markdown, { async: false }) as string;
+  // 1. Markdown → HTML (strip a leading YAML frontmatter block first; marked
+  //    has no frontmatter awareness and would otherwise render it as a literal
+  //    paragraph of body text on its own first page).
+  const rawHtml = marked.parse(stripFrontmatter(opts.markdown), { async: false }) as string;
 
   // 2. Sanitize
   const cleanHtml = sanitizeUntrustedHtml(rawHtml);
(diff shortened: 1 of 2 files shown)


time-attack:
@jbetala7 Rewrite the frontmatter detection so legitimate leading Markdown is preserved; the current implementation can remove rendered thematic breaks and visible text.

PR #1914, opened 2026-06-08
fix(make-pdf): stop content before the first H1 from forcing a blank page 1

### LINKED ISSUE #1904: make-pdf: spurious blank first page when content (frontmatter or leading <style>) precedes the first H1

jbetala7:
Fixes #1904

## Problem

`make-pdf` emitted a spurious blank/near-blank **first page** whenever any content preceded the first `<h1>`. Two triggers:

1. **YAML frontmatter** — `marked` has no frontmatter awareness, so a leading `---\n...\n---` block rendered as a literal paragraph of body text on its own page.
2. **A leading top-of-file `<style>` block** — invisible, but still produced an empty first page.

## Root cause

`wrapChaptersByH1()` wraps everything before the first `<h1>` in a preamble `.chapter`. That section becomes `:first-of-type`, claiming the `break-before: auto` exception, so the first *real* chapter inherits `break-before: page` and starts on page 2. `preamble.trim().length > 0` was true even when the preamble was only non-rendering markup (`<style>`/comments).

## Fix

- Strip a leading YAML frontmatter block before `marked.parse` (anchored to the document start, so a `---` thematic break elsewhere is untouched).
- Don't give a non-rendering preamble (style/script/comment-only) its own chapter. Fold the markup into the first real chapter so its styling still applies but it no longer forces a page break.

## Testing

`bun test make-pdf/test/render.test.ts` → **66 pass / 0 fail**.

Adds render-unit regression tests for both triggers, plus two invariants:
- a real text preamble still gets its own chapter (unchanged behavior)
- a non-frontmatter `---` thematic break is left intact as `<hr>`

Both new trigger tests **fail on `main`** (2 chapters / leaked frontmatter) and pass with the fix.

### DIFF
--- a/make-pdf/src/render.ts
+++ b/make-pdf/src/render.ts
@@ -57,8 +57,10 @@ export interface RenderResult {
  * Pure renderer. No side effects.
  */
 export function render(opts: RenderOptions): RenderResult {
-  // 1. Markdown → HTML
-  const rawHtml = marked.parse(opts.markdown, { async: false }) as string;
+  // 1. Markdown → HTML (strip a leading YAML frontmatter block first; marked
+  //    has no frontmatter awareness and would otherwise render it as a literal
+  //    paragraph of body text on its own first page).
+  const rawHtml = marked.parse(stripFrontmatter(opts.markdown), { async: false }) as string;
 
   // 2. Sanitize
   const cleanHtml = sanitizeUntrustedHtml(rawHtml);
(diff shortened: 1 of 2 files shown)


jbetala7:
Closing as an exact duplicate of #1913 — same fix at the same commit (`ced81536`), both filed against #1904 within two minutes. Consolidating on #1913 as the canonical PR for this issue. No code difference between the two.

PR #1915, opened 2026-06-08
feat(browse): add GSTACK_HEADED_NOGPU opt-in for headed mode on virtual displays (Xvnc/Xvfb)

auroracapital:
## Problem

On headless Linux servers using virtual displays (Xvnc, Xvfb), Playwright's bundled Chromium crashes with SIGTRAP during GPU process initialization when headed mode is requested. This makes `/browse --headed` completely non-functional on remote Linux dev boxes.

Root cause: virtual display drivers (Xvnc, Xvfb) have no real GPU. Chromium's GPU process init path raises SIGTRAP when it can't initialize hardware acceleration, crashing the browser before the page loads.

## Solution

Add an opt-in env var `GSTACK_HEADED_NOGPU=1` that appends GPU-disabling Chromium flags at launch:

- `--disable-gpu`
- `--disable-gpu-compositing`
- `--disable-software-rasterizer`
- `--disable-dev-shm-usage`

This bypasses the GPU process entirely, keeping headed mode alive on VNC-backed virtual displays.

## Usage

```bash
GSTACK_HEADED_NOGPU=1 browse --headed https://example.com
```

Or export it for a session:

```bash
export GSTACK_HEADED_NOGPU=1
browse --headed ...
```

## Safety

- **Opt-in only** — users on real displays (macOS, physical Linux desktop) are unaffected; they never set this flag
- **No behavioral change** when the env var is unset (default)
- Tested on Amazon Linux 2023 aarch64 with Xvnc :2 using system Brave 148 as the Chromium path

## Context

This came up while using gstack `/browse --headed` on an EC2 dev box with a VNC display for interactive browser automation. The fix unblocks headed mode for the growing number of users running Claude Code on remote Linux servers.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -442,6 +442,17 @@ export class BrowserManager {
       // Sites like Google and NYTimes check this to block automation browsers.
       '--disable-blink-features=AutomationControlled',
     ];
+    // Headless virtual displays (Xvnc/Xvfb on EC2) have no real GPU; Playwright's
+    // bundled Chromium crashes (SIGTRAP) during GPU process init. Disabling the
+    // GPU process keeps headed mode alive on VNC. Opt-in via GSTACK_HEADED_NOGPU.
+    if (process.env.GSTACK_HEADED_NOGPU === '1') {
+      launchArgs.push(
+        '--disable-gpu',
+        '--disable-gpu-compositing',
+        '--disable-software-rasterizer',
+        '--disable-dev-shm-usage',
+      );
+    }
     if (extensionPath) {
       // Skip --load-extension when running against a custom Chromium build
       // that already bakes the extension in as a component extension


jbetala7:
Directionally useful for VNC-backed boxes. Two notes:

**1. Coverage gap on the handoff path.** The NOGPU flags are only added to the `launchArgs` array in `launchHeaded()` (~L439). The headless→headed **handoff relaunch** builds its own `launchArgs = ['--hide-crash-restore-bubble']` (~L1559) and never gets them, so a `/browse` session that starts headless and is later promoted to headed on a Xvnc/Xvfb display would still SIGTRAP — the exact crash this PR targets. Adding the same `GSTACK_HEADED_NOGPU` block to that second `launchArgs` would close the path.

**2. `--disable-dev-shm-usage` isn't a GPU flag** — it controls `/dev/shm` usage, not GPU init. Fine to keep (it's a common headless-stability flag), but it widens the opt-in beyond what the env var name advertises; a one-line note in the comment/docs would help.

Coordination: #1933 adds a general `GSTACK_CHROMIUM_ARGS` passthrough that could express these GPU flags directly, and both it and #1968 touch this same launch path — flagging for merge-order / possible consolidation.

time-attack:
Thanks for this. The opt-in flags apply to only one headed launch path, and disabling software rendering could produce blank output. Please centralize justified flags across every headed path and add Xvfb/Xvnc rendering, screenshot, PDF, and handoff coverage.

PR #1919, opened 2026-06-08
v1.57.7.0 fix(ios-qa): sliding tunnel cache so device sessions survive past 30s

sternryan:
## What

Make the iOS QA daemon's device-tunnel cache a **sliding idle-timeout** instead of a flat 30s window, with a configurable default.

## Why

`getTunnel()` in `ios-qa/daemon/src/index.ts` cached the CoreDevice tunnel for a flat **30s measured from bootstrap**, and never refreshed the timer on use. A vision-driven agent loop (`/ios-qa`) reads a screenshot, reasons, then taps — that round trip routinely exceeds 30s. When the cache expires the daemon re-bootstraps, but **re-bootstrap can't succeed against a still-running app**: `StateServer` deletes its on-disk boot token at `/auth/rotate` (`templates/StateServer.swift.template` → `removeItem(atPath: bootTokenPath)`), so `bootstrapTunnel` fails with `boot_token_unavailable` until the app is cold-relaunched.

Observed live: the bridge died roughly every 30 seconds mid-session and could only recover by killing + relaunching the app on the device.

## Fix

Refresh `cachedTunnelAt` on every cache hit (sliding idle-timeout), so an active session holds the same live tunnel and rotated bearer for as long as it keeps working. Default idle window raised to **5 min** (covers reason-then-act gaps) and made configurable via **`GSTACK_IOS_TUNNEL_CACHE_MS`**. No security change — the boot token still dies at rotate; this only avoids needless re-bootstrap of a healthy tunnel.

The default bump (30s → 5min) is the debatable knob; with the sliding behavior even a short default survives active use, but the reason-then-act *gap* between requests is what needs the longer window. Easy to dial back if you'd prefer.

## Test

Adds a regression test in `daemon-integration.test.ts` asserting a burst of requests bootstraps the tunnel exactly once. `bun test test/daemon-integration.test.ts` → 5 pass, 0 fail.

## Notes
- Heads-up (pre-existing, not touched here): `daemon-integration.test.ts` imports `afterEach` at the bottom of the file, and the `tailnet listener` describe block errors with `afterEach is not a function` under bun — reproduces on `main` verbatim, so that block's tests don't run. Worth a separate fix (move the import to the top).
- Fork PR: eval/E2E CI jobs won't get base-repo secrets, so they'll fail on empty-env auth. The daemon test + lint don't need secrets. Move base-side if you want full CI.

### DIFF
--- a/ios-qa/daemon/src/index.ts
+++ b/ios-qa/daemon/src/index.ts
@@ -64,9 +64,21 @@ export async function startDaemon(opts: DaemonOptions): Promise<RunningDaemon | 
   let tunnel: DeviceTunnel | null = null;
   let cachedTunnelAt = 0;
 
+  // Tunnel cache TTL — sliding idle-timeout. Every reuse refreshes the timer,
+  // so an active agent session never re-bootstraps. That matters because
+  // re-bootstrap is fragile: bootstrapTunnel reads the on-disk boot token,
+  // which StateServer DELETES at /auth/rotate (StateServer.swift), so a
+  // re-bootstrap against an already-rotated app fails with
+  // boot_token_unavailable until the app is cold-relaunched. Keeping a healthy
+  // tunnel alive avoids that trap. Default 5 min covers an agent's
+  // reason-then-act gaps; override via GSTACK_IOS_TUNNEL_CACHE_MS for long idles.
+  const tunnelCacheMs = Math.max(0, parseInt(process.env.GSTACK_IOS_TUNNEL_CACHE_MS ?? '300000', 10) || 300000);
+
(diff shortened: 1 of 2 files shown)


sternryan:
Rebased onto current main to clear the conflict — the PR is now mergeable. The only conflict was a stale `chore(release): v1.57.7.0` commit (main is already at 1.58.5.0), which I dropped since versioning is the maintainer's call at merge time; the actual code fix applied with zero conflicts. Branch is now a single commit touching only `ios-qa/daemon/src/index.ts` + its integration test. Ran the affected test locally: 5 pass / 0 fail.

PR #1920, opened 2026-06-08
feat(design-review): capture the design system before auditing + domain/register copy checks

tomdinh24:
## What

Two improvements to `/design-review`, both targeting how it handles a project that **already has a design language**:

**1. Capture the design system *first*, and never invent a new aesthetic over it.**
Today the skill has a single "Check for DESIGN.md" step, and if none exists it falls back to "universal design principles" — i.e. it's free to invent. But Phase 2 already extracts real tokens from the live DOM, and most projects encode their system in `tailwind.config` / CSS vars / a theme file rather than a `DESIGN.md`. This replaces that step with a priority-ordered capture (live DOM → code tokens → DESIGN.md → sibling apps) treated as the source of truth, plus two hard rules:
   - **Never invent a new aesthetic when one exists** — reuse the captured tokens verbatim in any mock/redesign/variant; introducing a new theme is itself a HIGH-impact finding (it makes the artifact unusable as a reference and silently re-litigates a settled decision). This is reinforced inline at the `$D generate --brief` call, so it binds at the moment of generation, not only in the preflight.
   - **Design from the shipped app, not the spec**, when a real implementation exists — specs go stale.

**2. Two new Content & Microcopy checklist items.**
   - **Domain vocabulary** — flag generic app-speak where the product's own domain has a real term ("Available balance" not "Account value"; "In transit" not "Status: 2"). Verify against the real schema/spec, never invent. HIGH-impact, because the user reads "what does this mean?" and stalls.
   - **Human register** — write like a knowledgeable colleague, not a system log; raw schema/field identifiers leaking into UI copy (snake_case keys, enum values, status codes) is the tell.

## Why

The existing checklist is already strong on AI-slop and visual hygiene, but it has a gap on the *content* side (domain fit, register) and, more importantly, nothing stops the mockup/redesign step from generating a brand-new palette for a product that already has one — the single most damaging design-review failure. These changes close that loop.

## Notes

- The two checklist items are **semantic/intent checks**, so they live in the design-review methodology, not the code-level `review/design-checklist.md` subset (which is for grep-able patterns).
- Source-only edits: `design-review/SKILL.md.tmpl` + `scripts/resolvers/design.ts`. Regenerated all hosts with `bun run gen:skill-docs --host all`; `design-review/SKILL.md` is the generated output.
- `bun test` introduces no new failures vs. `main` (the only failures locally — `gbrain-detect-install`, `user-slug-fallback` — reproduce identically on a clean `main` checkout and are environment-specific to a sandboxed clone). Doc-drift dry-run is clean for the changed files.

### DIFF
--- a/design-review/SKILL.md
+++ b/design-review/SKILL.md
@@ -785,9 +785,36 @@ $B status 2>/dev/null | grep -q "Mode: cdp" && echo "CDP_MODE=true" || echo "CDP
 ```
 If `CDP_MODE=true`: skip cookie import steps — the real browser already has cookies and auth sessions. Skip headless detection workarounds.
 
-**Check for DESIGN.md:**
+**Capture the design system FIRST (before any audit, mock, or fix):**
 
-Look for `DESIGN.md`, `design-system.md`, or similar in the repo root. If found, read it — all design decisions must be calibrated against it. Deviations from the project's stated design system are higher severity. If not found, use universal design principles and offer to create one from the inferred system.
+A design system is rarely just a `DESIGN.md`. Capture what the project actually
+ships, in priority order, and treat what you find as the source of truth:
+
+1. A running instance / the live DOM — extract real tokens (Phase 2 does this:
+   computed colors, fonts, spacing, radii).
+2. Code-level tokens — `globals.css` `:root` custom properties, `tailwind.config`,
+   a theme/`tokens` file, CSS variables, RN `StyleSheet` / Gluestack theme.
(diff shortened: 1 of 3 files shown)


time-attack:
Thanks for this. The blocker is that live-DOM-first precedence can treat accidental drift as canonical when design sources conflict. Please keep the copy checks, define normative-versus-observed precedence and currentness rules, and add contradictory-source fixtures.

Issue #1899, opened 2026-06-07
The latest security review agent in ship is triggering usage policy API denials from Anthropic


bmajewski:
The Usage Policy block happens because the repo’s security regression tests use deliberate attack payloads (rm -rf ~, command substitution, injection strings). When /ship runs adversarial review, red team, or the security specialist subagent, Opus tries to “think like an attacker” against that material and gets blocked by real-time cyber safeguards.

There needs to be a way to segregate those kinds of tests so that they are manually run and approved

garrytan:
Fixed in #1911 (shipping in v1.57.6.0). The Claude adversarial subagent now carries an authorized-defensive-testing preamble and reads test/fixture files in summary mode, so it no longer trips Anthropic's usage policy when reviewing the repo's own attack-payload fixtures. Fixed at the prompt source (`scripts/resolvers/review.ts`). Thanks for the report. Will auto-close when the PR merges.

PR #1921, opened 2026-06-08
v1.57.7.0 fix(review): extend #1899 defensive framing to red-team + security passes

bmajewski:
## Summary

Follow-up to #1899. That fix gave the **always-on adversarial pass** an "authorized defensive-security review" framing + summary-mode fixture handling, so reviewing a repo's own attack-payload regression corpus stopped getting denied by Anthropic's real-time cyber safeguards.

But the framing lived inline in one resolver, and the **red-team** and **security-specialist** dispatches — which *also* reason adversarially over the full diff — were left un-framed. On a repo whose test corpus is shell-injection / path-traversal / symlink-escape fixtures, those two passes still get **denied**. That gap is exactly what pushes such repos toward opting out of AI review entirely.

## What this does

- Extracts the framing into a single source of truth: `scripts/resolvers/defensive-framing.ts`
  - `DEFENSIVE_REVIEW_FRAMING` — prepended to **every** specialist dispatch (universally safe; only reframes intent + suppresses novel attack-content generation) and the red-team dispatch.
  - `FIXTURE_SUMMARY_MODE` — applied to the **attacker-framed** passes only (adversarial ✓, red-team, security specialist), so fixture bytes never enter adversarial reasoning.
- Refactors `generateAdversarialStep` to consume the shared constant — **byte-identical output**, the adversarial golden is unchanged.
- The **testing specialist keeps full fixture reads** — summary mode is scoped to the adversarial passes, so test-quality review is not blinded.

## Why a shared constant

The #1899 framing was inline in one place, which is *why* red-team and security drifted out of coverage. A single source of truth means the three Claude-dispatched passes can't diverge again.

## Tests

- New `test/defensive-framing.test.ts` locks the invariant: every Claude-dispatched adversarial pass (adversarial, red-team, security specialist) carries the framing; codex host strips both; a future un-framed dispatch fails CI.
- `test/fixtures/golden/factory-ship-SKILL.md` regenerated (red-team + specialist prompts now carry the framing). Claude + Codex ship goldens unchanged (carved / Review-Army-stripped respectively).
- Full local suite green except pre-existing, environment-dependent failures unrelated to this change (gbrain-detect, resolve-user-slug, v1.53 parity baseline).

VERSION/CHANGELOG bumped to 1.57.7.0 — adjust to fit your release flow.

Reported/authored by @bmajewski.


time-attack:
Thanks for this. The change is outdated, and prompt-string coverage does not demonstrate the intended fixture visibility boundaries between specialist passes. Please rebase and regenerate, then add controlled end-to-end traces showing adversarial fixture visibility while preserving full reads only where intended.

PR #1924, opened 2026-06-09
Fix AskUserQuestion hook pass-through

jacksondc:
## Summary

- stop returning `permissionDecision: "defer"` for normal AskUserQuestion pass-through paths
- keep auto-decision `deny` behavior unchanged
- preserve memory `additionalContext` by emitting context without a permission decision

## Why

Claude Code treats `permissionDecision: "defer"` as an explicit SDK pause/resume flow, not as "no opinion". In SDK hosts such as Conductor, Gstack's no-op defer response caused `mcp__conductor__AskUserQuestion` to be emitted but not dispatched to the MCP handler.

The correct pass-through behavior is exit 0 with no stdout. When Gstack has context to inject, it can emit `additionalContext` without a `permissionDecision`.

## Test plan

- `bun test test/question-preference-hook.test.ts test/memory-cache-injection.test.ts test/auq-error-fallback-hook.test.ts test/gstack-settings-hook-schema-aware.test.ts`
- SDK repro from the original Conductor workspace against this PR checkout: exact Gstack hooks invoked the AUQ MCP handler in `none`, `pre`, `post`, and `both` hook modes; orphaned AUQs: 0.

### DIFF
--- a/docs/spikes/claude-code-hook-mutation.md
+++ b/docs/spikes/claude-code-hook-mutation.md
@@ -51,7 +51,8 @@ Optional in subagent context: `agent_id`, `agent_type`.
 - `"deny"` — block (feedback to Claude, NOT a synthetic answer per Codex
   correction in D-prefixed decisions)
 - `"ask"` — escalate to user
-- `"defer"` — let permission flow continue
+- No output — let permission flow continue
+- `"defer"` — pause a non-interactive SDK/tool caller so it can resume later
 
 **`updatedInput` semantics:** shallow merge of fields present in the returned
 object onto the original `tool_input`. Only valid with
@@ -88,7 +89,7 @@ required for our hook to fire there.
   accepting.
 
 **`permissionDecision` precedence (when multiple hooks decide):**
-`deny > ask > allow > defer` — most restrictive wins.
+Claude Code treats `defer` as an explicit pause/resume decision, not as the normal pass-through path. Hooks that do not need to decide should exit 0 with no output.
 
 ## Implementation hookSpecificOutput examples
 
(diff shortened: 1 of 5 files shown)


garrytan (approved):
Reviewed against main, and this one has extra significance: the `defer`-as-pause bug is almost certainly the root cause of the flaky `mcp__conductor__AskUserQuestion` calls we've been seeing in Conductor sessions ("[Tool result missing due to internal error]" — hit it live today). The diagnosis matches: the hook returned `permissionDecision: "defer"` on every pass-through path, the SDK treated it as a pause/resume decision, and the AUQ tool_use was emitted but never dispatched to the MCP handler.

**Verified:**
- `passThrough()` correctly emits nothing on the no-context path (exit 0, no stdout) and emits `additionalContext` WITHOUT a `permissionDecision` when memory nuggets exist — both are the documented correct shapes.
- The deny/auto-decide path is untouched, so plan-tune enforcement semantics don't change.
- The spike doc updates keep the protocol contract honest (the old `deny > ask > allow > defer` precedence line was describing behavior that doesn't exist).
- Tests updated across all four affected suites, and the SDK repro matrix (none/pre/post/both, orphaned AUQs: 0) is exactly the right verification.

Thanks for chasing this to the protocol level instead of patching symptoms.

jbajou:
This would be very helpful to be merged to main for folks using conductor and having prose instead of AUQ

Frully:
Hit this exact bug today (gstack 1.60.1.0) — every `AskUserQuestion` call in an interactive Claude Code session was failing (`Tool result missing due to internal error` / `Tool execution was interrupted`) until we traced it to `permissionDecision: "defer"` in `question-preference-hook.ts`'s pass-through path. Independently arrived at the same root cause as #2006 before finding this PR already had the fix approved.

This PR looks stuck on a merge conflict against `main` (currently showing `CONFLICTING`/needs rebase) rather than anything substantive — any chance it can get merged soon?

time-attack:
@16francej — recommend closing this PR as superseded by #2165. #1924 has the correct neutral pass-through behavior, but it conflicts with current Conductor logic; #2165 is the active implementation for the same hook defect. Please carry over the protocol explanation, passThrough naming, exact-empty-stdout assertions, and contributor attribution without overwriting current behavior.

PR #1928, opened 2026-06-09
fix(gstack-upgrade): rerun setup for detected hosts

### LINKED ISSUE #1925: gstack-upgrade only re-registers Claude Code; Codex / OpenCode installs silently drift
## Summary

`/gstack-upgrade` (and the `./setup` it runs) re-registers skills only for **Claude Code**. On a machine where gstack was also installed for **Codex** (and potentially OpenCode), those agents are never refreshed by the upgrade, so they silently drift to older versions. Over time this produces multiple divergent gstack installs across agents.

## Environment

- macOS (Apple Silicon)
- gstack primary (Claude Code): `~/.claude/skills/gstack` — **v1.57.0.0**
- gstack clone backing Codex: `~/gstack` (Codex symlinks `~/.codex/skills/gstack-* → ~/gstack/.agents/skills/`) — **v1.26.3.0** (31 versions behind)
- OpenCode: no gstack skills registered

I run gbrain + gstack across Claude Code, Codex and OpenCode. gbrain handles multi-agent cleanly (single binary + per-agent MCP; one `bun install` updates all). gstack exposes multi-agent `setup` flags, but the *upgrade* path isn't host-aware, so the agents fall out of sync after every upgrade with no warning.

## Root cause (from reading `setup` + the gstack-upgrade skill)

- `setup` defaults to `HOST="claude"`; multi-agent registration only happens with an explicit `--host codex|opencode|auto`.
- `--host auto` already auto-detects installed agents (`command -v claude|codex|kiro-cli|droid|opencode`) and registers all of them.
- The `/gstack-upgrade` skill runs **bare `./setup`** (no `--host`), so every upgrade re-registers only Claude Code. Codex / OpenCode keep pointing at whatever clone they were first set up from and drift.
- Side note: `setup` already has `migrate_direct_codex_install`, so the project clearly anticipates multi-clone Codex installs — but the upgrade path doesn't keep them current.

## Expected

The upgrade re-registers every host gstack was installed for, not just Claude — and ideally warns when a registered agent's version differs from the primary.

## Possible fixes

1. **Quick:** have `/gstack-upgrade` run `./setup --host auto` instead of bare `./setup`. Safe — `auto` falls back to claude when no other agent CLI is present. Trade-off: auto-detect can register gstack into an agent the user didn't deliberately choose.
2. **Proper:** persist the set of installed hosts at setup time (e.g. in `~/.gstack/config.yaml`) and have the upgrade re-run setup for exactly those hosts; warn on per-host version drift.

Happy to send a PR for whichever direction you prefer.

maxpetrusenkoagent:
## Summary
- update gstack-upgrade install refresh paths to rerun ./setup --host auto
- cover template and generated skill docs with a regression test

Fixes #1925.

## Verification
- bun test test/gstack-upgrade-host-auto.test.ts

## Notes
- Branch is based on the fork main because the agent token lacks GitHub workflow scope for pushing upstream workflow history changes.

### DIFF
--- a/gstack-upgrade/SKILL.md
+++ b/gstack-upgrade/SKILL.md
@@ -131,7 +131,7 @@ cd "$INSTALL_DIR"
 STASH_OUTPUT=$(git stash 2>&1)
 git fetch origin
 git reset --hard origin/main
-./setup
+./setup --host auto
 ```
 If `$STASH_OUTPUT` contains "Saved working directory", warn the user: "Note: local changes were stashed. Run `git stash pop` in the skill directory to restore them."
 
@@ -142,7 +142,7 @@ TMP_DIR=$(mktemp -d)
 git clone --depth 1 https://github.com/garrytan/gstack.git "$TMP_DIR/gstack"
 mv "$INSTALL_DIR" "$INSTALL_DIR.bak"
 mv "$TMP_DIR/gstack" "$INSTALL_DIR"
-cd "$INSTALL_DIR" && ./setup
+cd "$INSTALL_DIR" && ./setup --host auto
 rm -rf "$INSTALL_DIR.bak" "$TMP_DIR"
 ```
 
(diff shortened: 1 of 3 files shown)


maxpetrusenkoagent:
Autoreview: CLEAN

Verified:
- bun test test/gstack-upgrade-host-auto.test.ts
- bun run gen:skill-docs --dry-run

time-attack:
@maxpetrusenkoagent Rewrite this to preserve installed targets rather than relying on setup --host auto PATH discovery, and cover the independent-clone case.

Issue #1917, opened 2026-06-08
buildGbrainEnv leaves PGLite users exposed to .env poisoning → broken-db false-positive


techguyisme2:
# `buildGbrainEnv` leaves PGLite users exposed to `.env` poisoning → `broken-db` false-positive

## Summary

PR #1583 (CHANGELOG line 1359, 1410) fixed `lib/gbrain-local-status.ts:freshClassify` so the probe's env is routed through `buildGbrainEnv`. The fix is **complete for Postgres/Supabase brains** but **incomplete for PGLite brains**, because `buildGbrainEnv` works by *seeding* `DATABASE_URL` from `~/.gbrain/config.json` — and a PGLite config has no `database_url` field.

For PGLite users, line 111 of `lib/gbrain-exec.ts` fires `if (!cfg.database_url) return out;` and returns the caller env **unmodified**. Bun's auto-load of the cwd's `.env` then flows straight through to the `gbrain sources list --json` probe, gbrain treats the leaked `DATABASE_URL` as a Postgres override, connection fails, classifier reads `"Cannot connect to database"` from stderr, and returns `broken-db`. The 60s cache then propagates that poisoned negative to clean directories.

Net effect: every PGLite user with a `DATABASE_URL` in any project's `.env` (sqlite, mysql, app DB on a different port, etc.) sees `broken-db` from gstack — and brain-aware blocks get silently suppressed in their planning-skill `SKILL.md` files.

## Reproduction

Setup: gbrain 0.42.x with `engine: "pglite"` (e.g. `~/.gbrain/config.json` lacks `database_url`).

```bash
# Working: $HOME, no .env in scope, no DATABASE_URL leaked
rm -f ~/.gstack/.gbrain-local-status-cache.json
( cd "$HOME" && env -u DATABASE_URL -u GBRAIN_DATABASE_URL \
    ~/.claude/skills/gstack/bin/gstack-gbrain-detect ) | jq .gbrain_local_status
# → "ok"

# Broken: any project repo with DATABASE_URL in .env
cd /path/to/some-project   # contains .env with DATABASE_URL=sqlite:///app.db
rm -f ~/.gstack/.gbrain-local-status-cache.json
~/.claude/skills/gstack/bin/gstack-gbrain-detect | jq .gbrain_local_status
# → "broken-db"

# Direct gbrain works fine — proving the brain isn't actually broken:
gbrain sources list --json   # exit 0, valid JSON, empty stderr
```

[...]

PR #1931, opened 2026-06-09
fix(gbrain): strip leaked db env for PGLite probes

maxpetrusenkoagent:
## Summaryn- strip leaked DATABASE_URL values when a PGLite gbrain config has no database_urln- keep explicit config database_url and GSTACK_RESPECT_ENV_DATABASE_URL behavior unchangedn- add regressions for buildGbrainEnv and local status probingnnFixes #1917.nn## Verificationn- bun test test/build-gbrain-env.test.ts test/gbrain-local-status.test.tsnn## Notesn- Branch is based on the fork main because the agent token lacks GitHub workflow scope for pushing upstream workflow history changes.n


maxpetrusenkoagent:
Autoreview: CLEANnnVerified:n- bun test test/build-gbrain-env.test.ts test/gbrain-local-status.test.tsn

garrytan (approved):
Reviewed against main. Correct fix for #1917, well-tested.

**Verified:**
- The PGLite-config branch strips both `DATABASE_URL` and `GBRAIN_DATABASE_URL` so Bun's project `.env` autoload can't make the probe hit an unrelated app database and misclassify a healthy file-backed brain as `broken-db`. The `engine === "pglite" || database_path` detection covers both config shapes.
- The `GSTACK_RESPECT_ENV_DATABASE_URL=1` opt-out still works (early return precedes the strip), and the legacy no-signal config keeps today's pass-through behavior — the regression test rewrite makes that explicit.
- The `fail-on-database-url` fake-gbrain behavior is a nice direct proof of the leak path.

**Coordination note (no action needed from you):** the fix-wave PR from `garrytan/triage-high-priority-prs` touches the same neighborhoods — it removes the GBRAIN_PREPARE auto-set block just below your hunk in `lib/gbrain-exec.ts` (#1965) and extends the same `makeEnv`/`makeFakeGbrainScript` unions in `test/gbrain-local-status.test.ts` (a `"slow"` behavior for #1964). Whichever lands second has small mechanical conflicts; resolution keeps both sides. One heads-up for the docblock: the wave deletes the "sets GBRAIN_PREPARE=true" paragraph your hunk sits next to — keep the wave's version of that paragraph, your PGLite paragraph above it.

Thanks — this closes #1917.

time-attack:
@maxpetrusenkoagent Rewrite this against current main and validate the current GBrain configuration paths. The environment cleanup is still needed, but this branch conflicts and is far behind.

PR #1933, opened 2026-06-09
feat(browse): env-driven Chromium launch overrides

jeanfw:
## What

Two no-op-by-default launch overrides for the browse server's Chromium, read from the environment so users can customize the browser **without patching the source** (which `/gstack-upgrade` wipes via `git reset --hard`):

- **`GSTACK_CHROMIUM_ARGS`** — extra Chromium command-line flags appended to the launch args. Accepts a JSON array of strings (`'["--mute-audio", "--use-fake-ui-for-media-stream"]'`) or a whitespace-separated string (`'--flag-a --flag-b'`).
- **`GSTACK_HTTP_CREDENTIALS`** — HTTP basic-auth as `"user:pass"`, auto-answers 401 challenges (e.g. a dev environment behind a shared gate). Splits on the first colon so passwords may contain colons.

Both default to a no-op, so out-of-the-box launch behavior is unchanged.

## Why

There's currently no supported way to add Chromium flags or basic-auth credentials to the headed/headless browser. The only option is editing `browser-manager.ts` directly, and that edit is lost on every upgrade. This mirrors the existing `GSTACK_CLAUDE_BIN_ARGS` env-override pattern (JSON-array-or-scalar) so the convention is consistent.

## How

- Parsing extracted into a small pure module `browse/src/launch-overrides.ts` (`parseExtraChromiumArgs`, `parseHttpCredentials`) with unit tests.
- Wired into all three launch paths: headless `launch()`, headed `launchHeaded()`, and the headless→headed handoff re-launch.

## Tests

`bun test browse/test/launch-overrides.test.ts` — 13 assertions covering JSON-array, whitespace-split, single-flag, empty-drop, colon-in-password, and the unset/blank/no-colon no-op cases.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -22,6 +22,7 @@ import { emitActivity } from './activity';
 import { validateNavigationUrl } from './url-validation';
 import { TabSession, type RefEntry } from './tab-session';
 import { resolveChromiumProfile, cleanSingletonLocks } from './config';
+import { parseExtraChromiumArgs, parseHttpCredentials } from './launch-overrides';
 import { withCdpSession } from './cdp-bridge';
 import type { MemorySnapshot, MemoryStructureStats, MemoryTabSnapshot, MemoryProcess } from './memory-snapshot';
 
@@ -369,6 +370,9 @@ export class BrowserManager {
       console.log(`[browse] Extensions loaded from: ${extensionsDir}`);
     }
 
+    // User-supplied extra Chromium flags (GSTACK_CHROMIUM_ARGS), no-op when unset.
+    launchArgs.push(...parseExtraChromiumArgs());
+
     this.browser = await chromium.launch({
       headless: useHeadless,
       // On Windows, Chromium's sandbox fails when the server is spawned through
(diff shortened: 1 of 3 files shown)


jbetala7:
The parsing helpers are clean and well-covered by `launch-overrides.test.ts`. Two things before this lands:

**1. Security — `GSTACK_HTTP_CREDENTIALS` is unscoped.** Setting `httpCredentials` without an `origin` means Playwright will answer a 401 from *any* origin the browse session reaches with these basic-auth creds, not just the intended dev gate. If a session navigates to (or is redirected to) an untrusted site that returns 401, the credentials leak to that origin. Playwright's `httpCredentials` accepts an optional `origin` to restrict sending to a single origin — worth either threading an origin through the env var or, at minimum, documenting in the helper comment that creds are sent to any 401-issuing origin. A test asserting the origin is honored would lock it in.

**2. Coordination.** This overlaps #1968 (`executablePath` in the headless `launch()`) and #1915 (`GSTACK_HEADED_NOGPU`) on the same launch paths. #1915's hard-coded `--disable-gpu ...` is exactly the kind of thing a user could pass through your `GSTACK_CHROMIUM_ARGS`, so there may be room to consolidate — flagging so the three don't conflict on merge.

time-attack:
@16francej not needed close agent does this for you

PR #1935, opened 2026-06-09
fix: tool rewrites catch all 'Agent tool' phrasings across agent-runtime hosts

katlun-lgtm:
## Problem

Each agent-runtime host's `toolRewrites` keyed the Agent→host-tool rewrite on the exact literal `'use the Agent tool'`, applied via case-sensitive `replaceAll` in `gen-skill-docs.ts`. That exact phrase appears **0 times** in generated content. The real phrasings in the templates/resolvers are:

- `using the Agent tool` (ship, cso)
- `Dispatch via the Agent tool` / `via the Agent tool`
- `Use the Agent tool` (capitalized, sentence start)
- bare `Agent tool` — `(via Agent tool)`, `(Agent tool,`, `Claude Code's Agent tool` (autoplan, plan-eng-review)

So the host's subagent tool name never rendered and the Claude-ism "Agent tool" leaked into generated docs. In the Hermes tree: `Agent tool` appeared **16×** and `delegate_task` **0×**. Same gap in gbrain/openclaw (→ `sessions_spawn`) and factory.

## Fix

Add ordered rules to each host: the article-bearing form first (drops "the" so there's no dangling article), then a bare sweep for the article-less forms.

- **hermes** → `delegate_task`
- **gbrain**, **openclaw** → `sessions_spawn`
- **factory** → `delegation`. Factory's rewrites are plain prose, not tool names, so a literal mirror (`'dispatch a subagent'`) would produce "using dispatch a subagent" and the noun "a subagent" would echo the surrounding wording ("Dispatch this step as a subagent using a subagent…"). The plain noun `delegation` reads clean in every slot ("using delegation with `subagent_type`", "via delegation", "multiple delegation calls"). A possessive-first rule handles `Claude Code's Agent tool`.

## Verification

Regenerated all four host trees — zero `Agent tool` leaks, target tokens present:

| host | `Agent tool` after | target token |
|------|--------------------|--------------|
| hermes | 0 | `delegate_task` ×16 |
| gbrain | 0 | `sessions_spawn` ×16 |
| openclaw | 0 | `sessions_spawn` ×16 |
| factory | 0 | `delegation` ×31 |

The factory ship golden baseline (`test/fixtures/golden/factory-ship-SKILL.md`) was updated; its diff is exclusively `the Agent tool` → `delegation`, reviewed line-by-line.

## Test plan

- [x] `bun test test/host-config.test.ts test/gen-skill-docs.test.ts` → **433 pass, 0 fail**
[...]

PR #1937, opened 2026-06-09
docs(hermes): correct cliCommand and GBrain non-suppression comments

katlun-lgtm:
Two documentation-accuracy fixes to the Hermes host config (`hosts/hermes.ts`), from a review. **Comments only — no behavioral change**, generated output is unchanged.

## 1. `cliCommand: 'hermes'` — documented as verified

The binary name had been flagged as an unverified guess. It is correct: the installed executable is `hermes`. The repo/product is `nousresearch/hermes-agent` — the package name differs from the executable, which is the likely source of the doubt. Added an inline note so it isn't re-flagged.

## 2. GBrain non-suppression comment — corrected

The comment claimed the un-suppressed `GBRAIN_CONTEXT_LOAD` / `GBRAIN_SAVE_RESULTS` resolvers "handle GBrain-not-installed gracefully (proceed without brain context)", which reads as if the rendered output were benign/empty. It isn't: the blocks instruct `gbrain search` / `get_page` / `put_page` commands and only fall back at the end. Reworded to describe the actual behavior — commands are emitted, each block ends with a fallback ("if GBrain is not available, proceed without / skip this step"), and a no-GBrain agent recovers after a failed probe rather than getting stuck.

The design itself (not suppressing these, because Hermes is brain-capable and can carry GBrain as a mod) is correct and unchanged — only the comment was inaccurate.

## Test plan

- [x] `bun run scripts/host-config-export.ts validate` → `All 10 configs valid` (config still loads/parses)
- [x] Comments only; `gen:skill-docs --host hermes` output is byte-identical

Independent of #1935 and #1936.

### DIFF
--- a/hosts/hermes.ts
+++ b/hosts/hermes.ts
@@ -3,7 +3,7 @@ import type { HostConfig } from '../scripts/host-config';
 const hermes: HostConfig = {
   name: 'hermes',
   displayName: 'Hermes',
-  cliCommand: 'hermes',
+  cliCommand: 'hermes',  // installed executable is `hermes`; the repo/product is `hermes-agent` (nousresearch/hermes-agent)
   cliAliases: [],
 
   globalRoot: '.hermes/skills/gstack',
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej not useful
close pls

marcusLewi:
Unsubscribe

PR #1940, opened 2026-06-09
Fix override injection false positives

### LINKED ISSUE #1934: INJECTION_PATTERNS /override[:\s]/i false-rejects legitimate learnings/decisions containing the word "override"
### Summary

`lib/jsonl-store.ts` `INJECTION_PATTERNS` includes `/override[:\s]/i`, which matches the bare English/code word **"override"** followed by any whitespace. This silently rejects legitimate `gstack-learnings-log` / `gstack-decision-log` writes whose free-text discusses override behavior — very common in architecture and code-review learnings:

- "prose **overrides** the deterministic table on key overlap"
- "never **override** the flag when stale"
- "the renderer **override** of the table"

All of these get rejected as prompt-injection.

### Why the pattern adds ~no protection

A genuine injection like `Override: ignore all previous instructions` is **already** caught by `/ignore\s+(all\s+)?previous\s+(instructions|context|rules)/i`. So the bare `override` token carries the false-positives without unique safety value. (Confirmed: that injection string trips both patterns; removing `override` loses nothing.)

### The failure is silent

`bin/gstack-learnings-log` runs the validator under `set -euo pipefail`:

```bash
VALIDATED=$(printf '%s' "$INPUT" | bun -e "…" 2>/dev/null)
if [ $? -ne 0 ] || [ -z "$VALIDATED" ]; then
  exit 1
fi
```

Under `set -e`, the failing command substitution exits the script **at the assignment line**, so the following `if … exit 1` is effectively dead code. The validator's rejection reason goes to the `2>/dev/null` void, and callers (the skills) invoke the bin with their own `2>/dev/null`. Net effect: a rejected learning vanishes with zero signal — it looks like a successful no-op.

### Repro

```bash
gstack-learnings-log '{"skill":"review","type":"pitfall","key":"x","insight":"never override the flag","confidence":7,"source":"observed"}'
echo $?   # 1, silently — nothing written, no reason surfaced
```

### Proposed fixes

1. **Narrow the pattern to injection context** (keeps real coverage, drops the prose false-positive):
   ```js
   /\boverride\b(?:\s*:|\s+(?:the\s+)?(?:all|every|previous|prior|system|instructions?|rules?|safety|checks?|gate|guard|review)\b)/i
   ```
   Verified: passes "prose overrides the table" / "never override the flag" / "method override"; still rejects "Override: …", "override all rules", "override the system", "override safety checks".

2. **Surface the rejection reason** in the bins: capture the validator's stderr instead of `2>/dev/null`, and echo it / persist it to a discoverable log so a dropped write isn't invisible (especially since callers swallow stderr). Wrapping the validator assignment in `set +e` / `set -e` is also needed so the rejection-handling block actually runs.

Happy to send a PR if useful.

JonasFocus:
Fixes #1934

What changed:
- Narrowed the shared `override` injection pattern so normal prose like `never override the flag` can be stored.
- Kept the guard for instruction-like override attempts such as `Override:` and `override all rules`.
- Updated `gstack-learnings-log` so validation failures reach the error handler and print the validator message instead of silently disappearing.
- Added regression coverage for the shared sanitizer, learnings log path, and decision validation path.

Verification:
- `bash -n bin/gstack-learnings-log`
- `git diff --check`
- `bun test test/jsonl-store.test.ts test/learnings.test.ts test/gstack-decision.test.ts test/gstack-decision-bins.test.ts`
- Manual CLI repro: benign `never override the flag` now writes to `learnings.jsonl`; `ignore all previous instructions` still exits 1 and prints the rejection reason.
- `bun run test` was also run. It exited 0, but the log includes a pre-existing isolated failure in `test/user-slug-fallback.test.ts` expecting `user_slug_at_<hash>` while current code writes `user_slug_at_local`. I confirmed the same isolated failure on a clean `origin/main` worktree.

### DIFF
--- a/bin/gstack-learnings-log
+++ b/bin/gstack-learnings-log
@@ -14,6 +14,9 @@ mkdir -p "$GSTACK_HOME/projects/$SLUG"
 INPUT="$1"
 
 # Validate and sanitize input
+VALIDATE_ERR="$(mktemp)"
+trap 'rm -f "$VALIDATE_ERR"' EXIT
+set +e
 VALIDATED=$(printf '%s' "$INPUT" | bun -e "
 import { hasInjection } from '$SCRIPT_DIR/../lib/jsonl-store.ts';
 const raw = await Bun.stdin.text();
@@ -63,9 +66,12 @@ if (!j.ts) j.ts = new Date().toISOString();
 j.trusted = j.source === 'user-stated';
 
 console.log(JSON.stringify(j));
-" 2>/dev/null)
+" 2>"$VALIDATE_ERR")
+VALIDATE_STATUS=$?
+set -e
 
-if [ $? -ne 0 ] || [ -z "$VALIDATED" ]; then
+if [ "$VALIDATE_STATUS" -ne 0 ] || [ -z "$VALIDATED" ]; then
+  cat "$VALIDATE_ERR" >&2
   exit 1
 fi
 
--- a/lib/jsonl-store.ts
+++ b/lib/jsonl-store.ts
@@ -27,7 +27,7 @@ export const INJECTION_PATTERNS: readonly RegExp[] = [
   /you\s+are\s+now\s+/i,
   /always\s+output\s+no\s+findings/i,
   /skip\s+(all\s+)?(security|review|checks)/i,
-  /override[:\s]/i,
+  /\boverride\b(?:\s*:|\s+(?:the\s+)?(?:all|every|previous|prior|system|instructions?|rules?|security|safety|checks?|gate|guard|review)\b)/i,
   /\bsystem\s*:/i,
   /\bassistant\s*:/i,
   /\buser\s*:/i,
--- a/test/gstack-decision.test.ts
+++ b/test/gstack-decision.test.ts
@@ -60,6 +60,16 @@ describe("validateDecide", () => {
     expect(r.ok).toBe(false);
     if (!r.ok) expect(r.error).toContain("injection");
   });
+  it("accepts benign override prose (regression: #1934)", () => {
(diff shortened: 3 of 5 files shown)


garrytan (approved):
Reviewed against main + tested the pattern locally. This is the right fix, calibrated the right way.

**Verified:**
- The tightened pattern keeps every true positive we could construct ("Override: ignore all previous instructions", "override all rules", "override the system", "override safety checks") and releases the false-positive class that was silently eating legitimate learnings ("never override the flag when stale", "prose overrides the deterministic table", "the renderer override of the table"). The instruction-noun allowlist after `\boverride\b` is a smart middle ground — strictly better than both the old pattern and a bare-colon-only match.
- The learnings-log stderr surfacing kills the silent-exit-1 path. Good catch that the bare `$?` check was dead under `set -euo pipefail`.

**Two coordination notes (no action needed from you):**
1. The fix-wave PR from this branch (`garrytan/triage-high-priority-prs`) independently made the same stderr-surfacing change to `bin/gstack-learnings-log` plus a Windows cygpath guard (#1950), and added tests in the same region of `test/learnings.test.ts` — whichever lands second has a small mechanical conflict; resolution is keep both (your pattern fix + the cygpath guard).
2. `bin/gstack-question-log` carried a local duplicate of INJECTION_PATTERNS with the same `/override[:\s]/i` bug, so this fix alone didn't reach it — the wave dedupes question-log to import `hasInjection` from `lib/jsonl-store.ts`, so it inherits your fix automatically whichever order these land.

Thanks for the fix and the order-independent test cases — this closes #1934.

time-attack:
@JonasFocus Rewrite this to contain only the predicate fix and resolve current main conflicts; #2077 already shipped the Windows path and error-reporting changes.

PR #1954, opened 2026-06-10
feat: persist imported cookies across browse daemon restarts

eric-pamba:
## Problem

Cookies imported into the browse daemon (`/setup-browser-cookies`, `cookie-import`, `cookie-import-browser`, or the picker UI) live only in the ephemeral Playwright context. Every daemon restart loses them, so QA flows against auth-gated apps (e.g. a localhost app behind Auth0) need a manual cookie re-import each session. Re-importing isn't always possible: `browse cookies` output redacts session-token values by design, and the source browser may not have the cookie anymore.

## Change

New `browse/src/cookie-persistence.ts` with two helpers:

- `persistCookies(context)` — snapshots `context.cookies()` to `~/.gstack/saved-cookies.json`. Atomic write (tmp + rename), file mode `0600` since the contents are session tokens.
- `loadPersistedCookies(context)` — restores them on context creation. Skips files older than 30 days, drops expired cookies, keeps session cookies (`expires === -1`).

Wire-up:

- `write-commands.ts`: persist after every `addCookies` in the `cookie`, `cookie-import`, and both `cookie-import-browser` direct modes (`--domain`, `--all`).
- `cookie-picker-routes.ts`: persist after picker imports, and after domain removal so deletions stick.
- `browser-manager.ts`: auto-load in `launchBrowser()` after `setExtraHTTPHeaders`, before the first tab.

All persistence calls are try/catch-wrapped warn-only — a failed disk write never breaks the cookie operation itself.

## Testing

Verified on macOS (Bun runtime): imported a real Auth0 `__session` for a localhost app, stopped the daemon, relaunched headless — `[browse] Auto-loaded N persisted cookies from previous session` and the app's SSR loader authenticated without a redirect to login. Domain removal via the picker re-persists correctly.

## Notes

- Mirrors what some users patch in locally today (a community "persistent-cookies" skill); upstreaming removes the need to re-apply that patch after every gstack upgrade.
- To clear persisted state: `rm ~/.gstack/saved-cookies.json`.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -24,6 +24,7 @@ import { TabSession, type RefEntry } from './tab-session';
 import { resolveChromiumProfile, cleanSingletonLocks } from './config';
 import { withCdpSession } from './cdp-bridge';
 import type { MemorySnapshot, MemoryStructureStats, MemoryTabSnapshot, MemoryProcess } from './memory-snapshot';
+import { loadPersistedCookies } from './cookie-persistence';
 
 /**
  * Detect whether GSTACK_CHROMIUM_PATH points at a custom Chromium build that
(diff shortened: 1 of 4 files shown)


eric-pamba:
Closing — keeping this as a local patch on our end.

PR #1969, opened 2026-06-11
fix: classify slow gbrain probes separately

### LINKED ISSUE #1964: gbrain-local-status: 5s probe timeout misclassifies healthy-but-slow engines as broken-config; /sync-gbrain skips code+memory
## Symptom

On a brain whose Postgres sits behind a Supabase **transaction-mode pooler** (port 6543), `/sync-gbrain` permanently skips the code + memory stages:

```
  SKIP  code         skipped — local engine broken-config — config at ~/.gbrain/config.json is malformed; see /setup-gbrain Step 1.5
  SKIP  memory       skipped — local engine broken-config — ...
  OK    brain-sync   curated artifacts pushed
```

…while the engine is in fact healthy: `gbrain doctor` is green and `gbrain sources list --json` returns valid JSON with exit 0. The config file is well-formed.

## Root cause

`lib/gbrain-local-status.ts` classifies the engine by probing `gbrain sources list --json` with `PROBE_TIMEOUT_MS = 5_000`. On this setup a fresh CLI connection through the pooler takes **6.9–10.7s** (measured across runs), so the probe times out every time. The timeout error's stderr matches neither `"Cannot connect to database"` nor `"config.json"`, so the defensive default (codex #8) classifies it as `broken-config` — and the 60s cache then propagates the misdiagnosis to the orchestrator and every skill preamble.

The user-facing message ("config at ~/.gbrain/config.json is malformed") points at a file that is fine, so the remediation steps in /setup-gbrain Step 1.5 can't fix it.

## Reproduce

Any gbrain on a connection where `time gbrain sources list --json` > 5s (e.g. Supabase transaction pooler from a residential connection). Then run `/sync-gbrain` → code+memory SKIP with broken-config.

## Suggested fix (any of)

1. Raise `PROBE_TIMEOUT_MS` (20s covered my worst case), and/or
2. Honor an env override (e.g. `GSTACK_GBRAIN_PROBE_TIMEOUT_MS`) so slow-pooler users can tune it without patching the lib, and/or
3. Classify the timeout case distinctly (e.g. `slow-engine`) instead of falling into `broken-config`, with a message that names the real condition ("probe exceeded Ns — engine may be reachable but slow; raise the timeout").

Locally patching `PROBE_TIMEOUT_MS` 5000→20000 fully resolves it (sync runs all stages), but the patch is reverted by every `git reset --hard` upgrade.

RudraMalvankar:
## Related Issue

Closes #1964

## Summary

This PR fixes an issue where healthy but slow gbrain engines could be incorrectly classified as `broken-config` when the probe operation exceeded the default timeout threshold.

In environments where the gbrain CLI takes longer than 5 seconds to respond (for example, when using remote databases or connection poolers), the probe would time out and fall back to the generic `broken-config` classification. As a result, `/sync-gbrain` skipped the code and memory stages even though the engine itself was healthy and operational.

### 1. Added a dedicated `slow-engine` status

Introduced a new `LocalEngineStatus` value:

```ts
"slow-engine"
```

This allows timeout-related failures to be distinguished from actual configuration problems.

### 2. Improved timeout handling

The classifier now explicitly detects timeout conditions (such as `ETIMEDOUT`) and returns `slow-engine` instead of incorrectly falling back to `broken-config`.

### 3. Made probe timeout configurable

Replaced the hardcoded timeout value with an environment-variable-based configuration:

```ts
GSTACK_GBRAIN_PROBE_TIMEOUT_MS
```

This allows users with slower database connections or remote infrastructure to adjust the probe timeout without modifying source code.

### 4. Increased the default timeout

Updated the default probe timeout from:

```ts
5000ms
```

to:

```ts
20000ms
```

This better accommodates slower but healthy environments while still preventing indefinite waits.

### 5. Added a user-friendly sync message

When a slow engine is detected, users now receive a clear explanation indicating that the engine may still be healthy and suggesting increasing the timeout or improving database connectivity, rather than being told their configuration is malformed.

## Why This Change?

The original behavior treated timeout failures as configuration failures, which could mislead users into troubleshooting the wrong component. This PR separates slow-response scenarios from actual configuration issues, making diagnostics more accurate and improving the overall user experience.

## Testing

* Verified that timeout-related failures are classified as `slow-engine`.
* Verified that existing `broken-config` and `broken-db` classifications remain unchanged.
* Verified that the timeout can be configured using `GSTACK_GBRAIN_PROBE_TIMEOUT_MS`.
* Verified that sync skip messages now provide more accurate guidance for slow environments.

## Checklist

* [x] Code follows project style
* [x] No unrelated changes included
* [x] Existing behavior preserved for valid configurations
* [x] Added clearer handling for slow-engine scenarios


16francej:

PR #1978, opened 2026-06-11
fix: support Codex auth env and AUQ fallback

hsuperman:
## Summary
- load Codex auth env for non-interactive gstack Codex automation from `.env`, `.env.local`, machine-local `~/.config/recruitmagic/cli.env`, and Codex `env_key` config
- preserve caller shell state when sourcing `bin/gstack-codex-probe`
- keep upstream AUQ failure fallback behavior after merging current `main`, with test coverage adjusted for interactive prose fallback vs headless blocking

## Why
Codex can be installed and still fail gstack automation when the auth key only exists in a repo `.env` file. `codex exec` / `codex -p` need auth in the process environment, and Codex can also use a configured Azure/OpenAI `env_key` from `~/.codex/config.toml`. The probe should mirror that reality.

This also keeps the new AskUserQuestion fallback behavior compatible with Codex-style interactive chat hosts: no AUQ tool means render the decision as prose and wait, while headless sessions still block.

## Tests
- ALL PASSED: `bun test test/codex-hardening.test.ts test/resolver-ask-user-format.test.ts test/preamble-compose.test.ts test/gen-skill-docs-idempotency.test.ts`
- ALL PASSED: `bun run gen:skill-docs --dry-run`


time-attack:
AUQ is nice but doesnt work in practice ( gstack supports a lot of different ai models and they get the premise AUQ already)

also arbritrary .env reading isnt secure

@16francej

PR #1979, opened 2026-06-11
fix(browse): stop isProcessAlive from flashing a Windows console every watchdog tick

### LINKED ISSUE #1952: terminal-agent watchdog flashes a console window every tick on Windows — isProcessAlive() spawns tasklist instead of signal-0
## TL;DR

On Windows, a connected headed `browse` session blinks a console window
(`conhost.exe`) on a fixed interval for the entire session — no CLI command, no
user action. The terminal-agent watchdog calls `isProcessAlive()` once per
`setInterval` tick to check the agent PID, and on Windows `isProcessAlive()`
shells out to `tasklist`, which Windows gives its own console window.
macOS/Linux stay silent because their path is a `process.kill(pid, 0)` syscall
that spawns nothing. It is purely cosmetic — each window self-closes — but it
blinks on a fixed interval for the whole session with no indication of what it
is, which is confusing and distracting.

## Root Cause

`isProcessAlive()` (`browse/src/error-handling.ts`) has a Windows-only branch
that spawns `tasklist`:

```ts
export function isProcessAlive(pid: number): boolean {
  if (IS_WINDOWS) {
    try {
      const result = Bun.spawnSync(
        ['tasklist', '/FI', `PID eq ${pid}`, '/NH', '/FO', 'CSV'],
        { stdout: 'pipe', stderr: 'pipe', timeout: 3000 }
      );
      return result.stdout.toString().includes(`"${pid}"`);
    } catch { return false; }
  }
  try { process.kill(pid, 0); return true; } catch { return false; }
}
```

The terminal-agent watchdog (`browse/src/server.ts`) registers a `setInterval`
that fires every `AGENT_WATCHDOG_TICK_MS` (default `60000`) on the normal headed
`connect` path (`ownsTerminalAgent === true`) and calls `isProcessAlive()` on
every tick:

```ts
agentWatchdogInterval = setInterval(() => {
  if (isShuttingDown) return;
  if (agentRespawnGuardTripped) return;
  const record = readAgentRecord(stateDir);
  if (record && isProcessAlive(record.pid)) return;   // ← spawns tasklist on Windows, every tick
  // ...respawn dead agent...
}, AGENT_WATCHDOG_TICK_MS);
```

Each tick → `Bun.spawnSync(['tasklist', …])` → a console window. The **parent
watchdog** two functions away already avoids exactly this: it calls
`process.kill(BROWSE_PARENT_PID, 0)` directly
(`// signal 0 = existence check only, no signal sent`). Only `isProcessAlive()`
takes the spawn path.

## Evidence

Reproduced on Windows 11, gstack `1.57.6.0`:

1. `browse connect` (headed), then issue no further commands.
2. A console window flashes open and immediately closes on a fixed interval,
   indefinitely, for as long as the session stays connected. Each one
   self-closes — it does not pile up — but the steady periodic blink is the
   problem.
3. The only process-spawning call on the watchdog tick is `isProcessAlive()` →
   `tasklist`; macOS shows no window for the same watchdog (its branch is
   `process.kill(pid, 0)`).

## Proposed Fix

Use signal-0 on all platforms — what the parent watchdog already does. Node/Bun
implement `process.kill(pid, 0)` on Windows via `OpenProcess`: a pure existence
check, no child process, no window.

```diff
 export function isProcessAlive(pid: number): boolean {
-  if (IS_WINDOWS) {
-    try {
-      const result = Bun.spawnSync(
-        ['tasklist', '/FI', `PID eq ${pid}`, '/NH', '/FO', 'CSV'],
-        { stdout: 'pipe', stderr: 'pipe', timeout: 3000 }
-      );
-      return result.stdout.toString().includes(`"${pid}"`);
-    } catch {
-      return false;
-    }
-  }
-  try {
-    process.kill(pid, 0);
-    return true;
-  } catch {
-    return false;
-  }
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (err: any) {
+    // ESRCH → gone; EPERM → exists but not signallable → still alive.
+    return err?.code === 'EPERM';
+  }
 }
```

### Why `EPERM` ⇒ alive

`process.kill(pid, 0)` throws `ESRCH` when the process is gone (→ false) and
`EPERM` when it exists but can't be signalled (→ alive, → true). The old Unix
branch collapsed every error to false; unifying on signal-0 corrects that edge
for free.

## Verification

Applied the patch and restarted the server:

- Patched `isProcessAlive()` to the signal-0 form in `src/error-handling.ts`
  **and** in the running `dist/server-node.mjs` — on Windows the server is
  launched from the bundle, not `src/` (`resolveNodeServerScript()` in
  `cli.ts`), so a `src/`-only edit has no runtime effect until re-bundled.
- After restart, the periodic console-window flashes stopped completely —
  observed over several minutes on Windows 11.
- The watchdog's alive/dead branch is unchanged; signal-0 returns the same
  boolean it already keys on, so dead-agent detection and respawn still work.

The same `isProcessAlive` is also compiled into the `browse` CLI binary, which
spawns `tasklist` on every stale-detection call — the one-line fix covers it on
rebuild.

---

Windows 11. gstack `1.57.6.0`. Bun.

jbetala7:
## Problem

On Windows, a connected headed `browse` session blinks a `conhost.exe` console window on a fixed interval for the entire session — no CLI command, no user action. The terminal-agent watchdog calls `isProcessAlive()` once per `setInterval` tick (default 60s) to check the agent PID, and on Windows `isProcessAlive()` shelled out to `tasklist`, which Windows gives its own console window. macOS/Linux stayed silent because their branch is a `process.kill(pid, 0)` syscall that spawns nothing.

It is purely cosmetic — each window self-closes — but it blinks on a fixed interval for the whole session with no indication of what it is.

## Root cause

`isProcessAlive()` (`browse/src/error-handling.ts`) had a Windows-only branch that spawned `tasklist`:

```ts
if (IS_WINDOWS) {
  const result = Bun.spawnSync(['tasklist', '/FI', `PID eq ${pid}`, '/NH', '/FO', 'CSV'], {...});
  return result.stdout.toString().includes(`"${pid}"`);
}
```

Every watchdog tick → a `tasklist` spawn → a console window. The parent watchdog two functions away already avoids exactly this by calling `process.kill(pid, 0)` directly; only `isProcessAlive()` took the spawn path.

## Fix

Unify on signal-0 across all platforms. Node/Bun implement `process.kill(pid, 0)` on Windows via `OpenProcess` — a pure existence check, no child process, no window. The Windows-only branch (and its now-unused `IS_WINDOWS` const) are removed.

This also corrects a latent edge: the old branch collapsed every error to `false`, so a live-but-unsignalable process (`EPERM`) read as dead. Now `ESRCH → dead`, `EPERM → alive`.

```ts
export function isProcessAlive(pid: number): boolean {
  try {
    process.kill(pid, 0);
    return true;
  } catch (err: any) {
    // ESRCH → gone (false); EPERM → exists but not signallable → still alive.
    return err?.code === 'EPERM';
  }
}
```

The same `isProcessAlive` is compiled into the `browse` CLI binary's polling loops, so the one-file fix covers those on rebuild too. The watchdog's alive/dead branch is unchanged — signal-0 returns the same boolean it already keys on, so dead-agent detection and respawn still work.

## Testing

`bun test browse/test/error-handling.test.ts` → **9/9 pass**. Added:
- an `EPERM`-as-alive regression case (probes PID 1, which a non-root user can't signal), and
- a static tripwire asserting the `isProcessAlive` source never reintroduces `tasklist`/`spawn` — the idiomatic gstack guard for a Windows-only cosmetic bug that can't be exercised on macOS CI. **Proven to fail on the old `tasklist` body** before the fix.

The watchdog/terminal-agent suites (`terminal-agent-watchdog.test.ts`, `watchdog.test.ts`, etc.) stay green. One unrelated pre-existing failure in `terminal-agent.test.ts` (a source-guard grepping for `spawnClaude(`, since renamed to `maybeSpawnPty`) reproduces on clean `main` with this branch stashed — not touched here (this PR only changes `browse/src/error-handling.ts` + its test).

Fixes #1952

### DIFF
--- a/browse/src/error-handling.ts
+++ b/browse/src/error-handling.ts
@@ -7,8 +7,6 @@
 
 import * as fs from 'fs';
 
-const IS_WINDOWS = process.platform === 'win32';
-
 // ─── Filesystem ────────────────────────────────────────────────
 
 /** Remove a file, ignoring ENOENT (already gone). Rethrows other errors. */
@@ -36,23 +34,24 @@ export function safeKill(pid: number, signal: NodeJS.Signals | number): void {
   }
 }
 
-/** Check if a PID is alive. Pure boolean probe — returns false for ALL errors. */
+/**
+ * Check if a PID is alive. Pure boolean probe — never throws.
+ *
+ * Uses signal-0 on every platform. Node/Bun implement `process.kill(pid, 0)`
+ * on Windows via `OpenProcess` — a pure existence check that spawns no child
+ * process and opens no console window. The old Windows branch shelled out to
+ * `tasklist`, which Windows gives its own console window; on the terminal-agent
+ * watchdog's per-tick existence check that flashed a `conhost.exe` window every
+ * 60s for the whole session (#1952). The parent watchdog already uses signal-0
+ * directly; this unifies on it.
+ */
 export function isProcessAlive(pid: number): boolean {
-  if (IS_WINDOWS) {
-    try {
-      const result = Bun.spawnSync(
-        ['tasklist', '/FI', `PID eq ${pid}`, '/NH', '/FO', 'CSV'],
-        { stdout: 'pipe', stderr: 'pipe', timeout: 3000 }
-      );
-      return result.stdout.toString().includes(`"${pid}"`);
-    } catch {
-      return false;
-    }
-  }
   try {
     process.kill(pid, 0);
     return true;
-  } catch {
-    return false;
+  } catch (err: any) {
+    // ESRCH → gone (false); EPERM → exists but not signallable → still alive.
+    // The old Unix branch collapsed every error to false, missing the EPERM edge.
+    return err?.code === 'EPERM';
   }
 }
--- a/browse/test/error-handling.test.ts
+++ b/browse/test/error-handling.test.ts
@@ -2,6 +2,7 @@ import { describe, test, expect } from 'bun:test';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
+import { fileURLToPath } from 'url';
 import { safeUnlink, safeKill, isProcessAlive } from '../src/error-handling';
 
 describe('safeUnlink', () => {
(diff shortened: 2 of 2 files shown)


time-attack:
@jbetala7 Please rewrite this narrow fix to handle compiled Windows behavior safely instead of assuming signal zero works there.

PR #1981, opened 2026-06-11
fix(gbrain): parse refresh detection without Python path dependency

### LINKED ISSUE #1967: `gstack-config gbrain-refresh` on Windows always reports `local-status: unknown` — Windows Python cannot open the MSYS-style detection-file path
## Environment

- Windows 11, Git Bash (MSYS2)
- gstack v1.57.9.0, global-git install at `~/.claude/skills/gstack`
- `python3` resolves to the native Windows Python (`/c/Users/<user>/AppData/Local/Microsoft/WindowsApps/python3`)
- gbrain 0.35.8.0, PGLite engine, fully healthy (`gbrain doctor` ok, `gbrain sources list --json` exits 0)

## Symptom

`gstack-config gbrain-refresh` always bails with:

```
gbrain not detected (local-status: unknown) → brain-aware blocks will be suppressed in planning-skill SKILL.md files.
```

…even though the detection file it just wrote (`~/.gstack/gbrain-detection.json`) contains `"gbrain_local_status": "ok"`. The render step is never reached, so the new-in-1.57.9.0 brain-aware blocks can't be enabled on Windows via the documented command.

## Root cause

`bin/gstack-config` (gbrain-refresh case, ~line 388) parses the detection JSON with:

```sh
STATUS=$("$PYTHON_CMD" -c "import json,sys; d=json.load(open('$DETECTION_FILE')); print(d.get('gbrain_local_status','unknown'))" 2>/dev/null || echo unknown)
```

`$DETECTION_FILE` is an MSYS-style path (`/c/Users/<user>/.gstack/gbrain-detection.json`). `command -v python3 || command -v python` finds **native Windows Python**, which does not understand `/c/...` paths and raises `FileNotFoundError`. The `2>/dev/null` swallows the traceback, the `|| echo unknown` kicks in, and `STATUS=unknown` falls through to the "not detected" branch.

Repro (any Windows box with Git Bash + Windows Python):

```sh
$ python3 -c "import json; json.load(open('/c/Users/<user>/.gstack/gbrain-detection.json'))"
FileNotFoundError: [Errno 2] No such file or directory: '/c/Users/<user>/.gstack/gbrain-detection.json'
```

The grep fallback a few lines down would have parsed the file correctly, but it's only used when no Python is found.

## Suggested fix (any of)

1. On Windows/MSYS, convert the path first: `DETECTION_FILE_NATIVE=$(cygpath -w "$DETECTION_FILE" 2>/dev/null || echo "$DETECTION_FILE")` and pass that to Python.
2. Prefer the existing grep fallback over Python (it's sufficient for these two flat string fields and has no path-translation problem).
3. Parse with `bun` instead — it's already a hard requirement for the render step in the same code path, and Bun handles MSYS paths fine.

## Workaround

`cd ~/.claude/skills/gstack && bun run gen:skill-docs:user --host claude` renders the brain-aware blocks correctly (gen-skill-docs reads the detection file itself via `--respect-detection` and isn't affected).

## Adjacent (separate) issue, mentioning for context

On the same machine, the `freshClassify` probe (`lib/gbrain-local-status.ts`, `PROBE_TIMEOUT_MS = 5_000`) intermittently times out on a 633MB PGLite brain's cold open and caches a false `broken-config` for the TTL window — which is what `./setup` hits when it prints "gbrain not detected" during install. `--no-cache` immediately returns `ok`. That looks like another flavor of #1792 (probe misclassifies a healthy brain); happy to file separately with timings if useful.

maxpetrusenkoagent:
## Summary
- fixes `gstack-config gbrain-refresh` so the detection summary no longer depends on Python opening the detection file path
- parses the flat `gbrain-detection.json` status/version fields with shell `grep`/`sed` first, then keeps Python only as a fallback
- adds a regression test with a fake PATH `python3` that fails like native Windows Python does on MSYS-style paths

Closes #1967

## Verification
- `bun test test/gstack-config-gbrain-refresh.test.ts` (failed before the fix, passes after)
- `bash -n bin/gstack-config`
- `bun test test/gstack-config-gbrain-refresh.test.ts test/gstack-config-redact-keys.test.ts`
- Windows CI subset equivalent plus new regression:
  `bun test test/gstack-paths.test.ts browse/test/claude-bin.test.ts test/test-free-shards.test.ts browse/test/file-permissions.test.ts browse/test/security.test.ts browse/test/server-sanitize-surrogates.test.ts test/setup-windows-fallback.test.ts test/build-script-shell-compat.test.ts test/docs-config-keys.test.ts test/brain-sync-windows-paths.test.ts make-pdf/test/browseClient.test.ts make-pdf/test/pdftotext.test.ts test/gstack-config-gbrain-refresh.test.ts`
- Local autoreview subagent: clean, no blocking findings

## Notes
- I also tried `bun run test:windows`; it failed in existing `browse/test/terminal-agent.test.ts` on a source-level guard unrelated to this diff. The actual Windows workflow's curated file list plus the new regression passes locally.

### DIFF
--- a/bin/gstack-config
+++ b/bin/gstack-config
@@ -381,18 +381,20 @@ case "${1:-}" in
     fi
     mv "$DETECTION_FILE.tmp" "$DETECTION_FILE"
 
-    # Summarize for the user. Use python (already required elsewhere) to
-    # parse the JSON portably; fall back to grep if python is unavailable.
-    PYTHON_CMD=$(command -v python3 || command -v python || true)
-    if [ -n "$PYTHON_CMD" ]; then
-      STATUS=$("$PYTHON_CMD" -c "import json,sys; d=json.load(open('$DETECTION_FILE')); print(d.get('gbrain_local_status','unknown'))" 2>/dev/null || echo unknown)
-      VERSION=$("$PYTHON_CMD" -c "import json,sys; d=json.load(open('$DETECTION_FILE')); print(d.get('gbrain_version') or 'unknown')" 2>/dev/null || echo unknown)
-    else
-      STATUS=$(grep -o '"gbrain_local_status":[[:space:]]*"[^"]*"' "$DETECTION_FILE" | sed 's/.*"\([^"]*\)"$/\1/')
-      VERSION=$(grep -o '"gbrain_version":[[:space:]]*"[^"]*"' "$DETECTION_FILE" | sed 's/.*"\([^"]*\)"$/\1/')
-      [ -z "$STATUS" ] && STATUS=unknown
-      [ -z "$VERSION" ] && VERSION=unknown
+    # Summarize for the user. Prefer the shell parser because Git Bash can put
+    # native Windows Python first on PATH, and that Python cannot open MSYS-style
+    # /c/... paths. Keep Python only as a fallback if the flat-field grep misses.
+    STATUS=$(grep -o '"gbrain_local_status":[[:space:]]*"[^"]*"' "$DETECTION_FILE" | sed 's/.*"\([^"]*\)"$/\1/' | head -1 || true)
+    VERSION=$(grep -o '"gbrain_version":[[:space:]]*"[^"]*"' "$DETECTION_FILE" | sed 's/.*"\([^"]*\)"$/\1/' | head -1 || true)
+    if [ -z "$STATUS" ] || [ -z "$VERSION" ]; then
+      PYTHON_CMD=$(command -v python3 || command -v python || true)
+      if [ -n "$PYTHON_CMD" ]; then
+        [ -z "$STATUS" ] && STATUS=$("$PYTHON_CMD" -c "import json,sys; d=json.load(open('$DETECTION_FILE')); print(d.get('gbrain_local_status','unknown'))" 2>/dev/null || echo unknown)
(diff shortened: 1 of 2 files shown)


maxpetrusenkoagent:
## Hermes verification note

Autoreview against the exact PR diff is clean.

Verification run locally:
- `bun test test/gstack-config-gbrain-refresh.test.ts` failed before the fix and passes after
- `bash -n bin/gstack-config`
- `bun test test/gstack-config-gbrain-refresh.test.ts test/gstack-config-redact-keys.test.ts`
- Windows workflow subset equivalent plus this regression:
  `bun test test/gstack-paths.test.ts browse/test/claude-bin.test.ts test/test-free-shards.test.ts browse/test/file-permissions.test.ts browse/test/security.test.ts browse/test/server-sanitize-surrogates.test.ts test/setup-windows-fallback.test.ts test/build-script-shell-compat.test.ts test/docs-config-keys.test.ts test/brain-sync-windows-paths.test.ts make-pdf/test/browseClient.test.ts make-pdf/test/pdftotext.test.ts test/gstack-config-gbrain-refresh.test.ts`

Reviewer state:
- GitHub reports no checks on this fork branch, so I ran explicit Codex autoreview against `gh pr diff #1981`.
- Codex verdict: `clean`, no blocking findings.

Known unrelated local gate note:
- `bun run test:windows` currently fails in existing `browse/test/terminal-agent.test.ts` source-level guard. This PR does not touch `browse/src/terminal-agent.ts` or that test, and the actual Windows workflow's curated file list plus the new regression passes locally.

PR #1986, opened 2026-06-12
feat(codex): document native image generation in /codex skill

seungwonme:
## TL;DR

Codex CLI ships a native image-generation tool (`image_gen.imagegen`), but the /codex skill's hard "read-only, never modify files" rule makes image-generation consults silently fail — the tool call succeeds, the generated PNG just can't be saved. This documents the capability and the one sanctioned exception.

## What changed

- New `## Image Generation` section in `codex/SKILL.md.tmpl` (after Model & Reasoning):
  - swap `-s read-only` → `-s workspace-write` when the consult asks Codex to MAKE an image (not analyze one)
  - scope writes to a scratch dir under `$TMP_ROOT` via `-C` + `--skip-git-repo-check`, so nothing lands in the user's repo
  - read + open the result for the user, report the file path
- `Important Rules` first bullet now names this as the sole exception to "Never modify files"
- `codex/SKILL.md` regenerated via `bun run gen:skill-docs`

## Verified

Asked Codex to enumerate its callable tools and generate an image with code-drawing fallbacks (SVG/matplotlib) explicitly forbidden: it listed `image_gen.imagegen` and produced a real 1254×1254 PNG. Saving worked under `workspace-write`; the skill's default `read-only` sandbox blocks it.

### DIFF
--- a/codex/SKILL.md
+++ b/codex/SKILL.md
@@ -1514,6 +1514,29 @@ or `/codex challenge -m gpt-5.2`), pass the `-m` flag through to codex.
 
 ---
 
+## Image Generation
+
+Codex ships a native image-generation tool (`image_gen.imagegen`) — it produces real
+raster images (illustrations, photo-style renders, diagrams) from a text prompt, not
+just code that draws SVGs. Verified by having Codex enumerate its tools and generate
+a PNG. Image *input* is separate and always available (`codex -i screenshot.png`).
+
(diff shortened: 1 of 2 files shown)


time-attack:
@seungwonme Please rewrite this to use the public image-generation name and define artifact confinement before granting nested workspace-write, with regression coverage.

PR #1992, opened 2026-06-13
Fix local gbrain trust policy config keys

### LINKED ISSUE #1988: bug: gstack-config key validator rejects brain_trust_policy@local for stdio engines
## Summary

`gstack-config set brain_trust_policy@local personal` fails with a validation error when gbrain is configured as a stdio (local PGLite) engine, making `/setup-gbrain` report `DONE_WITH_CONCERNS` and silently skip writing the trust policy.

## Root cause

`endpoint_hash()` in `bin/gstack-config` intentionally falls back to the literal string `"local"` when no MCP URL is found (stdio engines have no URL to hash):

```bash
# bin/gstack-config ~L161
# MCP server URL. Falls back to the literal 'local' when no MCP is configured.
endpoint_hash() {
  ...
  printf '%s' "local"   # ← stdio fallback
}
```

The comment at L138 even documents this intent:
```
#   brain_trust_policy@<hash>  — unset on fresh install; setup-gbrain
#                                writes 'personal' for local engines,
```

However, the key validator on both `get` and `set` only allows hex characters after `@`:

```bash
# get (L265) and set (L280) — same pattern in both
if ! printf '%s' "$KEY" | grep -qE '^[a-zA-Z0-9_]+(@[a-f0-9]+)?$'; then
  echo "Error: key must contain only alphanumeric characters, underscores, and an optional @<hex-hash> suffix" >&2
  exit 1
fi
```

`local` contains `l` and `o` which are outside `[a-f0-9]`, so the key `brain_trust_policy@local` is rejected even though `endpoint_hash()` produces it by design.

The same hex-only pattern also appears in the `get` value-lookup grep (L270).

## Affected lines

| Line | Code |
|------|------|
| 265 | `grep -qE '^[a-zA-Z0-9_]+(@[a-f0-9]+)?$'` (get validator) |
| 270 | `grep -E "^${KEY%@*}(@[a-f0-9]+)?:"` (get value lookup) |
| 280 | `grep -qE '^[a-zA-Z0-9_]+(@[a-f0-9]+)?$'` (set validator) |

## Fix

Change `[a-f0-9]+` → `[a-zA-Z0-9]+` at all three sites. This covers both sha8/sha16 hex hashes (remote endpoints) and the `local` fallback (stdio engines) without opening up anything unsafe.

```bash
# get + set validators
grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'

# get value lookup
grep -E "^${KEY%@*}(@[a-zA-Z0-9]+)?:"
```

Error message wording should also drop "hex-" from "optional @\<hex-hash\> suffix".

## Observed behaviour

`/setup-gbrain` on a stdio/local-PGLite engine reports:

```
DONE_WITH_CONCERNS — minor:
gstack-config set brain_trust_policy@local personal fails validation because
the @<hash> suffix regex requires hex-only ([a-f0-9]+) and endpoint-hash
returns the literal 'local' for stdio engines.
```

The trust policy is never written; the concern is non-fatal but visible on every `/setup-gbrain` run.

## Verified in

`~/.claude/skills/gstack` commits `a5833c4` (v1.57.10.0) and `8241949` (v1.57.9.0) — present in both.

JonasFocus:
## Summary
- Allow `gstack-config` endpoint suffixes to use alphanumeric endpoint IDs, so `brain_trust_policy@local` is valid for stdio/local gbrain engines.
- Update related config tests for `@local` and punctuation rejection.

## Testing
- `bash -n bin/gstack-config`
- `bun test browse/test/gstack-config.test.ts`
- `bun test test/user-slug-fallback.test.ts`
- Manual check: `GSTACK_STATE_DIR=<tmp> bash bin/gstack-config set brain_trust_policy@local personal && ... get brain_trust_policy@local` returns `personal`.

Fixes #1988

### DIFF
--- a/bin/gstack-config
+++ b/bin/gstack-config
@@ -260,14 +260,14 @@ resolve_user_slug() {
 case "${1:-}" in
   get)
     KEY="${2:?Usage: gstack-config get <key>}"
-    # Validate key (alphanumeric + underscore + optional @<hash> suffix for
+    # Validate key (alphanumeric + underscore + optional @<endpoint-id> suffix for
     # endpoint-namespaced keys introduced by the brain-aware planning layer)
-    if ! printf '%s' "$KEY" | grep -qE '^[a-zA-Z0-9_]+(@[a-f0-9]+)?$'; then
-      echo "Error: key must contain only alphanumeric characters, underscores, and an optional @<hex-hash> suffix" >&2
+    if ! printf '%s' "$KEY" | grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'; then
+      echo "Error: key must contain only alphanumeric characters, underscores, and an optional @<endpoint-id> suffix" >&2
       exit 1
     fi
-    # Use literal match for keys containing @ (sha hashes), regex otherwise
-    VALUE=$(grep -F "${KEY}:" "$CONFIG_FILE" 2>/dev/null | grep -E "^${KEY%@*}(@[a-f0-9]+)?:" | grep -F "${KEY}:" | tail -1 | awk '{print $2}' | tr -d '[:space:]' || true)
+    # Use literal match for keys containing @ (endpoint ids), regex otherwise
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej Please close this PR. The local GBrain trust-policy key fix already shipped in #2222 and includes the original issue attribution. There is no remaining change to take from this branch.

PR #1993, opened 2026-06-13
fix: honor terse skill rendering in user installs

### LINKED ISSUE #886: GENERAL: Reduce Token Overhead
Session initialisation hook on every skill invocation pays a ~3000-4,200 token tax before the actual skill content. This makes using gstack more expensive than it should be.

maxpetrusenkoagent:
## Summary
- honor user-local `explain_level: terse` when setup/session-update regenerates installed Claude skills
- keep canonical checked-in SKILL.md output full-prose while using `gen:skill-docs:user` for local installs
- add regression coverage for terse user-local rendering and setup's non-gbrain path

## Verification
- `git diff --check`
- `bash -n setup`

Closes #886


time-attack:
@maxpetrusenkoagent Please rewrite this from draft form and include completed CI results and human approval for the generator, setup, and test changes.

PR #2010, opened 2026-06-15
fix(gstack-slug): resolve slug from git remote, not a stale cache

steven1370:
## Problem
`bin/gstack-slug` resolves a project's slug **cache-first** (`~/.gstack/slug-cache/<encoded-path>`) and never re-validates it against the git remote once cached. If the slug is first computed while a repo's `origin` is temporarily wrong — most commonly a project scaffolded by copying another repo before its remote is repointed — the wrong slug gets cached and then **shadows the corrected remote forever**. Every consumer (ship metrics, `/retro`, artifacts, project state) then files under the wrong project slug indefinitely.

## Root cause
Lines 16-19 trust a present cache file verbatim, and the remote is only consulted on a cache miss (line 23). There is no invalidation path, so a stale entry is permanent.

## Fix
Flip resolution to **remote-first**:
1. Derive the slug from `git remote get-url origin` on every run.
2. Fall back to the cached slug (then basename) only when there is no remote.
3. Always refresh the cache to match the resolved slug.

A stale or poisoned cache now self-heals on the next run. For a remote-less repo or a non-repo dir, behavior is unchanged: the cache still provides cross-session stability and is still sanitized on read.

Invariants preserved: `[a-zA-Z0-9._-]` sanitization on every path, atomic cache write via `mktemp` + `mv`, and `pipefail`-safe remote read.

## Verification
- `bash -n` clean.
- Correct resolution for repos with a remote, no-remote dirs (basename fallback), and a repo whose slug had previously been mis-cached.
- **Self-heal:** writing a wrong value into the cache file, then running once, returns and re-caches the correct remote-derived slug.
- **Injection safety:** a malicious cached value (only reachable on a no-remote dir) is sanitized to `[a-zA-Z0-9._-]`, not executed.

### DIFF
--- a/bin/gstack-slug
+++ b/bin/gstack-slug
@@ -13,33 +13,37 @@ PROJECT_DIR="$(pwd)"
 CACHE_KEY=$(printf '%s' "$PROJECT_DIR" | tr '/' '_')
 CACHE_FILE="${CACHE_DIR}/${CACHE_KEY}"
 
-# 1. Try cached slug first (guarantees consistency across sessions)
-if [[ -f "$CACHE_FILE" ]]; then
-  SLUG=$(cat "$CACHE_FILE")
+# 1. Compute the slug from the git remote — the source of truth. Run separately
+#    from the pipeline so `set -o pipefail` can't swallow the error into an empty
+#    slug. Computing from the remote on EVERY run (not just on a cache miss) is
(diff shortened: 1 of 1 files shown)


time-attack:
there are a lot of bugs with the slug but this doesnt fix them @16francej pls close

steven1370:
Thank You!

PR #2012, opened 2026-06-15
fix(make-pdf): correct CJK rendering (URL sentinel leak, JP-first fonts, CJK quotes)

rssprivacy-commits:
Three real defects surface when rendering Simplified-Chinese documents with `make-pdf`. All reproduced on v1.57.10.0, fixed at source, and verified.

### 1. Bare URLs corrupt the document (worst)
Any bare URL leaks internal `SMARTPANTS_PRESERVED_N` sentinels as visible text **and** leaves the `<a>`/`<p>` unclosed — so every heading/paragraph after the URL becomes one giant hyperlink.

**Root cause:** `smartypants.ts` carves tags into NUL-delimited `SMARTPANTS_PRESERVED_N` placeholders, then `URL_RE = /\bhttps?:\/\/\S+/g` runs. `\S+` (a NUL is non-whitespace) greedily swallows the adjacent `</a>`/`</p>` placeholders; the single-pass restore then cannot un-nest them.
**Fix:** stop the URL match at the NUL boundary — `[^\s]+`.

### 2. Simplified-Chinese renders in Japanese glyphs
`print-css.ts` `CJK_STACK` lists `Hiragino Kaku Gothic ProN` (Japanese) before any Chinese font, so Chinese text falls back to Japanese glyph variants (直/骨/角/没/别) — `pdffonts` shows a mix of `HiraKakuProN` + `PingFangSC` in one document.
**Fix:** put `PingFang SC, Heiti SC, Noto Sans CJK SC, Source Han Sans SC, Microsoft YaHei` first; JP fonts demoted to last resort.

### 3. Opening quotes after CJK punctuation render as closing quotes
A double/single quote directly after CJK punctuation (：，。（「) rendered as a *closing* quote. The opening-quote heuristic only treats a quote as opening after whitespace/brackets.
**Fix:** add CJK punctuation/openers `：，。、；！？（【「『〈《` to the opening-quote context. Ideographs are intentionally excluded, so a closing quote after a Han character stays closing.

### Verification
- `pdffonts`: PingFang-only, zero Hiragino
- `pdftotext`: no `SMARTPANTS_PRESERVED` leak; correct curly quotes
- Visual render: no anchor bleed, uniform Chinese glyphs
- `bun test make-pdf/test/`: **91 pass / 0 fail** (unchanged before/after)

Total change: 5 insertions / 5 deletions across 2 files. Opened as draft for maintainer review.

### DIFF
--- a/make-pdf/src/print-css.ts
+++ b/make-pdf/src/print-css.ts
@@ -35,8 +35,8 @@
 // Metric-compatible sans stack: Helvetica (macOS), Liberation Sans (Linux,
 // ships via fonts-liberation), Arial (Windows). Shared by every text surface.
 const SANS_STACK = `Helvetica, "Liberation Sans", Arial`;
-// CJK fallback families, appended to the body stack only.
-const CJK_STACK = `"Hiragino Kaku Gothic ProN", "Noto Sans CJK JP", "Microsoft YaHei"`;
+// CJK fallback families (Simplified-Chinese first), appended to the body stack only.
+const CJK_STACK = `"PingFang SC", "Heiti SC", "Noto Sans CJK SC", "Source Han Sans SC", "Microsoft YaHei", "Hiragino Kaku Gothic ProN", "Noto Sans CJK JP"`;
 // Color-emoji families: Apple (macOS), Segoe (Windows), Noto (Linux).
 const EMOJI_FAMILIES = `"Apple Color Emoji", "Segoe UI Emoji", "Noto Color Emoji"`;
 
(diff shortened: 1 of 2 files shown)


time-attack:
Thanks for this. The URL repair duplicates focused work, while a globally Chinese-first font stack can render Japanese incorrectly and the quote scope lacks adequate coverage. Please split the URL change, make font selection language-aware, align the punctuation scope, and add real CJK PDF and font probes.

PR #2013, opened 2026-06-15
fix(retro): count polyglot test files and exclude vendored dirs

### LINKED ISSUE #1999: /retro test-file count uses a JS-only glob, undercounts Python repos (and overcounts if naively fixed)
## Problem

`/retro` Step 1 command 10 counts repository test files with a JavaScript/TypeScript-shaped glob:

`find . -name '*.test.*' -o -name '*.spec.*' -o -name '*_test.*' -o -name '*_spec.*' 2>/dev/null | grep -v node_modules | wc -l`

Python's dominant convention is `test_*.py`, which none of those four patterns match. On a Python repo the `test_health.total_test_files` metric silently lands on whatever stray `*_test.*` files happen to exist, never the real count. Because Step 13 persists this number and Step 12 trends it, the test-health trend line measures noise, not test growth.

There's a second failure mode hiding behind the obvious fix. Naively adding `test_*.py` over-corrects, because the only directory excluded is `node_modules`, while Python projects keep their dependencies in a virtualenv (`.venv/`, `venv/`, `site-packages/`) full of third-party test suites.

Measured on a real Python repo this week:

| Method | Count | Why |
|--------|-------|-----|
| current gstack glob | 54 | misses `test_*.py` entirely |
| naive `find . -name 'test_*.py'` | 1,496 | 1,188 of those live in `.venv/` (dependency tests) |
| correct project count (`tests/`) | 297 | the actual project test files |

So the live metric reported 54 where the truth is 297, a 5.5x undercount.

## Proposed fix

Two parts, both required:

1. Extend the glob with common non-JS conventions: `test_*.py`, `*_test.py`, `*_test.go`, `*_test.rb`, `*Test.java`, `*_spec.rb`. The existing JS/TS patterns stay.
2. Exclude virtualenv and vendored dependency dirs, not just `node_modules`: add `.venv`, `venv`, any `*-venv`, and `site-packages` to the prune/grep filter. Otherwise the Python fix sweeps in thousands of dependency test files.

A `-prune`-based find (or a `grep -vE '(/node_modules/|/\.venv/|/venv/|/site-packages/)'` filter) handles both the count (command 10) and the test-files-changed metric (command 12) consistently.

## Why it matters

`test_health` is one of the few quality signals the retro trends over time. On every Python repo it's been reporting a number 5x too low, so "tests added this period" deltas are meaningless. The fix is a few characters in two `find` commands and makes the metric correct for the largest non-JS ecosystem.

genisis0x:
## Problem

`/retro` Step 1 command 10 counts repository test files with a JS/TS-only glob and excludes only `node_modules`:

```bash
find . -name '*.test.*' -o -name '*.spec.*' -o -name '*_test.*' -o -name '*_spec.*' 2>/dev/null | grep -v node_modules | wc -l
```

Python's dominant convention is `test_*.py`, which none of those patterns match, so on a Python repo the `test_health.total_test_files` metric never reflects the real count. Step 13 persists it and Step 12 trends it, so the test-health trend line tracks noise. As #1999 notes, a naive widening over-corrects, because virtualenv / vendored dependency trees (`.venv`, `venv`, `site-packages`, `vendor`) carry their own test suites and only `node_modules` was pruned.

## Fix

- Add the common non-JS conventions: `test_*.py` and `*Test.java`. The existing `*_test.*` / `*_spec.*` already cover Go (`*_test.go`), Ruby (`*_test.rb` / `*_spec.rb`), and Python's `*_test.py`, so the genuine gaps are Python's `test_` prefix and Java's separator-less `FooTest.java`.
- Prune vendored and virtualenv directories (`node_modules`, `.venv`, `venv`, `site-packages`, `vendor`, `.git`) inside `find` instead of post-filtering only `node_modules`, so dependency test suites no longer inflate the count.

## Verification

On a polyglot fixture (5 real project test files across ts/js/go/py/java, plus dependency test files under `.venv/` and `node_modules/`):

| glob | count |
|------|-------|
| old (JS-only, node_modules filter) | 4 (misses `test_*.py` + `*Test.java`, counts a `.venv` test) |
| new | 5 (exact project count) |

`bun test test/gen-skill-docs.test.ts test/skill-validation.test.ts` — 734 pass, 0 fail. SKILL.md regenerated from the template.

The same glob shape also lives in the shared `scripts/resolvers/testing.ts` (the `/ship` before/after counts); left out of scope here to keep this focused on the `/retro` trend metric in #1999.

Closes #1999

### DIFF
--- a/retro/SKILL.md
+++ b/retro/SKILL.md
@@ -996,8 +996,8 @@ cat ~/.gstack/greptile-history.md 2>/dev/null || true
 # 9. TODOS.md backlog (if available)
 cat TODOS.md 2>/dev/null || true
 
-# 10. Test file count
-find . -name '*.test.*' -o -name '*.spec.*' -o -name '*_test.*' -o -name '*_spec.*' 2>/dev/null | grep -v node_modules | wc -l
+# 10. Test file count (polyglot conventions; prunes vendored/virtualenv dirs)
+find . \( -name node_modules -o -name .venv -o -name venv -o -name site-packages -o -name vendor -o -name .git \) -prune -o \( -name '*.test.*' -o -name '*.spec.*' -o -name '*_test.*' -o -name '*_spec.*' -o -name 'test_*.py' -o -name '*Test.java' \) -type f -print 2>/dev/null | wc -l
 
 # 11. Regression test commits in window
 git log origin/<default> --since="<window>" --oneline --grep="test(qa):" --grep="test(design):" --grep="test: coverage"
--- a/retro/SKILL.md.tmpl
+++ b/retro/SKILL.md.tmpl
@@ -196,8 +196,8 @@ cat ~/.gstack/greptile-history.md 2>/dev/null || true
 # 9. TODOS.md backlog (if available)
 cat TODOS.md 2>/dev/null || true
 
-# 10. Test file count
-find . -name '*.test.*' -o -name '*.spec.*' -o -name '*_test.*' -o -name '*_spec.*' 2>/dev/null | grep -v node_modules | wc -l
+# 10. Test file count (polyglot conventions; prunes vendored/virtualenv dirs)
+find . \( -name node_modules -o -name .venv -o -name venv -o -name site-packages -o -name vendor -o -name .git \) -prune -o \( -name '*.test.*' -o -name '*.spec.*' -o -name '*_test.*' -o -name '*_spec.*' -o -name 'test_*.py' -o -name '*Test.java' \) -type f -print 2>/dev/null | wc -l
 
 # 11. Regression test commits in window
 git log origin/<default> --since="<window>" --oneline --grep="test(qa):" --grep="test(design):" --grep="test: coverage"


jbetala7:
Heads-up on an overlap with #2037 (opened 06-18). Both PRs rewrite the same Step-1 test-counting glob — command 10 at `retro/SKILL.md:1000` and the mirrored line in `retro/SKILL.md.tmpl`:

```
find . -name '*.test.*' -o -name '*.spec.*' -o -name '*_test.*' -o -name '*_spec.*' 2>/dev/null | grep -v node_modules | wc -l
```

— replacing the JS/TS-only glob + node_modules-only filter with a polyglot pattern and vendored-dir exclusion. Because both touch the identical lines in both `SKILL.md` and `SKILL.md.tmpl`, whichever merges second will conflict.

This PR is the narrower, issue-linked fix (Closes #1999, command 10 only). #2037 is effectively a superset: it also fixes the per-commit test glob in commands 12/13 (`retro/SKILL.md:1009`, still `grep -E '\.(test|spec)\.'` on `main`, which this PR doesn't touch) and adds a Step 1.5 anti-fabrication guard + regression test.

Not flagging either as wrong — just so the authors/maintainer can sequence them: if #2037 lands it subsumes this change (worth carrying `Closes #1999` and crediting this PR); if this lands first, #2037 only needs to drop the now-duplicate command-10 hunk and keep its command-12/13 + plausibility-guard value.

genisis0x:
Thanks for the careful triage @jbetala7 — that's an accurate summary. This one is deliberately the narrow, issue-linked fix (Closes #1999, command 10 only); #2037 is the broader change with the command 12/13 glob and the anti-fabrication guard on top.

I'm easy either way: if the maintainer would rather take #2037 as the superset, I'll close this in its favor — worth carrying the `Closes #1999` link, and the command-10 hunk is identical so no behavior is lost. If the smaller one is easier to land first, #2037 can drop the now-duplicate command-10 lines and keep its 12/13 + guard value. Whatever sequencing is least work on your side.

time-attack:
@16francej — recommend closing this PR. It is superseded by #2037 because it improves only one total-count command, leaving changed-test and shared counters inconsistent. Credit #2013, and retain its safe total-count change plus the root and nested vendor cases in #2037.

genisis0x:
Closing in favor of #2037, which takes the more complete approach (it covers changed-test and shared counters, not just the single total-count command this PR touched). Happy to help land #2037 instead.

PR #2014, opened 2026-06-15
fix(autoplan): surface scope-detection counts and stop silent phase skips

### LINKED ISSUE #1957: /autoplan Phase 0: silent phase-skip when scope detection misses by one term; no diagnostic
Filed downstream at aviraldua93/gstack-copilot#9. Bug in canonical upstream gstack — exposed during smoke test of `/autoplan`.

## Summary

`/autoplan` Phase 0 detects whether to run Phase 2 (Design Review) and Phase 3.5 (DX Review) by grepping the plan file with `grep -ciE "\b($TERMS)\b"` and a threshold of `≥ 2 matches`. When count is below threshold, the phase is silently skipped — no diagnostic, no opportunity to override.

## Failure modes

1. **Silent miss** — plan IS UI/DX-relevant but uses hyphenated terms (`form-control`, `api-endpoint`), terms in code fences grep skips, or terms synonymous-but-not-listed (`screen reader` matches, `chrome accessibility tree` does not). User gets 2 phases instead of 4 and never knows why.
2. **Borderline trip** — exact-threshold matches give no warning. Lose one match anywhere and the phase silently disappears next run.

## Suggested fixes

1. **Print matched terms, not just count**:
```
UI scope detection:
  matched: dashboard (×2)
  threshold: 2
  → UI_SCOPE: yes
```
When count equals threshold, add inline warning `⚠ borderline match (count = threshold); confirm or override`.

2. **Confirm at the gate when borderline** — single AskUserQuestion before phases run.

3. **Expand term list and version it** — move to `scripts/scope-terms.json` so community can contribute. Current list is 10 UI terms, 30 DX terms — short for the domain.

---
*Filed via GitHub Copilot CLI by @aviraldua93 during stress-testing a downstream port.*

genisis0x:
## Problem

`/autoplan` Phase 0 decides whether to run Phase 2 (Design Review) and Phase 3.5 (DX Review) by grepping the plan for scope terms and requiring 2+ matches. When a scope lands below the threshold, the phase is dropped silently — the run reports only a yes/no, with no match count, no near-miss warning, and no way to override.

Two failure modes follow:

1. **Silent miss** — a plan that is genuinely UI- or developer-facing but phrases terms with hyphens (`form-control`, `api-endpoint`), uses synonyms outside the list, or buries them in code fences grep skips, loses two whole review phases and the user never learns why.
2. **Borderline trip** — a scope at exactly 2 matches gives no warning; one fewer match on a re-run makes the phase silently disappear.

## Fix

In Phase 0 Step 2:

- Report the match **count** per scope, not just yes/no, and never skip a phase silently.
- Treat a below-threshold scope (0-1 matches) that the plan plausibly touches as a **near-miss** rather than a clean no, naming the known false-negative triggers (hyphenated terms, synonyms, code-fenced terms).
- Call out the **borderline** exactly-two case.
- Offer an **override** — the user can tell the skill to force a phase regardless of the count.

The Phase 0 output line now carries the counts and the override note.

This is scoped to `/autoplan` and the change is prose-only in the skill template plus its regenerated `SKILL.md`; it touches no shared preamble.

## Verification

Added `test/autoplan-scope-diagnostic.test.ts` (gate tier, 6 checks) pinning the count reporting, near-miss flag, borderline flag, override offer, the output-line format, and template/generated parity. All six assertions return 0 occurrences on `main`, so the test fails before the fix and passes after. `bun test test/gen-skill-docs.test.ts test/skill-validation.test.ts` — 734 pass, 0 fail.

Closes #1957

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -1038,6 +1038,14 @@ Then prepend a one-line HTML comment to the plan file:
   a developer tool (the plan describes something developers install, integrate, or build
   on top of) or if an AI agent is the primary user (OpenClaw actions, Claude Code skills,
   MCP servers).
+- Record the match COUNT for each scope, not just yes/no, and never skip a phase
+  silently. A scope under the 2-match threshold means the phase is dropped, so make the
+  miss visible: when a scope lands at 0-1 matches yet the plan plausibly touches it
+  (hyphenated terms like `form-control`/`api-endpoint`, synonyms not in the list, or
+  terms inside code fences that grep skips), flag it as a near-miss rather than a clean
+  no. A scope sitting at exactly 2 matches is borderline — one fewer match on a re-run
+  would silently drop the phase — so call that out too. In every case offer the override:
+  the user can tell you to force the phase and you run it regardless of the count.
 
 ### Step 3: Load skill files from disk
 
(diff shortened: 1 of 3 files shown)


time-attack:
@genisis0x Please rewrite this into a focused Autoplan change that keeps source and generated output synchronized and covers counts, near misses, exactly-two cases, and forced paths.

PR #2020, opened 2026-06-16
v1.32.0.1 fix: browse daemon stop/restart crash-loop - defer shutdown after response

devkd111:
## Problem

`meta-commands.ts` case `'stop'` and `'restart'` call `await shutdown()` BEFORE returning the HTTP response. `shutdown()` tears down the HTTP listener and calls `process.exit()`, so the response never flushes. The CLI sees `ECONNRESET`, its crash-retry logic re-sends the command, and the daemon logs "crashed twice in a row" before exit.

## Fix

Replace `await shutdown()` with `setTimeout(() => { shutdown(); }, 0)` in both `'stop'` and `'restart'` cases. This defers the shutdown until after the current execution context completes, allowing the HTTP response to flush before the listener is torn down.

## Test

Added regression tests in `browse/test/commands.test.ts`:
- `Server Control > stop returns response before shutdown fires` — verifies shutdown calls are deferred (not synchronous)
- `Server Control > restart returns response before shutdown fires` — same for restart
- `Server Control > status returns healthy` — baseline

All 226 existing tests pass, zero regressions.

## Files changed

- `browse/src/meta-commands.ts` — defer shutdown via setTimeout(0)
- `browse/test/commands.test.ts` — regression tests (3 new, 226 total pass)
- `VERSION` — bump 1.32.0.0 to 1.32.0.1
- `package.json` — bump 1.32.0.0 to 1.32.0.1
- `CHANGELOG.md` — entry for 1.32.0.1


time-attack:
@16francej — recommend closing this PR. It is superseded by #1732 because its timer-ordering test does not show that a real HTTP response is flushed before exit. Credit #2020 for the response-ordering idea, while requiring bounded response-body, process-identity, state, and restart-behavior coverage in #1732.

PR #2021, opened 2026-06-16
fix(autoplan): bind commit before split-array index in tasks aggregator (#2018)

### LINKED ISSUE #2018: /autoplan aggregator silently returns 0 tasks: jq scope bug in commit filter
## Summary

`/autoplan`'s Phase 4 Implementation Tasks aggregator returns zero tasks for every invocation because of a jq scope bug. The error is swallowed by `2>/dev/null`, so users see an empty aggregated list with no diagnostic, even when the per-phase `tasks-<phase>-*.jsonl` files contain valid records on the current branch/commit.

Repro'd against `v1.58.1.0` (also affects `v1.52.2.0`).

## Location

`autoplan/SKILL.md:1641` (current `main`):

```jq
select(.branch == $branch and ($commits | split("|") | index(.commit) != null))
```

## Root cause

Inside `index(.commit)`, `.commit` is evaluated against jq's **current input**, which at that point is the array produced by `$commits | split("|")` — NOT the outer record. jq tries to "index the split array by the string `commit`" and errors out:

```
jq: error (at <stdin>:1): Cannot index array with string "commit"
```

The error is suppressed by `2>/dev/null` on the surrounding `jq ... >> "$ALL_JSONL"` invocation, so the aggregator silently drops every record and the user sees:

> _No actionable tasks emitted from any phase._

…even when the JSONL files clearly contain matching records on the current branch and recent commits.

## Repro

```bash
echo '{"branch":"dev","commit":"abc123","phase":"ceo-review"}' | \
  jq --arg branch "dev" --arg commits "abc123|def456" \
  'select(.branch == $branch and ($commits | split("|") | index(.commit) != null))'
# jq: error (at <stdin>:1): Cannot index array with string "commit"
```

Triggered today against `MHVFCU/mhv-member-manager` with two valid JSONL files (12 CEO tasks + 20 Eng tasks, both on `dev`, both with `commit` values present in the last-5-commits filter set). Aggregator output: 0 tasks. After applying the fix below: 32 tasks rendered correctly.

## Fix

Bind the record before the inner pipeline so `.commit` resolves to the record's field, not the split array:

```jq
. as $r | select($r.branch == $branch and ($commits | split("|") | index($r.commit) != null))
```

Equivalent alternative using `IN/1`:

```jq
select(.branch == $branch and (.commit | IN($commits | split("|") | .[])))
```

Both verified working against the same input.

## Suggested patch

Diff against `autoplan/SKILL.md:1641`:

```diff
-      'select(.branch == $branch and ($commits | split("|") | index(.commit) != null))' \
+      '. as $r | select($r.branch == $branch and ($commits | split("|") | index($r.commit) != null))' \
```

## Why this matters

Every `/autoplan` run is currently shipping an empty `### Implementation Tasks (aggregated across phases)` section at the Final Approval Gate. Users either:
1. Don't notice (and lose the aggregated checklist value that /autoplan promises), or
2. Notice and have to re-aggregate by hand.

The `2>/dev/null` masks the actual jq error from anyone running with verbose shells, which makes this near-impossible to diagnose without dropping the redirect.

## Related: defensive improvement

While you're in there, consider dropping the `2>/dev/null` on the per-file `jq` invocation OR teeing stderr to a debug file. The current silent-error pattern hides not just this bug but any future malformed-JSONL case.

Repro environment: macOS 15.5, jq 1.7.1, bash 5.x, zsh 5.x. Bug is jq-version-independent (it's a scope/parsing issue, not a version-specific behavior change).

0xDevNinja:
## Problem

`/autoplan`'s Phase 4 implementation-tasks aggregator returned **zero tasks on every run**, even when valid per-phase `tasks-<phase>-*.jsonl` records existed for the current branch and recent commits. Users saw `No actionable tasks emitted from any phase` with no diagnostic.

## Root cause

The branch+commit filter was:

```jq
select(.branch == $branch and ($commits | split("|") | index(.commit) != null))
```

Inside `($commits | split("|") | ...)` the pipe's input is the split **array**, so `index(.commit)` tried to index that array with the string `"commit"`:

```
jq: error (at <stdin>:1): Cannot index array with string "commit"
```

The surrounding `jq ... >> "$ALL_JSONL"` swallowed the error via `2>/dev/null`, so the aggregator silently dropped every record. Repro'd on `v1.58.1.0` (also affects `v1.52.2.0`).

## Fix

Bind the record's `.commit` to a variable **before** entering the split-array pipe, so `index()` receives the commit string:

```jq
select(.branch == $branch and (.commit as $c | ($commits | split("|") | index($c)) != null))
```

One-line change in the source resolver (`scripts/resolvers/tasks-section.ts`), `autoplan/SKILL.md` regenerated via `bun run gen:skill-docs`. Added a short comment so the binding isn't "simplified" back into the bug later.

## Tests

New free regression test `test/autoplan-tasks-aggregator.test.ts` extracts the **shipped** `select(...)` filter from `autoplan/SKILL.md` and runs it through `jq`:

- matching branch+commit → kept, no jq error
- non-matching commit → excluded, no error
- wrong branch → excluded, no error
- record missing `commit` field → excluded, no error
- asserts the filter no longer contains the bare `index(.commit)` form

5/5 pass. Full free suite green (`bun test` gen-skill-docs + skill-validation: 734 pass), gen-skill-docs freshness check clean.

Fixes #2018.

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -1637,8 +1637,11 @@ if command -v jq >/dev/null 2>&1; then
       # Filter to current branch + recent commits, then keep records for the
       # latest run_id only. (Single phase may have multiple files if the user
       # re-ran the review; aggregator takes the newest.)
+      # Bind the record's .commit to $c BEFORE entering the split-array pipe —
+      # inside ($commits | split("|") | ...) the input is the array, so a bare
+      # index(.commit) would index the array with the string "commit" and error.
       jq -c --arg branch "$BRANCH" --arg commits "$COMMITS_RECENT" \
-        'select(.branch == $branch and ($commits | split("|") | index(.commit) != null))' \
+        'select(.branch == $branch and (.commit as $c | ($commits | split("|") | index($c)) != null))' \
         "$f" 2>/dev/null >> "$ALL_JSONL" || true
     done < <(find "$TASKS_DIR" -maxdepth 1 -name "tasks-$phase-*.jsonl" 2>/dev/null | sort)
     # Reduce to latest run_id per phase
--- a/scripts/resolvers/tasks-section.ts
+++ b/scripts/resolvers/tasks-section.ts
@@ -118,8 +118,11 @@ if command -v jq >/dev/null 2>&1; then
       # Filter to current branch + recent commits, then keep records for the
       # latest run_id only. (Single phase may have multiple files if the user
       # re-ran the review; aggregator takes the newest.)
+      # Bind the record's .commit to \$c BEFORE entering the split-array pipe —
+      # inside (\$commits | split("|") | ...) the input is the array, so a bare
+      # index(.commit) would index the array with the string "commit" and error.
       jq -c --arg branch "$BRANCH" --arg commits "$COMMITS_RECENT" \\
-        'select(.branch == $branch and ($commits | split("|") | index(.commit) != null))' \\
+        'select(.branch == $branch and (.commit as $c | ($commits | split("|") | index($c)) != null))' \\
         "$f" 2>/dev/null >> "$ALL_JSONL" || true
     done < <(find "$TASKS_DIR" -maxdepth 1 -name "tasks-$phase-*.jsonl" 2>/dev/null | sort)
     # Reduce to latest run_id per phase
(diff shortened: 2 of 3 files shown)


time-attack:
@16francej ensures commits are real and dont get split can merge

PR #2022, opened 2026-06-16
fix(preamble): gate upgrade-handling prose on update_check opt-out (#2001)

### LINKED ISSUE #2001: update_check: false gates the binary but not the preamble instruction text — upgrade-handling prose ships unconditionally
## Summary

`update_check: false` correctly silences the `gstack-update-check` binary (it exits early and emits nothing), but the **instruction text** that tells the agent how to interpret update-check output still ships in every generated `SKILL.md` preamble. The runtime layer honors the opt-out; the instruction layer does not.

This is the same bug *class* as #1150 (preamble content not gated by a config opt-out), for a different flag (`update_check`) and different content (upgrade-handling prose).

## Why it's a real opt-out, not a misconfig

`update_check` is a first-class config key:
- `bin/gstack-config:41` — documented: `# update_check: true   # false = suppress version check notifications`
- `bin/gstack-config:120` — registered default `true`
- Present in the active-values / known-keys lists (`bin/gstack-config:347,363`)

## The layer disagreement

**Binary (honors the flag):** `bin/gstack-update-check:32-34` reads `update_check`, and on `false` exits 0 with no output before any version logic.

**Instruction text (ignores the flag):** `scripts/resolvers/preamble/generate-upgrade-check.ts:8,10` unconditionally emit:

```
If output shows `UPGRADE_AVAILABLE <old> <new>`: read .../gstack-upgrade/SKILL.md and follow the "Inline upgrade flow" ...
If output shows `JUST_UPGRADED <from> <to>`: print "Running gstack v{to} (just updated!)" ...
```

These render into every skill's preamble regardless of the flag.

## Inconsistency with every other config flag

Every other runtime flag follows an **echo-then-gate** convention — `generate-preamble-bash.ts` echoes the value (`PROACTIVE:`, `SKILL_PREFIX:`, `EXPLAIN_LEVEL:`, `QUESTION_TUNING:`) and the instruction layer gates on it. `update_check` is the only flag with **no echoed signal and no instruction-layer gate**.

## Impact

Functionally the upgrade check *is* suppressed — the binary emits nothing, so the conditional is vacuously false and the agent takes no upgrade action. The residual cost is:

- ~13 lines of dead upgrade-handling instructions in every preamble for users who opted out.
- The model still reads and visibly "considers" upgrade state, which reads to the user as the check still firing.
- Convention drift: opt-out should mean opt-out at both layers (the principle #1150 is built on).

## Suggested fix (surgical — ~4 lines)

`generateUpgradeCheck()` is misnamed: it bundles four concerns (PROACTIVE, SKILL_PREFIX, upgrade-availability/just-upgraded, feature discovery), so gating the whole function would wrongly disable the proactive/skill-prefix prose. Instead:

1. Echo `UPDATE_CHECK: $_UPDATE_CHECK` in `generate-preamble-bash.ts` alongside the other flags.
2. Gate **only lines 8 and 10** of `generate-upgrade-check.ts` on `UPDATE_CHECK == "true"`.
3. Regenerate via `bun run gen:skill-docs`.

## Secondary note

`update_check: false` also suppresses the post-upgrade `JUST_UPGRADED` "just updated!" confirmation (the binary exits before `:119`). That arguably differs from a "version check notification" — maintainer may want to scope whether the just-upgraded message should survive the opt-out.

0xDevNinja:
## Problem

`update_check: false` correctly silences the `gstack-update-check` binary (exits early, emits nothing), but the **instruction text** telling the agent how to act on update-check output still shipped in every generated `SKILL.md` preamble. The runtime layer honored the opt-out; the instruction layer ignored it. Same bug class as #1150, different flag.

## Fix — echo-then-gate, like every other runtime flag

Every other flag (`PROACTIVE`, `SKILL_PREFIX`, `EXPLAIN_LEVEL`, `QUESTION_TUNING`) is echoed in the preamble bash and gated in the instruction layer. `update_check` now follows suit:

- **`scripts/resolvers/preamble/generate-preamble-bash.ts`** — reads `update_check` and echoes `UPDATE_CHECK: <value>` alongside the other config flags.
- **`scripts/resolvers/preamble/generate-upgrade-check.ts`** — prepends a gate: *If `UPDATE_CHECK` is `"false"`, skip the next two lines — the binary emits nothing, so there is no `UPGRADE_AVAILABLE` / `JUST_UPGRADED` output to act on.*

All 49 preamble-bearing `SKILL.md` regenerated via `bun run gen:skill-docs`.

### Out of scope (intentionally)

`gstack-upgrade/SKILL.md` documents its **own** inline upgrade flow (hand-authored in the `.tmpl`). A user invoking that skill is explicitly upgrading, so gating its prose would be wrong. The 50-vs-49 count difference is exactly this one file.

## Tests

New invariant in `test/gen-skill-docs.test.ts` iterates every generated SKILL.md with the config-echo cluster and asserts it (a) echoes `UPDATE_CHECK`, (b) reads `update_check`, (c) gates the upgrade prose on the flag. Guards against the scope filter matching nothing.

735 pass (gen-skill-docs + skill-validation), gen-skill-docs freshness check clean.

Fixes #2001.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -68,6 +68,8 @@ if [ "$_EXPLAIN_LEVEL" != "default" ] && [ "$_EXPLAIN_LEVEL" != "terse" ]; then 
 echo "EXPLAIN_LEVEL: $_EXPLAIN_LEVEL"
 _QUESTION_TUNING=$(~/.claude/skills/gstack/bin/gstack-config get question_tuning 2>/dev/null || echo "false")
 echo "QUESTION_TUNING: $_QUESTION_TUNING"
+_UPDATE_CHECK=$(~/.claude/skills/gstack/bin/gstack-config get update_check 2>/dev/null || echo "true")
+echo "UPDATE_CHECK: $_UPDATE_CHECK"
 mkdir -p ~/.gstack/analytics
 if [ "$_TEL" != "off" ]; then
 echo '{"skill":"gstack","ts":"'$(date -u +%Y-%m-%dT%H:%M:%SZ)'","repo":"'$(_repo=$(basename "$(git rev-parse --show-toplevel 2>/dev/null)" 2>/dev/null | tr -cd 'a-zA-Z0-9._-'); echo "${_repo:-unknown}")'"}'  >> ~/.gstack/analytics/skill-usage.jsonl 2>/dev/null || true
@@ -139,6 +141,8 @@ If the user invokes a skill in plan mode, the skill takes precedence over generi
 If `PROACTIVE` is `"false"`, do not auto-invoke or proactively suggest skills. If a skill seems useful, ask: "I think /skillname might help here — want me to run it?"
 
 If `SKILL_PREFIX` is `"true"`, suggest/invoke `/gstack-*` names. Disk paths stay `~/.claude/skills/gstack/[skill-name]/SKILL.md`.
+
+If `UPDATE_CHECK` is `"false"`, skip the next two lines — the update-check binary emits nothing in that mode, so there is no `UPGRADE_AVAILABLE` / `JUST_UPGRADED` output to act on.
 
 If output shows `UPGRADE_AVAILABLE <old> <new>`: read `~/.claude/skills/gstack/gstack-upgrade/SKILL.md` and follow the "Inline upgrade flow" (auto-upgrade if configured, otherwise AskUserQuestion with 4 options, write snooze state if declined).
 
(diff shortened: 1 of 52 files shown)


time-attack:
@0xDevNinja this should just be linked across all files and not an individual update to each file. 
if you fix that i can merge this for you

time-attack:
@16francej Please merge this version because it independently fixes update_check opt-out behavior, regenerates every affected output, and passes both enabled and disabled configuration tests.

PR #2026, opened 2026-06-16
fix(security-dashboard): fail closed instead of reporting "0 attacks" on backend error

### LINKED ISSUE #1947: community-pulse / security dashboard fail open: backend error or missing jq reports "0 attacks" instead of "unknown"
## Summary

The community **security** dashboard can display a reassuring "0 attacks / all clear" when it is actually erroring — a fail-open on a security-signaling surface. Found during a security review of v1.57.7.0.

### 1. `community-pulse` returns all-zeros (HTTP 200) on any error

`supabase/functions/community-pulse/index.ts:194-211`:
```ts
} catch {
  return new Response(
    JSON.stringify({
      weekly_active: 0, change_pct: 0, top_skills: [], crashes: [], versions: [],
      security: {
        attacks_last_7_days: 0,
        top_attack_domains: [],
        top_attack_layers: [],
        verdict_distribution: [],
      },
    }),
    { status: 200, ... }
  );
}
```
Any internal failure (DB outage, malformed cache, query error) is masked as a healthy "0 attacks" response with HTTP 200 — indistinguishable from a genuine zero.

### 2. `gstack-security-dashboard` reports 0 when `jq` is missing

`bin/gstack-security-dashboard:51-64` — when `jq` is unavailable it falls back to a lossy `grep -o '"security":{[^}]*}'` that breaks on the nested arrays (`top_attack_domains`, etc.), and the documented behavior is to "degrade gracefully to '0 attacks'" (the machine-readable fallback emits `attacks_last_7_days: 0`). So a host without `jq` silently under-reports attacks as zero.

### 3. (related, latent) telemetry `error_message` is transmitted unredacted

`bin/gstack-telemetry-sync:83-84` strips only `_repo_slug`/`_branch` before upload, while `bin/gstack-telemetry-log:227` includes a raw `error_message` field in the synced record. The current edge function's `TelemetryEvent` interface (`supabase/functions/telemetry-ingest/index.ts:7-18`) omits `error_message` and persists only `error_class` (sliced to 100 chars) — so it isn't stored in the DB today — but the raw string still leaves the machine in the POST body and would be persisted if that field is ever added server-side. Free-text telemetry fields should run through the redaction engine before upload.

## Impact

A user glancing at the shield icon / dashboard sees "0 attacks" and concludes there's no prompt-injection activity, when the real state is "backend errored" or "`jq` not installed." For a security indicator, the safe default is "unknown / degraded," not "0 / clear." (This is the same fail-open pattern the project's own `9cc41b71` commit — "4 security guards failing open" — was correcting elsewhere.)

## Suggested fix

- `community-pulse`: on `catch`, return a non-200 status or include an explicit `{ "error": true }` / `{ "stale": true }` flag instead of all-zero aggregates.
- Dashboards: render "data unavailable" distinctly from "0 attacks"; when `jq` is missing, say so rather than reporting 0.
- Redact (or drop) `error_message` before it is written to the synced payload.

jbetala7:
## Problem

The community **security** dashboard fails **open** on a security-signaling surface. A backend error, an unreachable backend, or a host without `jq` all render a reassuring:

```
Attacks detected last 7 days: 0
  (No attack attempts reported by the community yet. Good news.)
```

This is **indistinguishable from a genuine zero**. A user glances at the shield, sees "0 / Good news", and concludes there is no prompt-injection activity when the real state is "backend errored" or "`jq` not installed". For a security indicator the safe default is "unknown / degraded", not "0 / clear". This is the same fail-open class the project's own `9cc41b71` ("4 security guards failing open") corrected elsewhere.

Reproduce on `main` (backend unreachable):

```bash
GSTACK_SUPABASE_URL="http://127.0.0.1:9/x" GSTACK_SUPABASE_ANON_KEY=k bash bin/gstack-security-dashboard
# -> "Attacks detected last 7 days: 0 / Good news."   (wrong: backend is down)
```

## Fix — fail closed at both layers

**`supabase/functions/community-pulse/index.ts`** — the `catch` returned all-zero aggregates with HTTP 200. It now tags the body with an explicit `error: true` flag (top level and inside `security`) so consumers can distinguish an internal failure (DB outage, malformed cache, query error) from a real zero. Status stays `200` so existing `curl -sf` clients still receive the body and can read the flag.

**`bin/gstack-security-dashboard`** — now captures curl's exit code and detects the degraded states up front:
- backend unreachable / empty / `{}`
- `"error": true` flag from the edge function
- `security` section missing from the response
- `jq` not installed (the old grep fallback silently emitted zeros on the nested arrays)

In any degraded state it renders:
- human: `Attacks detected last 7 days: unknown  (Degraded: <reason>. This is NOT a confirmed zero...)`
- `--json`: `{"error":"security_data_unavailable","security":null}` instead of a fake all-zero security object.

A **genuine** zero still reads as `0 / Good news` (no over-degrade). The count is now read via an exact `jq` path (`.security.attacks_last_7_days`) instead of a two-stage grep that also mis-matched the `7` inside `last_7_days` (a fixture of `4` printed `7` on `main`).

Scope note: the latent telemetry `error_message` redaction concern raised in the issue is a separate subsystem and is intentionally left for a follow-up.

## Testing

New `test/security-dashboard.test.ts` (5 tests), each **proven to fail on `main`** and pass with this change:

| Scenario | Asserts |
|---|---|
| Unreachable backend | "unknown / Degraded", **not** "0 / Good news" |
| Explicit `error:true` | "unknown", "internal error" |
| `--json` degraded | `error: "security_data_unavailable"`, `security: null`, no `"attacks_last_7_days":0` |
| Genuine zero | "0 / Good news", **not** degraded |
| Healthy non-zero | exact count reported verbatim |

```
bun test test/security-dashboard.test.ts   # 5 pass
bun test test/skill-validation.test.ts      # 337 pass (unchanged)
```

Fixtures are served to the bin via a `file://` URL (`GSTACK_SUPABASE_URL`), so the tests need no network and no live Supabase.

Fixes #1947

### DIFF
--- a/test/security-dashboard.test.ts
+++ b/test/security-dashboard.test.ts
@@ -1,0 +1,143 @@
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { execSync, ExecSyncOptionsWithStringEncoding } from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+
+// Regression tests for #1947: the community security dashboard must FAIL CLOSED.
+// A backend error, an unreachable backend, or unparseable data must render
+// "unknown", never a reassuring "0 attacks / Good news" — reporting 0
+// when the real state is unknown is a fail-open on a security-signaling surface
+// (the same class the project's "4 security guards failing open" wave corrected).
+//
+// The dashboard fetches `${GSTACK_SUPABASE_URL}/functions/v1/community-pulse` via
+// curl. Tests put a fake curl first on PATH: when a fixture exists it writes the
+// body and prints HTTP 200; when no fixture exists it prints HTTP 000.
+
+const ROOT = path.resolve(import.meta.dir, '..');
+const BIN = path.join(ROOT, 'bin', 'gstack-security-dashboard');
+const HAS_JQ = (() => {
+  try {
+    execSync('command -v jq', { stdio: 'ignore' });
+    return true;
+  } catch {
+    return false;
+  }
+})();
+
+let tmpDir: string;
+let fixtureFile: string;
+let fakeBin: string;
+
+beforeEach(() => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'gstack-secdash-'));
+  const fnDir = path.join(tmpDir, 'functions', 'v1');
+  fs.mkdirSync(fnDir, { recursive: true });
+  fixtureFile = path.join(fnDir, 'community-pulse');
+  fakeBin = path.join(tmpDir, 'bin');
+  fs.mkdirSync(fakeBin, { recursive: true });
+  const curl = path.join(fakeBin, 'curl');
+  fs.writeFileSync(
+    curl,
+    `#!/bin/sh
+out=""
+while [ "$#" -gt 0 ]; do
+  if [ "$1" = "-o" ]; then
+    shift
+    out="$1"
+  fi
+  shift
+done
+if [ -f "${fixtureFile}" ]; then
+  cp "${fixtureFile}" "$out"
+  printf "200"
+  exit 0
+fi
+: > "$out"
+printf "000"
+exit 7
+`,
+  );
+  fs.chmodSync(curl, 0o755);
+});
+
+afterEach(() => {
+  fs.rmSync(tmpDir, { recursive: true, force: true });
+});
+
+function run(args: string, env: Record<string, string>): string {
+  const execOpts: ExecSyncOptionsWithStringEncoding = {
+    env: {
+      ...process.env,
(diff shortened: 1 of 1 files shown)


jbetala7:
Rebased onto current main (9fd03fae). The functional fail-closed behavior from this PR has already been absorbed into main, so the rebased branch now keeps only regression coverage for #1947 against the current response contract: unknown/backend_error for unavailable data and status ok for authoritative zero/nonzero fixtures.

Validation: ~/.bun/bin/bun test test/security-dashboard.test.ts - 5 pass, 0 fail.

time-attack:
@16francej Please close this PR. The security dashboard already fails closed on backend errors through #2077. This PR now contains only duplicate tests for behavior that is already on main.

PR #2029, opened 2026-06-16
v1.59.0.0 feat(learn): semantic dedup for the learnings store (/learn refine)

tonyjzhou:
## What

`gstack-learnings-search` collapses learnings by exact `key+type` only, so the same
lesson relogged under a different key accumulates forever. This adds
`bin/gstack-learnings-refine` and a `/learn refine` subcommand that find *semantic*
near-duplicates and merge them, keeping the highest-confidence survivor and unioning
their `files` lists. It is the "grow-and-refine" step the log/search pair never had.

## How

Deterministic and file-only, so it runs with gbrain off (no embeddings DB, no model,
no network). It blends word-level TF-IDF cosine (precision: rare shared terms carry
the signal) with character-trigram cosine (recall: survives rewording and `sessions`
vs `session`). Neither signal alone catches a reworded duplicate. The larger of the
two does.

Safe by default:

- Dry-run unless `--apply`. `--apply` writes atomically, leaves a `.bak`, and never
  rewrites a file that has an unparseable line.
- A `--review` band surfaces gray-zone pairs in `[--review-floor, --sim)` for manual
  inspection and never auto-merges them. Because `--apply` deletes a row, the 0.70
  default bar is tuned for precision and the REVIEW band catches the rest without
  deleting on a guess.
- Scope mirrors the sibling tools: current project via `gstack-slug` by default,
  `--slug SLUG`, or `--all`. Survivor selection uses the same confidence-decay model
  as `gstack-learnings-search`.

## /learn refine

New subcommand in the learn skill: runs the dry-run, presents the auto-mergeable and
REVIEW bands, and applies only after AskUserQuestion confirmation.

## Tests

`test/gstack-learnings-refine.test.ts` adds 10 tests (mirroring the
`gstack-learnings-search` harness): near-dup clustering, exact-dup compaction,
cross-type gating, `--apply` (file union, confidence bump, idempotency, parse-error
safety), the `--review` band, and `--min-entries` lazy-on-overflow.

Every suite this change touches is green: learnings, skill-validation, gen-skill-docs,
and the package.json-to-VERSION sync. The remaining failures in a full local
`bun test test/` run are pre-existing and environmental (resolve-user-slug sandbox
timeouts, an iOS XCTest-toolchain test, a hook-install test). They fail identically on
`main` (c7ae6320) without this branch.

## Numbers

On the test fixtures a reworded duplicate scores 0.58 on words alone, under any
threshold safe enough to use, but 0.77 blended once the character-trigram signal lifts
it. Unrelated same-domain pairs stay under 0.20. Reproduce with
`bun test test/gstack-learnings-refine.test.ts` and
`gstack-learnings-refine --all --review --debug-scores`.

VERSION 1.58.1.0 -> 1.59.0.0 (MINOR, new capability), with a CHANGELOG entry.

## Note for CI

This is a fork PR, so the eval and E2E jobs will not receive base-repo secrets and
will fail on empty-env auth. The free unit tier (`bun test`, skill validation,
gen-skill-docs, the new dedup tests) runs without secrets. Happy to have the branch
re-pushed to a base-repo branch if you would rather run the full eval suite.


time-attack:
@tonyjzhou Redesign as schema-lossless, dry-run-first conservative refinement with stable IDs, containment, locking, unique durable atomic writes, recovery, and adversarial chain/concurrency probes.

PR #2031, opened 2026-06-17
fix(gbrain): pass --confirm-destructive on drift re-register (#1985)

### LINKED ISSUE #1985: /sync-gbrain code stage fails on existing sources (gbrain ≥0.42 --confirm-destructive guard); use reindex-code in place
## Summary

`/sync-gbrain`'s **code stage fails for any already-registered source** on gbrain `>= 0.42`. Both `--full` and `--code-only` (and plain incremental) error with a destructive-delete guard. The orchestrator (`gstack-gbrain-sync.ts`) refreshes a code source by **removing and re-adding** it, but gbrain 0.42 now gates `gbrain sources remove` behind `--confirm-destructive`. The orchestrator doesn't pass that flag, so the remove is refused and the whole code stage errors.

The source is **not** deleted (the guard blocks the remove), and the memory + brain-sync stages still succeed — but the code index never gets refreshed through the orchestrator. The only working refresh path is calling `gbrain reindex-code` directly.

## Environment

- gstack: **1.57.10.0**
- gbrain: **0.42.40.0**
- Engine: PGLite (local-stdio MCP), single-machine macOS (darwin)
- Source already registered (created by a prior `/setup-gbrain` / `--full` run)

## Repro

1. `/setup-gbrain` → local PGLite, repo policy `read-write` (registers source `gstack-code-<repo>` with N pages). This first run works because the source doesn't exist yet (register = pure add).
2. Make any change (or none) and run `/sync-gbrain --full` **or** `/sync-gbrain --code-only` again.
3. Code stage errors; memory + brain-sync still pass.

## Actual output

```
gstack-gbrain-sync (full):
  ERR   code         source registration failed: gbrain sources remove gstack-code-autoleadai-6b2cc40d failed:
⚠️  This will permanently delete: 768 pages, 6,786 chunks, 6,786 embeddings

To proceed, pass --confirm-destructive (or use soft-delete: gbrain sources archive gstack-code-autoleadai-6b2cc40d).
To preview without side effects: --dry-run
 (1.9s)
  OK    memory       gbrain import: 3 imported, 112 unchanged, 0 failed (3.7s)
  OK    brain-sync   curated artifacts pushed (2.1s)
  SKIP  dream        call-graph state unknown (doctor unavailable)
  3 ok, 1 error, 0 skipped
```

Identical `ERR code` for `--code-only`.

## Root cause

The orchestrator's code stage re-registers an existing source via `gbrain sources remove <id>` + `gbrain sources add`, instead of refreshing in place. gbrain `>= 0.42` added a `--confirm-destructive` guard on `sources remove` (it would drop all pages/chunks/embeddings). The orchestrator predates that guard and doesn't pass the flag, so re-registration of any existing source fails.

This is distinct from #1734 (autopilot destructive auto-recovery) — here it's the orchestrator's own remove+add registration path tripping gbrain's new safety guard, on a non-autopilot manual sync.

## Impact

`/sync-gbrain --full` and `--code-only` can no longer refresh the code index for a repo that was already indexed once. Users on gbrain 0.42+ get a confusing "permanently delete" error on a routine re-sync. The index is safe (remove is blocked), but the documented refresh verb is broken.

## Workaround (confirmed working)

Refresh in place, no remove:

```
gbrain reindex-code --source <id> --yes
```

For memory + artifacts only: `/sync-gbrain --no-code` (those stages are unaffected).

## Suggested fix

In `gstack-gbrain-sync.ts`'s code stage, for an **existing** source skip the remove+add and refresh in place with `gbrain reindex-code --source <id> --yes` (already the path `--full` advertises). Only `sources add` for a genuinely new source. If a true re-register is ever needed, prefer `gbrain sources archive` or pass `--confirm-destructive` explicitly with a clear log line — never an implicit destructive remove on a routine sync.

jbetala7:
## Problem

Following `/sync-gbrain`, the **code stage fails for any already-registered source whose path has drifted** on gbrain `>= 0.42`:

```
gstack-gbrain-sync (full):
  ERR   code   source registration failed: gbrain sources remove gstack-code-<id> failed:
⚠️  This will permanently delete: ... To proceed, pass --confirm-destructive
```

The memory and brain-sync stages still pass, so the code index silently stops refreshing while the run reports a partial success.

## Root cause

`ensureSourceRegistered()` (`lib/gbrain-sources.ts`) handles the match-but-different-path case by removing the old source and re-adding it at the new path (gbrain has no `sources update`). The remove was issued as:

```ts
gbrain sources remove <id> --yes
```

gbrain `>= 0.42` now gates `sources remove` behind `--confirm-destructive`; `--yes` alone no longer suppresses the data-loss prompt. The remove fails, `ensureSourceRegistered()` throws, and the orchestrator reports `source registration failed`, aborting the code stage.

The orchestrator's own `safeSourcesRemove()` (`bin/gstack-gbrain-sync.ts`) already passes `--confirm-destructive` — this just brings the lib helper in line with that established convention.

## Fix

Pass `--confirm-destructive` on the drift re-register remove (keeping `--yes` for older gbrain).

## Testing

- Extended the fake `gbrain` shim in `test/gbrain-sources.test.ts` to simulate the gbrain `>= 0.42` guard (a `sources remove` without `--confirm-destructive` exits 1 with the real error text).
- Updated the existing drift re-register assertion to expect `--confirm-destructive`.
- Added a regression test asserting the drift path no longer throws and ends with a single source at the new path.
- Both tests **fail on `main`** with the exact `To proceed, pass --confirm-destructive` error and **pass with the fix**. Full `test/gbrain-sources.test.ts` (10) + adjacent `gbrain-sources-parse` / `gstack-gbrain-sync` / `gstack-gbrain-source-wireup` suites (64) green.

Fixes #1985

### DIFF
--- a/lib/gbrain-sources.ts
+++ b/lib/gbrain-sources.ts
@@ -124,7 +124,7 @@ export function probeSource(id: string, env?: NodeJS.ProcessEnv): SourceState {
  * Behavior:
  *   - status=absent  → `gbrain sources add <id> --path <path> [--federated]`, returns changed=true.
  *   - status=match + same path → no-op, returns changed=false.
- *   - status=match + different path → `sources remove` + `sources add`, returns changed=true.
+ *   - status=match + different path → `sources remove --confirm-destructive` + `sources add`, returns changed=true.
  *     (Skip when reregister_on_drift=false; returns changed=false.)
  *
  * Caller is responsible for catching errors. The function uses withErrorContext for
@@ -157,8 +157,15 @@ export async function ensureSourceRegistered(
     }
 
     // For drift, remove first.
+    //
+    // #1985: gbrain >= 0.42 gates `sources remove` behind --confirm-destructive
+    // (`--yes` alone no longer suppresses the data-loss prompt). Without it the
+    // remove fails with "To proceed, pass --confirm-destructive", which surfaces
+    // as "source registration failed" and aborts the whole /sync-gbrain code
+    // stage for any source that has drifted to a new path. This matches the
+    // flag the orchestrator's own safeSourcesRemove() already passes.
     if (state.status === "drift") {
-      const rm = spawnSync("gbrain", ["sources", "remove", id, "--yes"], {
+      const rm = spawnSync("gbrain", ["sources", "remove", id, "--yes", "--confirm-destructive"], {
         encoding: "utf-8",
         timeout: 30_000,
         env,
--- a/test/gbrain-sources.test.ts
+++ b/test/gbrain-sources.test.ts
@@ -32,8 +32,11 @@ interface FakeGbrainSetup {
  * Build a temp dir with a fake `gbrain` shell script on PATH. The fake honors:
  *   gbrain sources list --json     → cat $STATE_PATH
  *   gbrain sources add <id> --path <p> [--federated]  → append to state, log
- *   gbrain sources remove <id> --yes                  → drop from state, log
- *   gbrain --version                                  → echo "gbrain 0.25.1"
+ *   gbrain sources remove <id> --confirm-destructive  → drop from state, log
+ *                                                       (#1985: remove WITHOUT
+ *                                                       --confirm-destructive
+ *                                                       fails like gbrain >= 0.42)
+ *   gbrain --version                                  → echo "gbrain 0.42.40.0"
  * Anything else exits 1.
  */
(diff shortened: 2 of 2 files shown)


time-attack:
autonomously destry your local gbrain before confirming the replacement works
no rollback

close

time-attack:
@16francej — recommend closing this PR because it does not safely resolve #1985. It enables destructive replacement without the current source safeguards or verification that replacement succeeded, so a later failure can leave existing data removed; the included coverage does not demonstrate preservation or rollback.

PR #1638, opened 2026-05-21
design: bump per-image OpenAI timeout from 120s to 240s

ionianFury:
## Summary

The hardcoded 120s `setTimeout(() => controller.abort(), 120_000)` in `design/src/{generate,iterate,variants,evolve}.ts` fires too often against `gpt-image-1` at the default `quality: "high"` + `size: "1536x1024"`. Real session today (mid-2026): all 3 `design variants` calls timed out at exactly 120s while the OpenAI API was actively processing the request. Direct `curl --max-time 240` to the same endpoint with the same payload returned 200 in 33s.

```
✗ variant-A.png: Timeout (120s)
✗ variant-B.png: Timeout (120s)
✗ variant-C.png: Timeout (120s)
0/3 variants generated (123.0s)
```

Bumping each call to 240s.

## Why 240s

- Observed gpt-image-1 wall times at `high + 1536x1024`: typically 60-180s, occasionally 200s+ under provider load.
- 240s gives ~2x headroom over the typical worst-case while staying under the threshold where users walk away.
- Lower options I considered: 180s would have caught most spikes but still cliffs in the tail. 240s is the sweet spot.

## Sites touched

- `design/src/generate.ts:40`
- `design/src/iterate.ts:85, 133`
- `design/src/variants.ts:61`
- `design/src/evolve.ts:55`

Each carries a brief comment pointing back to `generate.ts` for the rationale.

## What I deliberately did NOT change

- **Quality default stays `high`.** That's what produces the polished mockups skill users expect. The timeout was the cliff, not the quality.
- **No shared-constant refactor.** Could promote to `design/src/constants.ts` or a `GSTACK_DESIGN_TIMEOUT_MS` env var, but I'm keeping this PR small. Happy to follow up if you want the env-var or shared-constant change in a separate PR.

## Verification

```
bun install               # 230 packages, no issues
bun build --compile design/src/cli.ts --outfile /tmp/test-bin
# [120ms] compile /tmp/test-bin → succeeds
```

No test changes — the timeout itself isn't easily unit-testable without injecting a stub clock or a slow mock fetch, and the existing `design/test/` suite doesn't cover it today.

## Repro

```sh
$D variants --brief "<any dense brief>" --count 3 --output-dir /tmp/
# Reliably reproduces 120s timeouts on busy OpenAI hours
```

(Filed by ionianFury — happened to me mid-`/plan-design-review` today.)

PR #2033, opened 2026-06-17
fix(design/variants): guard --count against NaN/zero/negative input

### LINKED ISSUE #2032: design variants: --count silently generates zero variants on NaN/0/negative input
### Observed problem

`design variants --count <bad value>` silently produces **zero** variants and emits a confusing JSON result instead of failing or clamping to a sane value.

The CLI parses the flag with `parseInt` and no validation (`design/src/cli.ts:200`):

```ts
count: flags.count ? parseInt(flags.count as string) : 3,
```

`variants()` then only clamps the **upper** bound (`design/src/variants.ts:156`):

```ts
const count = Math.min(options.count, 7); // Cap at 7 style variations
```

There is no lower-bound or NaN guard, so non-numeric / zero / negative counts flow straight through.

### Current behavior on `main` (c7ae6320)

Reproduced against the real clamp + JSON-emit path:

| `--count` | parsed | clamped | variants generated | JSON `count` field |
|-----------|--------|---------|--------------------|--------------------|
| `abc`     | `NaN`  | `NaN`   | 0                  | `null`             |
| `0`       | `0`    | `0`     | 0                  | `0`                |
| `-2`      | `-2`   | `-2`    | 0                  | `-2`               |
| `3`       | `3`    | `3`     | 3                  | `3`                |
| `10`      | `10`   | `7`     | 7                  | `7`                |

A typo like `--count e` (NaN) makes the command exit "successfully" having created nothing, and the structured stdout reports `"count": null` — actively misleading for any caller parsing that JSON.

### Expected behavior

A bad `--count` should resolve to a valid number in the supported range, never a silent no-op:
- non-finite (`NaN`) -> fall back to the default of 3
- numeric but out of range -> clamp to `[1, 7]`

So `--count abc` -> 3, `--count 0`/`-2` -> 1, `--count 10` -> 7 (unchanged).

- Open PRs touching `design/`: #1774 (timeout label only), #1638 (timeout 240s bump), #1528 (`--api-timeout` flag + `parseApiTimeoutMs` guard — validates the *timeout* flag, not `--count`), #1737 (variant preservation). None guard `--count`.
- `gh pr list --search "design count"` -> no matches; `--search "variants"` -> only the timeout PRs above.
- No open/closed issue covers the `--count` validation gap.

### Candidate fix shape

Normalize the count in `variants.ts` via a small exported, unit-testable helper (mirrors the `parseApiTimeoutMs` helper pattern proposed in #1528):

```ts
export function normalizeVariantCount(count: number): number {
  if (!Number.isFinite(count)) return 3;        // unparseable -> default
  return Math.max(1, Math.min(Math.trunc(count), 7)); // clamp to [1, 7]
}
```

Use it where `Math.min(options.count, 7)` is today. Add a regression test asserting the table above.

jbetala7:
## Problem

`design variants --count <bad value>` silently produces **zero** variants and emits a misleading JSON result.

`--count` is parsed with `parseInt` and never validated (`design/src/cli.ts:200`), and `variants()` only clamped the **upper** bound (`design/src/variants.ts`):

```ts
const count = Math.min(options.count, 7); // Cap at 7 style variations
```

So bad inputs flow through. Measured against the real clamp + JSON-emit path on `main` (c7ae6320):

| `--count` | generated | JSON `count` |
|-----------|-----------|--------------|
| `abc`     | 0         | `null`       |
| `0`       | 0         | `0`          |
| `-2`      | 0         | `-2`         |
| `3`       | 3         | `3`          |
| `10`      | 7         | `7`          |

`Math.min(NaN, 7)` is `NaN`, so `for (i = 0; i < NaN; i++)` never runs and the command exits "successfully" having created nothing, reporting `"count": null`. A single typo (`--count e`) is enough.

## Root cause

No lower-bound or NaN guard on the requested count — only the `<= 7` cap.

## Fix

Add a small exported, pure `normalizeVariantCount` helper and use it in place of the bare cap:

```ts
export function normalizeVariantCount(count: number): number {
  if (!Number.isFinite(count)) return 3;            // unparseable -> default
  return Math.max(1, Math.min(Math.trunc(count), 7)); // clamp to [1, 7]
}
```

- non-finite (`NaN`/`Infinity`) -> default 3
- `0` / negative -> 1
- `> 7` -> 7 (unchanged)
- in-range -> unchanged

This mirrors the exported-helper validation shape proposed for `--api-timeout` in #1528, but is independent of it (different flag).

## Testing

`bun test design/test/variants-retry-after.test.ts` -> 10 pass / 0 fail (5 existing + 5 new). Full `bun test design/test/` passes. The helper is pure, so the regression is pinned without an `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`.

## Scope / collision check

- Touches only `design/src/variants.ts` + its test. No behavior change for valid counts.
- Open design PRs are about timeouts and unrelated concerns (#1774 timeout label, #1638 240s bump, #1528 `--api-timeout` guard, #1737 variant preservation) — none guard `--count`. Searched `variants`, `design count`, `design timeout`; no overlap.

Fixes #2032

### DIFF
--- a/design/src/variants.ts
+++ b/design/src/variants.ts
@@ -19,6 +19,23 @@ export interface VariantsOptions {
   viewports?: string; // "desktop,tablet,mobile" — generates at multiple sizes
 }
 
+/**
+ * Normalize a requested variant count to the supported range.
+ *
+ * `--count` reaches us via `parseInt`, which yields `NaN` for non-numeric
+ * input and happily passes zero/negative values through. The generation loop
+ * only capped the upper bound, so any of those silently produced zero variants
+ * and emitted `"count": null`/`0`/negative in the JSON result. Clamp to the
+ * supported `[1, 7]` range; fall back to the default of 3 when the value is not
+ * a finite number (an unparseable flag, not an explicit out-of-range request).
+ *
+ * Exported for direct unit testing.
+ */
+export function normalizeVariantCount(count: number): number {
+  if (!Number.isFinite(count)) return 3;
+  return Math.max(1, Math.min(Math.trunc(count), 7));
+}
+
 const STYLE_VARIATIONS = [
   "", // First variant uses the brief as-is
   "Use a bolder, more dramatic visual style with stronger contrast and larger typography.",
@@ -153,7 +170,7 @@ export async function variants(options: VariantsOptions): Promise<void> {
     return;
   }
 
-  const count = Math.min(options.count, 7); // Cap at 7 style variations
+  const count = normalizeVariantCount(options.count); // Clamp to [1, 7] style variations
   const size = options.size || "1536x1024";
 
   console.error(`Generating ${count} variants...`);
--- a/design/test/variants-retry-after.test.ts
+++ b/design/test/variants-retry-after.test.ts
@@ -2,7 +2,7 @@ import { describe, test, expect, beforeEach, afterEach } from "bun:test";
 import fs from "fs";
 import os from "os";
 import path from "path";
-import { generateVariant } from "../src/variants";
+import { generateVariant, normalizeVariantCount } from "../src/variants";
 
 // 1x1 transparent PNG, base64 — valid bytes that fs.writeFileSync can write.
 const TINY_PNG_BASE64 =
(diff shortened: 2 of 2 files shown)


time-attack:
@16francej ensures numers are real. merge

PR #2034, opened 2026-06-17
fix(skill-check): skip missing template validation for skipped primar…

dammanhdungvn:
…y host skills

### DIFF
--- a/scripts/skill-check.ts
+++ b/scripts/skill-check.ts
@@ -13,6 +13,7 @@ import { discoverTemplates, discoverSkillFiles } from './discover-skills';
 import * as fs from 'fs';
 import * as path from 'path';
 import { execSync } from 'child_process';
+import { claude } from '../hosts/index';
 
 const ROOT = path.resolve(import.meta.dir, '..');
 const ROOT_REALPATH = fs.realpathSync(ROOT);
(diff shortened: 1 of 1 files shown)


dammanhdungvn:
no comment

fiestyelf:
## Logic bug: skip guard is unreachable for missing templates

The guard fires **after** the `!fs.existsSync(tmplPath)` check, so skipped skills with no template still print `⚠️ — no template` and the skip block is never reached.

**Current (broken order):**
```ts
if (!fs.existsSync(tmplPath)) {
  console.log(`  ⚠️  ${output.padEnd(30)} — no template`);
  continue;  // exits here before skip guard
}
const dirName = path.dirname(tmpl);
if (dirName !== '.' && claudeSkipSkills.has(dirName)) {  // dead for missing templates
  ...
}
```

**Fix — move skip guard above the existence check:**
```ts
const dirName = path.dirname(tmpl);
if (dirName !== '.' && claudeSkipSkills.has(dirName)) {
  console.log(`  ✅ ${tmpl.padEnd(30)} → (skipped for primary host ${claude.displayName})`);
  continue;
}
if (!fs.existsSync(tmplPath)) {
  console.log(`  ⚠️  ${output.padEnd(30)} — no template`);
  continue;
}
```

Two other minor nits while here:
- `|| []` → `?? []` (nullish coalescing is safer for optional arrays)
- Three separate `import` statements from `'../hosts/index'` — can consolidate to one: `import { claude, getExternalHosts, ALL_HOST_CONFIGS } from '../hosts/index';`

time-attack:
@dammanhdungvn Please rewrite the validation so skill:check does not require claude/SKILL.md when the Claude host intentionally skips that primary-host output.

PR #1502, opened 2026-05-14
fix(setup): mirror support files + asset dirs alongside SKILL.md in link_claude_skill_dirs

NikhileshNanduri:
## Summary

Closes #1499.

`link_claude_skill_dirs` in `./setup` previously only symlinked `SKILL.md` into each skill directory. Skills like `/review` reference sibling files — `checklist.md`, `greptile-triage.md`, `design-checklist.md`, `TODOS-format.md`, and `specialists/` — via the `.claude/skills/review/` path. Those files were never linked, so `/review` hit its own STOP point (`"If the file cannot be read, STOP and report the error"`) on every global install. 20 releases since the regression was introduced.

This is **Option B** from the issue — the durable fix. It teaches the linker to mirror all sibling assets, not just SKILL.md. No SKILL.md.tmpl path hacks needed.

## What changed

**`setup` — `link_claude_skill_dirs()`**: after the existing `SKILL.md` symlink, two new loops:

1. **Supporting `.md` files** — iterates `$gstack_dir/$dir_name/*.md`, skips `SKILL.md` (already linked), links the rest (`checklist.md`, `greptile-triage.md`, `TODOS-format.md`, etc.)
2. **Asset directories** — iterates `$gstack_dir/$dir_name/*/`, skips build/source dirs (`dist`, `src`, `test`, `tests`, `scripts`, `node_modules`), links the rest (`specialists/`, `bin/`, `references/`, `templates/`, `migrations/`, etc.)

Both loops are idempotent (`rm` then `ln -snf`) — `./setup` re-runs cleanly over existing installs.

**`test/gen-skill-docs.test.ts`**: two new unit tests that verify the linker body contains the new loops and the correct exclusion list.

## Affected skills (verified by functional test)

| Skill | Previously missing | Now linked |
|-------|--------------------|------------|
| `review/` | `checklist.md`, `greptile-triage.md`, `design-checklist.md`, `TODOS-format.md`, `specialists/` | ✅ All 5 |
| `qa/` | `references/`, `templates/` | ✅ Both |
| `careful/` | `bin/` | ✅ |
| `freeze/` | `bin/` | ✅ |
| `plan-devex-review/` | `dx-hall-of-fame.md` | ✅ |
| `cso/` | `ACKNOWLEDGEMENTS.md` | ✅ |
| `setup-gbrain/` | `memory.md` | ✅ |
| `gstack-upgrade/` | `migrations/` | ✅ |

[...]

PR #2036, opened 2026-06-17
fix(setup): link all skill aux files next to SKILL.md, not just sections/

adamascode:
## Problem

`link_claude_skill_dirs()` registers each top-level skill dir with only `SKILL.md` (plus `sections/` since the v2 carve). But a skill's `SKILL.md` reads its *other* aux files by a path **relative to its own dir**:

- `/review` → `Read .claude/skills/review/checklist.md`, dispatches `review/specialists/*.md`, reads `greptile-triage.md` / `design-checklist.md`
- carved skills → `Read sections/<name>.md` (already handled)

The non-`sections/` aux files stay behind in `gstack/<skill>/`, so the registered target dir holds only `SKILL.md` and those reads 404. `/review` aborts at *"Read `.claude/skills/review/checklist.md` … STOP and report the error."* Any aux-bearing skill breaks the same way.

It's most acute on **Windows**, where `_link_or_copy` makes `SKILL.md` a real **copy** whose base dir is the target (not the gstack source), so nothing resolves back. The existing `sections/` special-case shows this 404 class was already known — it just wasn't generalized.

## Fix

Generalize the `sections/`-only handling: route **every** sibling aux entry (except `SKILL.md` and `node_modules`) through `_link_or_copy` — symlink on Unix, fresh copy on Windows that re-syncs on every `./setup`. Subsumes the `sections/` case.

## Testing (Windows / Git Bash)

Deleted `~/.claude/skills/review/checklist.md`, re-ran `./setup`, confirmed it was restored and `~/.claude/skills/review/` then mirrored `gstack/review/` exactly (checklist + all 7 specialists). `/review` runs its full pipeline again. `bash -n setup` clean.

### DIFF
--- a/setup
+++ b/setup
@@ -569,14 +569,22 @@ link_claude_skill_dirs() {
       # Validate target isn't a symlink before creating the link
       if [ -L "$target/SKILL.md" ]; then rm "$target/SKILL.md"; fi
       _link_or_copy "$gstack_dir/$dir_name/SKILL.md" "$target/SKILL.md"
-      # Link the sections/ subdir for carved skills (v2 plan T9). The prefixed
-      # Claude skill dir otherwise holds only SKILL.md, so a runtime
-      # "Read sections/<name>.md" 404s. Route through _link_or_copy so Windows
-      # gets a fresh copy (and re-copies on every ./setup, refreshing staleness).
-      if [ -d "$gstack_dir/$dir_name/sections" ]; then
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej — recommend closing this PR. It is superseded by #1502 because copying every sibling asset would also install templates, source, tests, scripts, and other development directories. Credit #2036 for the Windows refresh intent, and retain it through an explicit runtime-asset manifest with flat, prefixed, idempotent, and stale-removal coverage.

PR #2037, opened 2026-06-18
fix(retro): language-agnostic test detection + anti-fabrication guard

SholtoMc:
## Problem

While verifying a generated weekly retro against git history, two headline metrics turned out to be fabricated:

- **"Bootstrap commit #2203 — 5,165 files · 1.4M lines · 1,455 test files."** The cited commit was actually a 2-file, 187-line docs PR. No bootstrap had occurred (the repo was 8+ months / 1,200+ commits old).
- **"0 new tests added."** In reality ~28 test files were added that week.

## Root cause

The `/retro` Step-1 data collection counts tests with JS/TS-only patterns:

```bash
# "Test files changed in window"
git log ... --name-only | grep -E '\.(test|spec)\.' | sort -u | wc -l
```

`\.(test|spec)\.` only matches `foo.test.ts` / `foo.spec.js`. It returns **0** for:
- Python `test_*.py` / `*_test.py`
- Terraform `*.tftest.hcl`
- Bats `*.bats`
- path-based `tests/…`

The total-test `find` (command 10) likewise missed the `test_*.py` **prefix** convention. So the generator saw "N total tests exist" but "0 added this period," and reconciled that contradiction by **inventing a non-existent "bootstrap commit"** with made-up file/LOC figures — there was no guard forcing per-commit numbers to come from the commit itself.

## Fix

- **Commands 10 / 12 / 13** now use a language-agnostic test pattern (Python / JS-TS / Ruby / Terraform / Bats / path-based), exclude vendored dirs, and split tests **ADDED** (`--diff-filter=A`) from **TOUCHED**. Command 10 uses `git ls-files` (respects `.gitignore`).
- **New Step 1.5 "plausibility guard":** per-commit figures must come from `git show --shortstat <hash>`; never invent a "bootstrap/foundation/initial-import" narrative; reconcile count contradictions by re-checking command 12, not by narrating an unverified event.
- Applied to both `retro/SKILL.md.tmpl` (source) and `retro/SKILL.md` (runtime).
- `test/regression-retro-test-detection.test.ts` locks in both invariants (21 tests).

## Verification

On the affected repo's window the patched commands return **28 tests added / 47 touched** (was `0`). `bun test test/regression-retro-test-detection.test.ts` → 21/21 pass; existing `regression-1624-retro-stale-base.test.ts` → 13/13 pass.

### DIFF
--- a/retro/SKILL.md
+++ b/retro/SKILL.md
@@ -1,5 +1,5 @@
 ---
-name: retro
+name: gstack-retro
 preamble-tier: 2
 version: 2.0.0
 description: Weekly engineering retrospective. (gstack)
@@ -972,7 +972,8 @@ git log origin/<default> --since="<window>" --format="%H|%aN|%ae|%ai|%s" --short
 
 # 2. Per-commit test vs total LOC breakdown with author
 #    Each commit block starts with COMMIT:<hash>|<author>, followed by numstat lines.
-#    Separate test files (matching test/|spec/|__tests__/) from production files.
+#    Separate test files from production files using the language-agnostic TEST pattern
+#    below (commands 10/12) — NOT just JS/TS. Python test_*.py, *.tftest.hcl, *.bats count.
 git log origin/<default> --since="<window>" --format="COMMIT:%H|%aN" --numstat
 
 # 3. Commit timestamps for session detection and hourly distribution (with author)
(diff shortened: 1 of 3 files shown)


jbetala7:
Overlap heads-up: this shares its command-10 change with the older, issue-linked #2013 (Closes #1999). Both rewrite the same glob at `retro/SKILL.md:1000` and the mirrored line in `retro/SKILL.md.tmpl` — JS/TS-only `find … | grep -v node_modules` → polyglot pattern + vendored-dir exclusion — so whichever merges second will conflict on those exact lines.

The deltas here are real and worth keeping: this PR additionally fixes commands 12/13 (`retro/SKILL.md:1009`, still `grep -E '\.(test|spec)\.'` on `main`, which #2013 doesn't touch) and adds the Step 1.5 plausibility guard + regression test — the headline anti-fabrication value. #2013 is narrower but carries the issue link.

Suggestion: since this is a superset, consider adding `Closes #1999` and crediting #2013, so the maintainer can land this one and close the narrow PR without dropping the issue. If #2013 lands first instead, this just needs a rebase to drop the duplicate command-10 / vendored-exclusion hunk.

time-attack:
Thanks for this. The blocker is incomplete language coverage, fragile root-level exclusions, duplicated detection patterns, and tests that do not exercise the real shell and Git path. Please rebase, incorporate the Java and exclusion coverage from PR #2013, centralize the taxonomy, and add Git-history tests.

Issue #1163, opened 2026-04-23
Codex generated skills still carried Claude defaults


giaphutran12:
@garrytan heads up: I opened PR #1162 for this Codex-host issue: https://github.com/garrytan/gstack/pull/1162

Problem: the Codex generated skill output under `.agents/skills/gstack-*` existed, but it was still carrying Claude-host behavior in important places:

- `MODEL_OVERLAY: claude` showed up in Codex-generated skills.
- The generated body injected the Claude model-specific behavioral patch instead of a GPT/Codex patch.
- Codex-host skills still contained Claude-only phrasing like `Agent tool`, `Claude subagent`, and `subagent_type: "general-purpose"`.
- Small no-preamble skills like `/freeze`, `/careful`, `/guard`, `/unfreeze`, and `/gstack-upgrade` missed host/model adaptation entirely.

PR #1162 fixes that by:

- defaulting Codex-host generation to `gpt-5.4` unless `--model` is explicit;
- adding a Codex host runtime patch to every generated Codex skill;
- rewriting Claude-only tool/subagent vocabulary for Codex output;
- making spawned skill-review agents load the full named `SKILL.md`, not a tiny prompt excerpt;
- adding regression tests that iterate every generated Codex skill and assert the Codex/GPT patches are present and Claude defaults are absent.

Verification on the PR branch:

- `bun run gen:skill-docs --host all`
- `bun run skill:check`
- `bun test test/gen-skill-docs.test.ts`
- `bun test test/host-config.test.ts test/gen-skill-docs.test.ts`
- `bun test`

PR: https://github.com/garrytan/gstack/pull/1162

time-attack:
@giaphutran12 Removing Claude defaults from generated Codex skills remains unresolved until the selected implementation lands.

PR #2045, opened 2026-06-18
v1.58.1.1 fix(askuserquestion): document the JSON tool-call shape to prevent malformed calls

mruderman:
## Problem

A model following the AskUserQuestion Format directive emitted `questions` as a **JSON string** instead of an array (observed crashing a host TUI: the renderer spread `questions[0].options` → `undefined` → brick-on-startup, persisted across restarts).

Root cause in the directive itself: `generateAskUserFormat` described the **prose decision-brief format** (D<N>, ELI10, pros/cons) but **never specified the actual JSON tool-call schema**. The array shape was left for the model to infer — exactly the seam where double-encoding (`questions: "[{...}]"` instead of `questions: [{...}]`) slipped in. Hosts render the prompt *before* validating tool args, so a malformed shape can crash the session before the tool implementation rejects it.

## Fix

Adds a **"Tool-call shape (JSON) — schema-critical"** subsection to the AskUserQuestion Format directive:

- `questions` MUST be a true JSON **array of objects** — never a string, never a stringified array. 1–4 questions (matches the tool's `maxItems: 4`).
- Each question MUST carry a non-empty `options` array of 2–4 `{ label, description }` (matches `minItems: 2, maxItems: 4`).
- If the schema can't be satisfied, fall back to prose per the existing failure-fallback rule — never emit a bad call.

Plus two **self-check** items pinning the array / non-string requirement.

## Why prompt-level (not a runtime guard)

This hardens the *emission* side. A separate companion PR to the host project (`letta-ai/letta-code`) hardens the *renderer* side with runtime validation so any malformed payload falls through to a generic approval instead of crashing. The two are complementary: this one reduces the chance a model produces a bad call; the renderer fix makes any that slips through non-fatal.

## Changes

| File | Change |
|---|---|
| `scripts/resolvers/preamble/generate-ask-user-format.ts` | + new "Tool-call shape (JSON)" section + 2 self-check items |
| `test/resolver-ask-user-format.test.ts` | +1 locking test for the new section |
| `*/SKILL.md` (Claude + Codex hosts) | regenerated (required by freshness CI) |
| `VERSION` | micro bump `1.58.1.0 → 1.58.1.1` |

## Verification

- `bun test test/resolver-ask-user-format.test.ts` → 38 pass (37 existing + 1 new)
- `bun run skill:check` → Claude ✅ + Codex ✅ fresh (gitignored host flavors stale, as expected — not tracked)
- `detect-bump 1.58.1.0 1.58.1.1` → `micro`

## Notes

- The 2–4 options / 1–4 questions bounds in the new section match the existing tool schema (`src/tools/schemas/AskUserQuestion.json` in the host project) and the existing "Handling 5+ options — split, never drop" rule already in this directive.
- Generated `SKILL.md` files are tracked in this repo (not gitignored), so the regen is included to satisfy the `skill-docs.yml` freshness gate.


time-attack:
@16francej — recommend closing this PR. It is superseded by #1163 because it includes an invalid pseudo-JSON example, a damaged version edit, stale generated files, and host-specific fields in shared guidance; #2267 is withdrawn. Credit #2045 for establishing that questions must be an array rather than a string, and carry that into valid host-aware schemas.

PR #2050, opened 2026-06-19
v1.58.3.0 feat: GBrowser Layer C JS-layer stealth by default

adamouabakar:
## What ships

- Layer C stealth on all 4 context paths (7 JS tell categories, 80 tests)
- Windows CI: Playwright chromium + stealth test suite
- CHANGELOG/BROWSER.md with honest JS-layer vs CDP framing
- TODOS.md CDP context updated for v1.58.3.0 baseline

## NOT in scope (deferred)

- CDP patches, Google captcha E2E, kill switch, /health stealth status

## Next

1. Wait for CI (windows-free-tests + main gate)
2. /land-and-deploy after merge, if you use that workflow


jbetala7:
Heads up: the always-on Layer C stealth this PR ships already landed on `main` via maintainer PR #2047 ("gbrowser anti-detection Layer C stealth"), merged 2026-06-18 (commit `a861c00c`, also tagged `v1.58.3.0`).

The overlap is at the file level, not just the title:

- `browse/src/stealth.ts` on `main` already opens with the Layer C doc block and `buildStealthScript` (webdriver mask, `window.chrome.{runtime,app,csi,loadTimes}`, `Notification.permission` alignment, per-install `hardwareConcurrency`/`deviceMemory`, the depth-3 `Function.prototype.toString` proxy, Selenium/Phantom/Playwright global sweep).
- `browse/test/stealth-layer-c.test.ts` already exists on `main`. This PR's diff adds it from `/dev/null`, which means the branch was cut before #2047 merged — so `stealth.ts`, `browser-manager.ts`, and the three stealth test files will conflict head-on rather than apply cleanly.
- `VERSION` here is `v1.58.3.0`, the version #2047 already shipped.

Two things that would make this salvageable instead of a duplicate:

1. Rebase onto current `main` and drop the Layer C re-implementation, keeping only what's genuinely additive over #2047 (e.g. the `windows-free-tests.yml` Windows CI stealth job, if that isn't already covered). If nothing remains additive once #2047 is on the base, this can be closed.
2. Split `hosts/grok-build.ts` + `hosts/index.ts` + the `grok-build-ship-SKILL.md` golden into their own PR — that grok-build host is unrelated to stealth, isn't on `main` yet, and bundling it makes the conflict harder to resolve.

Not trying to discount the work — the Layer C direction was right, it just already merged. Pointing at #2047 so this doesn't get reviewed as a fresh feature.

time-attack:
@16francej Please close this PR. Its browser stealth Layer C files match the implementation already merged in #2047. Any unrelated Windows or host work should be proposed separately rather than keeping this duplicate release PR open.

PR #2054, opened 2026-06-19
fix(context-restore): prefer the current branch's own checkpoint (#2052)

### LINKED ISSUE #2052: /context-restore can load the wrong worktree checkpoint (shared origin-derived slug + newest-across-all-branches restore)
## Summary
With multiple git worktrees of the same repo, `/context-restore` in one worktree
can load a checkpoint saved by a *different* worktree. Two design choices combine.

## Root cause
1. `bin/gstack-slug` derives the slug from `git remote get-url origin`. All
   worktrees of a repo share one origin → same slug → same
   `~/.gstack/projects/<slug>/checkpoints/` dir. (The per-cwd slug cache just
   stores that same shared value under each worktree's path key.)
2. `/context-restore` by default "loads the most recent saved context across ALL
   branches" and explicitly "does NOT filter the candidate set by current branch"
   (per `context-restore` SKILL.md — intentional, for Conductor handoff).

Together: each worktree's `/context-save` writes into the same shared dir, and
`/context-restore` picks the newest file across all branches — which may be
another worktree's.

## Repro
1. Repo with `origin` set; `git worktree add ../repo-b feature-b`.
2. Worktree B: `/context-save` (checkpoint B).
3. Worktree A (older/no checkpoint for its branch): `/context-restore`.
4. A restores B's checkpoint.

## Impact
Cross-worktree context contamination — the agent resumes from the wrong
worktree's state. Most visible in concurrent multi-session/multi-worktree usage.

## Proposed fix (options)
- Default `/context-restore` to branch-scoped selection (filter candidates by
  frontmatter `branch:` == current branch), with an explicit `--all-branches`
  opt-in to keep the cross-branch handoff use case. Checkpoints already record
  `branch:` in frontmatter.
- And/or make the slug worktree-aware so each worktree gets its own checkpoints dir.
- A `--branch <name>` / explicit-path selector for deterministic pinning.

## Related
#1125 (slug derivation), #1191 (concurrent same-repo state), #1161 (repo-local artifacts)

Environment: gstack 1.48.0.0

jbetala7:
## Problem

All worktrees of a repo share one `origin`-derived slug (`bin/gstack-slug` keys off `git remote get-url origin`), so they all resolve to the same `~/.gstack/projects/<slug>/checkpoints/` directory. `/context-restore` Step 1 then loaded the **newest checkpoint across the whole directory, regardless of branch**:

```bash
FILES=$(find "$CHECKPOINT_DIR" -maxdepth 1 -name "*.md" -type f | sort -r | head -20)
```

With multiple live worktrees, that means `/context-restore` in worktree A could silently restore a checkpoint saved by worktree B — cross-worktree context contamination (issue #2052).

## Fix

Step 1 now orders candidates **current-branch-first**, reading each file's `branch:` frontmatter, and keeps other-branch checkpoints as a *fallback* rather than dropping them. A git branch is checked out in at most one worktree, so preferring the current branch's own checkpoint cleanly separates worktrees.

Crucially this **preserves the documented Conductor cross-branch handoff**: when the current branch has no checkpoint of its own, the full newest-first set across all branches is still used — exactly the old behavior. The change only stops a *newer sibling-worktree save* from shadowing a branch that does have its own checkpoint.

Details:
- scan the 200 newest before partitioning, so a current-branch checkpoint sitting below a burst of sibling saves is still found; the result is still capped at 20
- non-git / detached-HEAD / branchless legacy saves fall back to the old newest-first behavior (back-compat), and no checkpoint is ever dropped from the candidate set

This implements the issue's first proposed option (branch-scoped default + cross-branch fallback) without flipping the default to a hard branch filter, so the Conductor handoff use case keeps working with no new flag.

## Testing

`bun test test/context-save-hardening.test.ts` → **26 pass** (21 existing + 5 new).

New regression tests in the existing `context-restore: ...` harness (they exercise the exact Step-1 shell pipeline):
- the bug: a current-branch save is **not** shadowed by a newer sibling-worktree save (fails on the old pipeline)
- fallback: current branch has no checkpoint → newest across all branches (Conductor handoff)
- back-compat: empty/non-git current branch → newest across all
- branchless legacy saves still rank as fallback, never lost
- within the current branch, ordering stays newest-first

Regenerated `context-restore/SKILL.md` and `scripts/proactive-suggestions.json` via `bun run gen:skill-docs` (all three hosts); generated docs are fresh.

Branched from `main` at `a861c00c`.

Fixes #2052

### DIFF
--- a/context-restore/SKILL.md
+++ b/context-restore/SKILL.md
@@ -23,8 +23,8 @@ triggers:
 ## When to invoke this skill
 
 Loads the most recent
-saved state (across all branches by default) so you can pick up where you
-left off — even across Conductor workspace handoffs.
+saved state (preferring the current branch, falling back across branches) so
+you can pick up where you left off — even across Conductor workspace handoffs.
 Use when asked to "resume", "restore context", "where was I", or
 "pick up where I left off". Pair with /context-save.
 Formerly /checkpoint resume — renamed because Claude Code treats /checkpoint
@@ -762,13 +762,19 @@ context and present it clearly so the user can resume work without losing a beat
 **HARD GATE:** Do NOT implement code changes. This skill only reads saved
 context files and presents the summary.
 
-**Default: load the most recent saved context across ALL branches.** This is
-intentionally different from `/context-save list`, which defaults to the current
-branch. `/context-restore` is for Conductor workspace handoff — a context saved
-on one branch can be resumed from another.
+**Default: prefer the most recent checkpoint saved on the CURRENT branch; if
+this branch has none, fall back to the most recent across ALL branches.** The
+fallback is for Conductor workspace handoff — a context saved on one branch can
+be resumed from another. The current-branch preference exists because every
+worktree of a repo shares one checkpoints directory (same origin-derived slug),
+so without it `/context-restore` in one worktree could silently load a sibling
+worktree's newer checkpoint.
 
-**Do NOT filter the candidate set by current branch.** The `list` flow does
-that; `/context-restore` does not.
+**Do NOT hard-filter the candidate set to the current branch** — other-branch
+checkpoints stay in the set as a fallback. They are just ordered *after* the
+current branch's own, so a current-branch save is never shadowed by a newer
+sibling-worktree save. (`/context-save list` is the flow that hard-scopes to one
+branch.)
 
 ---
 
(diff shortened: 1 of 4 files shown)


time-attack:
@jbetala7 Please rewrite checkpoint selection to derive the current branch directly instead of trusting CURRENT_BRANCH, which may be stale and select another branch’s checkpoint.

PR #2058, opened 2026-06-19
Feat/cons 003 localhost binding validation

seab-group:


16francej:
Closing — this adds content unrelated to gstack (a separate product or project). If there's a gstack-specific fix in here, please open a focused PR for just that change.

PR #2059, opened 2026-06-19
fix: respect GBrain-aware Claude skill render in skill check

furrwilson-collab:
## Summary
- make skill-check respect host generation exclusions when validating Claude Code templates
- allow the Claude Code freshness check to accept the local GBrain-aware user render
- point stale Claude Code repair output at `bun run gen:skill-docs:user`

## Test plan
- `HOME=/Users/majestic12 bun run skill:check`

## Risk
- Low: scoped to the validation/check script; generated skill docs are not changed by this commit.


time-attack:
@furrwilson-collab Please add and commit a regression test covering the broader GBrain-aware Claude rendering behavior, with validation showing the fix works.

PR #2062, opened 2026-06-20
v1.58.4.0 fix: scope /health bootstrap token to pinned extension + patch dependency CVEs

Saimoguloju:
## What

Two related daemon-security fixes:

1. **`/health` token leak (privilege escalation).** `/health` returned the
   root `AUTH_TOKEN` to *any* caller with a `chrome-extension://` Origin,
   regardless of which extension. On a multi-extension machine a second
   extension could scrape it and drive every browse-server endpoint. The
   extension-origin path now requires an exact
   `chrome-extension://<BROWSE_EXTENSION_ID>` match when the id is pinned,
   mirroring the existing `/ws` origin gate in `terminal-agent.ts`.
   **Backward-compatible:** unset id = unchanged behavior.

2. **Dependency CVEs.** `bun audit` flagged 37 advisories. Patched the direct
   deps we control (`@anthropic-ai/sdk` 0.78.0 → 0.91.1 insecure file perms,
   `diff` 7 → 8 parsePatch DoS) and refreshed transitive deps within range.
   Added a non-blocking `dependency-audit` CI workflow so new CVEs surface at
   review time.

## Why

The `/health` behavior is a real, exploitable privilege-escalation path on
shared/multi-extension machines (it was already filed internally as a deferred
P2). The remaining transitive advisories live behind pinned upstream packages
and are documented as follow-up.

## Testing

- `bun test browse/test/server-auth.test.ts` — 39/39 pass (added a regression
  test pinning the extension-id gate).
- `bun audit` before/after confirms the two direct-dep CVEs are cleared.

## Notes

- No behavior change unless `BROWSE_EXTENSION_ID` is set.
- Pure additive CI; the audit job is non-blocking to start.


Saimoguloju:
trunk merge below.

time-attack:
@16francej — recommend closing this PR as superseded by #2226. Unresolved security problems remain, and the branch also combines unrelated dependency and release changes; #2226 owns the complete local-auth and legacy-profile migration. Preserve #2062’s authentication-hardening attribution in #2226, and move any still-current dependency remediation to a separate compatibility-reviewed change.

PR #2064, opened 2026-06-20
v1.59.0.0 feat(browse): device / geo / locale / timezone emulation

Saimoguloju:
## Design

- Extracts a shared **`buildContextOptions()`** so the three headless
  `newContext()` sites stop duplicating the viewport/scale/UA block — the new
  overrides are added in one place and can't drift.
- `buildEmulationExtras()` (locale/timezone/geo, no viewport dependency) also
  spreads into the two headed `launchPersistentContext` sites, so those carry
  into headed mode and across the headless→headed handoff.
- Device viewport sizing is **headless-only** (headed keeps the real window),
  documented in the command help.
- All overrides persist across navigations, context rebuilds, and new tabs —
  same pattern as `useragent`/`viewport`.

## Testing

- `bun test browse/test/emulation-command.test.ts` — 10/10 pass (state +
  option builders, no real Chromium, runs on Windows).
- `bun run gen:skill-docs` regenerated; usage-consistency validation passes.
- Build-checked `browser-manager.ts` / `write-commands.ts`.


Saimoguloju:
trunk merge below.

time-attack:
@Saimoguloju Please resolve the conflicts and add runtime tests for device, geolocation, locale, and timezone emulation. Current checks only verify title and version synchronization.

PR #2071, opened 2026-06-21
[codex] Add consumer session ingest contract

MattJWard1986:
Linear: MAT-14

## Summary
- Adds a normalized consumer AI chat session contract for future provider importers.
- Extends `gstack-memory-ingest` with an explicit `consumer-session` source, metadata-only dry-run, deterministic rendering/chunking, and per-conversation incremental state.
- Keeps consumer sessions out of the default ingest source set until provider importers and consent UX land.

## Adversarial review hardening
- `consumer-session` ingest now requires explicit `--sources consumer-session` opt-in.
- `--limit` does not record partially staged multi-chunk conversations as ingested.
- Probe accounting compares consumer-session content hashes against state.
- Dry-run output uses sanitized relative roots and does not expose absolute local paths or private storage filenames.

## Verification
- `bun test test/gstack-memory-ingest.test.ts test/gstack-gbrain-sync.test.ts test/gbrain-source-gitignore.test.ts` - pass, 72 tests.
- `git diff --check origin/main...HEAD` - pass.
- `bun run build` - blocked before this change by missing local dependency artifact: `node_modules/xterm/lib/xterm.js`.

PR #2073, opened 2026-06-21
[codex] Add ChatGPT consumer session import

MattJWard1986:
Linear: MAT-15

## Summary
- Adds a ChatGPT official export normalizer for extracted exports, direct `conversations.json`, and ZIP exports.
- Writes normalized ConsumerSession JSON under `consumer-sessions/normalized/chatgpt`.
- Preserves ordered turns, attachment metadata, receipts, account hash, host, completeness, and MAT-14 renderer compatibility.

- Replaced `unzip` extraction with safe Python zip extraction that rejects symlinks and path escapes.
- Preserves non-active ChatGPT branch messages instead of silently importing only the current lineage.
- Redacts CLI dry-run paths and avoids leaking local export folders.
- Uses stable fallback account identity for ZIPs without `user.json`.
- Strips raw `download_url` values from attachment identity.
- Writes normalized outputs atomically and removes stale outputs via a manifest.

## Verification
- `bun test test/consumer-session-chatgpt.test.ts test/gstack-memory-ingest.test.ts` - pass, 39 tests.
- `git diff --check` - pass.

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -66,6 +66,13 @@ import {
 } from "../lib/gstack-memory-helpers";
 import { execGbrainText, spawnGbrainAsync } from "../lib/gbrain-exec";
 import { checkOwnedStagingDir, STAGING_MARKER } from "../lib/staging-guard";
+import {
+  consumerSessionStateKey,
+  discoverConsumerSessionFiles,
+  parseNormalizedConsumerSessionExport,
+  renderConsumerSessionPages,
+  walkConsumerSessionNormalizedFiles,
+} from "../lib/consumer-sessions";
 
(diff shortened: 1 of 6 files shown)


MattJWard1986:
Orchestrator audit note: this PR is stacked on #2071. Worker implementation was followed by an adversarial challenge pass. I agreed with and patched the material findings: safe ZIP extraction rejects symlinks/path escapes, non-active ChatGPT branch messages are preserved, dry-run paths are redacted, ZIP fallback account identity is stable, raw download URLs are not stored as attachment identifiers, and normalized outputs are written atomically with stale-output cleanup. Verification: `bun test test/consumer-session-chatgpt.test.ts test/gstack-memory-ingest.test.ts` passes (39 tests), and `git diff --check` passes.

PR #2074, opened 2026-06-21
[codex] Add Gemini Takeout consumer session import

MattJWard1986:
Linear: MAT-17

## Summary
- Adds a Gemini Apps Google Takeout normalizer.
- Supports extracted folders plus ZIP/TGZ/TAR.GZ archives through a safe extractor.
- Handles synthetic Gemini Apps conversation JSON, embedded My Activity conversations, Gems metadata, uploads/generated-media metadata, recurring export overlap, and MAT-14 renderer compatibility.

- Safe archive extraction rejects symlinks, hardlinks, special files, absolute paths, and path escapes.
- Keeps raw provider conversation IDs distinct and uses hashed filename suffixes to avoid slug collisions.
- CLI dry-run output redacts paths, account hash, and conversation IDs.
- Writes normalized outputs atomically and removes stale outputs via a manifest.
- Tightened text extraction so unknown payload-like fields such as inline data are not copied into turn text.

## Verification
- `bun test test/consumer-session-gemini.test.ts test/gstack-memory-ingest.test.ts` - pass, 38 tests.
- `git diff --check` - pass.

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -66,6 +66,13 @@ import {
 } from "../lib/gstack-memory-helpers";
 import { execGbrainText, spawnGbrainAsync } from "../lib/gbrain-exec";
 import { checkOwnedStagingDir, STAGING_MARKER } from "../lib/staging-guard";
+import {
+  consumerSessionStateKey,
+  discoverConsumerSessionFiles,
+  parseNormalizedConsumerSessionExport,
+  renderConsumerSessionPages,
+  walkConsumerSessionNormalizedFiles,
+} from "../lib/consumer-sessions";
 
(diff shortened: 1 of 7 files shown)


MattJWard1986:
Orchestrator audit note: this PR is stacked on #2071. Worker implementation was followed by an adversarial challenge pass. I agreed with and patched the material findings: safe archive extraction rejects symlinks/hardlinks/special files/path escapes, raw provider conversation IDs remain distinct, dry-run output redacts paths/account/conversation IDs, normalized writes are atomic with stale-output cleanup, and text extraction avoids unknown payload-like fields such as inline data. Verification: `bun test test/consumer-session-gemini.test.ts test/gstack-memory-ingest.test.ts` passes (38 tests), and `git diff --check` passes.

PR #2075, opened 2026-06-21
[codex] Add consumer export bridge inbox sync

MattJWard1986:
Linear: MAT-20

## Summary
- Extends the artifacts bridge for explicit consumer AI export inboxes:
  `consumer-sessions/raw/<provider>/<account>/<device>/inbox/`.
- Adds per-device `freshness.json`, additive discovery/sync, archive `.complete` handoff guards, dry-run discovery, and best-effort local permission hardening.
- Documents that raw exports stay outside Obsidian and outside default GBrain search.

- Drain-time staging now enforces the same temp-name and archive completion-marker checks as discovery.
- Consumer raw paths are validated by exact path shape, not just broad `fnmatch` globs.
- Discovery is privacy-mode aware and does not advance the cursor for behavioral exports hidden by `artifacts-only`.
- Dry-run output redacts root, account, device, and export filenames.

## Verification
- `bash -n bin/gstack-brain-sync`
- `bash -n bin/gstack-artifacts-init`
- `HOME=/private/tmp/gstack-mat20-test-home bun test test/brain-sync.test.ts` - pass, 33 tests.
- `git diff --check` - pass.

Note: running `bun test test/brain-sync.test.ts` with the real sandboxed `HOME` fails because the test path writes `~/.gstack-artifacts-remote.txt`; isolating `HOME` exercises the intended behavior.

### DIFF
--- a/bin/gstack-artifacts-init
+++ b/bin/gstack-artifacts-init
@@ -249,6 +249,13 @@ builder-profile.jsonl
 # is skipped; brain admin pulls + indexes into the remote brain.
 transcripts/run-*/*.md
 transcripts/run-*/**/*.md
+# Consumer AI export bridge inboxes. Raw exports are synced only when the
+# user explicitly places them under:
+#   consumer-sessions/raw/<provider>/<account>/<device>/inbox/
+# The per-device freshness marker lets the receiving machine distinguish
+# "device reported no new exports" from "device has not reported".
+consumer-sessions/raw/*/*/*/freshness.json
+consumer-sessions/raw/*/*/*/inbox/**
 # NOT synced (machine-local UX state):
 #   projects/*/question-preferences.json (per-machine UX preferences)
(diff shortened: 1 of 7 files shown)


MattJWard1986:
Orchestrator audit note: this PR is stacked on #2071. Worker implementation was followed by an adversarial challenge pass. I agreed with and patched the material findings: drain-time staging enforces archive completion markers and temp-name checks, consumer raw paths are validated by exact path shape, discovery does not advance cursor for behavioral exports hidden by `artifacts-only`, and dry-run output redacts root/account/device/export filenames. Verification: `bash -n bin/gstack-brain-sync`, `bash -n bin/gstack-artifacts-init`, `HOME=/private/tmp/gstack-mat20-test-home bun test test/brain-sync.test.ts` (33 tests), and `git diff --check` pass. Running the same test with real sandboxed `HOME` fails because init writes `~/.gstack-artifacts-remote.txt`; isolated `HOME` is the valid harness.

Issue #1950, opened 2026-06-10
`gstack-learnings-log` silently fails on Windows (git-bash) — all AI-logged learnings dropped


okyhm520:
## Summary

On Windows running gstack under git-bash, `bin/gstack-learnings-log` always exits 1 without writing anything, AND without surfacing any error to the caller. The bin swallows stderr with `2>/dev/null`, so neither the user nor the AI agent sees that the call failed.

Root cause: the bin computes `SCRIPT_DIR` via `pwd`, which returns a POSIX path (`/c/Users/...`) under git-bash. The bin then interpolates that into a bun `import` statement (`import { hasInjection } from '$SCRIPT_DIR/../lib/jsonl-store.ts'`). Bun on Windows can't resolve POSIX-style paths as ES module specifiers and throws `Cannot find module`, which gets swallowed.

**Impact:** every Windows user has `/learn`, `/review`, `/ship`, `/investigate`, etc. silently losing every AI-captured learning. The feature is structurally non-functional on Windows.

`gstack-decision-log` does not appear to share this code path, so cross-session decision memory still works.

## Environment

| | |
|---|---|
| gstack | 1.57.7.0 (just upgraded from 1.56.0.0) |
| OS | Windows 11 Pro 26200 |
| Shell | MINGW64_NT bash (git-for-windows 2.54.0) |
| bun | 1.3.14 (Windows x64) |
| Claude Code | (host) |

## Repro

From git-bash, with gstack installed at `~/.claude/skills/gstack`:

```bash
~/.claude/skills/gstack/bin/gstack-learnings-log '{"skill":"learn","type":"pitfall","key":"test-key","insight":"hello world","confidence":5,"source":"observed"}'
echo "exit=$?"
wc -l ~/.gstack/projects/$(~/.claude/skills/gstack/bin/gstack-slug | sed 's/SLUG=//;s/ .*//')/learnings.jsonl
```

**Expected:** `exit=0`, line count increases by 1.

**Actual:** `exit=1`, line count unchanged. No stderr output.

## Root cause (verified)

Running the validation pipeline manually with stderr unsuppressed reveals:

```
error: Cannot find module '/c/Users/okyhm/.claude/skills/gstack/bin/../lib/jsonl-store.ts'
       from 'C:\Users\okyhm\[eval]'

Bun v1.3.14 (Windows x64)
```

[...]

PR #2076, opened 2026-06-21
fix(learnings-log): bun -e import fails on Windows Git Bash (MSYS path mangling)

DanCanadian:
## Problem

On **Windows Git Bash (MSYS)**, `bin/gstack-learnings-log` exits 1 with no output, so `/learn` cannot append learnings — they have to be hand-appended to `learnings.jsonl`.

## Root cause

bun is a **native Windows binary**. The script runs:
```bash
bun -e "import { hasInjection } from '$SCRIPT_DIR/../lib/jsonl-store.ts'; ..."
```
MSYS unreliably mangles the absolute `/c/Users/...` import path embedded inside the `bun -e` script string, so the static import fails with `Cannot find module ... from 'C:\...\[eval]'` → empty stdout → `exit 1`, no write. (Confirmed via `bash -x` trace + isolating the snippet.)

It is **not** the bun-shim PATH issue — the shim resolves fine.

## Fix

Resolve the lib to a `file://` URL (`cygpath -m` on MSYS, plain absolute path on POSIX), pass it via env var `GSTACK_LIB_URL` (env values are **not** path-mangled), and `await import()` it dynamically (the script already uses top-level await). One-file change in `bin/gstack-learnings-log`.

POSIX/macOS/Linux are unaffected (no `cygpath`; `file://` + absolute path resolves correctly).

## Verification (Windows Git Bash)

- Probe entry logged → `exit=0`, written to `learnings.jsonl` with auto-injected `ts` and `trusted:false`.
- Reject paths intact: invalid `type` and invalid `key` both exit 1 and write nothing.
- `gstack-learnings-search` reads the file cleanly.

Only `gstack-learnings-log` used this inline-absolute-path-in-`-e`-string pattern; other bin scripts use relative imports (`from "../lib/..."`) which bun resolves from the module location regardless of MSYS.

### DIFF
--- a/bin/gstack-learnings-log
+++ b/bin/gstack-learnings-log
@@ -7,6 +7,19 @@
 # by gstack-learnings-search ("latest winner" per key+type).
 set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+# bun is a native binary. On Git Bash (Windows) MSYS unreliably mangles an
+# absolute /c/... path embedded inside the `bun -e` script string, breaking
+# the static import ("Cannot find module"). Resolve the lib to a file:// URL
+# and hand it to bun via an env var (env values are NOT path-mangled), then
+# dynamic-import it. cygpath -m yields a Windows path on MSYS; on POSIX the
+# canonical absolute path is used directly.
+LIB_TS="$(cd "$SCRIPT_DIR/../lib" && pwd)/jsonl-store.ts"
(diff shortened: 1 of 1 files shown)


DanCanadian:
Superseding: upstream already fixed this exact Windows Git Bash bug in v1.58.4.0 (issue #1950) via `cygpath -m` on SCRIPT_DIR — cleaner than this env-var/dynamic-import approach. Confirmed the shipped fix works on my Windows install after upgrading. Closing as redundant. Thanks!

PR #2083, opened 2026-06-22
fix: scope benchmark skill to web performance

creatornader:
## Summary
- rename the web performance benchmark skill to `gstack-benchmark`
- narrow benchmark triggers so bare benchmark requests do not match this skill
- add generator coverage for benchmark scope and valid `llms.txt` skill links

## Tests
- `bun test test/gen-skill-docs.test.ts test/llms-txt-shape.test.ts`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,7 +61,7 @@ Invoke them by name (e.g., `/office-hours`).
 | `/learn` | Manage what gstack learned across sessions. |
 | `/retro` | Weekly retro with per-person breakdowns and shipping streaks. |
 | `/health` | Code quality dashboard (type checker, linter, tests, dead code). |
-| `/benchmark` | Performance regression detection (page load, Core Web Vitals). |
+| `/web-performance-benchmark` | Performance regression detection (page load, Core Web Vitals). |
 | `/benchmark-models` | Cross-model benchmark for skills (Claude, GPT, Gemini side-by-side). |
 | `/cso` | OWASP Top 10 + STRIDE security audit. |
 | `/setup-gbrain` | Set up gbrain for cross-machine session memory sync. |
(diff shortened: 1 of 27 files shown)


time-attack:
Thanks for this. The command rename leaves active references, lacks a compatibility plan, and combines an independent link-generation repair with incomplete routing coverage. Please rebase, choose one canonical command, preserve compatibility, separate the link fix, update all references, and add real routing tests.

creatornader:
Thanks. I am addressing this review as three focused PRs:

- #2083, this PR: the canonical /web-performance-benchmark command, compatibility aliases, active references, and routing coverage.
- #2279: the independent generated gstack/llms.txt link repair and its link-validity coverage.
- #1795: the unrelated voice-guardrail feedback and regenerated outputs.

The rename exposes the generator's invalid path assumption, so #2279 needs to land first. I will then rebase #2083 onto main and remove the copied link-repair diff from this branch. That leaves this PR limited to the command migration and its compatibility behavior.

The canonical command is /web-performance-benchmark. /benchmark and /gstack-benchmark remain compatibility aliases. I have updated the remaining active browser-design reference and added regression coverage for the canonical route. I will follow up after the rebase with the final verification rather than marking this feedback resolved early.

creatornader:
Update: the two live routing selections now pass locally. A performance-regression request invokes web-performance-benchmark, and an explicit /benchmark request resolves to the same canonical workflow.

The remaining dependency is #2279. Trunk rejected my submission from this fork because I do not have write access to the repository. Once a maintainer submits and lands #2279, I will rebase this PR onto main, remove the duplicated link repair, and post the final focused diff for review.

creatornader:
Can we please bump this fix? My agents are making this mistake almost a dozen times a day where they call the "benchmark" skill thinking it will help with model benchmarks etc only to find out it's completely irrelevant. A skill this narrow should not simply be treated as a generic "benchmark" skill, because it's not. I will have to uninstall this skill and/or gstack if not fixed soon, it's extremely disruptive.

PR #2106, opened 2026-06-24
feat(context-save): persist plan file before saving checkpoint when in plan mode

ruth-zhangx:
When GSTACK_PLAN_MODE=active and CLAUDE_PLAN_FILE exists, /context-save now
persists the plan to .claude/plans/ before writing the checkpoint. This
prevents plan loss when the user saves and exits without approving the plan
(via ExitPlanMode) — common on unstable hardware or when the user needs to
rest.

The plan is saved with a timestamp and also copied to active-plan.md as a
stable symlink target for /context-restore.

Fixes: checkpoint referencing a plan path that does not exist on disk.


ruth-zhangx:
Closing: no upstream response after 9 days. Local fork retains the patch as 防御层 B.

PR #2108, opened 2026-06-25
v1.59.2.0 fix(make-pdf): stop smartypants URL carve from swallowing the next placeholder (#2084)

### LINKED ISSUE #2084: make-pdf: bare/visible http(s):// URLs become SMARTPANTS_PRESERVED tokens and bleed link-blue into the rest of the doc
## Summary

`make-pdf generate` mangles any bare/visible `http(s)://` URL into literal `SMARTPANTS_PRESERVED_<n>` tokens, and leaves an unclosed `<a>` that bleeds link-blue into **every following block** for the rest of the document.

Version: gstack `1.55.0.0` (binary `make-pdf/dist/pdf`).

## Repro

```markdown
# t

See https://example.com here.

This paragraph after the URL should be black.
```

```bash
make-pdf generate /tmp/ex.md /tmp/ex.pdf
```

## Actual

- URL renders as `https://example.comSMARTPANTS_PRESERVED_4 here.` (the smartypants protect-placeholder is never restored).
- The next paragraph ("This paragraph after the URL should be black.") renders **blue/underlined** — the broken `<a>` is never closed, so link styling bleeds into all subsequent blocks (headings + body) to the end of the doc.

## Expected

URL renders intact as a clickable link; following content keeps its normal (black) styling.

## Scope / triggers

Reproduces whenever a raw URL is **visible text**, in any of:
- bare URL (`https://example.com`), line-end or mid-line
- bold (`**https://example.com**`)
- markdown link whose **display text is also a URL** (`[https://example.com](https://example.com)`)

Only clean case: markdown link with non-URL display text → `[example](https://example.com)`. That suggests the smartypants/typography pre-pass tokenizes the URL to protect it, then fails to restore the placeholder when the visible run is itself a URL.

## Workaround

Use `[label](https://…)` with non-URL display text, or drop the `://` scheme and write the host as plain text. Posting in case the protect/restore step is an easy fix.

genisis0x:
Closes #2084.

`make-pdf generate` mangled any bare/autolinked `http(s)://` URL into a literal `SMARTPANTS_PRESERVED_<n>` token and left an unclosed `<a>` that bled link-blue into every following block to the end of the document.

## Root cause

`make-pdf/src/smartypants.ts` carves code/tags/URLs into U+0000-delimited placeholder tokens, transforms the remaining text, then restores the tokens. The URL pattern was `/\bhttps?:\/\/\S+/g`. `\S` matches non-whitespace, and the placeholder delimiter U+0000 is non-whitespace, so for an autolinked URL (`<a href="URL">URL</a>`) the link-text URL match greedily swallowed the trailing `</a>` placeholder. The nested placeholder never restored (the restore is a single left-to-right pass), so `SMARTPANTS_PRESERVED_<n>` leaked into the text and the `</a>` was dropped.

## Fix

Exclude U+0000 from the URL character class (`[^\s]+`) so the URL match stops at the placeholder boundary. Real URLs never contain U+0000, so the match is identical to `\S+` except that it no longer crosses a carve boundary.

Before / after on the issue's repro (via the real `smartypants()`):

- before: `<a href="https://example.com">https://example.com SMARTPANTS_PRESERVED_2  here.` (leak, no `</a>`)
- after: `<a href="https://example.com">https://example.com</a> here.`

## Test

Added a regression test in `make-pdf/test/render.test.ts` (the existing `smartypants` block): an autolinked URL keeps its `<a>...</a>` intact and leaks no placeholder. Verified it fails on the old `\S+` regex and passes on the fix. Full `render.test.ts`: 67 pass. Quotes, em dashes, ellipses, code/pre zones, and href-only URLs are unaffected.


time-attack:
@16francej — recommend closing this PR as superseded by #2084. Its URL boundary and intact-anchor test are useful, but the branch has stale version edits and unresolved safety problems; #2084 owns the complete current-source renderer fix. Credit #2108 for the narrow boundary work, preserve its test, and add safe input handling without version changes.

genisis0x:
Closing this as recommended in triage. The branch has diverged (stale version-file edits now conflict with main), so it is not worth rebasing in place. The URL-boundary fix and the intact-anchor regression test are still the right shape for #2084 and can be salvaged into a clean branch if a fresh PR is wanted.

PR #2114, opened 2026-06-26
feat: add gbrain context-load and save surfaces to 16 skills

aleksanderstevens191-sudo:
## Summary

- Adds a **Brain Context Load** block to 16 skills: before running, the skill extracts 2–4 keywords from the user's request, searches gbrain, reads the top 3 results, and uses that context to inform analysis. Skipped gracefully when `gbrain` is not on PATH.
- Adds a **Save Results to Brain** block to the same skills: after completing, the skill writes its output as a structured page under a feature-namespaced slug, creates stub pages for person/org entities found in the output, and handles throttle/rate-limit errors non-fatally.
- Full protocol details are delegated to `docs/gbrain-write-surfaces.md` to keep individual skill files thin.

## Skills updated

`cso`, `design-consultation`, `design-review`, `investigate`, `office-hours`, `plan-ceo-review`, `plan-design-review`, `plan-devex-review`, `plan-eng-review`, `qa`, `retro`, `ship` — plus their relevant section files.

## What a reviewer should know

- The gbrain blocks are **opt-in by environment**: both sections open with `**Skip this entire section if gbrain is not on PATH.**` so users without gbrain set up are unaffected.
- No existing skill logic was changed — only additive blocks prepended/appended.
- The `docs/gbrain-write-surfaces.md` doc (referenced in both blocks) is assumed to exist in the repo; if it doesn't yet, that doc should land before or alongside this PR.
- Untracked files (`.feature-prompted-*`, `ruvector.db`) were intentionally excluded from the commit.

## Test plan

- [ ] Run any updated skill (`/plan-eng-review`, `/investigate`, etc.) with `gbrain` on PATH — confirm context is loaded pre-run and output is saved post-run
- [ ] Run the same skill with `gbrain` absent from PATH — confirm both blocks are silently skipped with no errors
- [ ] Confirm no regressions in skill behaviour for users without gbrain

### DIFF
--- a/cso/SKILL.md
+++ b/cso/SKILL.md
@@ -755,7 +755,17 @@ Replace `SKILL_NAME`, `OUTCOME`, and `USED_BROWSE` before running.
 
 Skills that run plan reviews (`/plan-*-review`, `/codex review`) include the EXIT PLAN MODE GATE blocking checklist at the end of the skill, which verifies the plan file ends with `## GSTACK REVIEW REPORT` before ExitPlanMode is called. Skills that don't run plan reviews (operational skills like `/ship`, `/qa`, `/review`) typically don't operate in plan mode and have no review report to verify; this footer is a no-op for them. Writing the plan file is the one edit allowed in plan mode.
 
+## Brain Context Load
 
+**Skip this entire section if `gbrain` is not on PATH.**
+
+Extract 2-4 keywords from the user's request. Search the brain:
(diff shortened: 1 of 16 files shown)


time-attack:
@16francej Please close this PR. GBrain context loading and saving already shipped through the generated architecture in #1742. This branch edits generated output only, so regeneration would erase it and it should not be merged.

PR #2119, opened 2026-06-26
fix(relink): don't clobber foreign skills that share a gstack name

smblight:
## Problem

Installing gstack with `--prefix` (or any later `./setup` run) silently deletes a user's **own** flat-named skills when they share a name with a gstack skill. Concretely: a personal `/qa` or `/review` skill symlinked into `~/.agents/skills/` gets removed.

It's not a one-time hit — `setup` calls `gstack-relink` as a self-healing step on every run, so each upgrade re-deletes them.

## Root cause

`bin/gstack-relink` → `_cleanup_skill_entry()` removed **any** flat-named entry matching a gstack skill name, with no check on where the symlink points:

```bash
if [ -L "$entry" ]; then
  rm -f "$entry"            # deletes foreign symlinks too
```

In prefix mode the deletion isn't even necessary — gstack installs under `gstack-*` names — so the removed flat entry was pure collateral.

## Fix

Scope the cleanup to **gstack-owned** entries only: remove a flat entry just when its symlink (or its dir's `SKILL.md` symlink) resolves into the gstack install dir. This mirrors the provenance check `setup`'s `cleanup_old_claude_symlinks()` already uses, and also handles installs not literally named `gstack/` via an `$INSTALL_DIR`-prefix match.

Result: foreign skills sharing a gstack name are preserved; genuinely stale flat-mode gstack entries are still cleaned.

## Tests

Two regression tests added to `test/relink.test.ts`:

- `prefix mode preserves foreign flat skills sharing a gstack name` — covers both the directory-symlink shape and the real-dir-with-foreign-`SKILL.md`-symlink shape.
- `prefix mode still removes gstack-owned stale flat entries` — guards against over-correcting.

Verified the first test fails on `main` (foreign `/qa` deleted → `ENOENT`) and passes with the fix. Full suite: `26 pass, 0 fail`.

```
bun test test/relink.test.ts
```

### DIFF
--- a/bin/gstack-relink
+++ b/bin/gstack-relink
@@ -37,12 +37,24 @@ SKILLS_DIR="${GSTACK_SKILLS_DIR:-$(dirname "$INSTALL_DIR")}"
 PREFIX=$("$GSTACK_CONFIG" get skill_prefix 2>/dev/null || echo "false")
 
 # Helper: remove old skill entry (symlink or real directory with symlinked SKILL.md)
+# ONLY removes gstack-owned entries — i.e. ones whose symlink (or whose dir's
+# SKILL.md symlink) resolves into the gstack install dir. A flat entry of the
+# same name that points elsewhere (e.g. a personal /qa or /review symlink into
+# ~/.agents/skills/) is left untouched, so switching prefix modes never clobbers
+# unrelated user skills that happen to share a gstack skill name.
 _cleanup_skill_entry() {
   local entry="$1"
+  local dest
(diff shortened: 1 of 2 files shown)


time-attack:
@smblight replace lexical cases with link parent aware path resolutin pls, there is potential for destructive execution with this

time-attack:
@smblight Please replace lexical ownership checks with path-safe containment checks. Current logic can treat paths outside the installation root as owned and delete them.

PR #2122, opened 2026-06-27
fix: make builder-profile checks windows-safe

TheOneDarkHorse:
## Summary
- make builder-profile checks Windows-safe
- adjust setup/test behavior for Windows path handling

## Verification
- git diff --check HEAD~1..HEAD


TheOneDarkHorse:
Closing at our request; we will keep this local/fork-only for now.

PR #2126, opened 2026-06-28
Fix grammar in SKILL.md checklist item

aniruddh10124:
Past tense to assertive present tense


time-attack:
@16francej — recommend closing this PR. It edits generated output that will be overwritten, and the proposed tense conflicts with the surrounding completed-action checklist. Any wording change should begin in the resolver source.

PR #2128, opened 2026-06-28
fix(browse): skip crash-retry when user invokes stop or restart

TTmo123:
## What

`browse stop` and `browse restart` no longer trigger the "Server crashed twice in a row — aborting" warning.

## Why

Today's `sendCommand` ECONNRESET handler (`browse/src/cli.ts:586`) treats every dropped connection as a crash. But `stop` and `restart` are user-initiated shutdowns (daemon-side at `browse/src/meta-commands.ts:423`, `:428`): the daemon intentionally exits before the HTTP response body flushes. The restart path then sends `stop` again, the new daemon exits again, and `retries >= 1` throws the misleading crash message.

## How

One 4-line branch at the top of the ECONNRESET handler:

```ts
if (command === 'stop' || command === 'restart') {
  process.exit(0);
}
```

`command` is already in closure from `sendCommand`'s function signature (`cli.ts:527`) — no new param passing. `process.exit(0)` (not a bare `throw`) is intentional: `main()`'s `.catch()` (`cli.ts:1357-1362`) would print `[browse] fetch failed` and exit 1 on a thrown error. Clean exit 0 is what the user intended.

### Verification
- [x] `bun test test/send-command-stop-skip-restart.test.ts` — 4 new cases pass (2 integration + 2 regression-guard)
- [x] `bun test` — full Tier 1 suite passes
- [x] `bun build browse/src/cli.ts` — TypeScript compiles cleanly
- [ ] `bun run test:e2e` (Tier 2) — requires EVALS secret, maintainer-side gate
- [ ] Manual: `browse stop` — exits 0, no stderr output
- [ ] Manual regression: kill the daemon externally, then `browse goto <url>` — crash-retry still fires (unchanged)

### Notes for reviewer
- Independent of PR #NNNN (PR 1, setup symlink). Can be reviewed and merged separately.
- Daemon-side stop/restart handlers (`meta-commands.ts:423-433`) are unchanged. The `"Server stopped"` / `"Restarting..."` strings rarely reach the client because `shutdown()` exits before the response body flushes; the fix respects that as an intentional daemon behavior.
- No `chore: rebuild` commit: `.gitignore` excludes `browse/dist/`, and CI's `windows-setup-e2e.yml` rebuilds binaries on every push touching `browse/src/cli.ts` (trigger path at `.github/workflows/windows-setup-e2e.yml:22`).
- Test design: cases 1-2 are integration tests gated by `describe.skipIf(!HAS_BINARY)` so shards without a built binary gracefully skip; cases 3-4 are code-shape regression guards that always run, catching future refactors that accidentally widen the short-circuit or strip the crash-retry path.

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -585,6 +585,15 @@ async function sendCommand(state: ServerState, command: string, args: string[], 
     }
     // Connection error — server may have crashed, OR may just be busy.
     if (err.code === 'ECONNREFUSED' || err.code === 'ECONNRESET' || err.message?.includes('fetch failed')) {
+      // User-initiated shutdown (stop / restart) intentionally closes the daemon.
+      // Both meta-commands.ts:423 ('stop') and :428 ('restart') call shutdown(),
+      // which closes the HTTP connection before the response body reaches the
+      // client. ECONNRESET here is the expected exit signal, not a crash.
+      // 'command' is already in closure from sendCommand's signature — no new
+      // param passing.
+      if (command === 'stop' || command === 'restart') {
+        process.exit(0);
+      }
       const oldState = readState();
       // #1781 busy-vs-dead: a single-threaded daemon under beacon/extension load
       // can briefly stop answering HTTP while still alive. Before declaring a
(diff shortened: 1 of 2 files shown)


time-attack:
Just reports succesful whenevr the connection fails. Even if the daemon never recieved the stop command 
close

time-attack:
@16francej — recommend closing this PR because it can report a successful stop or restart without proving the daemon processed the command. The available coverage checks only the client exit behavior, not handler execution or resulting process state, so it does not safely resolve #1732.

PR #2129, opened 2026-06-28
fix(setup): prefer symlinks on Windows when SeCreateSymbolicLinkPrivilege is available

TTmo123:
## What

Lets `_link_or_copy()` try `ln -sn` before falling back to `cp` on Windows, when Developer Mode or elevation is present.

## Why

Today's helper unconditionally takes the `cp` branch on Windows to avoid "frozen copies" — `ln -snf` without `SeCreateSymbolicLinkPrivilege` silently produces a copy that doesn't refresh after `git pull`. But this is over-defense: users with Developer Mode get the worst of both worlds (frozen copies) despite having the capability for true symlinks. macOS/Linux users get live symlinks; Windows users with Developer Mode should too.

## How

Three changes to `_link_or_copy()`, all inside the existing IS_WINDOWS=1 branch:

1. Try `ln -sn src dst` first (stderr redirected to /dev/null).
2. Verify the result is a real symlink (`[ -L "$dst" ]` postcondition — catches races where `ln` exits 0 but writes a regular file under `dst`).
3. On failure, run a second `rm -rf "$dst"` and fall through to `cp -R` / `cp -f` (preserves today's behavior for non-Developer-Mode users).

The second `rm -rf` is load-bearing: without it, `cp -R` would fail with "cannot overwrite non-directory ... with directory" when a half-successful `ln` left a regular file at `dst` and `src` is a directory.

All raw `ln` calls stay inside the helper body, preserving the D7 static invariant (`test/setup-windows-fallback.test.ts:25-44`). The `if [ ! -e "$src" ]; return 0` early-return is preserved from the original helper, not introduced here.

### Verification

- [x] `bun test test/setup-windows-fallback.test.ts` — D7 invariant + 3 new symlink-fallback cases pass
- [x] `bun test` — full Tier 1 suite passes
- [ ] `bun run test:e2e` (Tier 2) — requires EVALS secret, maintainer-side gate
- [ ] Manual: Developer Mode ON → `ls -la ~/.claude/skills/qa/SKILL.md` shows symlink (`lrwxrwxrwx`)
- [ ] Manual: Developer Mode OFF → `ls -la` shows regular file (`-rw-r--r--`)

### Notes for reviewer

- No `chore: regenerate llms.txt` commit: the local modification produces no content diff in `gstack/llms.txt` (the file is generated from SKILL.md frontmatter, not from `setup`).
- No Windows path edge cases covered (spaces, non-ASCII, UNC). Follow-up if reported.
- CI's `windows-setup-e2e.yml` will rebuild binaries on this branch's push (paths trigger includes `setup`).

### DIFF
--- a/setup
+++ b/setup
@@ -44,7 +44,6 @@ _link_or_copy() {
   local src="$1"
   local dst="$2"
   if [ "$IS_WINDOWS" -eq 1 ]; then
-    rm -rf "$dst"
     # Unix `ln -snf` accepts a name-only or relative-path source even when the
     # target doesn't resolve from CWD (e.g. the connect-chrome alias points at
     # the sibling-relative "gstack/open-gstack-browser"). On Windows the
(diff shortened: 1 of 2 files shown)


TTmo123:
trunk merge

TTmo123:
trunk merge



At 2026-06-28 22:26:55, "trunk-io[bot]" ***@***.***> wrote:

trunk-io[bot] left a comment (garrytan/gstack#2129)

Merging to main in this repository is managed by Trunk.

To merge this pull request, check the box to the left or comment /trunk merge below.

After your PR is submitted to the merge queue, this comment will be automatically updated with its status. If the PR fails, failure details will also be posted here

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

PR #2131, opened 2026-06-28
Clean fixture terminology

apex-system:
## Summary
- removes old vertical-specific fixture terminology from security bench response fixtures

## Verification
- fixture-only text cleanup


time-attack:
@16francej manually changes text inside a security benchmark no generating timestamps hasshes signals or anything. pls close

PR #2132, opened 2026-06-29
fix(memory-ingest): render Codex rollout message bodies

### LINKED ISSUE #2105: memory-ingest renders all Codex rollout sessions as empty bodies (parser expects payload.message; current format is payload.type=='message' + payload.content)
## Summary
`gstack-memory-ingest` renders **every Codex rollout session with an empty body** (`message_count: 0`, zero-length content). All Codex transcripts import as empty shells.

## Root cause
`parseTranscriptJsonl` (bin/gstack-memory-ingest.ts) detects Codex correctly (`first.type === 'session_meta'`) but its message loop extracts content via:
```js
} else if (isCodex && rec?.payload?.message) { const msg = rec.payload.message; ... }
```
Current Codex rollout JSONL has no `payload.message`. Records are shaped:
```
{ type: 'response_item', payload: { type: 'message', role: 'user'|'assistant'|'developer', content: [...] } }
```
So the branch never fires → 0 messages → empty page body for every Codex session.

## Repro
Ingest any `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`; the resulting `transcripts/codex/...` page has `message_count: 0` and an empty body. Confirmed across 243/243 Codex sessions on this machine.

## Suggested fix
Handle the `response_item` / `payload.type === 'message'` shape: read `payload.role` and flatten `payload.content` (array of `{type, text}` items) into the body, mirroring the Claude Code branch.

## Env
gstack 1.58.4.0, Codex CLI rollout format (June 2026), Windows 10.

genisis0x:
Closes #2105.

## Problem

`gstack-memory-ingest` renders every Codex rollout session with an empty body (`message_count: 0`). Current Codex rollout JSONL carries each turn as a `response_item` record with the message on `payload.{type:"message", role, content[]}`:

```
{ "type": "response_item", "payload": { "type": "message", "role": "user"|"assistant"|"developer", "content": [...] } }
```

but `parseTranscriptJsonl` (`bin/gstack-memory-ingest.ts`) only extracts content when `rec.payload.message` is present (the older shape). That branch never fires on the current format, so the message loop counts zero turns and stages an empty page for every Codex transcript.

## Fix

Handle the current `payload.type === "message"` shape alongside the legacy `payload.message` wrapper: read `payload.role` and flatten `payload.content` through the existing `extractContentText` helper. Legacy sessions keep working.

## Test

Adds a product-path regression in `test/gstack-memory-ingest.test.ts` that ingests a `response_item`-format Codex session through the full `--bulk` path (fake-gbrain staging shim, same pattern as the NUL-byte test) and asserts the staged transcript renders both the user and assistant turns. It fails on the old parser (staged body empty, `message_count: 0`) and passes with the fix.

`bun test test/gstack-memory-ingest.test.ts` -> 24 pass. No template/SKILL.md change, so no regeneration; no VERSION bump (same as other bin/skill bug-fix PRs).

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -541,6 +541,12 @@ interface ParsedSession {
   partial: boolean;
 }
 
+// Roles whose content may be ingested as memory. `developer` (and any other
+// system-level role) carries instructions to the agent, not user memory, so it
+// must never be captured — otherwise an injected developer instruction in a
+// rollout would resurface later as trusted memory content.
+const INGESTIBLE_ROLES = new Set(["user", "assistant"]);
+
 function parseTranscriptJsonl(path: string): ParsedSession | null {
   // Best-effort tolerant parser. Handles truncated last lines (D10 partial-flag).
   let raw: string;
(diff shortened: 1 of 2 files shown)


time-attack:
@genisis0x Please enforce a role boundary in the current-schema parser so developer instructions cannot be ingested as memory content.

PR #2139, opened 2026-06-30
fix: Chrome numbered-profile cookie discovery + headed window stability on macOS 26

### LINKED ISSUE #2138: headed mode: Chromium window closes immediately on macOS 26 (Sequoia)
## Bug report

**gstack version:** 1.34.1.0 (upgrade to 1.58.5.0 available but not yet applied)
**OS:** macOS Darwin 25.5.0 (macOS 26 / Sequoia)
**Running inside:** Claude Code VSCode extension

---

### What happens

Using `$B connect` (headed mode / `connect-chrome` skill), gstack reports a successful connection, but the visible Chromium window disappears within seconds of opening. There is no time to interact with it.

### Reproduction steps

1. Run `$B connect` inside Claude Code (VSCode extension).
2. gstack prints "Connected to real Chrome", shows a PID, reports status healthy.
3. A Chromium window appears briefly on screen.
4. The window closes on its own within a few seconds (no user interaction).
5. On the next `$B disconnect` attempt, gstack reports "server was unresponsive — force cleaned".

### What works

- `$B connect` returns success (PID shown, status healthy).
- `$B goto <url>` navigates successfully (returns HTTP 200).
- `$B focus` returns "Browser window activated".
- Everything works at the API level; only the visible window disappears.

### Additional context: Chrome profile mismatch

Regular Google Chrome is installed with multiple profiles, but none of them is the `Default` profile. The existing profiles are numbered: Profile 4, 9, 10, 11, 12, 13.

gstack appears to look for `Default/Cookies` (or similar), which does not exist. This may be contributing to the crash or early exit if gstack cannot find or load the expected profile directory and then abandons the process silently.

### Expected behavior

The Chromium window should stay open and remain interactive for the duration of the session, consistent with the behavior documented for headed / `connect-chrome` mode.

### Attempted workarounds

Reproduced multiple times with identical results. The window never stays open long enough to interact with.

---

Happy to provide any additional diagnostic output (e.g., gstack server logs, Chromium crash logs from `~/Library/Logs/DiagnosticReports/`) if that would help narrow this down.

chrisreedbates:
## Summary

Fixes two bugs reported in issue #2138, both affecting macOS 26 (Sequoia).

### Bug 1 — Headed Chromium window closes immediately on macOS 26

**Root cause (two interacting issues):**

1. macOS 26 tightened window activation policy. Chromium windows spawned by a
   non-app process (the Bun server) are subject to a new OS-level window-hiding
   behavior that causes them to close within seconds unless properly registered
   via `NSApplicationActivateIgnoringOtherApps`.

2. `raiseHeadedWindowMacOS()` in `cli.ts` tried to activate "Google Chrome for
   Testing" via osascript, but `launchHeaded()` patches the Info.plist to rename
   the app to "GStack Browser". The activation silently failed, leaving the window
   unraised and subject to the OS hiding it.

**Fixes in `browse/src/browser-manager.ts` and `browse/src/cli.ts`:**

- Add `--disable-features=MacControlledWindowBehavior` to headed launch args on
  darwin. This flag suppresses the new OS-level window-hiding behavior introduced
  in macOS 26 while Playwright's NSApp startup sequence handles activation. No-op
  on macOS < 26 and all non-macOS platforms.
- Fix `raiseHeadedWindowMacOS()` to try "GStack Browser" first (post-patch name)
  with fallback to "Google Chrome for Testing" (pre-patch / patch skipped).

### Bug 2 — `cookie-import-browser` fails when no Default profile exists

**Root cause:** Chrome on macOS 26 and some multi-profile installations never
create a `Default` directory. Every account lands in a numbered profile (`Profile
4`, `Profile 12`, etc.). `cookie-import-browser.ts` only looked for
`Default/Cookies`, throwing "Chrome is not installed" even when cookies existed.

**Fix in `browse/src/cookie-import-browser.ts`:**

- When `getBrowserMatch()` can't find `Default`, call new
  `findFirstNumberedProfile()` which scans for `Profile N` directories, sorts by
  number (lowest first so the primary account wins), and returns the first one
  with a Cookies file.
- Same fallback logic applies on macOS, Linux, and Windows.
- Explicit non-default profile names (e.g. `importCookies('chrome', [...], 'Profile 4')`)
  still fail fast as before — the fallback only fires for the default `'Default'` case.

## Test plan

- [ ] `$B connect` on macOS 26: headed Chromium window stays open and
  `$B screenshot` succeeds (no "Cannot take screenshot with 0 width")
- [ ] `$B disconnect` on macOS 26 reports clean shutdown, not "server was
  unresponsive — force cleaned"
- [ ] `cookie-import-browser chrome` on a Mac with only numbered profiles (no
  `Default` directory): succeeds and imports cookies from the lowest-numbered
  profile
- [ ] `cookie-import-browser chrome` on a standard install with `Default`
  profile: still works (no regression)
- [ ] New unit tests pass: `bun test browse/test/cookie-import-browser.test.ts`

## Files changed

- `browse/src/cookie-import-browser.ts` — `getBrowserMatch` + new `findFirstNumberedProfile`
- `browse/src/browser-manager.ts` — macOS 26 `--disable-features` flag in `launchHeaded` and `handoff`
- `browse/src/cli.ts` — fix `raiseHeadedWindowMacOS` to try both app names
- `browse/test/cookie-import-browser.test.ts` — 3 new tests for numbered-profile fallback

Closes #2138

---

## Local test evidence (macOS 26 / Darwin 25.5.0)

**Environment:** macOS 26 Sequoia (Darwin 25.5.0), running inside Claude Code VSCode extension.

**Test session:** After applying the `--disable-features=MacControlledWindowBehavior` fix, launched
a headed Chromium window via `$B connect` and left it running for 10+ minutes. The window
remained open and interactive for the entire session. 23 Google Search Console indexing requests
were submitted successfully through the headed window with no crashes or premature closures.

**Before this fix:** The window closed within seconds of opening every time.

---

## Second problem found: `handoff()` path missing the macOS 26 flag

After the main fix was verified working, a code audit found that `handoff()` in
`browser-manager.ts` (the headless-to-headed re-launch path, lines ~1560-1600) builds
its own `launchArgs` array independently from `launchHeaded()` and was NOT including
`--disable-features=MacControlledWindowBehavior`.

A user who triggers a handoff (e.g., via the gstack extension "hand off to user" UI flow)
on macOS 26 would have seen the same immediate window-close bug, even after the
`launchHeaded()` fix.

**Fix applied in commit `5c7058bb`:** Added the same `if (process.platform === 'darwin')` guard to
the `handoff()` launchArgs block, immediately after the initial `launchArgs` array is built:

```typescript
// macOS 26: same window-hiding suppression as launchHeaded(). The handoff
// path launches a fresh headed Chromium and is subject to the same
// MacControlledWindowBehavior that causes the window to close immediately
// on Sequoia. No-op on macOS < 26 and on non-macOS platforms.
if (process.platform === 'darwin') {
  launchArgs.push('--disable-features=MacControlledWindowBehavior');
}
```

Build verified: `bun run build` passes cleanly after the change.

Note: the fix commit `5c7058bb` is local only — push access to `garrytan/gstack` was denied
(403). The fix is in the local worktree at `/tmp/gstack-pr/browse/src/browser-manager.ts`
and needs to be pushed by a repo collaborator.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -1569,6 +1569,13 @@ export class BrowserManager {
       // STEALTH_LAUNCH_ARGS the handed-off browser kept the AutomationControlled
       // tell that the other two paths strip.
       const launchArgs: string[] = ['--hide-crash-restore-bubble', ...STEALTH_LAUNCH_ARGS, ...buildGStackLaunchArgs()];
+      // macOS 26: same window-hiding suppression as launchHeaded(). The handoff
+      // path launches a fresh headed Chromium and is subject to the same
+      // MacControlledWindowBehavior that causes the window to close immediately
+      // on Sequoia. No-op on macOS < 26 and on non-macOS platforms.
+      if (process.platform === 'darwin') {
+        launchArgs.push('--disable-features=MacControlledWindowBehavior');
+      }
       if (extensionPath) {
         launchArgs.push(`--disable-extensions-except=${extensionPath}`);
         launchArgs.push(`--load-extension=${extensionPath}`);
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -575,6 +575,17 @@ export class BrowserManager {
     // and the chrome-runtime shape changes Playwright otherwise triggers) and
     // three more (--disable-popup-blocking, --disable-component-update,
     // --disable-default-apps — each a documented automation tell per Patchright).
+    // macOS 26 (Sequoia) tightened window activation policy: Chromium windows
+    // spawned by a non-app process close immediately unless the app is properly
+    // registered with the window server via NSApplicationActivateIgnoringOtherApps.
+    // --disable-features=MacControlledWindowBehavior suppresses the new OS-level
+    // window-hiding behavior introduced in macOS 26 while Playwright handles
+    // activation via its own NSApp startup sequence.
+    // This flag is a no-op on macOS < 26 and on non-macOS platforms.
+    if (process.platform === 'darwin') {
+      launchArgs.push('--disable-features=MacControlledWindowBehavior');
+    }
+
     const { STEALTH_IGNORE_DEFAULT_ARGS } = await import('./stealth');
     this.context = await chromium.launchPersistentContext(userDataDir, {
       headless: false,
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -275,14 +275,32 @@ export function buildRestartEnv(
 }
 
 /** macOS only: pull the headed Chromium window to the user's current Space.
- * "Google Chrome for Testing" frequently opens behind the active window or on
- * another Space — the first thing users read as "I can't see the browser"
- * (#1781). Best-effort, fire-and-forget, never throws. The app name is a fixed
- * literal (no interpolation). */
+ * The window frequently opens behind the active window or on another Space —
+ * the first thing users read as "I can't see the browser" (#1781).
+ *
+ * launchHeaded() patches the Chromium .app's Info.plist to rename it from
+ * "Google Chrome for Testing" to "GStack Browser" for better branding. That
+ * patch means the old hard-coded name no longer works for the osascript
+ * activate call. We try both names: "GStack Browser" first (post-patch),
+ * then "Google Chrome for Testing" as a fallback (pre-patch / patch skipped).
+ *
+ * Best-effort, fire-and-forget, never throws. App names are fixed literals —
+ * no interpolation. */
 function raiseHeadedWindowMacOS(): void {
   if (process.platform !== 'darwin') return;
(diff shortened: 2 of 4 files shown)


time-attack:
@chrisreedbates Please separate profile discovery from headed-launch changes, address the signed-bundle and watchdog failures, and make tests use isolated fixtures rather than real user files.

garrytan:
The headed-stability half is the same root cause as #2242 (the rebrand invalidated the signed Chromium bundle) — fixed in fork-port wave 2 by `a30ca53c` (rebrand removal + poisoned-bundle self-heal) and `fb20ada4` (v1.65 migration re-fetch), pinned by `browse/test/rebrand-signed-bundle.test.ts`.

Numbered-profile cookie discovery (Profile 1, Profile 2, ...) already landed on main — `browse/src/cookie-import-browser.ts` scans numbered profiles.

Ships in #2577 (v1.65.0.0).

PR #2141, opened 2026-06-30
review: catch loosened-input consumers and stale user-facing strings

aversini:
## What

Two additions to the pre-landing review checklist (`review/checklist.md`):

1. Broadens the **Enum & Value Completeness** critical category to **Enum, Value & Input-Acceptance Completeness**. It already tells the reviewer to trace a new value through every consumer *outside the diff*, but the same blind spot applies when a change *loosens* what an input accepts (a new MIME type, file extension, flag, or a widened validator). The code that breaks is usually unchanged, so it's invisible to a diff-scoped pass. One new bullet covers that case, with a concrete example.

2. Adds a small informational category, **Stale User-Facing Strings**: when the logic guarding a message changes but the string doesn't, the string can quietly become wrong. Same family as the existing stale-CHANGELOG note, applied to in-code copy.

## Why

Both came out of a real review miss. A client validator was loosened to accept a typeless `.pdf`, but an unchanged submit helper still forced every typeless file to `text/plain`, so the file got sent as garbage text instead of natively. A separate "Total file size…" message also went stale after the logic behind it changed to count only some files. The checklist pass missed both because the broken code (and the stale string) never showed up in the diff. An external reviewer caught them.

Docs only, no behavior change.

### Note
If you take the heading rename in (1), `review/SKILL.md.tmpl` lists the critical categories by name ("…Shell Injection, Enum & Value Completeness") and would want the same rename to stay in sync. Happy to fold that in, or to keep the old heading and just add the bullet if you'd rather not touch two files.

### DIFF
--- a/review/checklist.md
+++ b/review/checklist.md
@@ -58,11 +58,12 @@ Be terse. For each issue: one line describing the problem, one line with the fix
 - `os.system()` with variable interpolation — replace with `subprocess.run()` using argument arrays
 - `eval()` / `exec()` on LLM-generated code without sandboxing
 
-#### Enum & Value Completeness
-When the diff introduces a new enum value, status string, tier name, or type constant:
+#### Enum, Value & Input-Acceptance Completeness
+When the diff introduces a new enum value, status string, tier name, or type constant — OR loosens what an input accepts (a newly-accepted MIME type, file extension, or flag; a widened validator/parser/guard):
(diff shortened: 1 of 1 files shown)


time-attack:
Thanks for this. The renamed category is not reflected in the template and generated outputs, leaving the documentation inconsistent. Please either update and regenerate every affected reference or retain the existing heading while adding the new bullets.

PR #2146, opened 2026-07-01
fix: pre-empt broken AskUserQuestion on the Claude Desktop host

kalpajit279:
## Problem

In the **Claude Desktop app**, `AskUserQuestion` is advertised and enabled (`CLAUDE_CODE_ENABLE_ASK_USER_QUESTION_TOOL=true`, `CLAUDE_CODE_ENTRYPOINT=claude-desktop`) but its handler returns `[Tool result missing due to internal error]` — the same class of host bug already documented for Conductor. The model gets a raw dead-end error every time it tries to ask the user a question.

## Why the existing fallbacks don't catch it

- `question-preference-hook` (PreToolUse) only runs its transport-avoidance redirect for `isConductor()`, so a desktop session falls through to `defer` and the doomed call proceeds.
- `auq-error-fallback-hook` (PostToolUse) is designed to inject the prose-fallback directive on this error — but **Claude Code does not fire PostToolUse hooks on the missing-result/transport-error path** (the hook's own header comment flags this as unverified; it's now confirmed on `claude-desktop`). So the only place that can intercept is PreToolUse.

## Fix

Extend the existing PreToolUse transport-avoidance to also fire when `CLAUDE_CODE_ENTRYPOINT === 'claude-desktop'`, redirecting to the same prose decision brief used for Conductor. The reason label is generalized (`[conductor]` / `[claude-desktop]`); all the existing prose-brief instructions are unchanged. `isConductor()` stays semantically pure (its test still asserts it only keys on `CONDUCTOR_*`).

## Tests

- Added a `Claude Desktop prose redirect` describe block in `test/question-preference-hook.test.ts`, paralleling the Conductor cases (two-way deny, one-way deny, never-ask auto-decide still wins, non-AUQ still defers).
- Mirrored the existing ambient-Conductor-strip with `delete env.CLAUDE_CODE_ENTRYPOINT` in the two hook test runners (`question-preference-hook.test.ts`, `memory-cache-injection.test.ts`), so running the suite **inside** Claude Desktop doesn't flip every `defer` into the prose deny.

## Verification

On `claude.exe 2.1.197` (Windows): the identical `AskUserQuestion` call that previously returned `[Tool result missing due to internal error]` now returns the `[claude-desktop]` prose-redirect `deny` from PreToolUse, so the model renders a prose decision brief instead of hitting the error. Direct hook invocation confirms `deny` for AUQ and `defer` for non-AUQ tools on the desktop host.

### DIFF
--- a/hosts/claude/hooks/question-preference-hook.ts
+++ b/hosts/claude/hooks/question-preference-hook.ts
@@ -451,15 +451,25 @@ async function main(): Promise<void> {
     return;
   }
 
-  // Not fully auto-decidable. In Conductor, AskUserQuestion is unreliable
-  // (native is disabled, the mcp__conductor__AskUserQuestion variant is flaky),
-  // so deny the tool and redirect to a prose decision brief. This is TRANSPORT
-  // AVOIDANCE, not preference enforcement: it fires regardless of marker,
-  // preference, or door type — including one-way doors, which must reach the
-  // human via prose rather than the unreliable tool.
-  if (isConductor()) {
-    const conductorReason =
-      '[conductor] AskUserQuestion is unreliable in Conductor (native disabled, MCP variant flaky). ' +
+  // Not fully auto-decidable. On some hosts AskUserQuestion is unreliable and
+  // must be avoided entirely:
+  //   - Conductor: native AUQ is disabled and the mcp__conductor__AskUserQuestion
+  //     variant is flaky.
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej Closing because blanket Desktop denial unnecessarily suppresses native questions; the neutral pass-through correction in #2165 addresses the failure without changing that flow.

PR #2150, opened 2026-07-02
fix: skill:check no longer fails on host-skipped skill templates

srmcno:
## What was broken

`bun run skill:check` exits 1 on a pristine checkout of main. The template-coverage loop in `scripts/skill-check.ts` flags `claude/SKILL.md` as "generated file missing! Run: bun run gen:skill-docs" — but running that command doesn't (and shouldn't) create the file: `hosts/claude.ts` declares `generation.skipSkills: ['claude']`, since the `/claude` outside-voice wrapper (added in #2078) is only generated for non-Claude hosts.

## The fix

The template loop now consults `getHostConfig('claude').generation.skipSkills` (same registry already imported for `getExternalHosts`) and prints an informational line for host-skipped skills instead of an error:

```
  -  claude/SKILL.md                — not generated for claude host (skipSkills)
```

## Verification

In a fresh worktree of main with all host files generated (`gen-skill-docs --host <h>` for each external host):

- **Without the fix:** `skill:check` exits 1 with exactly one ❌ (`claude/SKILL.md — generated file missing`)
- **With the fix:** exits 0, zero ❌, informational skip line present
- `bun test` (free tier) passes

## Note on CI

This is a fork PR, so per the repo's CLAUDE.md, eval/E2E jobs will fail with empty-env auth errors — that's the fork-secrets issue, not this change. check-freshness, workflow-lint, and windows-tests should pass.

### DIFF
--- a/scripts/skill-check.ts
+++ b/scripts/skill-check.ts
@@ -65,7 +65,17 @@ for (const file of SKILL_FILES) {
 console.log('\n  Templates:');
 const TEMPLATES = discoverTemplates(ROOT);
 
+// Skills the claude host never generates (e.g. the /claude outside-voice
+// wrapper, which only exists for non-Claude hosts) — a missing output for
+// these is intentional, not staleness.
+const CLAUDE_SKIP_SKILLS = new Set(getHostConfig('claude').generation.skipSkills ?? []);
+
 for (const { tmpl, output } of TEMPLATES) {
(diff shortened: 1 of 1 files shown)


time-attack:
@16francej — recommend closing this PR as superseded by #2059. This patch is a narrow primary-host skip guard without review, checks, or a focused regression, while #2059 implements the broader policy-aware behavior. Preserve #2150 contributor attribution and any clearer diagnostics in the positive and negative required-output coverage.

Issue #2019, opened 2026-06-15
Orphan terminal-agent.ts bun processes + visible conhost windows accumulate on Windows


kolibriprojekt:
### Summary

On Windows, every gstack session that spawns a `terminal-agent.ts` daemon leaves behind a visible `conhost.exe` window. When a Claude Code / gstack session ends abruptly (window closed via X, crash, force-kill), the `bun.exe` daemon and its `conhost.exe` survive and accumulate over time. After a few days of daily use I had 3 orphan processes from old sessions, each with a visible terminal window.

### Repro

1. Windows 11, Claude Code with gstack installed.
2. Run any `/browse` command — gstack spawns `bun run terminal-agent.ts` via `terminal-agent-control.ts:71` (`Bun.spawn(['bun', 'run', script], { stdio: ['ignore', 'ignore', 'ignore'] })` with `proc.unref()`).
3. A visible `conhost.exe` window opens (Windows default for console subprocesses).
4. Close the Claude Code window without `/exit` (or let it crash).
5. `Get-Process bun` shows the agent still running; the conhost window is still visible.
6. Repeat over several sessions → orphans pile up.

In my install:

```
> Get-Content ~/.gstack/terminal-agent-pid
{"pid":27492,"gen":"...","startedAt":1780844984663}   # tracked agent (8 days old, long dead)

> Get-CimInstance Win32_Process -Filter "Name='bun.exe'"
PID 31700, 37116, 37272   # all running terminal-agent.ts, none tracked
```

Three live daemons, none in the pid-record. The current spawn site only kills the one PID recorded in `terminal-agent-pid` (via `killAgentByRecord` in `terminal-agent-control.ts:65-68`), so orphans from prior crashed sessions are never reaped.

### Two issues, two fixes

**1. Orphan accumulation.** `spawnTerminalAgent` reads the pid-record and kills only that one PID. Anything from a prior session that wasn't cleanly torn down survives. Suggested fix: at boot, after killing the recorded PID, scan for additional bun processes running `terminal-agent.ts` whose owner matches (current uid / SID) and whose parent process is dead, and reap those too. Identity-based (the v1.44 design) is preserved — we'd still never touch a sibling gstack session whose parent is alive.

[...]

PR #2153, opened 2026-07-02
fix(windows): node-path console flashes + watchdog split-brain — polyfill windowsHide, signal-0 isProcessAlive, trippable respawn guard

### LINKED ISSUE #1952: terminal-agent watchdog flashes a console window every tick on Windows — isProcessAlive() spawns tasklist instead of signal-0

### LINKED ISSUE #2151: Windows: bun-polyfill.cjs spawn/spawnSync drop windowsHide, and the terminal-agent watchdog duplicates live agents (node-path gaps #1952/#2019 won't cover)
## TL;DR

On Windows the browse server always runs under **Node.js** (`dist/server-node.mjs` + `dist/bun-polyfill.cjs` — `cli.ts` hard-fails without the bundle because "the Bun path is known broken"). Two node-path defects that the open Windows-flash issues don't cover:

1. **`browse/src/bun-polyfill.cjs` `spawn()`/`spawnSync()` drop `windowsHide` (and every other option) on the floor.** They forward only `stdio`/`env`/`cwd`(/`timeout`). Node's `child_process` default is `windowsHide: false`, so every process the console-less daemon spawns gets a visible console window. Critically, this means the fix proposed in #2019 (add `windowsHide: true` to the `Bun.spawn` call in `terminal-agent-control.ts:71`) **will be silently swallowed on Windows**, because on Windows that call never reaches Bun — it reaches this polyfill, which won't forward the option.

2. **The v1.44 terminal-agent watchdog respawns around a live agent and never reaps it (split-brain), observed live.** I caught the daemon spawning a second `bun run terminal-agent.ts` exactly one watchdog tick (67s) after the first, with **both agents alive simultaneously** minutes later. Each respawn = a fresh visible `bun.exe` console window (defect 1), which is how it surfaced: my report was literally "bun.exe windows keep flashing open and closing".

## Live incident timeline (Windows 11, gstack v1.58.4.0)

A `/qa`-style browse session against a local app left this state:

```
17:07:07  node …\browse\dist\server-node.mjs starts (PID 6048, parent already exited)
17:09:24  bun.exe spawned: bun run …\browse\src\terminal-agent.ts   (PID 100448)
17:10:31  bun.exe spawned: bun run …\browse\src\terminal-agent.ts   (PID 111648)  ← +67s ≈ one AGENT_WATCHDOG_TICK_MS
17:11+    BOTH bun agents alive concurrently (verified via Win32_Process); PID 100448 was never killed
17:10:32  <project>\.gstack\claude-available.json written (agent 2 boot) — matches terminal-agent.ts main() step 1
```

The 17:10:31 respawn means the watchdog tick evaluated `record && isProcessAlive(record.pid)` as false while agent 1 (100448) was demonstrably alive. Then in `spawnTerminalAgent` (`terminal-agent-control.ts:63-68`) the prior-kill is gated on the **same** false signal — `killAgentByRecord` → `isProcessAlive` → false → returns without signalling — so the live agent survives as an orphan, its record is wiped by `clearAgentRecord`, and the duplicate registers in its place. That is exactly the split-brain the v1.44 comment in `server.ts` says the identity-based design exists to prevent ("prevents respawning around a slow-but-alive agent").

Standalone verification: `bun run terminal-agent.ts` with a fresh `BROWSE_STATE_FILE` on this same machine boots correctly and writes `terminal-port`, `terminal-agent-pid`, and `terminal-internal-token` within ~1s — the agent itself is fine on Windows; the supervision path is what breaks.

### 1. Polyfill spawn surface (`browse/src/bun-polyfill.cjs`)

```js
spawn(cmd, options = {}) {
  const [command, ...args] = cmd;
  const stdio = options.stdio || ['pipe', 'pipe', 'pipe'];
  const proc = spawn(command, args, { stdio, env: options.env, cwd: options.cwd });  // no windowsHide, no detached
  ...
},
spawnSync(cmd, options = {}) {
  ...
  const result = spawnSync(command, args, { stdio: [...], timeout: options.timeout, env: options.env, cwd: options.cwd });  // same
```

Every daemon-side spawn on Windows goes through this: the watchdog's `bun run terminal-agent.ts` respawn, `tasklist` liveness probes (#1952), `git rev-parse`, chrome `--version` probing. Each one pops a console window from a daemon that has no console of its own.

### 2. False-dead probes funnel into destructive respawn

Two paths make a live agent read as dead, and both end in "clear its record + spawn a duplicate + skip the kill":

- **Probe timeout → false.** The polyfill's `spawnSync` passes `timeout: 3000`. Node's `spawnSync` does **not throw** on timeout — it returns `result.error = ETIMEDOUT` with `status: null` — and the polyfill returns `stdout || Buffer.from('')`, so `isProcessAlive`'s `.includes()` check quietly returns false. `tasklist` on a loaded Windows box (this one had ~40 node processes, dev servers, and eval runs going) can exceed 3s.
- **Slow registration.** `writeAgentRecord` runs late in `terminal-agent.ts` `main()` (after `Bun.serve` binds, after two `writeSecureFile` calls that each shell out to `icacls` on Windows — more hidden-console subprocesses). Any boot slower than the gap to the next tick reads as "no record → dead".

Compounding it: the crash-loop guard (`RESPAWN_GUARD_MAX = 3` within `RESPAWN_GUARD_WINDOW_MS = 60_000`) can never trip, because the tick interval is itself 60s — at most ~1 respawn ever lands inside the window. The cycle repeats for the daemon's whole lifetime.

## Why this is distinct from the open issues

- #1952 covers the `tasklist` console flash per tick and proposes signal-0 — correct, and it removes that one spawn. It doesn't cover the polyfill option-dropping, so every *other* `Bun.spawn`/`Bun.spawnSync` on the node path keeps flashing.
- #2019 covers orphan accumulation across *crashed sessions* and proposes `windowsHide: true` at the Bun-runtime call site. This issue is (a) duplication by the watchdog *within a single healthy session*, and (b) the fact that #2019's fix as proposed won't take effect on Windows because `bun-polyfill.cjs` doesn't forward the option.
- #1784 / #1989 are the Bun-runtime and vendored-playwright spawn surfaces respectively; this is the third surface (node polyfill).

## Proposed fixes

1. **`bun-polyfill.cjs`**: add `windowsHide: true` unconditionally (both `spawn` and `spawnSync`), and forward `detached` — the daemon's children are all headless by design, so there's no downside. This is a prerequisite for #2019's call-site fix to work on Windows.
2. **`isProcessAlive`**: unify on `process.kill(pid, 0)` with `EPERM ⇒ alive` per #1952 — kills the probe subprocess, the 3s-timeout false-dead path, and the per-tick flash in one line.
3. **Watchdog/`spawnTerminalAgent`**: don't gate the prior-kill on the same probe that just returned false — `safeKill` already swallows `ESRCH`, so signalling a genuinely-dead PID is harmless, while skipping a live one creates split-brain. And widen `RESPAWN_GUARD_WINDOW_MS` to ≥ 3 ticks so the guard can actually trip.

Happy to send a PR for (1) and (3) if useful — (2) presumably lands with #1952.

## Env

- gstack: v1.58.4.0 local install (polyfill unchanged on main @ v1.58.5.0)
- Bun: 1.3.14, Node: 22.x (`server-node.mjs` path)
- Host: Claude Code

osouthgate:
Fixes #2151. Fixes #1952.

## What

Three small fixes for the Windows node path (`server-node.mjs` + `bun-polyfill.cjs`), one commit each:

1. **`bun-polyfill.cjs`: pass `windowsHide: true` in `spawn`/`spawnSync`, forward `detached`.** The polyfill forwarded only `stdio`/`env`/`cwd`, and node's `child_process` defaults `windowsHide` to false — so every subprocess the console-less daemon spawned (terminal-agent respawns, `tasklist` probes, `git`, `icacls`) flashed a visible console window. This is also why call-site fixes like the one proposed in #2019 would have been silently swallowed on Windows: on the node path `Bun.spawn` is this polyfill, and it dropped the option.

2. **`isProcessAlive`: signal-0 on all platforms, replacing the Windows `tasklist` branch** — the fix proposed in #1952 (credit to that report). Beyond the per-tick console flash, the `tasklist` branch had a correctness bug: node's `spawnSync` reports a timeout via `result.error` *without throwing*, so a slow `tasklist` (>3s on a loaded box) returned empty stdout and read as **"dead" for a live agent**. The watchdog then cleared the live agent's record and spawned a duplicate around it, while `killAgentByRecord` consulted the same false probe and skipped the kill — the split-brain documented in #2151 (two live agents one tick apart, prior never reaped). Signal-0 removes the subprocess, the flash, and the false-dead path in one line, and fixes the old Unix branch collapsing EPERM (exists-but-unsignallable) to "dead".

3. **Respawn guard window spans `guard-max` ticks.** The crash-loop guard (3 respawns in 60s → stop, require manual restart) could never trip: respawn attempts land at most once per tick and the window equaled the tick interval, so the rolling history never held more than one entry. A crash-looping agent respawned once a minute indefinitely. Now `RESPAWN_GUARD_WINDOW_MS = Math.max(60_000, AGENT_WATCHDOG_TICK_MS * RESPAWN_GUARD_MAX)`.

`killAgentByRecord`'s liveness gate is deliberately **untouched**: with a reliable probe (fix 2) the gate's PID-reuse defense is harmless again, so the design pinned by `terminal-agent-pid-identity.test.ts` tests 5–6 stands.

## Verification

- **Signal-0 on Windows** (Windows 11, node 22 / bun 1.3.14): alive PID → `true`, dead PID → `false`, PID 4 (System, unsignallable) → EPERM → `true`. Verified under both runtimes.
- **Live incident** (from #2151): with the unpatched install, the daemon respawned a duplicate agent one tick after the first, both stayed alive, and every spawn flashed a console window. With these patches applied to the local install (`src` + `dist` bundle), spawn/probe subprocesses no longer open windows and the polyfill smoke-tests pass under node.
- **Tests**: new static tripwires pin `windowsHide: true` in both polyfill spawn paths and reject any `tasklist` invocation in `error-handling.ts`; new EPERM⇒alive case; watchdog static test updated for the new window expression. On this Windows machine the branch *improves* the suite vs upstream main (e.g. `server-embedder-terminal-port.test.ts` goes 1 pass/3 fail → 3 pass/1 fail; every remaining failure reproduces identically on clean upstream main — pre-existing `new URL(import.meta.url).pathname` path mangling on Windows in the static tests, unrelated to this diff).

## Not covered (left to their own issues)

- Cross-session orphan reaping from #2019 (this PR fixes the *within-session* duplication trigger; stale agents from crashed sessions still need the boot-time sweep proposed there). #2019's call-site `windowsHide` for the Bun-runtime path also composes cleanly on top of this.
- The Bun-runtime `Bun.spawnSync` flash surfaces in #1784/#1989.

### DIFF
--- a/browse/src/bun-polyfill.cjs
+++ b/browse/src/bun-polyfill.cjs
@@ -75,6 +75,9 @@ globalThis.Bun = {
       timeout: options.timeout,
       env: options.env,
       cwd: options.cwd,
+      // The browse server runs console-less on Windows; without this every
+      // subprocess (tasklist, git, icacls) pops a visible console window.
+      windowsHide: true,
     });
 
     return {
@@ -91,6 +94,10 @@ globalThis.Bun = {
       stdio,
       env: options.env,
       cwd: options.cwd,
+      detached: options.detached,
+      // Same as spawnSync: detached daemons (terminal-agent respawns) must
+      // not flash a console window on Windows.
+      windowsHide: true,
     });
 
     return {
--- a/browse/src/error-handling.ts
+++ b/browse/src/error-handling.ts
@@ -7,8 +7,6 @@
 
 import * as fs from 'fs';
 
-const IS_WINDOWS = process.platform === 'win32';
-
 // ─── Filesystem ────────────────────────────────────────────────
 
 /** Remove a file, ignoring ENOENT (already gone). Rethrows other errors. */
@@ -36,23 +34,26 @@ export function safeKill(pid: number, signal: NodeJS.Signals | number): void {
   }
 }
 
-/** Check if a PID is alive. Pure boolean probe — returns false for ALL errors. */
+/**
+ * Check if a PID is alive. Pure boolean probe — never throws.
+ *
+ * signal-0 on every platform: Node and Bun both implement
+ * `process.kill(pid, 0)` on Windows via an OpenProcess existence check —
+ * no child process, no console window. The previous Windows branch spawned
+ * `tasklist`, which (a) flashed a visible console window from console-less
+ * daemons on every call (#1952), and (b) silently returned false for LIVE
+ * processes when tasklist exceeded its 3s timeout — node's spawnSync
+ * reports timeout via `result.error` without throwing, so the empty-stdout
+ * path read as "dead". A false "dead" makes the terminal-agent watchdog
+ * respawn around a live agent and orphan it (split-brain, #2151).
+ *
+ * EPERM ⇒ alive: the process exists but can't be signalled by this user.
+ */
 export function isProcessAlive(pid: number): boolean {
-  if (IS_WINDOWS) {
-    try {
-      const result = Bun.spawnSync(
-        ['tasklist', '/FI', `PID eq ${pid}`, '/NH', '/FO', 'CSV'],
-        { stdout: 'pipe', stderr: 'pipe', timeout: 3000 }
-      );
-      return result.stdout.toString().includes(`"${pid}"`);
-    } catch {
-      return false;
-    }
-  }
   try {
     process.kill(pid, 0);
     return true;
-  } catch {
-    return false;
+  } catch (err: any) {
+    return err?.code === 'EPERM';
   }
 }
--- a/browse/src/server.ts
+++ b/browse/src/server.ts
@@ -1504,9 +1504,9 @@ export function buildFetchHandler(cfg: ServerConfig): ServerHandle {
   // (which would create split-brain — two agents writing the port file,
   // tokens diverging between them, mystery PTY upgrade failures).
   //
-  // Crash-loop guard: 3 respawn attempts inside 60s → stop trying and emit
-  // a one-line error. Manual `forceRestart` from the sidebar clears the
-  // history (the user is the explicit signal to retry).
+  // Crash-loop guard: 3 respawn attempts inside 3 ticks → stop trying and
+  // emit a one-line error. Manual `forceRestart` from the sidebar clears
+  // the history (the user is the explicit signal to retry).
   //
   // Only active when ownsTerminalAgent === true. Embedders that pre-launch
   // their own PTY server (gbrowser phoenix overlay) must not be auto-respawned
@@ -1517,8 +1517,11 @@ export function buildFetchHandler(cfg: ServerConfig): ServerHandle {
     process.env.GSTACK_AGENT_WATCHDOG_TICK_MS || '60000',
     10,
   );
-  const RESPAWN_GUARD_WINDOW_MS = 60_000;
   const RESPAWN_GUARD_MAX = 3;
+  // The guard must span at least RESPAWN_GUARD_MAX ticks or it can never
+  // trip: respawn attempts land at most once per tick, so a 60s window
+  // over a 60s tick sees a single attempt forever (see #2151).
+  const RESPAWN_GUARD_WINDOW_MS = Math.max(60_000, AGENT_WATCHDOG_TICK_MS * RESPAWN_GUARD_MAX);
   let agentRespawnGuardTripped = false;
 
   if (ownsTerminalAgent) {
--- a/browse/test/bun-polyfill.test.ts
+++ b/browse/test/bun-polyfill.test.ts
@@ -1,4 +1,5 @@
 import { describe, test, expect, afterAll } from 'bun:test';
+import * as fs from 'fs';
 import * as path from 'path';
 
 // Load the polyfill into a fresh object (don't clobber globalThis.Bun)
@@ -46,6 +47,23 @@ describe('bun-polyfill', () => {
     expect(lines[0]).toBe('HAS_PID');
     expect(lines[1]).toBe('HAS_KILL');
     expect(lines[2]).toBe('HAS_UNREF');
+  });
+
+  test('spawn and spawnSync pass windowsHide (no console flash from console-less daemons)', () => {
+    // Static tripwire (#2151): on Windows the browse server always runs
+    // under Node via this polyfill, and node's child_process defaults to
+    // windowsHide: false — so every subprocess the console-less daemon
+    // spawns (terminal-agent respawns, git, icacls) pops a visible console
+    // window. Both spawn paths must pass windowsHide: true, and spawn must
+    // forward `detached` for daemon children. Behavioral assertion isn't
+    // possible cross-platform (window creation is invisible to the child),
+    // hence source-level pinning.
+    const src = fs.readFileSync(polyfillPath, 'utf-8');
+    const spawnSyncBlock = src.slice(src.indexOf('spawnSync(cmd'), src.indexOf('spawn(cmd'));
+    const spawnBlock = src.slice(src.indexOf('spawn(cmd'));
+    expect(spawnSyncBlock).toContain('windowsHide: true');
+    expect(spawnBlock).toContain('windowsHide: true');
+    expect(spawnBlock).toContain('detached: options.detached');
   });
 
   test('Bun.serve creates an HTTP server that responds', async () => {
--- a/browse/test/error-handling.test.ts
+++ b/browse/test/error-handling.test.ts
@@ -44,4 +44,28 @@ describe('isProcessAlive', () => {
   test('returns false for non-existent process', () => {
     expect(isProcessAlive(99999999)).toBe(false);
   });
+
+  test('returns true for a live but unsignallable process (EPERM ⇒ alive)', () => {
+    // PID 4 = System on Windows, PID 1 = init/launchd on POSIX. Both are
+    // always alive; signalling them as an unprivileged user throws EPERM,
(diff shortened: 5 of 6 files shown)


time-attack:
@osouthgate Please rewrite this into behaviorally covered changes. It currently combines spawn-option forwarding, disputed process-liveness semantics, and watchdog recovery without enough tests to establish safe behavior.

PR #2158, opened 2026-07-03
fix(browse): wait for web fonts before screenshots

CodeMocsh:
## Problem

Every screenshot capture path fires `page.screenshot()` immediately with no wait for web fonts. On any page using `@font-face` / Google Fonts, the shot lands mid-FOIT (flash of invisible/fallback text) — the fonts haven't downloaded yet, so text renders in a system fallback or blank. This shows up as "bad images" from `/qa`, `/design-review`, and `responsive` captures, especially on the first capture after navigation.

The `screenshot-size-guard` work already touched these same callsites; this is the timing gap right before them.

## Fix

New shared helper `browse/src/screenshot-utils.ts`:

```ts
export async function waitForFonts(page, timeoutMs = 3000) {
  // await document.fonts.ready, raced against a cap so lazy/streaming
  // fonts can't hang the capture; best-effort, never throws
}
```

Called before every full-page screenshot:

| File | Path |
|---|---|
| `meta-commands.ts` | `screenshot` (disk / `--base64` / `--clip` / `--selector`) |
| `meta-commands.ts` | `responsive` (per viewport — a resize can re-trigger font loads) |
| `write-commands.ts` | `prettyscreenshot` |
| `snapshot.ts` | annotated (`-a`) + heatmap |

It's best-effort and capped at 3s via `Promise.race`, so a page whose fonts never settle can't hang a capture, and a font-wait failure (navigation/context teardown race) never aborts the screenshot.

## Testing

- `browse/test/screenshot-utils.test.ts` — verifies the helper passes a numeric cap, defaults to a finite timeout, and never throws when `evaluate` rejects; plus a static invariant pinning that all three full-page callsites wire `waitForFonts` (mirrors the `screenshot-size-guard` test pattern).
- `bun test browse/test/screenshot-utils.test.ts` → 4 pass.
- `bash browse/scripts/build-node-server.sh` bundles cleanly.

### DIFF
--- a/browse/src/meta-commands.ts
+++ b/browse/src/meta-commands.ts
@@ -20,6 +20,7 @@ import * as path from 'path';
 import { writeSecureFile, mkdirSecure } from './file-permissions';
 import { TEMP_DIR } from './platform';
 import { resolveConfig } from './config';
+import { waitForFonts } from './screenshot-utils';
 import type { Frame } from 'playwright';
 
 /** Tokenize a pipe segment respecting double-quoted strings. */
@@ -496,6 +497,9 @@ export async function handleMetaCommand(
         throw new Error('Cannot use --viewport with --clip — choose one');
       }
 
+      // Let web fonts settle so text renders in the intended typeface, not a fallback.
+      await waitForFonts(page);
+
       // --base64 mode: capture to buffer instead of disk
       if (base64Mode) {
         let buffer: Buffer;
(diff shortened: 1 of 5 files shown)


CodeMocsh:
Closing this — it's based on a wrong root-cause assumption, and I want to be upfront about it rather than leave a misleading PR open.

I dogfooded it against a synthetic font-heavy page (distinctive web font served behind a controllable delay) using the same bundled Chromium. The result: **Playwright's `page.screenshot()` already waits for web fonts.** With a font deliberately held for 4s, `page.screenshot()` blocked the full 4032ms on its own and captured the correct font — so `waitForFonts` is redundant (before/after screenshots came out byte-identical).

It also doesn't help the real failure mode: when a font genuinely can't load (404 / blocked / unreachable in the headless context), you get the fallback typeface, and adding a `document.fonts.ready` wait produces a **pixel-identical** bad image. That reachability case is the actual cause of "bad fonts in screenshots," and it's not fixable by waiting.

Apologies for the noise. Closing.

Issue #1989, opened 2026-06-12
Windows: residual browse console flash from vendored playwright-core processLauncher taskkill (missing windowsHide) — distinct from #1784


kaiwulff:
## Summary

On Windows, `browse` still flashes a focus-stealing **`cmd.exe` console window** during normal use — but from a **different source** than #1784. #1784 covers the cold-start `Bun.spawnSync` (node-launcher + git) flashes. This one is in the **vendored `playwright-core` process launcher**, on the **browser-teardown force-kill path**, and is missed by the cold-start fixes.

It's intermittent (only on Playwright's *force*-kill path, not graceful close), which is why it reads as a "fixed it but it came back" regression.

## Root cause

`node_modules/playwright-core/lib/server/utils/processLauncher.js` (playwright-core **1.58.2**, the version gstack vendors), in `killProcess()`:

```js
const taskkillProcess = childProcess.spawnSync(`taskkill /pid ${spawnedProcess.pid} /T /F`, { shell: true });
```

`shell: true` makes Node wrap this as `cmd.exe /d /s /c "taskkill /pid N /T /F"`. With no `windowsHide: true`, and because the **browse daemon runs console-less** on Windows, that `cmd.exe` allocates a **new, visible console window** → the flash. The sibling browser *launch* spawn in the same file has the same omission upstream.

This is upstream Playwright code: the current `microsoft/playwright` `main` still ships it unfixed — `packages/utils/processLauncher.ts:236` — and `processLauncher.ts` contains **no `windowsHide` at all** (neither launch nor kill). It only bites deployments whose parent process has no console (i.e. a detached daemon), so most Playwright users never see it; gstack's daemon does.

## When it fires

Only on the **force-kill** path (`killProcess()` → win32 `taskkill`). Graceful `browser.close()` sets `processClosed` and skips it, so it does **not** fire on every browse — hence the intermittency. It reliably reproduces under rapid overlapping `browse restart` / daemon teardown with a live browser.

## Evidence (red → green, real daemon)

Instrument: a tight (25 ms) sampler counting top-level **visible** windows whose owning PID resolves to a console host (`cmd`/`conhost`/`WindowsTerminal`/…) — note a `ConsoleWindowClass`-only filter reads 0 here because the visible windows are Windows-Terminal-hosted. Correlated against a `Win32_Process` start logger.

**Unpatched** (`{ shell: true }`), forced-race:
- Visible windows appeared: `cmd:46876`, `cmd:22524`
[...]

PR #2161, opened 2026-07-03
fix(windows): patch playwright-core processLauncher to pass windowsHide (chrome-headless-shell console flash)

### LINKED ISSUE #2160: Windows: playwright-core processLauncher launches chrome-headless-shell without windowsHide — console window per browser launch on the node path (follow-up to #2151, node_modules twin of #1989)
## TL;DR

Follow-up to #2151 / PR #2153. With every gstack-owned spawn site fixed (polyfill `windowsHide`, signal-0 probe), Windows browse sessions **still** flash console windows — now titled `…\ms-playwright\chromium_headless_shell-1208\…\chrome-headless-shell.exe`. The remaining spawn surface is **playwright-core's own `processLauncher.js`**, which gstack does not control: it launches the browser without `windowsHide`, and its force-kill path shells out to `taskkill` (a `cmd.exe` window — the same class as the vendored-copy finding in #1989, but on the node/`node_modules` path).

## Root cause

Verified live: a browse daemon (`node …\browse\dist\server-node.mjs`) had `chrome-headless-shell.exe` as a direct child; the launch goes through `node_modules/playwright-core/lib/server/utils/processLauncher.js`:

```js
const spawnOptions = {
  detached: process.platform !== "win32",
  env: options.env,
  cwd: options.cwd,
  shell: options.shell,
  stdio
};                                      // ← no windowsHide
const spawnedProcess = childProcess.spawn(options.command, options.args || [], spawnOptions);
```

`chrome-headless-shell` is a console-subsystem executable. The browse daemon is console-less on Windows (detached node process), so without `CREATE_NO_WINDOW` the child allocates a **new visible console window** on every browser launch. Same file, force-kill path:

```js
const taskkillProcess = childProcess.spawnSync(`taskkill /pid ${spawnedProcess.pid} /T /F`, { shell: true });  // ← cmd.exe window per kill
```

Upstream playwright doesn't set `windowsHide` here, and there's no config/env knob to make it do so — so this can't be fixed from gstack's call sites the way #2153 fixed the polyfill.

## Relationship to open issues

- #1835 reports chrome-headless-shell black console windows (bun server context); this issue pins the node-path root cause to playwright's launcher and proposes a concrete mechanism.
- #1989 found the identical `taskkill` gap in the **vendored** playwright-core inside the compiled binary; this is the `node_modules` twin of that finding (the node path resolves playwright from `node_modules`, not the vendored copy).
- #2151 / #2153 fixed gstack's own spawn surfaces; this is the last remaining one on the node path.

## Proposed fix (PR incoming)

Bun's native `patchedDependencies`: `bun patch playwright-core`, add `windowsHide: true` to `spawnOptions` and to the `taskkill` `spawnSync` options, `bun patch --commit`. The patch applies automatically on every `bun install` — nothing changes for contributors or `./setup`.

Known caveat: bun applies patches per exact version (`playwright-core@1.58.2`), so a playwright bump would silently orphan the patch. The PR includes a tripwire test that reads the **installed** `node_modules` file and fails if the patch didn't apply — a version bump then fails CI instead of silently re-flashing consoles, and the fix is re-running `bun patch` against the new version.

Verified on Windows 11 (gstack v1.58.5.0 fork, bun 1.3.14): fresh `bun install` re-applies the patch; the patched launcher file carries both `windowsHide` sites.

## Env

- gstack: main @ v1.58.5.0
- playwright-core: 1.58.2, Bun: 1.3.14, node 22 (server-node.mjs path)

osouthgate:
Fixes #2160. Follow-up to #2153 (#2151); node_modules twin of #1989's vendored-copy finding; addresses the chrome-headless-shell half of #1835 on the node path.

## What

After #2153 fixed every gstack-owned spawn surface on the Windows node path, browse sessions still flashed console windows — now `chrome-headless-shell.exe` itself. The remaining spawn site is inside **playwright-core**, which gstack doesn't control: `lib/server/utils/processLauncher.js` launches the browser without `windowsHide` (chrome-headless-shell is a console-subsystem exe, and the browse daemon is console-less, so each launch allocates a visible console window), and its force-kill path runs `taskkill` through `shell: true` — a `cmd.exe` window per kill. Upstream playwright exposes no knob for this.

Fix: a bun-native `patchedDependencies` patch (`bun patch playwright-core` → `patches/playwright-core@1.58.2.patch`) adding `windowsHide: true` at both sites. It applies automatically on every `bun install` — no workflow change for contributors or `./setup`.

## Version-bump safety

Bun applies patches per exact version, so a future playwright bump would silently orphan the patch key. `browse/test/playwright-launcher-patch.test.ts` guards this with three tripwires:

1. `package.json` declares a `patchedDependencies` entry matching the **installed** playwright-core version (a bump without re-patching fails here);
2. the patch file still covers both spawn sites;
3. the installed `node_modules/...processLauncher.js` actually carries the patch.

Re-patching after a bump is mechanical: `bun patch playwright-core`, re-apply the two-line change, `bun patch --commit`.

## Verification

- Live repro (Windows 11): a browse daemon (`node …\dist\server-node.mjs`) had `chrome-headless-shell.exe` as a direct child, spawned via the unpatched launcher — console window per launch.
- Fresh-install check: removed `node_modules/playwright-core`, ran `bun install`, confirmed the reinstalled `processLauncher.js` contains both `windowsHide` sites.
- The same edit applied to a live local install stops the popups on the next daemon start.
- `bun test browse/test/playwright-launcher-patch.test.ts`: 3 pass / 0 fail.

### DIFF
--- a/browse/test/playwright-launcher-patch.test.ts
+++ b/browse/test/playwright-launcher-patch.test.ts
@@ -1,0 +1,52 @@
+import { describe, test, expect } from 'bun:test';
+import * as fs from 'fs';
+import * as path from 'path';
+
+// Windows console-flash guard for the Playwright browser launcher (#2151
+// follow-up; siblings: #1835, #1989).
+//
+// On Windows the browse server runs console-less (node server-node.mjs,
+// detached from the CLI). chrome-headless-shell is a console-subsystem
+// exe, so playwright-core's processLauncher spawning it without
+// windowsHide allocates a visible console window per launch, and its
+// force-kill path shells out to `taskkill` (another window). Upstream
+// playwright doesn't set windowsHide, so we carry a bun
+// `patchedDependencies` patch. These tripwires pin:
+//   1. package.json declares the patch for the installed version — if a
+//      playwright bump orphans the patch key, test 3 catches the silent
+//      regression (bun applies patches per exact version).
+//   2. the patch file itself still contains both windowsHide sites.
+//   3. the INSTALLED node_modules file actually got patched.
+
+const ROOT = path.resolve(import.meta.dir, '..', '..');
+const INSTALLED_LAUNCHER = path.join(
+  ROOT, 'node_modules', 'playwright-core', 'lib', 'server', 'utils', 'processLauncher.js',
+);
+
+describe('playwright-core processLauncher windowsHide patch', () => {
+  test('1. package.json declares a patchedDependencies entry matching the installed playwright-core version', () => {
+    const pkg = JSON.parse(fs.readFileSync(path.join(ROOT, 'package.json'), 'utf-8'));
+    const installed = JSON.parse(fs.readFileSync(
+      path.join(ROOT, 'node_modules', 'playwright-core', 'package.json'), 'utf-8',
+    ));
+    const key = `playwright-core@${installed.version}`;
+    expect(pkg.patchedDependencies?.[key]).toBeDefined();
+    expect(fs.existsSync(path.join(ROOT, pkg.patchedDependencies[key]))).toBe(true);
+  });
+
+  test('2. the patch covers both spawn sites (browser launch + taskkill)', () => {
+    const pkg = JSON.parse(fs.readFileSync(path.join(ROOT, 'package.json'), 'utf-8'));
+    const patchRel = Object.entries(pkg.patchedDependencies ?? {})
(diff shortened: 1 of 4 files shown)


time-attack:
@osouthgate Please rewrite this to resolve conflicts and apply windowsHide only to intended headless launches. Unconditionally hiding every Playwright Windows process breaks intentional UI modes.

PR #2165, opened 2026-07-03
fix(hooks): question-preference-hook breaks AskUserQuestion in non-interactive sessions (explicit permissionDecision 'defer')

dpnascimento:
## Problem

After running gstack setup, `AskUserQuestion` stops working entirely in non-interactive Claude Code sessions (e.g. the desktop app): the question UI never renders.

Root cause: the `question-preference-hook` PreToolUse hook emits `permissionDecision: 'defer'` as its "no opinion" outcome. But `defer` is a real value in Claude Code with a specific meaning — "pause this tool call and hand control back" — and it is **honored in print/non-interactive mode**. Claude Code's own strings confirm the split:

- interactive: `returned permissionDecision=defer in interactive mode; ignoring (defer is print-mode only)`
- non-interactive: the tool call is actually deferred (`tengu_pre_tool_hook_deferred` / `hook_deferred_tool`)

So in any non-interactive session, every AUQ call the hook doesn't auto-decide gets swallowed. Reproduced on Claude Code 2.1.187 (desktop app).

## Fix

The neutral outcome for a PreToolUse hook is a **silent exit 0**. `defer()` now:

- emits nothing when there is no context to inject;
- emits only `hookSpecificOutput.additionalContext` (no `permissionDecision` at all) when plan-tune memory context exists.

The `deny` paths (never-ask auto-decide, Conductor prose redirect) are unchanged.

## Tests

Updated assertions across `test/question-preference-hook.test.ts`, `test/memory-cache-injection.test.ts` and `test/skill-e2e-plan-tune-cathedral.test.ts` (defer now = no `permissionDecision` in output; additionalContext-only output still covered).

`bun test test/question-preference-hook.test.ts test/memory-cache-injection.test.ts` → 27 pass / 0 fail.

### DIFF
--- a/hosts/claude/hooks/question-preference-hook.ts
+++ b/hosts/claude/hooks/question-preference-hook.ts
@@ -93,12 +93,23 @@ function readStdin(): Promise<string> {
 }
 
 function defer(additionalContext?: string): void {
-  const out: Record<string, unknown> = {
-    hookEventName: 'PreToolUse',
-    permissionDecision: 'defer',
-  };
-  if (additionalContext) out.additionalContext = additionalContext;
-  process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
+  // "No opinion" must be a SILENT exit 0 (optionally with additionalContext
+  // only), never an explicit `permissionDecision: 'defer'`. `defer` is a real
+  // value in Claude Code, but it means "pause this tool call and hand control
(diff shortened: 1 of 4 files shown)


time-attack:
@dpnascimento Please rewrite this version to include the accurate documentation and strict output coverage from PR 1924. Its implementation is otherwise the best current foundation.

PR #2167, opened 2026-07-03
fix(test): make chromium profile fallback path portable

daehyeonxyz:
## Summary

- make the Chromium profile fallback test compare with `path.join(...)`
- keep the assertion portable across POSIX and Windows path separators

## Verification

- `bun test browse/test/config.test.ts`: 41 pass, 0 fail
- `git diff --check`


time-attack:
@daehyeonxyz ill merge if you just make this an exception instead of hardcoding the assumption

PR #2169, opened 2026-07-04
fix(browse): grant Windows ACLs by SID to avoid owner lockout when username matches machine name

voltapix26:
## Problem

`restrictFilePermissions` / `restrictDirectoryPermissions` in `browse/src/file-permissions.ts` harden Windows ACLs with:

```
icacls <path> /inheritance:r /grant:r <os.userInfo().username>:(F)
```

Granting by **bare username** is ambiguous. On machines where the username collides with another account object — most commonly when the **machine name equals the username** (e.g. user `Eshan` on machine `ESHAN`) — `LookupAccountName` resolves the bare name to the machine's *domain* SID (`S-1-5-21-…` with **no RID**) instead of the user account. The resulting ACL has inheritance stripped and a single grant to a principal that isn't the user, so it **denies even the owner**.

Symptoms: `~/.gstack` becomes inaccessible and browse fails with `EPERM` on `mkdir` — a self-lockout caused by the hardening itself. This is easy to hit in the wild: OOBE suggests the user's first name as the machine name, so `Alice` on `ALICE` is a common single-user-laptop configuration.

## Fix

Add a `getWindowsGrantee()` helper (cached per process) that:

1. Resolves the current user's **SID** via `%SystemRoot%\System32\whoami.exe /user /fo csv /nh` and grants `*<SID>:(F)` — SID grants are unambiguous by construction.
   - The **absolute path** matters: a bare `whoami` can resolve to GNU coreutils' `whoami` (Git for Windows / MSYS2 put `usr/bin` on `PATH`), which rejects `/user`.
2. Falls back to `USERDOMAIN\USERNAME` (domain-qualified, still unambiguous) if SID resolution fails, before ever falling back to the bare username.

Both `restrictFilePermissions` and `restrictDirectoryPermissions` now use the helper.

## Tests

- Added a Windows-only test asserting the resolved grantee is a SID (`*S-1-…`) or a domain-qualified name — never a bare username — via a `__getWindowsGranteeForTests` export (same pattern as the existing `__resetWarnedForTests`).
- Full `browse/test/file-permissions.test.ts` suite passes on a Windows 11 machine that exhibits the original bug (username == machine name): 14 pass, 0 fail. The existing "file remains readable by the caller" test is the behavioral regression guard on such machines.

## Note for release

`dist/browse.exe` and `dist/server-node.mjs` embed this module, so a **dist rebuild is required** for the fix to reach releases — the source change alone doesn't affect already-built binaries.

### DIFF
--- a/browse/src/file-permissions.ts
+++ b/browse/src/file-permissions.ts
@@ -42,6 +42,42 @@ import * as os from 'os';
 
 let warnedOnce = false;
 
+let cachedGrantee: string | undefined;
+
+/**
+ * Resolve the identity to use in icacls `/grant` arguments.
+ *
+ * Granting by bare username (`os.userInfo().username`) is ambiguous:
+ * LookupAccountName can resolve the name to the wrong principal when it
+ * collides with another account object — e.g. when the machine name equals
+ * the username (user "alice" on machine "ALICE"), the bare name can resolve
+ * to the machine's *domain* SID (S-1-5-21-… with no RID), producing an ACL
+ * that denies even the owner.
+ * A SID grant (`*S-1-5-21-…-1001`) is unambiguous, so we resolve the
+ * current user's SID via `whoami /user` and fall back to the fully
+ * qualified `DOMAIN\username` (still unambiguous) only if that fails.
+ */
+function getWindowsGrantee(): string {
+  if (cachedGrantee !== undefined) return cachedGrantee;
+  try {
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej Close this as a duplicate of PR 2170. Its SID enforcement and fallback coverage are weaker; move its useful quiet-process behavior into PR 2170.

PR #2170, opened 2026-07-04
fix(browse): grant Windows ACLs to token SID, not bare username

IntegriGit:
## Problem

`browse/src/file-permissions.ts` hardens sensitive gstack state (auth/canary tokens, chat history, agent queue, device salt, per-tab security decisions) by breaking ACL inheritance and granting a single-entry ACL to the current user via `icacls`. It resolved that user with `os.userInfo().username` — the **bare** login name.

On a **domain-joined Windows host that also has a local account of the same login name**, `icacls /grant lmiller:...` resolves the bare name to the **local** SAM account (`HOST\lmiller`), not the domain account the process actually runs as (`DOMAIN\lmiller`). Combined with `/inheritance:r`, the new ACL is granted to an account the process is *not*, locking it out of the `.gstack` state directory it just created. The browse daemon then can't create its lock/state file and fails to start with a misleading:

> Another instance is starting… Timed out.

This reproduced reliably on a domain-joined Windows 11 box (`INTEGRIBILT\lmiller` process vs. same-named local `OFC01\lmiller`).

## Fix

Add `currentUserIcaclsPrincipal()`:

- Resolves the current process **token SID** via `System32\whoami.exe /user /fo csv /nh`, called by **absolute path** so a shadowing Unix `whoami` earlier on `PATH` (e.g. Git-for-Windows) can't hijack it and force the weaker fallback.
- Grants `icacls ... *S-1-5-…` (SID literal), which is immune to local-vs-domain name collisions.
- Falls back to `USERDOMAIN\username`, then bare `username`, if the SID lookup is unavailable. Cached per process.

Both `restrictFilePermissions` and `restrictDirectoryPermissions` now use it instead of the bare `os.userInfo().username`. POSIX paths are unchanged.

## Tests

`bun test browse/test/file-permissions.test.ts` — **15 pass / 0 fail**.

Adds a regression test asserting the resolved principal is a SID literal (`*S-1-5-…`) on Windows, and a non-empty name-based fallback off-Windows. Verified on a domain-joined Windows 11 host: the emitted `icacls` grant is now `*S-1-5-21-…:(F)` rather than a bare name.

Co-Authored-By: Paperclip <noreply@paperclip.ing>

### DIFF
--- a/browse/src/file-permissions.ts
+++ b/browse/src/file-permissions.ts
@@ -42,6 +42,51 @@ import * as os from 'os';
 
 let warnedOnce = false;
 
+/**
+ * Resolve an unambiguous icacls principal for the account this process runs as.
+ *
+ * `os.userInfo().username` returns the bare login name (e.g. "lmiller"). On a
+ * domain-joined Windows box that also has a *local* account of the same name,
+ * `icacls /grant lmiller:...` resolves the bare name to the LOCAL account
+ * (e.g. OFC01\lmiller), NOT the domain account the process actually runs as
+ * (e.g. INTEGRIBILT\lmiller). Combined with `/inheritance:r`, that hands the
+ * new ACL to an account this process is *not*, locking the process out of the
+ * directory it just created — the daemon then can't create its state/lock file
+ * and startup fails. See INT-50 / INT-48.
+ *
+ * Fix: grant to the current process token's SID, which is immune to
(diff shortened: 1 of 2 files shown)


time-attack:
@IntegriGit Please rewrite the fallback to prevent ambiguous usernames from recreating owner lockout, and suppress visible windows for the spawned Windows tools. This remains the best implementation foundation.

PR #2171, opened 2026-07-04
fix: stop gen-skill-docs from truncating descriptions with embedded periods

sneakygriff:
## Problem

`splitCatalogDescription` in `scripts/gen-skill-docs.ts` extracts the lead (first) sentence of each skill description with the regex `^([^.!?]*[.!?])(?:\s|$)`. The `[^.!?]*` class is greedy and cannot backtrack across a period, so when the first sentence contains an embedded terminator — a filename like `TODOS.md` / `CLAUDE.md`, a URL, or a version number like `v1.45.0.0` — the regex fails to match a real sentence boundary. On failure the code falls back to a blind first-20-words cut, producing a lead that either stops mid-token or runs past the true sentence end into the next clause.

## Fix

Rewrite the lead regex to `^((?:[^.!?]|[.!?](?!\s|$))*[.!?])(?:\s|$)`. A terminator that is **not** followed by whitespace/end-of-text (an embedded period) is consumed by the second, disjoint alternative and no longer ends the sentence; a terminator followed by a real boundary still closes the lead. The two alternatives match disjoint inputs, so there is no ambiguity and no catastrophic-backtracking risk.

This re-renders two previously mis-split skill descriptions:
- **diagram** — lead now runs to "…and rendered SVG + PNG." instead of truncating at "you can open".
- **sync-gbrain** — lead now stops at "…guidance in CLAUDE.md." instead of spilling the next clause in via the 20-word fallback.

(The diagram source description was also reworded to split an overlong first sentence.)

## Verification

- `bun test test/catalog-trim.test.ts` — **23 pass, 0 fail** (adds regression cases for embedded periods, URLs, and a >200-char first sentence with embedded periods).
- `bun test test/gen-skill-docs.test.ts test/skill-validation.test.ts test/llms-txt-shape.test.ts` — **741 pass, 0 fail**.

### DIFF
--- a/diagram/SKILL.md
+++ b/diagram/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: diagram
 version: 1.0.0
-description: "Turn an English description (or mermaid source) into a diagram triplet: the source, an editable .excalidraw file you can open (gstack)"
+description: "Turn an English description (or mermaid source) into a diagram triplet: the source, an editable .excalidraw file you can open on excalidraw.com, and rendered SVG + PNG. (gstack)"
 allowed-tools:
   - Bash
   - Read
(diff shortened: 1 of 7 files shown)


time-attack:
@16francej merge

PR #2172, opened 2026-07-04
fix: stop test teardowns from silently killing the bun test run

sneakygriff:
## Problem

Several `browse/` and `design/` test files armed `setTimeout(() => process.exit(0), 500)` inside `afterAll` — a workaround for a browser `close()` that can hang. That assumes each test file runs in its own process. It doesn't: `bun test` runs **every** file in one shared process. The 500ms timer therefore fired partway through a **later** test file and terminated the entire run with **exit code 0** and no summary.

Because the exit code was 0, the suite could not gate: only a fraction of the files ran, and every downstream failure was silently masked.

## Fix

- Remove the delayed `process.exit` from teardown; instead time-box resource cleanup. Browser teardowns race `close()` against a short timeout and abandon it if it hangs (the child is reaped at real process exit). The design daemon `/shutdown` test stubs `process.exit` around the shutdown timers so they fire harmlessly.
- Add `test/no-suicide-exit.test.ts`, a static guard that scans every `*.test.ts` and fails if any file schedules a delayed `process.exit`, so the pattern cannot silently return.
- Make `test/user-slug-fallback.test.ts` deterministic via HOME isolation.

## Verification

- `bun test test/no-suicide-exit.test.ts test/user-slug-fallback.test.ts` — **15 pass, 0 fail** (guard green + slug 14/14).
- `bun test browse/test/commands.test.ts` — **243 pass, 0 fail**, now runs to a full summary.
- `bun test design/test/daemon.test.ts` — **34 pass, 0 fail**, no early exit.

## Honest note

With the suite now able to run to completion, it surfaces **pre-existing** failures that the early exit had been hiding — for example `design/test/feedback-roundtrip.test.ts` fails identically on unmodified `origin/main` (a `session.clearLoadedHtml` stub gap), independent of this change. Those failures are **not introduced by this PR**; this change only makes them visible so they can be gated on and fixed in follow-ups.

### DIFF
--- a/browse/test/batch.test.ts
+++ b/browse/test/batch.test.ts
@@ -42,9 +42,14 @@ beforeAll(async () => {
   // The test needs to start a server. Let's use the existing server infrastructure.
 });
 
-afterAll(() => {
+afterAll(async () => {
   try { testServer.server.stop(); } catch {}
-  setTimeout(() => process.exit(0), 500);
+  // Close only this file's own browser — never process.exit(): bun test runs
+  // all files in one process, so a delayed exit kills the whole suite
+  // (see test/no-suicide-exit.test.ts). close() can hang when the browser
+  // already died, and its internal 5s timeout ties bun's 5s hook timeout —
+  // so race it at 3s and abandon; the child is reaped at process exit.
+  try { await Promise.race([bm?.close(), new Promise((resolve) => setTimeout(resolve, 3000))]); } catch {}
 });
 
 // We need a running browse server for HTTP tests.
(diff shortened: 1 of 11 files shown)


time-attack:
@16francej — recommend closing this PR as superseded by #2230. Its useful change is already included there, so landing both would duplicate teardown and isolation changes; this branch also conflicts with current main. In the rebased #2230, credit #2172 and retain its three-second close cap, design-daemon exit and HOME isolation, and no-suicide guard.

PR #2175, opened 2026-07-04
Use image_gen for Codex design-shotgun

### LINKED ISSUE #2174: Codex design-shotgun should use image_gen instead of requiring OpenAI API keys
## Summary

The Codex-hosted `gstack-design-shotgun` skill still routes visual generation through gstack's `design` binary (`$D generate`). That makes Codex users provide an OpenAI API key even though Codex already has a native `image_gen` tool available in the host.

## Current behavior

In the Codex-generated skill, `design-shotgun` still includes the normal design-binary flow:

- run `DESIGN SETUP`
- resolve `$D`
- ask variant subagents to run `$D generate`
- rely on `design/src/auth.ts` to find `~/.gstack/openai.json` or `OPENAI_API_KEY`

That path can fail or bill the wrong account:

- no gstack OpenAI key configured
- cwd `.env` contains `OPENAI_API_KEY`
- Codex design binary path resolution hits the existing fallback-path problem from #1159

## Expected behavior

When the skill name is `gstack-design-shotgun` / host is Codex, the skill should call Codex's native `image_gen` tool directly for each variant.

Expected Codex path:

1. Build the same design brief and variant directions.
2. Use `image_gen` directly to generate each visual variant.
3. Save or copy returned images into `~/.gstack/projects/$SLUG/designs/<session>/variant-A.png`, etc.
4. Present the images inline and collect feedback.
5. Do not call `$D generate` and do not require `OPENAI_API_KEY`.

## Why this matters

Codex users have an image generation capability through the host, not through gstack's OpenAI key chain. Requiring another OpenAI key adds setup friction and can accidentally use a project `.env` key. It also makes the advertised Codex skill behave like the Claude-hosted skill instead of using the host-native tool.

## Notes

I tested the host `image_gen` path locally from Codex and it can generate a multi-variant UI mockup without any gstack `design` binary or OpenAI API key setup. The generated image was saved under `~/.codex/generated_images/...` as expected.

This is related to, but not the same as:

- #1159: Codex design workflows can report `DESIGN_NOT_AVAILABLE` due to path/install-surface drift
- #990: support non-OpenAI image providers in `design`
- #1248: `design` can pick up `OPENAI_API_KEY` from cwd `.env`
- #1771: OpenAI image tool model pairing regression

jinzheio:
## Summary

Fixes #2174.

The Codex-generated `gstack-design-shotgun` skill now delegates image generation and editing to the documented `$imagegen` skill in its default built-in mode. This uses Codex's host `image_gen` tool without requiring `OPENAI_API_KEY`.

## Changes

- Adds a Codex-specific `$imagegen` generation and screenshot-editing path.
- Preserves `$B` resolution for existing-page screenshots.
- Uses `$D` only for the local, cross-platform `compare` and `serve` board commands. Codex never runs the design binary's generation, iteration, or vision-check commands.
- Falls back to inline images and `AskUserQuestion` when the board binary is unavailable.
- Adds executable setup coverage and a real `codex exec` routing E2E.

## Validation

- `bun test test/gen-skill-docs.test.ts --grep "Codex design-shotgun"`
- `bun test test/skill-validation.test.ts --grep "Codex skill validation|\$B commands in Codex"`
- `EVALS=1 EVALS_BASE=origin/main bun test test/codex-e2e.test.ts --grep "codex-design-shotgun-routing"`
- `bun run gen:skill-docs --host codex --dry-run`
- `git diff --check`


time-attack:
Thanks for this. The Codex path uses the wrong documented command, leaves browser resolution unavailable for screenshot evolution, and relies on non-portable board handling. Please switch to `$imagegen`, preserve browser resolution, make board handling portable, and add realistic Codex coverage.

jinzheio:
Thanks, addressed in `b2b5b5e7`.

- Generation and screenshot edits now invoke the documented `$imagegen` skill in default built-in mode.
- `$B` resolution is restored, and screenshot evolution loads the captured image before the `$imagegen` edit call.
- The macOS-only static board was removed. Codex now uses `$D compare` and `$D serve` for the portable local board, while all `$D` generation, iteration, and vision-check commands remain prohibited.
- Added executable setup coverage plus a real `codex exec` E2E that verifies `$imagegen`, `$B screenshot`, `$D compare`, and no `OPENAI_API_KEY` requirement.

Validated with the targeted generator and Codex skill tests, both generation dry-runs, and `git diff --check`.

PR #2183, opened 2026-07-06
Fix formatting issues in SKILL.md

aniruddh10124:
The prompt given to the sub-agent should be indented.


time-attack:
Thanks for this. The change edits generated documentation, so regeneration will overwrite it and leave the source unchanged. Please move the indentation fix into the resolver, regenerate every affected output, and add source-to-generated freshness coverage.

PR #2192, opened 2026-07-07
v1.58.6.0 feat(codex): auth probe recognizes custom OpenAI-compatible providers

Haijie-Lee:
## Summary

`_gstack_codex_auth_probe` in `bin/gstack-codex-probe` only checked three hardcoded signals (`$CODEX_API_KEY`, `$OPENAI_API_KEY`, `~/.codex/auth.json`). Users with self-hosted, proxy, or non-OpenAI OpenAI-compatible providers (mimo, internal gateways) configured their auth via `~/.codex/config.toml`'s `[model_providers.*].env_key`, which the probe ignored. Their `codex` worked fine in the terminal, but `/codex` (and `/autoplan`, which shares the probe) hit `AUTH_FAILED` in Step 0.5 and bailed out before the diff was read.

This release adds a TOML-parsing fallback: after the original three signals fail, the probe reads `~/.codex/config.toml`, strips comment lines, extracts every `env_key` value from `[model_providers.*]` tables, and treats any of those env vars being non-empty as `AUTH_OK`. New providers need zero probe edits — just add the `[model_providers.*]` block you already use.

## Why PATCH (1.58.5.0 → 1.58.6.0)

The probe was supposed to detect whether `codex` can authenticate. It failed to detect a valid authentication pathway. The fix makes it detect that pathway. No new API, no new user-facing feature, no new configuration surface. Reviewed independently by Codex + a Claude eng subagent; both recommended PATCH over the MINOR bump the original draft proposed.

## Changes

- **`bin/gstack-codex-probe`**: Added custom-provider fallback in `_gstack_codex_auth_probe` (+29 lines). Comment lines are stripped first (`grep -v '^[[:space:]]*#'`) so stale commented-out `# env_key = "OLD"` entries don't cause false `AUTH_OK`. OpenAI / ChatGPT paths are unchanged — their three original signals still short-circuit before the TOML parse.
- **`test/codex-hardening.test.ts`**: 7 new test cases in the existing `'auth probe'` describe block: env_key hit + env set, env_key hit + env unset, env_key hit + env whitespace-only, multiple providers with one match, config.toml present but env absent (negative path), commented-out env_key (must NOT match), and malformed TOML (must not crash). Total auth-probe tests: 15 (8 original + 7 new). All pass.
- **`VERSION`**: `1.58.5.0` → `1.58.6.0` (PATCH).
- **`CHANGELOG.md`**: Added user-facing entry at top, following the project's release-summary spec (2-line headline + lead paragraph + "no measurement yet" + "What this means for users with custom LLM providers" + itemized changes).

## Known limitation (documented)

`env_key` outside `[model_providers.*]` tables (e.g., in a `[some_other_feature]` section) would also match. Real Codex configs don't use `env_key` outside `model_providers`, so we accept this as a documented tradeoff rather than adding a TOML section tracker. A follow-up can add `awk`-based section tracking if real-world configs prove this matters.

## Test plan

- `bash -n bin/gstack-codex-probe` — syntax check, passes
- `bun test test/codex-hardening.test.ts --test-name-pattern "auth probe"` — 15 pass, 0 fail
- Manual end-to-end: pre-patch probe with `mimo` provider configured returned `AUTH_FAILED` and blocked `/codex`. Post-patch returns `AUTH_OK` and `/codex` runs end-to-end against the user's installed copy.

## Bisect

2 commits:
- `c86e4cf` — `feat(codex): auth probe recognizes custom OpenAI-compatible providers` (probe + 7 tests)
- `20e3ed3` — `chore(release): v1.58.6.0 — codex auth probe supports custom providers` (VERSION + CHANGELOG)


time-attack:
@Haijie-Lee Please rewrite this on the current branch without stale release files, and use structured TOML parsing so unrelated content cannot be mistaken for provider configuration.

Issue #778, opened 2026-04-03
[browse] Auth state is lost across separate invocations; next call restarts on about:blank with stale state


robertDouglass:
## Summary

The `browse` daemon can lose its authenticated browser context across separate CLI invocations. In my case, a login succeeds inside one `browse chain` call, but the next independent `browse` call starts a fresh server/session on `about:blank` and the authenticated state is gone.

A related symptom is that `.gstack/browse.json` can still exist with a PID, but that PID is already dead by the time the next command runs.

## Environment

- gstack version: `0.15.2.1`
- gstack commit: `c620de38e19fe63e8a6e361e11d706b5ed432b93`
- OS: macOS 15.6 (Darwin 24.6.0), arm64
- Install mode: global install under `~/gstack`, using the compiled `browse` binary
- Working directory: project-local state under `<repo>/.gstack/`

## Repro

1. From a repo directory, run `browse chain` against an authenticated site login flow.
2. In the same `chain` call, navigate to the login page, take a snapshot, fill username/password, and click the primary sign-in button.
3. Confirm the login succeeds and `browse` lands on an authenticated page.
4. Run a later independent command from the same cwd, for example `browse url` or `browse text`.

## Observed

- The login succeeds inside the original chain call.
- On the next independent invocation, `browse` prints `[browse] Starting server...` again and returns `about:blank` instead of the previously authenticated page.
- Navigating back to the app after that lands on the public/unauthenticated experience, so the session cookie/context appears lost.
- `.gstack/browse.json` may still be present even though the PID inside it is no longer alive.

## Expected

- Separate `browse` invocations from the same cwd should reconnect to the same live daemon/browser context while it is healthy.
- If the daemon is gone, the failure mode should either preserve persisted session state or at least cleanly invalidate stale state instead of silently starting a fresh blank session that looks healthy but has lost auth.

## Concrete output

Successful login inside one chain call:

```text
[click] Clicked @e8 → now at https://spec-kitty-dev.fly.dev/a/robert-team/dashboards/
[url] https://spec-kitty-dev.fly.dev/a/robert-team/dashboards/
```

Next independent invocation from the same cwd:

```text
$ browse url
[browse] Starting server...
about:blank
```

After navigating back to the same host in that fresh session, I get the public home page rather than the authenticated dashboard.

## Notes

- I reproduced this against a Fly.io-hosted Django app with cookie auth.
- `browse chain` is usable as a workaround because it keeps the whole flow inside one daemon run.
[...]

PR #2193, opened 2026-07-07
feat(browse): opt-in automatic cookie persistence for the headless daemon

theyool88-blip:
### Problem

Device-trust cookies set by a web app (the kind a site drops after you approve a
new device) do not survive a `browse` daemon restart. The daemon runs Playwright
with a non-persistent browser context, so every cookie — including long-lived
`__Host-`/`Secure` device-trust cookies — is gone the moment the context is torn
down. For any workflow that leans on `browse` against an authenticated site, this
means re-approving the device again and again after each restart. The server side
is fine; the gap is purely on the browse client.

### Root cause

Two things compound:

1. **Non-persistent context.** `browse` launches a fresh, in-memory browser
   context each boot. Nothing is written to disk, so cookies live only as long as
   the process.
2. **No auto-save.** There is no mechanism that captures cookies on shutdown or
   periodically and replays them on the next launch. Manual `cookies export` /
   `import` exists, but nothing does it automatically across the daemon lifecycle.

### Solution

Add **opt-in** automatic cookie persistence, off by default, gated behind
`BROWSE_AUTO_COOKIE_PERSIST=1`. When enabled, the daemon saves eligible cookies to
a per-workspace file and restores them into the context on the next launch. The
server itself is unchanged — this is entirely inside the browse client, and the
default behavior of every existing session is untouched.

When enabled, the feature emits a one-time plaintext-storage warning so the user
knows cookies are being written to disk.

### Design

- **Opt-in, default off.** `BROWSE_AUTO_COOKIE_PERSIST=1` turns it on; one-time
  plaintext warning on first save.
- **Persistent cookies only.** Only cookies with `expires !== -1 && expires > now`
  are saved. Session cookies are dropped. An optional
  `BROWSE_AUTO_COOKIE_DOMAINS` allowlist narrows what gets persisted, and
  internal/private-network hosts are excluded.
- **storageState-shaped file, cookies only.** Saved in Playwright's storageState
  shape but carrying cookies alone (no `localStorage`). `__Host-` prefixes and
  `partitionKey` are preserved verbatim so partitioned/`__Host-` cookies restore
  correctly.
- **Three save triggers.** Shutdown flush (just before context close), a debounce
  after mutating commands, and a periodic 60s flush. Crash-path emergency cleanup
  deliberately does **not** save (avoids persisting a torn state).
- **Single restore point.** State is injected once, at `launch()`, right before
  `newContext`. `recreateContext` is intentionally left alone to avoid a
  double-restore.
- **Per-workspace isolation + exclusive lock.** The state file is hashed from
  `realpath(projectDir)`, and an `mkdir`-based atomic lock (with dead-PID
  reclaim) prevents two daemons for the same workspace from clobbering each
  other's file.
- **Torn-write-safe writes.** State is written via a new `writeSecureFileAtomic`
  helper (temp file + `rename` + `fsync`, `0600` perms) so a crash mid-write never
  leaves a half-written state file.
- **Why not `launchPersistentContext`.** That would persist much more than
  cookies and collides with the existing `recreateContext` flow, so this stays on
  a targeted cookies-only storageState instead.

### Testing

`bun test browse/test/auto-cookie-persist.test.ts` → **20 pass, 0 fail**. Coverage
spans the unit level (eligibility filtering, `__Host-`/`partitionKey`
preservation, atomic write, lock acquire/dead-PID reclaim), a Playwright
round-trip (save cookies → restore into a fresh context), and a real
`BrowserManager.launch()` integration that confirms a persisted device-trust
cookie is restored on the next boot.

No regressions against `main`: the pre-existing failing tests (headed-browser
suites that need a Chromium build not installed in this environment) fail
identically on `main`, verified by comparing with the feature stashed.

### Commits (bisected)

- `feat(browse): add writeSecureFileAtomic for torn-write-safe state`
- `feat(browse): auto-cookie persistence module (opt-in, persistent-cookie-only)`
- `feat(browse): wire auto-cookie persistence into launch + daemon lifecycle`
- `test(browse): auto-cookie persistence — unit, round-trip, launch, wiring`
- `fix(browse): surface auto-cookie state read errors instead of swallowing`

### DIFF
--- a/browse/src/auto-cookie-persist.ts
+++ b/browse/src/auto-cookie-persist.ts
@@ -258,12 +258,21 @@ export function loadAutoCookieState(
 ): { cookies: Cookie[]; origins: [] } | null {
   if (!isAutoCookiePersistEnabled()) return null;
   const file = statePath(config);
-  let parsed: AutoCookieState;
+  let raw: string;
   try {
     if (!fs.existsSync(file)) return null;
-    parsed = JSON.parse(fs.readFileSync(file, 'utf-8')) as AutoCookieState;
+    raw = fs.readFileSync(file, 'utf-8');
+  } catch (err: any) {
+    // A read error here (e.g. EACCES) is NOT the benign corrupt-file case — the
+    // device cookie silently won't restore, so surface it rather than swallow.
+    console.warn(`[browse] auto-cookie state unreadable (${err?.code || err?.message}); starting without persisted cookies`);
+    return null;
+  }
+  let parsed: AutoCookieState;
+  try {
+    parsed = JSON.parse(raw) as AutoCookieState;
   } catch {
-    return null; // corrupt — ignore (a later save will overwrite it)
+    return null; // corrupt JSON — ignore (a later save overwrites it)
   }
   if (!parsed || parsed.kind !== STATE_KIND || parsed.version !== STATE_VERSION) return null;
   if (parsed.workspaceId !== computeProfileId(config)) return null;
--- a/browse/src/auto-cookie-persist.ts
+++ b/browse/src/auto-cookie-persist.ts
@@ -215,12 +215,39 @@ export function acquireLock(config: BrowseConfig): boolean {
   return true;
 }
 
+function readProcessStartTimeMs(pid: number): number | null {
+  if (process.platform === 'win32') return null;
+  try {
+    const result = Bun.spawnSync(['ps', '-p', String(pid), '-o', 'lstart='], {
+      stdout: 'pipe', stderr: 'pipe', timeout: 2000,
+    });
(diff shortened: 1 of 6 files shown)


garrytan:
Shipped in fork-port wave 2: opt-in automatic session persistence for the headless daemon. `BROWSE_PERSIST_STATE=1` snapshots cookies + per-tab URL/localStorage/sessionStorage to `<stateDir>/session-state.json` (0600) on an interval and at clean shutdown, and restores on the next launch. Same feature closes #778.

Commit: `80943401` feat(browse): opt-in session persistence — auth survives daemon restarts (#778, #2193)
Pinned by: `browse/test/session-persist.test.ts`

Ships in #2577 (v1.65.0.0).

PR #2197, opened 2026-07-08
docs: add French translation of the README (README.fr.md)

smartitemgalaxy:
Hi Garry, and thanks for open-sourcing gstack.

This PR adds a **French translation of the README** (`README.fr.md`) for francophone users, plus a small language link at the top of the English `README.md`.

**Scope — prose only.** Every slash-command, code block, install command, path, URL, image alt-text and the "See it work" transcript is kept **verbatim in English**, so nothing that the agent or the reader needs to copy/paste is altered. Technical terms follow a consistent glossary (PR, CI, QA, LOC… kept with a French gloss on first use; roles translated; a short "Glossaire pour les novices" section added at the end).

It's the repo's first translation, so I'm happy to adjust the format — e.g. move it under `docs/i18n/` or drop the README link — whatever fits your i18n preferences. Happy to maintain it as the English README evolves.

### DIFF
--- a/README.fr.md
+++ b/README.fr.md
@@ -1,0 +1,498 @@
+# gstack
+
+> "Je ne pense pas avoir tapé une seule ligne de code depuis décembre, en gros, ce qui représente un changement extrêmement important." — [Andrej Karpathy](https://fortune.com/2026/03/21/andrej-karpathy-openai-cofounder-ai-agents-coding-state-of-psychosis-openclaw/), podcast No Priors, mars 2026
+
+Quand j'ai entendu Karpathy dire ça, j'ai voulu comprendre comment. Comment une seule personne peut-elle livrer comme une équipe de vingt ? Peter Steinberger a construit [OpenClaw](https://github.com/openclaw/openclaw) — 247K étoiles GitHub — essentiellement seul avec des agents d'IA. La révolution est là. Un seul builder avec les bons outils peut aller plus vite qu'une équipe traditionnelle.
+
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej — recommend closing this PR. The project does not yet have localization conventions or durable maintenance ownership, and this translation is already stale while not fully preserving code blocks as claimed. It can be revisited after a localization policy is established.

smartitemgalaxy:
Thanks for the review — the technical points are fair, and I found a worse one you didn't mention. All fixed in 4d98e48.

**The install prompts.** This is the real bug. The two blockquotes in *Install → Step 1* and *OpenClaw* aren't prose for the reader — they're prompts the user pastes into Claude Code or an OpenClaw agent. I had translated them into French, which changes the instructions the agent actually executes. Both are now restored verbatim from `README.md`, with a French paragraph above each explaining that they must be pasted as-is and summarising what they do, so a francophone reader still knows what they're running. This one was my mistake and it was the most damaging thing in the diff.

**Code blocks.** You're right, and my PR description overstated it. Eight of the nine fenced blocks were byte-identical; in the manual-uninstall block I had translated the `#` comments. No executable line was ever changed, but "every code block kept verbatim in English" wasn't accurate. Those comments are back in English, and the steps are glossed in French in the prose above the block instead.

**Staleness.** The branch was one line behind: the `gstack-ios-qa-regen` row added to the CLI table in 7efb596c. That row is now translated and in place.

As of `main` today the file checks out: 9/9 fenced blocks byte-identical, both install prompts verbatim, 92/92 table rows, 18/18 URLs, 47/47 slash-commands.

**On the policy question** — I think that's the real decision here, and it isn't mine to make. The repo has no localization convention and no ownership model for translations, and I'd rather see that settled than worked around. Whatever direction fits best, I'm happy to follow: move the file under `docs/i18n/`, drop the link from `README.md`, or close this and revisit once a policy exists. I'm also willing to commit to keeping the French file in sync as `README.md` evolves, if that helps with the maintenance concern.

PR #2208, opened 2026-07-09
fix(setup): don't auto-install plan-tune AUQ hooks in Conductor (fixes #2207)

### LINKED ISSUE #2207: Conductor + Claude: plan-tune PreToolUse hook breaks native AskUserQuestion (auto-installed in Conductor)
## Summary

In **Conductor** sessions running **Claude Code**, gstack's plan-tune AUQ hooks **break Conductor's native `AskUserQuestion`**. Every AUQ call returns `[Tool result missing due to internal error]` (or the input is echoed with no rendered options). The same Conductor workspace running **Codex/GPT works fine**, because Codex doesn't load `~/.claude/settings.json` hooks.

The important part: the premise behind the v1.58.1.0 "Conductor prose AskUserQuestion" feature — *"Conductor disables native AUQ and its MCP variant is flaky"* — appears to be **self-inflicted**. Conductor's AUQ is not flaky. gstack's own PreToolUse hook is what breaks it, and then the code attributes the failure to Conductor and routes to prose.

## Impact

- Claude + Conductor users lose clickable `AskUserQuestion` entirely (every decision silently degrades to prose, or errors).
- It's auto-installed with no opt-in: `setup` treats a Conductor workspace as an implicit opt-in for the plan-tune hooks (see below), so a routine `/gstack-upgrade` silently plants the breakage.

## Root cause

`AskUserQuestion` in Conductor + Claude Code is resolved through the Agent-SDK `canUseTool` elicitation bridge (`CLAUDE_CODE_ENTRYPOINT=sdk-ts`; the `mcp__conductor__AskUserQuestion` tool is injected by Conductor's harness, not present in `~/.claude.json`).

gstack registers a **PreToolUse hook** (`hosts/claude/hooks/question-preference-hook.ts`, source `plan-tune-cathedral`) on the matcher `(AskUserQuestion|mcp__.*__AskUserQuestion)`. When that hook participates in the permission flow — **even when it returns `permissionDecision: "defer"`** — it corrupts the `canUseTool` round-trip, and the tool result comes back missing.

`setup` auto-installs it in Conductor:

```sh
# setup, ~line 1389
if [ "$PT_DECISION" = "prompt" ] && { [ -n "${CONDUCTOR_WORKSPACE_PATH:-}" ] || [ -n "${CONDUCTOR_PORT:-}" ]; }; then
  PT_DECISION="yes"
  _PT_CONDUCTOR_AUTO=1
fi
```

Introduced in **v1.58.1.0** (`c7ae632`, "feat: hermetic local E2E + Conductor prose AskUserQuestion").

## Reproduction

1. Open a Conductor workspace, Claude Code agent, with gstack's plan-tune hooks installed (the Conductor default).
2. Trigger any `AskUserQuestion` / `mcp__conductor__AskUserQuestion` call.
3. Observe: no options render; result is `[Tool result missing due to internal error]`.
4. Remove the AUQ hooks from `~/.claude/settings.json` (or `gstack-config set plan_tune_hooks no`, delete the `hooks` block, restart).
5. Trigger AUQ again → it renders and works. (Verified live, 3×.)

Cross-check: the identical Conductor workspace on Codex/GPT renders AUQ fine throughout — Codex never reads `~/.claude/settings.json`, so the hook is never in the path. This isolates the hook as the cause.

Note: gating the `deny` behind a flag so the hook returns `defer` does **not** fix it — the hook has to be fully **unregistered**. Its mere presence in the permission flow is enough to break the bridge.

## Suggested fix

At minimum, stop auto-installing the AUQ **PreToolUse** hook in Conductor, since it's what breaks native AUQ there:

```diff
-  # Conductor host reliability: ... Treat Conductor as an implicit opt-in ...
-  if [ "$PT_DECISION" = "prompt" ] && { [ -n "${CONDUCTOR_WORKSPACE_PATH:-}" ] || [ -n "${CONDUCTOR_PORT:-}" ]; }; then
-    PT_DECISION="yes"
-    _PT_CONDUCTOR_AUTO=1
-  fi
```

More fundamentally, it's worth re-validating the whole "Conductor prose AUQ" premise: native AUQ works in Conductor when this PreToolUse hook isn't registered, so the flakiness the feature works around looks like it's caused by the feature itself. If the PostToolUse capture/fallback hooks are confirmed harmless to the bridge (they fire after the result, so they likely are), the preference-enforcement value could be kept by only dropping the PreToolUse participation in Conductor — but that needs a test on real Conductor hardware.

## Workaround (for affected users)

```sh
# unregister the AUQ hooks and stop setup from re-adding them (even in Conductor)
gstack-config set plan_tune_hooks no
# then remove the plan-tune-cathedral / auq-error-fallback hooks from ~/.claude/settings.json
```

## Environment

- gstack `v1.58.5.0`
- Claude Code `2.1.201` via Agent SDK (`CLAUDE_CODE_ENTRYPOINT=sdk-ts`)
- Conductor (macOS, local mode)
- Regression introduced in gstack `v1.58.1.0` (`c7ae632`)

Happy to open a PR from a fork if the minimal `setup` change above looks right to you.

philhie:
Fixes the Conductor + Claude AskUserQuestion breakage described in #2207.

## Problem

In Conductor + Claude Code (Agent SDK, `CLAUDE_CODE_ENTRYPOINT=sdk-ts`), the plan-tune **PreToolUse** hook `question-preference-hook` participates in Conductor's `canUseTool` elicitation flow for `AskUserQuestion`. Its presence corrupts the round-trip — **even when it returns `defer`** — so AUQ comes back as `[Tool result missing due to internal error]` and no options render.

Native AUQ works fine in Conductor *until* this hook is registered. So the "flaky `mcp__conductor__AskUserQuestion`" that the v1.58.1.0 Conductor-prose backstop was built to work around appears to be caused by the backstop itself. Codex sessions never read `~/.claude/settings.json`, so they're unaffected — which is exactly why AUQ breaks in Claude but works in Codex on the same workspace.

`setup` currently treats any Conductor workspace as an **implicit opt-in** for these hooks, so a routine `/gstack-upgrade` silently plants the breakage.

## Change

`setup`: drop the Conductor implicit opt-in for the plan-tune AUQ hooks. A Conductor workspace with no explicit `--plan-tune-hooks` / `GSTACK_PLAN_TUNE_HOOKS` / `plan_tune_hooks` config now falls through to the normal non-interactive skip, keeping native AUQ working. Explicit opt-in still installs the hooks for anyone who wants the prose backstop.

Minimal, comment-heavy, no behavior change outside Conductor.

## Verification

- `bash -n setup` passes.
- Reproduced on real Conductor hardware: with the hooks registered, AUQ returns `[Tool result missing due to internal error]`; after unregistering them (equivalently: this change → no auto-install), AUQ renders and answers correctly (confirmed 3×).
- Cross-checked against Codex on the same workspace (works throughout), isolating the hook as the cause.

## Notes / open questions

- I kept the diff to the auto-install trigger only. If you'd prefer, the deeper question is whether the PostToolUse capture/fallback hooks are also unsafe on the Conductor bridge — they fire *after* the result so they're likely fine, but I didn't want to over-reach without a test you'd trust.
- Left `VERSION` / `CHANGELOG.md` untouched since versioning is your release call — happy to add a CHANGELOG entry if you want it in the PR.

Environment: gstack v1.58.5.0, Claude Code 2.1.201 (sdk-ts), Conductor/macOS. Regression introduced in v1.58.1.0 (c7ae632).

### DIFF
--- a/setup
+++ b/setup
@@ -1380,16 +1380,22 @@ if [ "$NO_TEAM_MODE" -ne 1 ] \
     *)                       PT_DECISION="prompt" ;;
   esac
 
-  # Conductor host reliability: the PreToolUse preference hook also carries the
-  # Conductor-prose enforcement (deny the flaky mcp__conductor__AskUserQuestion,
-  # redirect to a prose decision brief). A Conductor workspace setup otherwise
-  # falls through to "prompt" → the non-interactive skip below, leaving Conductor
-  # users without that backstop. Treat Conductor as an implicit opt-in — but
-  # only on the silent fall-through, never overriding an explicit --no-plan-tune-hooks.
-  if [ "$PT_DECISION" = "prompt" ] && { [ -n "${CONDUCTOR_WORKSPACE_PATH:-}" ] || [ -n "${CONDUCTOR_PORT:-}" ]; }; then
-    PT_DECISION="yes"
-    _PT_CONDUCTOR_AUTO=1
-  fi
+  # Conductor host: do NOT auto-install the plan-tune AUQ hooks here.
+  #
+  # The PreToolUse preference hook participates in the Agent-SDK `canUseTool`
+  # elicitation flow that Conductor uses to render AskUserQuestion. Its mere
+  # presence corrupts that round-trip — even when it returns `defer` — so the
+  # tool comes back as `[Tool result missing due to internal error]` and no
+  # options render. In other words, native AUQ works fine in Conductor UNTIL
+  # this hook is registered; the "flaky mcp__conductor__AskUserQuestion" this
+  # backstop was meant to work around is caused by the backstop itself.
+  # (Codex sessions never read ~/.claude/settings.json, so they're unaffected —
+  # which is why AUQ works in Codex but breaks in Claude on the same workspace.)
+  #
+  # Fall through to the normal decision below: a Conductor workspace with no
+  # explicit --plan-tune-hooks / GSTACK_PLAN_TUNE_HOOKS / plan_tune_hooks config
+  # takes the non-interactive skip and keeps native AUQ working. Users who
+  # genuinely want the prose backstop can still opt in explicitly.
 
   _install_plan_tune_hooks() {
     "$SETTINGS_HOOK" add-event \


time-attack:
@philhie Please rewrite setup migration so existing Conductor users also have the problematic hook removed. The current change only prevents future installation and leaves it in the ALREADY_INSTALLED path.

PR #2209, opened 2026-07-09
fix(review): correct put_page call shape in CRITICAL-finding task stub

anoopkansupada:
The Step B task-stub block passed `type:` and `frontmatter:` as sibling kwargs, which the live `mcp__gbrain__put_page` tool rejects — the silent-write class. Frontmatter now lives inside the `content: |` string between `---` fences. Fixed in the .tmpl (source of truth) and regenerated via `bun run gen:skill-docs`. Verified: `scan-write-schemas.py` exits 0 across the library.

### DIFF
--- a/LOCAL_PATCHES.md
+++ b/LOCAL_PATCHES.md
@@ -1,10 +1,13 @@
-# Local Patches — Triage & Upstream Plan
+# Local Patches — Triage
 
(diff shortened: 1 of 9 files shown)


anoopkansupada:
Filed against wrong repo by mistake — this belongs on my fork. Apologies for the noise.

PR #2217, opened 2026-07-10
fix: make required team hooks cross-platform

neallee2012:
## Summary

Required team mode currently generates a Bash hook and registers it with Bash environment-variable syntax. Windows and Copilot hook runners can select a different shell, turning an intentional policy denial into `Denied by preToolUse hook ... (hook errored)`.

This PR makes required-mode enforcement shell-neutral:

- Generate `.claude/hooks/check-gstack.cjs` with an explicit CommonJS extension and a Node 18+ implementation.
- Register one `node -e` launcher that reads `CLAUDE_PROJECT_DIR` through `process.env`; no `shell` field or Bash, PowerShell, or cmd environment expansion is used.
- Return neutral `{}` when the global install is present. The hook never emits `permissionDecision: "allow"`, so normal permission prompts remain intact.
- Return structured `PreToolUse` deny JSON with exit code 0 when gstack is absent, its install cannot be verified, the project directory is blank or stale, or the hook cannot be loaded. These are policy denials, not hook execution errors.
- Resolve home directories through `HOME`, then `USERPROFILE`, then `os.homedir()`.

Legacy required-mode projects migrate cleanly:

- Matching Bash registrations are replaced and duplicate gstack handlers are removed without touching unrelated hooks.
- The obsolete `.claude/hooks/check-gstack.sh` file is removed only after the CommonJS hook and settings registration succeed.
- If Git tracks the legacy file, the generated `git add` command includes its deletion so the suggested migration commit leaves no unstaged change.

## Test plan

- [x] `bun test --timeout 30000 test\team-mode.test.ts` twice: 32 passed, 0 failed, 114 assertions per run.
- [x] Default-shell execution covers neutral success, intentional denial, blank project context, and nonexistent or incomplete project paths.
- [x] Windows PowerShell execution covers blank project context, stale project paths, and `USERPROFILE` home fallback.
- [x] Injected filesystem verification failures remain structured deny decisions with exit code 0.
- [x] Legacy migration executes the exact printed `git add` command, commits it, and verifies a clean worktree while preserving an unrelated hook.
- [x] `bash -n bin/gstack-team-init`.
- [x] `git diff --check`.


time-attack:
@16francej — recommend closing this PR as superseded by #2229. It offers useful portability and migration groundwork, but omits Copilot response details and includes unrelated files; #2229 provides the narrower cross-host implementation. Credit #2217 for the CommonJS hook, shell-neutral launcher, fail-closed errors, and legacy migration, while leaving its unrelated files out.

PR #2218, opened 2026-07-10
fix(repo-mode): reuse canonical cached project slug (#2212)

### LINKED ISSUE #2212: gstack-repo-mode bypasses gstack-slug's cache (state written under a divergent slug) and skips slug sanitization before mkdir
**Version:** gstack v1.58.5.0 (macOS)

`bin/gstack-repo-mode` derives its own project slug instead of consuming `bin/gstack-slug`, which causes two related problems: project state written under a divergent slug (with a silent-history-loss failure mode), and a slug that skips the `[a-zA-Z0-9._-]` sanitization contract before being used in a `mkdir` path.

## Bug 1: `gstack-repo-mode` bypasses `gstack-slug`'s cache → state written under a divergent slug

`gstack-slug` is cache-first by design — the comment on its cache-first read says the per-PWD cache (`~/.gstack/slug-cache/<path-key>`) "guarantees consistency across sessions". Whatever slug was derived first (e.g. a basename slug cached before the repo had a remote) stays canonical, and all project state accumulates under `~/.gstack/projects/<cached-slug>/`.

But `gstack-repo-mode` ([lines 16–21](https://github.com/garrytan/gstack/blob/11de390be1be6849eb9a15f91ff4922dd16c589a/bin/gstack-repo-mode#L16-L21)) deliberately re-derives the slug from `git remote get-url origin` and never consults that cache:

```bash
# Compute SLUG directly (avoid eval of gstack-slug — branch names can contain shell metacharacters)
REMOTE_URL=$(git remote get-url origin 2>/dev/null || true)
...
SLUG=$(echo "$REMOTE_URL" | sed 's|.*[:/]\([^/]*/[^/]*\)\.git$|\1|;s|.*[:/]\([^/]*/[^/]*\)$|\1|' | tr '/' '-')
```

So in any repo whose **cached** canonical slug differs from the remote-derived form, every repo-mode classification writes `repo-mode.json` into a **twin directory** `~/.gstack/projects/<owner-repo>/`, while timeline/decisions/checkpoints live under `~/.gstack/projects/<cached-slug>/`.

### Why this is worse than clutter

If the slug-cache entry is ever lost (cache dir cleaned, path moved, new machine), `gstack-slug` re-derives the remote-form slug and resolution **flips to the twin**. Because the twin is non-empty (it holds `repo-mode.json`), anything that treats "non-empty project dir" as "initialized" sees a healthy project — and context restore silently returns a near-empty history instead of months of accumulated state. No error, no signal.

### Repro

1. `mkdir myproj && cd myproj && git init` (no remote yet)
2. Run `gstack-slug` → caches slug `myproj` (basename fallback) in `~/.gstack/slug-cache/`
3. Use any gstack feature that writes project state (context save, timeline, decision log) → real history accumulates under `~/.gstack/projects/myproj/`
4. `git remote add origin git@github.com:me/myproj.git`, make ≥5 commits
5. From the repo root, run `gstack-repo-mode` → creates `~/.gstack/projects/me-myproj/repo-mode.json`
6. `ls ~/.gstack/projects/` → both `myproj/` (canonical, holds the real history) and `me-myproj/` (twin)

Observed in the wild: 4 twin dirs on one machine (each containing a lone `repo-mode.json`), all in repos whose slug was cached before the remote existed.

### Suggested fix

Parse the `SLUG=` line from `gstack-slug` output instead of re-deriving — sed, **not** eval, which respects the exact injection concern in repo-mode's own comment:

```bash
SLUG=$("$SCRIPT_DIR/gstack-slug" | sed -n 's/^SLUG=//p' | head -1)
```

`gstack-slug` sanitizes its output to `[a-zA-Z0-9._-]` on every path before echoing, so consuming the line as data (never eval'd) is safe, and repo-mode inherits the cached canonical slug. Two details matter for the swap:

- **Keep the existing no-remote early exit** (`REPO_MODE=unknown` when `git remote get-url origin` is empty) *before* this line. `gstack-slug` falls back to the directory basename when there's no remote, so calling it unconditionally would make local-only repos eligible for classification they're excluded from today.
- **Invoke from the repo root.** `gstack-slug`'s cache is keyed on `$PWD`, so a subdirectory invocation reads/writes a different cache entry than the repo root's. Something like `SLUG=$(cd "$(git rev-parse --show-toplevel)" && "$SCRIPT_DIR/gstack-slug" | sed -n 's/^SLUG=//p' | head -1)`.

Existing stray `repo-mode.json` twins can be deleted or ignored — repo-mode recomputes its cache on the next classification either way.

## Bug 2: the re-derived slug skips sanitization before `mkdir ~/.gstack/projects/$SLUG`

`gstack-slug` enforces a `[a-zA-Z0-9._-]` output contract (`tr -cd 'a-zA-Z0-9._-'` on every path, including a final unconditional re-sanitize). `gstack-repo-mode`'s derivation only converts `/` → `-` (line 21) — no character filtering — and then uses the value in a filesystem path:

```bash
CACHE_DIR="$HOME/.gstack/projects/$SLUG"   # line 37
...
mkdir -p "$CACHE_DIR" 2>/dev/null || true  # line 87
```

A git remote URL is free-form local config (`git remote add origin` accepts any string — typos, templates, odd hosting URLs), so unusual URLs produce directory names containing spaces, `~`, `$`, quotes, or control characters — violating the `[a-zA-Z0-9._-]` contract that `gstack-slug` enforces when it creates entries in the same tree. To be precise about blast radius: `$SLUG` is quoted at every use and only the `REPO_MODE=` line is sourced by consumers, and the slash-to-dash conversion means no path separator survives into the name — so this is a filesystem-invariant violation (surprise directory names under `~/.gstack/projects/`), not shell injection or path traversal.

The Bug 1 fix subsumes this one: consuming `gstack-slug`'s sanitized output removes the unsanitized derivation entirely. A minimal standalone fix would be appending `| tr -cd 'a-zA-Z0-9._-'` to [line 21](https://github.com/garrytan/gstack/blob/11de390be1be6849eb9a15f91ff4922dd16c589a/bin/gstack-repo-mode#L21).

time-attack:
## Summary
gstack caches a project’s slug from the first time it’s seen (often the local folder name). If you add a GitHub origin later, gstack-repo-mode used to ignore that cache and invent a new slug from the remote URL. That wrote state into a second “twin” project folder, so history could look split or lost and you’d get duplicate untracked project dirs.

This fix makes gstack-repo-mode reuse the canonical cached slug from gstack-slug. Repos with no remote still return REPO_MODE=unknown. It runs from the git root so the PWD-keyed cache matches.

Fixes #2212

## Tests
We built local Git workspaces that mimic real public repos people use with gstack, reproduced the twin-slug bug in each, then verified the fix:

1. **facebook/react** — local folder `react-local-history`, cache slug first, then add `https://github.com/facebook/react.git` as origin. Before: wrote under `facebook-react`. After: keeps `react-local-history` and no twin dir.
2. **vercel/next.js** — same flow with `next-local-history` → `https://github.com/vercel/next.js.git`.
3. **kubernetes/kubernetes** — same flow with `kubernetes-local-history` → `https://github.com/kubernetes/kubernetes.git`.

Additional coverage:
- No-origin repo still returns `REPO_MODE=unknown` and creates no project dir
- Running from a subdirectory still writes under the git-root cached slug
- Unusual origin URLs (spaces/parens) stay sanitized before mkdir

All 6 cases in `test/gstack-repo-mode.test.ts` pass via `bun test test/gstack-repo-mode.test.ts`.

### DIFF
--- a/bin/gstack-repo-mode
+++ b/bin/gstack-repo-mode
@@ -12,13 +12,20 @@
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
-# Compute SLUG directly (avoid eval of gstack-slug — branch names can contain shell metacharacters)
+# Keep the no-remote early exit before consulting gstack-slug. gstack-slug falls
+# back to the directory basename when origin is missing, which would incorrectly
+# make local-only repos eligible for classification.
 REMOTE_URL=$(git remote get-url origin 2>/dev/null || true)
 if [ -z "$REMOTE_URL" ]; then
   echo "REPO_MODE=unknown"
   exit 0
 fi
-SLUG=$(echo "$REMOTE_URL" | sed 's|.*[:/]\([^/]*/[^/]*\)\.git$|\1|;s|.*[:/]\([^/]*/[^/]*\)$|\1|' | tr '/' '-')
+# Consume the canonical cached slug from gstack-slug (sed, never eval — branch
+# names can contain shell metacharacters). Invoke from the git root so the
+# PWD-keyed slug cache matches other gstack project-state writers.
+REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null || true)
+[ -z "${REPO_ROOT:-}" ] && { echo "REPO_MODE=unknown"; exit 0; }
+SLUG=$(cd "$REPO_ROOT" && "$SCRIPT_DIR/gstack-slug" | sed -n 's/^SLUG=//p' | head -1)
 [ -z "${SLUG:-}" ] && { echo "REPO_MODE=unknown"; exit 0; }
 
 # Validate: only allow known values (prevent shell injection via source <(...))
--- a/test/gstack-repo-mode.test.ts
+++ b/test/gstack-repo-mode.test.ts
@@ -1,0 +1,176 @@
+/**
+ * Project-slug consistency for gstack-repo-mode.
+ *
+ * These are end-to-end reproductions of #2212. Each case is a real local Git
+ * workspace which predates its public GitHub origin: gstack-slug caches the
+ * local directory name, then the origin is added. repo-mode must write its
+ * cache next to that canonical project state, not beside a newly derived
+ * owner-repo slug.
+ */
+
+import { describe, expect, test } from 'bun:test';
+import { spawnSync } from 'child_process';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+const ROOT = path.resolve(import.meta.dir, '..');
+const REPO_MODE_BIN = path.join(ROOT, 'bin', 'gstack-repo-mode');
+const SLUG_BIN = path.join(ROOT, 'bin', 'gstack-slug');
+
+type CommandResult = { stdout: string; stderr: string; status: number };
+
+function run(command: string, args: string[], cwd: string, home: string): CommandResult {
+  const result = spawnSync(command, args, {
+    cwd,
+    encoding: 'utf8',
+    env: { ...process.env, HOME: home, GSTACK_HOME: path.join(home, '.gstack') },
+  });
+  return {
+    stdout: result.stdout ?? '',
+    stderr: result.stderr ?? '',
+    status: result.status ?? -1,
+  };
+}
+
+function git(args: string[], cwd: string, home: string) {
+  const result = run('git', args, cwd, home);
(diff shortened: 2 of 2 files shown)


time-attack:
@16francej

marcusLewi:
Unsubscribe

PR #2221, opened 2026-07-10
fix(benchmark): gemini stream-json content/stats parse (#2159)

### LINKED ISSUE #2159: benchmark-models: GeminiAdapter.parseStreamJson reads obj.text but current gemini CLI emits obj.content — empty runs row as $0 success
## Symptom
`gstack-model-benchmark --models gemini` returns `output: "", tokens 0/0, costUsd 0` with NO error field — the row looks like a successful $0 run, silently mis-ranking gemini in any comparison table.

## Root cause
`test/helpers/providers/gemini.ts` `parseStreamJson` reads assistant text via:
```ts
if (obj.type === 'message' && typeof obj.text === 'string') { output += obj.text; }
```
but current gemini CLI `--output-format stream-json` emits `content` (with `role`), not `text`:
```json
{"type":"message","role":"assistant","content":"PONG","delta":true}
{"type":"result","status":"success","stats":{"input_tokens":19920,"output_tokens":32,...}}
```
Tokens are likewise under `result.stats` (`input_tokens`/`output_tokens`), not `result.usage.input_token_count`; the model name arrives on the `init` event.

Note `test/helpers/gemini-session-runner.ts` was already updated to read `event.content` (with the `role === 'assistant'` guard) — the benchmark provider adapter was missed.

## Repro
`gstack-model-benchmark --prompt 'Reply with exactly the word: PONG' --models gemini --output json` on gemini CLI ≥ 0.x current (emits content-shape events). Verified on v1.58.5.0.

## Suggested fix (verified locally)
Additive fallbacks in `parseStreamJson`: accept `obj.content` when `obj.role === 'assistant'` (the role guard matters — the CLI echoes the user prompt as a `role:'user'` message event, which would otherwise concat into output); `obj.usage ?? obj.stats` with `input_tokens`/`output_tokens` fallbacks; read `modelUsed` from `init`. With that patch the same probe returns `output: 'PONG', tokens 24946/32`.

## Hardening ask
Independent of the field fix: an adapter run that produces EMPTY output with exit 0 should surface as an error row (or a loud flag), never a $0-success row — empty-success is indistinguishable from a healthy cheap run in the table.

time-attack:
## Summary
Fixes [#2159](https://github.com/garrytan/gstack/issues/2159): `GeminiAdapter` was reading legacy `message.text` / `result.usage`, so current gemini CLI `content` + `stats` events produced empty output, 0 tokens, and no error — a silent $0 success that mis-ranked gemini in benchmark tables.

- Parse `message.content` when `role === 'assistant'` (ignore user-prompt echoes); keep legacy `text` fallback
- Read tokens from `result.stats` with legacy `usage` fallbacks; model from `init`
- Empty exit-0 output → `error` row (not a $0 success)
- `--skip-trust` for headless/temp workdirs (required by current gemini CLI)
- Recognize `GEMINI_API_KEY`; map IneligibleTier / auth failures cleanly
- Unit + post-CLI fixture tests; live-verified with API key (3/3 probes)

No `VERSION` / `CHANGELOG` / `package.json` bump — leaving release metadata to maintainers.

## Test plan
- [x] `bun test test/helpers/providers/gemini.test.ts` (+ related free tests)
- [x] Live: 3 separate `GeminiAdapter.run()` instances → real `PONG`/`ALPHA`/`OK READY`, positive tokens, no error
- [ ] Optional: `gstack-model-benchmark --prompt 'Reply with exactly the word: PONG' --models gemini --output json`

Closes #2159


time-attack:
basically gemini updated their cli so we had to adjust to that. Google has stated they will switch to anti gravity for cli and depreciate later but the current version is not compatible with model benchmark, when google does the full migration we can handle that

time-attack:
@16francej

PR #2222, opened 2026-07-10
fix(gstack-config): accept @local endpoint ids for brain_trust_policy

### LINKED ISSUE #1988: bug: gstack-config key validator rejects brain_trust_policy@local for stdio engines

### LINKED ISSUE #2210: gstack-config key validator rejects brain_trust_policy@local (endpoint-hash returns literal 'local' for local engines)
## What happened

`/setup-gbrain` Step 9.5 persists the brain trust policy with:

```bash
~/.claude/skills/gstack/bin/gstack-config set brain_trust_policy@$_HASH personal
```

where `_HASH=$(gstack-config endpoint-hash)`. For local engines (PGLite path), `endpoint-hash` returns the literal string `local`, but gstack-config's key validator only accepts hex suffixes after `@` (`@[a-f0-9]+`), so the `set` call is rejected and the policy is never persisted under the per-endpoint key.

## Impact

- Step 9.5 fails on every fresh local-PGLite setup (the default try-first path).
- On subsequent runs, `gstack-config get brain_trust_policy@local` returns unset, so skills see the policy as never-set and retry the auto-set each time (harmless but noisy, and the per-endpoint tracking the feature was built for never engages for local engines).
- `/sync-gbrain` Step 1's trust-policy gate hits the same mismatch.

## Workaround

Store under the plain key: `gstack-config set brain_trust_policy personal`. Works, but loses the per-endpoint scoping.

## Suggested fix (either works)

1. Allow `local` as a valid endpoint-hash value in the key validator (`@([a-f0-9]+|local)`), or
2. Have `endpoint-hash` return a stable hex digest for local engines instead of the literal `local`.

## Environment

- gstack 1.58.5.0, macOS (Darwin 25.5.0)
- gbrain 0.42.26.0, engine=pglite, mode=local-stdio
- Repro: fresh `/setup-gbrain` picking Path 3 (PGLite local), reach Step 9.5

time-attack:
## Summary
- `endpoint-hash` returns the literal `local` for stdio/PGLite engines, but key validation only allowed `@[a-f0-9]+`, so `/setup-gbrain` Step 9.5 and `/sync-gbrain` could not persist `brain_trust_policy@local`.
- Widen the `@` suffix to alphanumeric endpoint ids (hex hashes + `local`) in get/set validators and lookup.
- Add regression coverage for `@local` accept and punctuation reject; fix `user_slug_at_local` persist assertion.

Fixes #2210
Fixes #1988

## Test plan
- [x] `bun test browse/test/gstack-config.test.ts`
- [x] `bun test test/user-slug-fallback.test.ts`
- [x] Manual: origin/main binary rejects `set brain_trust_policy@local`; fixed binary persists `personal`
- [x] Consumer/control clones (gbrain, zotero-arxiv-daily, actionbook, plane, twenty): Step 9.5 before/after
- [x] plane (without-gstack) Path 3: init → MCP → `gbrain put`/search → policy persist with YC OpenAI key

### DIFF
--- a/bin/gstack-config
+++ b/bin/gstack-config
@@ -134,12 +134,12 @@ lookup_default() {
     redact_repo_visibility) echo "" ;; # empty → fall through to gh/glab detection
     redact_prepush_hook) echo "false" ;;
     # Brain-aware planning (v1.48 / T5+T10+T16). Defaults documented inline:
-    #   brain_trust_policy@<hash>  — unset on fresh install; setup-gbrain
+    #   brain_trust_policy@<endpoint-id>  — unset on fresh install; setup-gbrain
     #                                writes 'personal' for local engines,
     #                                asks the user for remote-ambiguous.
     #   salience_allowlist          — empty falls through to
     #                                SALIENCE_DEFAULT_ALLOWLIST (D9).
-    #   user_slug_at_<hash>         — empty triggers resolve-user-slug
+    #   user_slug_at_<endpoint-id>  — empty triggers resolve-user-slug
     #                                fallback chain (D4 A3) on first call.
     brain_trust_policy*) echo "unset" ;;
     salience_allowlist) echo "" ;;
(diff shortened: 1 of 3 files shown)


time-attack:
@16francej

PR #2224, opened 2026-07-10
Drop deprecated `--enable web_search_cached` from codex calls

kinoko-studio:
## What

Codex CLI (0.144.x) deprecated the `web_search_cached` feature — web search is now **on by default** — and prints this to stderr on every invocation that still passes `--enable web_search_cached`:

```
deprecated: `[features].web_search_cached` is deprecated because web search is enabled by default.
Set `web_search` to "live", "indexed", "cached", or "disabled" ... in config.toml if you want to override it.
```

gstack hardcodes `--enable web_search_cached` in every codex review/adversarial invocation, so that warning now fires on every `/ship`, `/review`, `/autoplan`, `/plan-*-review`, `/design-consultation`, `/office-hours`, and `/document-release` codex pass. The flag is a no-op now (search is default-on), so this removes it. Review behavior is unchanged — Codex still has web search.

## Changes

- **Sources:** drop ` --enable web_search_cached` from `scripts/resolvers/{review,design}.ts`, `autoplan/SKILL.md.tmpl`, `codex/SKILL.md.tmpl`.
- **Doc note:** reword the "Web search" paragraph in `codex/SKILL.md.tmpl` (search is default-on, no flag needed).
- **Regenerated** all skill docs via `bun run gen:skill-docs` (default + `--host codex` + `--host factory`).
- **Fixtures:** refresh `test/fixtures/golden/factory-ship-SKILL.md` (its ship skill inlines the codex blocks) and the unreferenced `test/fixtures/golden-ship-claude.md`.
- **Comment:** update the stale flag comment in `test/helpers/tool-map.ts`.

The Claude/Codex ship goldens were unaffected (those skills reference the section files via Read-pointers rather than inlining them, so the flag never appeared in `ship/SKILL.md`).

## Verification

- `bun test test/host-config.test.ts` → green (golden regression passes).
- Full `bun test test/` → exit 0.
- Doc-sync gate reproduced locally: after commit, `bun run gen:skill-docs` + `git diff --exit-code` is clean for the default host and `--host codex`.
- A live `codex exec` (v0.144.1) in the new no-flag shape returns cleanly with **no deprecation warning**.

## Notes

- `VERSION` left untouched — I assumed the bump is a maintainer step at release; happy to add one if your `version-gate` wants it on the PR.
- `.agents/` and `.factory/` are gitignored generated mirrors, so they're not in the diff; they regenerate flag-free.


kinoko-studio:
Closing in favor of #2249, which covers the same fix. Your diagnosis here was right — I live-tested it and confirmed `codex exec` performs real web searches with no search-related flag at all, so the old `--enable web_search_cached` really was just inert noise, not the thing keeping search on.

Went with replacing the flag with `--search` (in #2249) rather than dropping it outright, mainly because `codex-cli` has already changed this exact mechanism once (the old `features.web_search_cached` boolean → the new `web_search` config key), so pinning search on explicitly seemed more robust than depending on whatever the built-in default happens to be if it changes again. Open to revisiting if you'd rather keep this simpler bare-removal version instead — the two are otherwise equivalent in current behavior.

PR #2231, opened 2026-07-10
fix(browse): preserve live daemon during busy loads

### LINKED ISSUE #2219: browse: CLI kills a live-but-busy daemon during heavy page loads — ~7s health-probe window too short (follow-up to #1781); all session state lost
## Environment
- gstack: 1.58.4.0 (`9fd03fa`), CLI built from same tree
- OS: macOS 26.3.1 (arm64), Bun 1.3.9
- Playwright: chromium_headless_shell-1208
- Target app: local Next.js 15 dev server (heavy routes, cold compiles routinely 15–60s)

## Symptom
During a `/browse`-driven QA session the headless daemon "randomly" dies and gets replaced 4 times in ~15 minutes. Every replacement loses all session state (login cookies, viewport, tabs), forcing a re-login each time. Nothing in the visible CLI output explains it — the daemon is spawned with `stdio: 'ignore'`, and the `[browse] Server connection lost. Restarting...` / `[browse] Starting server...` lines are easy to lose when output is filtered (e.g. `| tail -1`).

## Root cause (evidenced, reproduced 3× with daemon stderr captured)
The daemon never crashes on its own. The CLI kills it.

Chain, with refs at v1.58.4.0:

1. A `goto` to a heavy dev route exceeds the 15s navigation timeout. Playwright throws, the daemon **responds normally** (audit log: `goto … status=error durationMs=15010`) — but the navigation keeps loading in Chromium.
2. The next CLI command arrives while the daemon/Chromium is still churning (CDP console/network event flood from the dev app). The fetch fails at connection level, and `probeHealthWithBackoff` (`browse/src/cli.ts:251` — 3 attempts × 250ms backoff, 2s timeout each, so roughly a 7s grace window) also fails.
3. `sendCommand`'s catch (`browse/src/cli.ts:586-605`) declares the daemon dead: `killServer(oldState.pid)` → SIGTERM → daemon replaced → **all cookies/login lost** (ephemeral profile).
4. The daemon's death is then silent/garbled for two separate reasons filed separately: SIGTERM "ignore" being cosmetic (see companion issue) and the `browser?.process()` TypeError in `resolveDisconnectCause` (#2085, secondary problem — confirmed on macOS, comment there).

Captured daemon stderr at the moment of death (identical in all 3 reproductions):

```
[browse] Received SIGTERM (ignoring — use /stop or Ctrl+C for intentional shutdown)
[browse] FATAL unhandled rejection: browser?.process is not a function. (In 'browser?.process()', 'browser?.process' is undefined)
```

Server-side audit log around one of the original deaths — note the daemon fully served the command, then was killed seconds later (next entry is the freshly spawned replacement):

```
15:11:11.135Z | viewport 1440x900 | ok  | dur: 27
15:11:26.233Z | goto /en/training | error | dur: 15010   <- daemon responded fine
15:11:37.532Z | screenshot        | ok  | dur: 34        <- answered by the REPLACEMENT daemon
```

## Why #1781's protections don't cover this
The #1781 fixes are present and working: the 30s AbortError path refuses to kill ("busy, not restarting"), and the connection-error path probes health before killing. But the probe window (~7s max) is shorter than how long a daemon stays unresponsive while Chromium loads a heavy dev-mode page with a timed-out navigation still in flight. So the "busy, not dead" case #1781 was designed to protect still ends in kill+restart — just via the connection-error branch instead of the AbortError branch.

Ruled out with evidence: kernel/jetsam kills (unified log clean), macOS crash reports (none), harness/session teardown (daemon is correctly detached: PPID 1, own pgid — and a manually session-isolated daemon died identically), `killOrphanChromium` (headed-profile lock only), terminal-agent watchdog (kills agents by identity record only), idle timeout.

## Repro (deterministic-ish, no special app needed)
1. Serve any route that takes >15s to finish loading (a cold-compiling Next.js dev route, or an endpoint that accepts and stalls).
2. Spawn the daemon with stderr captured: `BROWSE_PARENT_PID=0 bun run browse/src/server.ts 2>&1 | tee /tmp/browse.log`
3. `browse goto <slow-route>` → "Timeout 15000ms exceeded" (daemon healthy, nav still in flight).
4. Within ~5s, run any other browse command (`browse url`).
5. Observe: SIGTERM → tab closed → FATAL in the log; a fresh daemon answers the command; all session state gone.

Timing-dependent (needs the daemon to actually blow the ~7s probe window), so it fires reliably on heavy real pages and intermittently on synthetic ones. In my sessions it fired 4/4 times on cold dev routes.

## Suggested direction
- Lengthen/adapt the probe window when the last command was a navigation that timed out (the highest-probability "busy" state), or track "nav in flight" server-side and have `/health` report busy-but-alive so the CLI can distinguish.
- Consider reconnect-don't-respawn: if the PID is alive, keep waiting/retrying rather than killing — a lost session is strictly worse than a slow command for QA flows.
- On unavoidable restart, restore state automatically (the `state save/load` machinery already exists) so a kill doesn't cost the login.

time-attack:
During a heavy page load, browse could mistake a busy browser daemon for a dead one and restart it. That discarded open tabs, cookies, and logged-in sessions.

Browse now preserves an unresponsive daemon while its process is still alive and asks the user to retry. Automatic restart happens only after the daemon process is actually gone.

## Technical details

- Treat an OS-live daemon as busy when its HTTP health endpoint temporarily times out.
- Apply the safeguard during initial daemon discovery and command connection failures.
- Preserve existing recovery behavior for processes that are actually dead.
- Add regression coverage for live, dead, and missing daemon states.

## Evidence

The same live CLI probe ran in five isolated baseline/candidate repo pairs:

- gstack-auto: baseline failed; candidate passed
- zotero-arxiv-daily: baseline failed; candidate passed
- lacp: baseline failed; candidate passed
- stagehand: baseline failed; candidate passed
- plane: baseline failed; candidate passed

Every baseline killed the daemon and replaced its session state. Every candidate preserved the original PID and session token.

## Secondary checks

- 12 targeted tests passed
- `bun run build` passed
- Version and release files were not changed

Fixes #2219

### DIFF
--- a/browse/src/cli.ts
+++ b/browse/src/cli.ts
@@ -89,7 +89,7 @@ if (IS_WINDOWS && !NODE_SERVER_SCRIPT) {
   );
 }
 
-interface ServerState {
+export interface ServerState {
   pid: number;
   port: number;
   token: string;
@@ -103,6 +103,16 @@ interface ServerState {
   xvfbPid?: number;
   xvfbStartTime?: number;
   xvfbDisplay?: string;
+}
+
+/** A daemon whose process still exists may be busy even when HTTP health checks
+ * time out. Preserve it (and its browser session) instead of guessing that it
+ * is dead. The caller can retry once the current page load settles. */
+export function shouldPreserveUnresponsiveServer(
+  state: ServerState | null,
+  processAlive: (pid: number) => boolean = isProcessAlive,
+): boolean {
+  return !!state?.pid && processAlive(state.pid);
 }
 
 // ─── State File ────────────────────────────────────────────────
@@ -451,12 +461,14 @@ async function ensureServer(flags?: GlobalFlags): Promise<ServerState> {
     process.exit(1);
   }
 
-  // Guard: never silently replace a headed server with a headless one.
-  // Headed mode means a user-visible Chrome window is (or was) controlled.
-  // Silently replacing it would be confusing — tell the user to reconnect.
-  if (state && state.mode === 'headed' && isProcessAlive(state.pid)) {
-    console.error(`[browse] Headed server running (PID ${state.pid}) but not responding.`);
-    console.error(`[browse] Run '/open-gstack-browser' to restart.`);
+  // A live daemon can stop answering HTTP while Chromium finishes a heavy
+  // navigation. Restarting here destroys tabs, cookies, and login state. Only
+  // auto-restart after the daemon process itself is demonstrably gone.
+  if (shouldPreserveUnresponsiveServer(state)) {
+    console.error(`[browse] Server running (PID ${state!.pid}) but not responding — busy, not restarting.`);
+    console.error(state!.mode === 'headed'
+      ? `[browse] Retry shortly, or run '/open-gstack-browser' to restart intentionally.`
+      : '[browse] Retry shortly, or run `browse disconnect` to restart intentionally.');
     process.exit(1);
   }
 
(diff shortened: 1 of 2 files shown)


time-attack:
@time-attack Please rewrite daemon recovery so busy sessions remain protected while a live but permanently wedged daemon can still be replaced safely.

Issue #2144, opened 2026-07-01
gstack-memory-ingest silently ingests 0 pages on gbrain ≥0.42 (git-ignore fast path vs ignore-all ~/.gstack staging)


ValDola-stack:
## Summary

`gstack-memory-ingest` (bulk + incremental) silently ingests **0** transcript pages into a local PGLite brain when running against **gbrain ≥ 0.42**. The run reports success (`written: 86`) but `gbrain import` reports `0 imported, 0 unchanged, 0 failed`, so nothing ever reaches the index. Semantic search over transcripts/artifacts is silently empty.

Discovered during a fresh `/setup-gbrain` (PGLite local, Voyage embeddings) on macOS.

## Root cause (two-repo interaction)

1. `gstack-memory-ingest.ts → makeStagingDir()` stages pages at `~/.gstack/.staging-ingest-<pid>-<ts>/` — a **direct child of `$GSTACK_HOME`**. This is required by `lib/staging-guard.ts` (`checkOwnedStagingDir` enforces `dirname(canon) === home`) so the recursive-delete cleanup is safe.
2. `$GSTACK_HOME` (`~/.gstack`) is the **artifacts git repo**, whose `.gitignore` is `*` (ignore-everything; explicit allowlist via `git add -f`).
3. gbrain 0.42.x (#1159 `--respect-gitignore`) added a git-aware fast path in `collectSyncableFiles` (`src/commands/import.ts`): when the import dir is inside a git work tree it enumerates via `git ls-files --cached --others --exclude-standard`, which **honors `.gitignore`**.
4. Because the staging dir lives inside the ignore-all `~/.gstack` tree, `git ls-files --exclude-standard` returns **nothing** → `Found 0 markdown files` → `0 imported`.

### Reproduction
```
# inside ~/.gstack (gitignore = '*')
mkdir -p ~/.gstack/staging-proof && echo hi > ~/.gstack/staging-proof/x.md
git -C ~/.gstack ls-files --cached --others --exclude-standard -- ~/.gstack/staging-proof   # -> empty
gbrain import ~/.gstack/staging-proof --no-embed --json                                      # -> Found 0 markdown files
# same dir OUTSIDE any git tree imports fine
mkdir -p ~/staging-proof && echo hi > ~/staging-proof/x.md
gbrain import ~/staging-proof --no-embed --json                                              # -> imported 1
```

## Impact
[...]

PR #2232, opened 2026-07-10
fix(memory): enforce transcript trust, scope, and source isolation

### LINKED ISSUE #2140: memory-ingest: transcripts get no source isolation / repo-scope / trust enforcement → land in default (federated) or the code source; dream then mines them
## Summary

`gstack-memory-ingest` ingests coding-agent transcripts with **no source isolation, no repo/workspace scope, and no trust enforcement** — contrary to what `setup-gbrain` documents. The practical result: transcript pages land in the `default` source (often *federated*) or in the current repo's code source, and gbrain's `dream`/`extract` cycle then mines every one of them as if it were curated knowledge. On a real corpus this smears unrelated context across the brain and produced a **~21M-token blowup** the first time `dream` ran over the ingested transcripts (had to delete everything afterward).

Versions: gstack `1.58.5.0`, gbrain `0.42.42.0`.

## What the docs promise

- `setup-gbrain/SKILL.md:1456-1470` — transcript ingest default scope is **"current repo only, last 90 days"**, and **"Per-repo trust policies still apply."**
- `setup-gbrain/SKILL.md:1330` — `read-only` repos are skipped (but "enforced by the **future** auto-import hook … **not here**").
- `setup-gbrain/memory.md:35-37` — **"Repos under a `deny` trust policy … are skipped — neither code nor transcripts from those repos ingest."**

## What the code actually does

1. **Global walk, no repo/workspace filter.** `walkClaudeCodeProjects` / `walkCodexSessions` (`bin/gstack-memory-ingest.ts:362-423`) enumerate **all** of `~/.claude/projects/*` and `~/.codex/sessions/*`. `preparePages` (`:1130-1191`) filters only on mtime-dedup, optional secret-scan, and "unattributed" (no `cwd`/`git_remote`). There is no current-repo scope and no date floor beyond the 90-day window (`--all-history` removes even that).
2. **No trust enforcement.** The binary never reads `~/.gstack/gbrain-repo-policy.json`. `gstack-gbrain-repo-policy`'s own header says the tier is "**enforced at the caller level; this binary just stores the decision**." Nothing in `gstack-memory-ingest` or `gstack-gbrain-sync` consults it, so `deny`/`read-only` repos are **not** skipped for transcripts.
3. **No source isolation.** Ingest calls `gbrain import <stagingDir> --no-embed --json` (`bin/gstack-memory-ingest.ts:1419`) with **no `--source-id`**. gbrain's resolver only auto-adopts a source when exactly one non-default source exists (`src/commands/import.ts:131-146`, `sole_non_default` tier). With multiple registered sources, transcripts route to **`default`** (which is commonly *federated*) — or, when `GBRAIN_SOURCE`/`.gbrain-source` is active, to the **current code source**. Never to a dedicated, isolated transcript source.
4. **The cycle then mines them.** `dream`/`extract` run per-source and do a full walk with **no `type` filter** (`src/core/cycle/extract-facts.ts:184` → `engine.getAllSlugs()`; sourceId scoping at `:98`). Every `type: transcript` page is treated as fact/take source material → one LLM call per page/chunk → the token blowup, plus low-quality facts smeared into the brain.

Note: `gstack-config set transcript_ingest_mode off` does **not** stop this at sync time — `transcript_ingest_mode` is referenced only in the `setup-gbrain` skill files, never read by `bin/gstack-gbrain-sync.ts` or `bin/gstack-memory-ingest.ts`. The only effective off-switch is `gstack-gbrain-sync --no-memory` / `--code-only`.

## Repro (no writes)

```sh
# 1. Brain has >1 registered non-default source (typical multi-project setup).
gbrain sources list
# 2. A bare sync runs the global transcript walk and imports with no --source-id:
gstack-gbrain-sync --incremental   # -> runMemoryIngest -> gbrain import <dir> (no --source-id)
# 3. Resulting transcript pages carry source_id = 'default' (or the CWD code source),
#    not an isolated transcript source. dream/extract over that source then mine them.
```

## Suggested fixes

1. **Route transcripts to a dedicated `--source-id`** (default a **non-federated** source, e.g. `transcripts-<workspace>`), never `default`/the code source.
2. **Enforce repo-trust + an actual scope** *before* the walk: honor `gbrain-repo-policy.json` (skip `deny`/`read-only`), default to current-repo, and support a hard date floor — matching what `setup-gbrain` already promises.
3. **Let the cycle exclude `type: transcript`** (or make transcript ingest opt-in to mining), so `dream`/`extract` don't burn tokens treating raw transcripts as curated knowledge.
4. Make `transcript_ingest_mode=off` actually gate the sync memory stage (today it's docs-only).

## Related

Pairs with #2104 (staging inside `GSTACK_HOME`'s `gitignore=*` repo → git-aware import sees 0 files). Together: when the staging bug is fixed, transcripts *will* flow — and without this issue's scoping/isolation they'll flow into the wrong source and get mined.

time-attack:
Transcript syncing ignored the off switch and repository trust boundaries, so private sessions from unrelated or blocked projects could be imported into the wrong shared knowledge source.

Transcript syncing now stops when disabled, accepts only the current repository while honoring blocked and read-only policies, and imports sessions into a separate repository-specific transcript source.

## Changes

- honor `transcript_ingest_mode=off`
- limit transcript ingestion to the current repository
- exclude repositories marked `deny` or `read-only`
- separate transcript and artifact import batches
- route transcripts through a stable, gbrain-valid dedicated source ID
- add regression coverage for trust, scope, off mode, and source isolation

## Verification

The identical issue-specific probe ran against isolated `origin/main` baseline and candidate environments:

| Repository | Baseline | Candidate |
|---|---:|---:|
| actionbook/actionbook | fail | pass |
| ahtavarasmus/lightfriend | fail | pass |
| garrytan/gbrain | fail | pass |
| makeplane/plane | fail | pass |
| twentyhq/twenty | fail | pass |

Targeted tests: 70 passed, 0 failed.

## Out of scope

Transcript mining behavior tracked separately in #2144 was not changed.

Fixes #2140

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -520,11 +520,63 @@ function* walkGstackArtifacts(ctx: WalkContext): Generator<{ path: string; type:
 }
 
 function* walkAllSources(ctx: WalkContext): Generator<{ path: string; type: MemoryType }> {
-  if (ctx.args.sources.has("transcript")) {
+  if (ctx.args.sources.has("transcript") && transcriptIngestMode() !== "off") {
     yield* walkClaudeCodeProjects(ctx);
     yield* walkCodexSessions(ctx);
   }
   yield* walkGstackArtifacts(ctx);
+}
+
+function transcriptIngestMode(): string {
+  try {
+    const raw = readFileSync(join(GSTACK_HOME, "config.yaml"), "utf-8");
+    const match = raw.match(/^transcript_ingest_mode:\s*([^#\s]+)\s*(?:#.*)?$/m);
+    return match?.[1]?.toLowerCase() || "";
+  } catch {
+    return "";
+  }
+}
+
+let cachedCurrentRepoRemote: string | undefined;
+function currentRepoRemote(): string {
+  if (cachedCurrentRepoRemote !== undefined) return cachedCurrentRepoRemote;
+  try {
+    cachedCurrentRepoRemote = canonicalizeRemote(
+      execFileSync("git", ["-C", process.cwd(), "remote", "get-url", "origin"], {
+        encoding: "utf-8",
+        timeout: 5_000,
+        stdio: ["ignore", "pipe", "ignore"],
+      }).trim(),
+    );
+  } catch {
+    cachedCurrentRepoRemote = "";
+  }
+  return cachedCurrentRepoRemote;
+}
+
+let cachedRepoPolicy: Record<string, unknown> | undefined;
+function repoTrustTier(remote: string): string {
+  if (!remote) return "unset";
+  try {
+    cachedRepoPolicy ??= JSON.parse(
+      readFileSync(join(GSTACK_HOME, "gbrain-repo-policy.json"), "utf-8"),
+    ) as Record<string, unknown>;
+    const tier = cachedRepoPolicy[remote];
+    return typeof tier === "string" ? tier : "unset";
+  } catch {
+    cachedRepoPolicy = {};
+    return "unset";
+  }
+}
+
+function transcriptSourceId(): string {
+  const remote = currentRepoRemote();
+  const readable = repoSlug(remote)
+    .replace(/[^a-z0-9]+/gi, "-")
+    .replace(/^-+|-+$/g, "")
+    .toLowerCase() || "current-repo";
+  const hash = createHash("sha256").update(remote || process.cwd()).digest("hex").slice(0, 8);
+  return `transcripts-${readable.slice(0, 11)}-${hash}`;
 }
 
 // ── Renderers ──────────────────────────────────────────────────────────────
@@ -1171,6 +1223,16 @@ function preparePages(
           skippedUnattributed++;
           continue;
         }
+        const currentRemote = currentRepoRemote();
+        const tier = repoTrustTier(page.git_remote || "");
+        const explicitlyUnattributed = (!page.git_remote || page.git_remote === "_unattributed") && args.includeUnattributed;
+        if (
+          !explicitlyUnattributed &&
+          (!currentRemote || page.git_remote !== currentRemote || tier === "deny" || tier === "read-only")
+        ) {
+          skippedUnattributed++;
+          continue;
+        }
         if (page.partial) partialPages++;
       } else {
         page = buildArtifactPage(path, type);
@@ -1409,6 +1471,7 @@ export function resolveImportTimeoutMs(
 function runGbrainImport(
   stagingDir: string,
   timeoutMs: number,
+  sourceId?: string,
 ): Promise<{ status: number | null; stdout: string; stderr: string; timedOut: boolean }> {
   installSignalForwarder();
   return new Promise((resolve) => {
@@ -1416,7 +1479,9 @@ function runGbrainImport(
     // inside Next.js / Prisma / Rails projects with their own
     // .env.local (codex review #7 — defense in depth on top of the
     // parent gstack-gbrain-sync seeding the bun grandchild's env).
-    const child = spawnGbrainAsync(["import", stagingDir, "--no-embed", "--json"]);
+    const importArgs = ["import", stagingDir, "--no-embed", "--json"];
+    if (sourceId) importArgs.push("--source-id", sourceId);
+    const child = spawnGbrainAsync(importArgs);
     _activeImportChild = child;
     let stdout = "";
     let stderr = "";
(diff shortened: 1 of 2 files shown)


time-attack:
@16francej

chris-conte:
I filed the #2140 comment about the unsourced `gbrain import` spawn before finding this PR — it's yours and it predates my comment by two months. Two things I can offer rather than duplicate it.

**A version-safety gap.** The import spawn pushes the flag unconditionally:

```ts
if (sourceId) importArgs.push("--source-id", sourceId);
```

`gbrain import` has only accepted `--source-id` since v0.37.7.0, and `gstack-memory-ingest.ts` pins no gbrain version floor. On an older gbrain the flag is silently swallowed rather than rejected — I checked against gbrain at garrytan/gbrain@06f59468a, before `--source-id` existed (added in garrytan/gbrain#1167, shipped in the v0.37.7.0 wave): that parser's `flagValues` set excludes only `--workers`' value, so `--source-id` is never read, the run exits 0 with `status: success`, and the page lands in `default` anyway. That makes the probe more important, not less — the failure mode is a silent no-op wearing a success mask, which is the #2140 bug itself. The repo already has a pattern for exactly this, two lines in the sibling wireup:

```bash
# bin/gstack-gbrain-source-wireup:343-344
if gbrain sync --help 2>/dev/null | grep -q -- '--source'; then
  sync_cmd=(gbrain sync --source "$id")
```

Same shape here — probe `gbrain import --help | grep -q -- '--source-id'` and skip the flag when absent — so an old install falling back to today's behavior is a deliberate, visible decision rather than a flag quietly evaporating. (Note import's flag is `--source-id`, not `sync`'s `--source`.)

**Verification is currently impossible from the caller's side.** `gbrain import --json` returns `status / duration_s / imported / skipped / errors / chunks / total_files` and no destination, so nothing in the ingest can confirm the routing actually took effect — a silently-ignored `--source-id` and a working one produce byte-identical output. I've opened garrytan/gbrain#4779 to add `source_id` to that payload, which would let this PR's routing be asserted rather than assumed.

**Offer.** I have a hermetic scratch-brain harness for this — PGLite brain, two registered sources, staged pages imported from an unregistered dir, then `gbrain sources list` to show where they landed. It's what produced the before/after in my #2140 comment. Happy to run it against this branch and post the output here as independent verification, or hand you the recipe if you'd rather run it yourself. Just say which.

For what it's worth, the batch split in `runScopedIngest` is the part I'd underestimated — routing on the bare invocation rather than only under `--sources transcript` is the difference between fixing this for people who opt in and fixing it for everyone.

chris-conte:
Ran the harness against `6249074d`. Hermetic PGLite brain, two registered non-default sources so `sole_non_default` can't fire, one unattributed Claude Code transcript, everything under a scratch HOME.

Starting with the good part: **the routing design works.** Registering the source id your code targets and re-running:

```
[1] transcripts/claude-code/_unattributed/2026-09-01-aaaa1111
[memory-ingest] gbrain import: 1 imported, 0 unchanged, 0 failed
  written:               1
```

```
  default                           federated   0 pages
  source-a                          unset       0 pages
  source-b                          unset       0 pages
  transcripts-garrytan-gs-31d6f839  unset       1 pages
```

Transcript in the dedicated source, `default` empty, on a brain where the auto-route can't fire. That's #2140's ask, satisfied.

**The blocker is that nothing registers that source.** Without the manual `sources add`, the same run does this:

```
[memory-ingest] ERR: gbrain import exited 1
Ingest pass complete (bulk):
  written:               0
  failed:                1
```

Nothing lands anywhere, `default` included. The ingest only echoes the last stdout line, so I put a pass-through shim on `gbrain` to catch the actual error:

```
argv: import <staging> --no-embed --json --source-id transcripts-garrytan-gs-31d6f839
exit: 1
  Warning: skipped transcripts/claude-code/_unattributed/2026-09-01-aaaa1111.md:
  insert or update on table "pages" violates foreign key constraint "pages_source_id_fkey"
  {"imported":0,"skipped":1,"errors":1,...}
```

`transcriptSourceId()` computes the id, but `pages.source_id` is a foreign key onto `sources.id`, so every page is rejected. The transcripts aren't misrouted — they're dropped. Looks like a `gbrain sources add <id> --path <dir>` (non-federated) ahead of the import, or a create-if-missing at the spawn site.

One dead end in case it saves you the detour: my first counterfactual registered `transcripts-current-rep-…` and still failed. Wrong id — the readable half comes from the git remote (`garrytan/gstack` → `garrytan-gs`); `process.cwd()` is only the fallback when there's no remote.

**Correcting something I said above.** I claimed an older gbrain would give an unknown-flag failure. It doesn't. Checked against gbrain at garrytan/gbrain@06f59468a, before `--source-id` existed (added in garrytan/gbrain#1167, shipped in the v0.37.7.0 wave): that parser's `flagValues` set excludes only `--workers`' value, so `--source-id` is never read, `stagingDir` is still picked up as the positional, and the run exits 0 with `status: success` and the page in `default`. A silent no-op, not a break. I've fixed that paragraph in place. It makes the `--help` probe more worth adding than I first argued, since a hard error would at least be visible.

**Base drift, unrelated to any of the above:** this branch is on v1.60.1.0 and main is now v1.77.0.0. The import spawn has moved from `:1419` to `:1778` and a fair amount around it has changed, so a rebase will probably be more than a trivial one.

Harness is a single hermetic bash script — happy to hand it over, or to re-run it on any push and post the output here. Whichever is more useful to you.

PR #2233, opened 2026-07-10
fix(setup): bound Playwright bootstrap and preserve skill registration

time-attack:
On some Macs, setup could wait forever while checking or downloading Chromium. Retrying started more stuck installers, and setup stopped before registering commands such as /qa and /ship.

Setup now limits how long browser checks and downloads may run, allows only one setup at a time, cleans up timed-out child processes, and still registers commands when Chromium is unavailable.

## Technical changes

- use Node instead of Bun for Chromium launch probing on macOS
- bound launch probes and Chromium installation
- terminate timed-out descendant processes
- serialize setup with a portable stale-PID lock
- continue skill registration after Chromium failure
- add focused regression coverage

## Live verification

Identical baseline/candidate probes were run in isolated clones and HOME directories:

- TideDra/zotero-arxiv-daily: fail → pass
- loperanger7/gstack-auto: fail → pass
- garrytan/gbrain: fail → pass
- browserbase/stagehand normal-install control: pass → pass
- makeplane/plane normal-install control: pass → pass

The evidence gate passed and no marked child processes remained.

## Secondary tests

- `bash -n setup`
- `bun test test/setup-playwright-deadline.test.ts test/setup-windows-fallback.test.ts` — 12 passed

### DIFF
--- a/setup
+++ b/setup
@@ -30,6 +30,82 @@ case "$(uname -s)" in
   MINGW*|MSYS*|CYGWIN*|Windows_NT) IS_WINDOWS=1 ;;
 esac
 
+# Only one setup may mutate an install/HOME at a time. An atomic directory is
+# portable to macOS, Linux, and Git Bash (unlike flock), and the pid lets a
+# later run recover after a crash.
+SETUP_LOCK_DIR="${GSTACK_HOME:-$HOME/.gstack}/.setup.lock.d"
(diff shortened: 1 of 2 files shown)


time-attack:
@time-attack Please rewrite setup so registration survives missing Node, empty locks remain recoverable, and tests execute the actual setup behavior rather than only indirect helpers.

PR #2249, opened 2026-07-12
fix(codex): replace dead --enable web_search_cached flag with --search

kinoko-studio:
## Summary
- `codex-cli` 0.144.1 marks `web_search_cached` deprecated (`codex features list`) and prints, on every invocation that still passes `--enable web_search_cached`:
  ```
  deprecated: `[features].web_search_cached` is deprecated because web search is enabled by default.
  Set `web_search` to "live", "indexed", "cached", or "disabled" ... in config.toml if you want to override it.
  ```
- **Live-tested:** `codex exec` with *no* search-related flag at all still performs real web searches successfully. So the old flag wasn't keeping search on — it was just inert, printing a noisy warning on every gstack codex pass (autoplan, codex, design-consultation, design-review, document-release, office-hours, plan-ceo-review, plan-design-review, plan-devex-review, plan-eng-review, review, ship, plus two `scripts/resolvers/*.ts` files and the `.tmpl` counterparts).
- This PR replaces the dead flag with `--search` rather than dropping it outright (see #2224 for the bare-removal alternative). Rationale for the explicit flag over relying on codex's implicit default: `codex-cli` has already changed this exact mechanism once (the old `features.web_search_cached` boolean → the new `web_search` config key), so pinning search on explicitly is more robust to it changing again than depending on whatever the built-in default happens to be.
- Also corrects the "Web search" doc note in `codex/SKILL.md`(`.tmpl`), which described the old flag as "OpenAI's cached index" — inaccurate for `--search`'s live lookup via the native Responses `web_search` tool.

## Test plan
- [x] Verified `web_search_cached` is deprecated via `codex features list` and the live stderr warning on codex-cli 0.144.1
- [x] Verified `codex exec` performs successful real web searches with *no* search flag at all (rules out "the old flag was required for search to work")
- [x] `grep -r "enable web_search_cached"` returns zero matches post-fix
- [x] Confirmed `--search` is present at every prior call site
- [x] `bun test test/host-config.test.ts` (golden-file regression) — green
- [x] Full `bun test browse/test/ test/ make-pdf/test/ ...` — exit 0
- [x] `bun run skill:check` clean across all host targets after `bun run gen:skill-docs --host all`

Supersedes #2224 (bare removal) — closing that one with this rationale.


16francej:
Closing — verified against codex-cli 0.144.4: --search is not accepted by codex exec / codex review, so this change would break every codex invocation in gstack. Removal of the deprecated flag is being handled separately.

PR #2252, opened 2026-07-13
fix(test): delayed process.exit(0) in 8 test files silently truncates full bun-test runs

whd4:
## The bug

Eight test files end `afterAll` with `setTimeout(() => process.exit(0), 500)`:
snapshot, commands, batch, handoff, compare-board, content-security, security-live-playwright, and design/feedback-roundtrip.

In a full `bun test` run, that timer fires ~500ms into the **next** file and kills the runner with exit code 0 and no failure tally. A full-suite run therefore only ever executes the files before the first exit, and still reports success. Sharded per-file CI (like this repo's) structurally never sees it — but anyone running `bun test` across the whole suite gets silently truncated green results.

## The fix (2 commits)

- **fix(test): un-leak GSTACK_HOME across files, file-scoped process.exit guard + idle-timer stop, closeTab zero-tab invariant, align config tests to DEFAULTS contract** — cross-file test-env hygiene so files no longer poison each other.
- **fix(test): remove delayed process.exit(0) time bombs from 8 browser test files + exitFn seam in ServerConfig** — deletes the bombs and adds an `exitFn` injection seam so tests can assert exit behavior without killing the runner.

## Honest caveat

After this fix, a full run no longer truncates — which *reveals* pre-existing failures that were always being skipped (in our environment 200+, mostly browser `goto` timeouts under heavy load, plus some real breaks such as `session.clearLoadedHtml is not a function` in design tests). This PR does not claim to make the suite green; it makes the runner honest so failures are visible and fixable.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -803,15 +803,14 @@ export class BrowserManager {
     this.tabSessions.delete(tabId);
     this.tabOwnership.delete(tabId);
 
-    // Switch to another tab if we closed the active one
-    if (tabId === this.activeTabId) {
+    // Never leave zero tabs, regardless of which tab was closed — activeTabId
+    // can point at an already-removed id, and the old active-only guard then
+    // skipped auto-create when the last tab closed.
+    if (this.pages.size === 0) {
+      await this.newTab();
+    } else if (tabId === this.activeTabId) {
       const remaining = [...this.pages.keys()];
-      if (remaining.length > 0) {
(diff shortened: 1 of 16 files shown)


time-attack:
@16francej — recommend closing this PR because #2230 supersedes it as the complete test-runner fix. Although #2252 adds useful scoped exit and timer injection plus environment restoration, it still leaves a real daemon exit path, unsafe teardown limits, global stubs, and no recurrence guard. Port those ideas into #2230’s three-second, daemon-safe design.

PR #2261, opened 2026-07-14
feat: Add Google Antigravity ecosystem support

VKrishna04:
- Implements headless mode adapter for 100% native gstack execution inside Antigravity.

- Adds 1-click cross-platform setup script (Windows junctions / UNIX symlinks) for transparent AI skill discovery.

- Passes all uniqueness and catalog generation tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ bin/gstack-global-discover*
 .hermes/
 .gbrain/
 .gbrain-source
+.antigravity/
 .context/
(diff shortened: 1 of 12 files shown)


time-attack:
@16francej — recommend closing this PR. Thanks for putting this together. We aren’t accepting new standalone skills right now, so we’re going to pass on this proposal. We can revisit the area later if it becomes part of GStack’s core workflow.

PR #2262, opened 2026-07-14
fix(setup): link shared runtime libraries

fedster99:
## What changed

- Link `lib/` beside `bin/` in the shared agent sidecar and the Codex, Factory, OpenCode, and Kiro runtime roots.
- Add setup validation coverage for every affected runtime root.

## Why

Several scripts in `bin/` import shared modules through `../lib`. Minimal runtime installs exposed `bin/` without `lib/`, so commands such as `gstack-learnings-log` failed after a fresh Codex setup with `Cannot find module .../lib/jsonl-store.ts`.

Keeping the two directories together makes the generated runtime self-contained and prevents a later gstack upgrade from restoring the broken layout.

## Impact

Fresh and upgraded installs can run logging, memory, redaction, and other shared-library-backed commands from their host-specific runtime root. Unix symlink and Windows copy behavior both continue to use `_link_or_copy`.

## Validation

- `bash -n setup`
- Generated Codex fixtures, then ran the five affected setup assertions on Bun 1.3.14: 5 passed
- `bun test test/setup-windows-fallback.test.ts`: 8 passed
- Fresh temporary-HOME install with `./setup --host codex --no-prefix`
- End-to-end `gstack-learnings-log` and `gstack-learnings-search` smoke passed from that installed runtime

The broader free-test command reached an unrelated environment-dependent timeout in `global-discover.test.ts`; no changed setup/runtime test failed.

### DIFF
--- a/setup
+++ b/setup
@@ -748,8 +748,9 @@ create_agents_sidecar() {
   local agents_gstack="$repo_root/.agents/skills/gstack"
   mkdir -p "$agents_gstack"
 
-  # Sidecar directories that skills reference at runtime
-  for asset in bin browse review qa; do
+  # Sidecar directories that skills reference at runtime. bin scripts import
+  # shared modules via ../lib, so bin and lib must always travel together.
+  for asset in bin lib browse review qa; do
     local src="$SOURCE_GSTACK_DIR/$asset"
     local dst="$agents_gstack/$asset"
     if [ -d "$src" ] || [ -f "$src" ]; then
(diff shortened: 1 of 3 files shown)


time-attack:
@fedster99 Add tests that run a command requiring the installed sibling library across every affected host root, including Windows copy behavior.

fedster99:
Done in f412cf1 — added `test/setup-runtime-lib-command.test.ts`, which goes beyond static validation and **executes** a lib-importing command from every affected host root:

- Builds each runtime root (agents sidecar, Codex, Factory, OpenCode, Kiro) in a sandbox using the **real setup shell code** (the extracted `_link_or_copy` + each root-builder function/block).
- Runs `gstack-learnings-log` — the exact command from the bug report, which imports `$SCRIPT_DIR/../lib/jsonl-store.ts` via bun — from the installed root and asserts it exits 0 and appends the learning.
- Covers **Windows copy behavior**: every root is exercised twice, `IS_WINDOWS=0` (symlink install, asserts `lib` is a symlink) and `IS_WINDOWS=1` (copy install, asserts `lib` is a real directory copy), following the `setup-windows-fallback.test.ts` precedent (skipped on actual win32 runners where `ln -snf` silently degrades).
- Includes a **negative control**: a root with `bin/` but no `lib/` (the pre-fix layout) must fail on the `../lib` import — so the positive cells can't pass vacuously.

Verified the tests detect the regression: against upstream/main's `setup` (pre-fix), all 10 positive cells fail; with this PR's fix, 11/11 pass in ~3s.

marcusLewi:
Unsubscribe

PR #2263, opened 2026-07-14
fix(relink): discover skills structurally so `browse` is materialized

topcoder1:
## Problem

`gstack-relink` gates skill discovery on a hardcoded meta-directory name list:

```sh
case "$skill" in bin|browse|design|docs|extension|lib|node_modules|scripts|test|.git|.github) continue ;; esac
```

`browse` is in that list — but `browse` is a **real, live skill**: `browse/SKILL.md` exists with `name: browse`, `version: 1.1.0`, "Fast headless browser for QA testing and site dogfooding". Two shipped skills depend on it by name: `benchmark` ("Performance regression detection using the browse daemon") and `setup-browser-cookies` ("Import cookies … into the headless browse session").

The result: `gstack-relink` never materializes `skills/browse`, so the `browse` skill is unavailable. It only appeared to work historically because a downstream consumer tracked a `skills/browse` symlink that papered over the exclusion. Once that symlink is removed, `browse` is permanently missing on any fresh install (or after any `skills/` rebuild) and cannot self-heal via relink.

## Fix

`browse` is the **only** one of the nine non-dot excluded names that actually carries a `SKILL.md`. The eight others (`bin`, `design`, `docs`, `extension`, `lib`, `node_modules`, `scripts`, `test`) have none — so the structural guard already on the next line handles them for free:

```sh
[ -f "$skill_dir/SKILL.md" ] || continue
```

This PR drops the redundant name list and makes that structural test the sole gate. A directory is a skill iff it carries a `SKILL.md`.

This is not a new convention — it aligns relink with the two sibling discovery paths that are already structural:

- **`scripts/discover-skills.ts`** (canonical shared discovery, used by CI + doc-gen): its skip set is only `{ node_modules, .git, dist }` — it already treats `browse` as a skill.
- **`bin/gstack-patch-names`**: iterates the same dirs using only `[ -f "$skill_dir/SKILL.md" ]`, no name list.

It also future-proofs `design`/`docs`/`test` should any ever become real skills.

### Safety

relink's glob is top-level `*/` only and stats just `$dir/SKILL.md`. The three names in discover-skills' skip set are already handled: `.git` is excluded by the glob (dotdirs never match `*/`), and neither `node_modules/SKILL.md` nor a top-level `dist/SKILL.md` exists (`dist` only appears nested, e.g. `browse/dist`). No top-level `test/`/`docs/` `SKILL.md` exists, so no fixture is mistaken for a skill.

## Tests

`test/relink.test.ts` gains three regression tests (suite: 27 pass, 0 fail — was 25 pass + 2 fail before the fix):

- **real payload**: enumerates every top-level `*/SKILL.md` in the repo and asserts each materializes — fails on exactly `browse` without the fix.
- **future-proofing**: mocks an install with dirs named after every former-exclusion entry, each carrying a `SKILL.md`, and asserts all materialize.
- **companion invariant**: dirs *without* a `SKILL.md` are still skipped.

Before: `Relinked 53 skills`. After: `Relinked 54 skills`, with `browse/SKILL.md` present.

### DIFF
--- a/bin/gstack-relink
+++ b/bin/gstack-relink
@@ -62,8 +62,10 @@ SKILL_COUNT=0
 for skill_dir in "$INSTALL_DIR"/*/; do
   [ -d "$skill_dir" ] || continue
   skill=$(basename "$skill_dir")
-  # Skip non-skill directories
-  case "$skill" in bin|browse|design|docs|extension|lib|node_modules|scripts|test|.git|.github) continue ;; esac
+  # A directory is a skill iff it carries a SKILL.md. This structural test — not a
+  # hardcoded name list — is the sole gate, so real skills whose names happen to
+  # collide with support-dir names (e.g. browse) are never dropped, and support dirs
+  # (bin, lib, node_modules, …) are skipped because they have no SKILL.md.
   [ -f "$skill_dir/SKILL.md" ] || continue
 
   if [ "$PREFIX" = "true" ]; then
--- a/test/relink.test.ts
+++ b/test/relink.test.ts
@@ -406,6 +406,76 @@ describe('gstack-relink (#578)', () => {
     expect(fs.existsSync(path.join(skillsDir, 'gstack-qa'))).toBe(true);
     expect(fs.existsSync(path.join(skillsDir, 'gstack-ship'))).toBe(true);
   });
+
+  // REGRESSION: skill discovery must be structural (has SKILL.md), never name-based.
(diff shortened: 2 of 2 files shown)


time-attack:
@topcoder1 Restore reserved-directory exclusions while keeping structural discovery for `browse`, so dependency and build content cannot be exposed as agent skills.

Issue #2124, opened 2026-06-27
Windows Smart App Control blocks the unsigned browse.exe — sign release binaries


pstilwell90:
## Summary

On Windows 11 with **Smart App Control (SAC)** enabled, the bundled `browse` binary cannot execute, which breaks every browser-driven skill (`/canary`, `/qa`, `/browse`, live `/design-review`). Please **code-sign the Windows release binaries** so SAC permits them.

## Environment

- gstack `1.58.0.0`
- Windows 11, **Smart App Control: On (Enforce)** — `HKLM:\SYSTEM\CurrentControlSet\Control\CI\Policy\VerifiedAndReputablePolicyState = 1`
- Binary: `~/.claude/skills/gstack/browse/dist/browse.exe` (PE32+ x86-64)

## What happens

Running the binary fails at the OS policy layer:

- **Git-Bash:** `.../browse.exe: Permission denied` (x-bits are set; not a chmod issue)
- **PowerShell:** `Program 'browse.exe' failed to run: An error occurred trying to start process ... An Application Control policy has blocked this file.`

`file` reports it as `PE32+ executable for MS Windows`, and it is **unsigned**.

## Root cause

Smart App Control blocks unsigned / low-reputation executables. Critically, **SAC has no per-file allowlist** — there is no supported way for a user to make one binary an exception. The only user-side workaround is turning SAC **Off**, which is **irreversible** (Windows requires a reset/reinstall to re-enable SAC). Asking users to permanently disable a security feature to run a dev tool is not a reasonable workaround.

## Impact

Any Windows 11 machine with SAC enabled (the default on many new devices) cannot use the browser-driven skills at all. `bin/` shell scripts run fine — only the compiled binaries are blocked.

## Ask

1. **Code-sign the Windows release binaries** (Authenticode), and ideally submit them for Microsoft reputation so SAC trusts them without per-machine action. This is the real fix.
2. Until signed: document the SAC limitation + fallbacks (curl-based health checks, user-driven browser QA) so Windows users aren't stuck debugging "Permission denied".

## Minor secondary note

`dist/browse` (no `.exe` extension) cannot be exec'd from Git-Bash even setting SAC aside — Git-Bash won't run a PE without the `.exe` suffix. Only `dist/browse.exe` is invokable. Skills compute the binary path as `dist/browse`; on Windows they should prefer `dist/browse.exe`.

optimization2026:
This is a release provenance problem, not just a Windows support problem.

[...]

PR #2265, opened 2026-07-15
fix(windows): add Smart App Control safe browser fallback

cko32002:
## Summary

- keep the native Windows browse.exe intact for future signing and release workflows
- stage an exact extensionless Git Bash launcher after Windows builds
- discover a complete gstack source checkout and fail closed when source or dependencies are incomplete
- run the browser client through the installed Bun and Node toolchain without disabling Smart App Control
- add executable regression coverage for Windows staging, non-Windows no-op behavior, source discovery, argument forwarding, and browse.exe preservation

## Validation

- `bun test test/windows-appcontrol-launcher.test.ts test/setup-windows-fallback.test.ts`: 10 passed, 4 skipped, 0 failed
- `bash -n` on both launchers, build script, and setup: passed
- full Windows Git Bash build: passed
- generated extensionless launcher navigated to a local app with HTTP 200 and returned healthy browser status
- Windows Smart App Control remained enabled throughout the local smoke test
- independent recheck verdict: ACCEPT

## Scope

This is a functional fallback for Windows installations where Smart App Control blocks the unsigned Bun-compiled executable. It does not Authenticode-sign `browse.exe`, and it does not close the signing and release-provenance request in #2124.

The launcher is intentionally Git Bash-only, matching gstack's documented Windows runtime. No `.cmd` wrapper is installed, which avoids batch metacharacter re-parsing and command-injection risk.

Addresses #2124.

### DIFF
--- a/browse/bin/stage-windows-browse-launcher
+++ b/browse/bin/stage-windows-browse-launcher
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+PLATFORM="${1:-$(uname -s)}"
+case "$PLATFORM" in
+  MINGW*|MSYS*|CYGWIN*|Windows_NT) ;;
+  *) exit 0 ;;
+esac
+
+ROOT="${2:-$(cd "$(dirname "$0")/../.." && pwd -P)}"
+if command -v cygpath >/dev/null 2>&1; then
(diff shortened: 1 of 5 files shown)


time-attack:
@cko32002 Make the launcher discover source checkouts created by the documented Codex installation flow, and add tests for the required Windows behavior.

PR #2267, opened 2026-07-15
fix(codex): use request_user_input for workflow decisions

### LINKED ISSUE #1066: Codex skills that call AskUserQuestion silently fail in Default mode

isaiahratahi:
## Summary

- make gstack's interactive-question preamble host-aware
- generate native `request_user_input` guidance for Codex while preserving Claude output
- fail closed outside Codex Plan mode instead of silently auto-selecting
- preserve every four-or-more-choice decision through sequential 2-3-option controls
- reject generated Codex markdown that retains literal `AskUserQuestion` instructions

Closes #1066.

## Codex behavior

- The complete decision brief stays in the structured question text; labels remain concise.
- One control is issued at a time, followed by a hard stop for the answer.
- In Default mode or whenever `request_user_input` is unavailable, the workflow reports `BLOCKED`, directs the user to Plan mode, and stops.
- No experimental Default-mode feature flag is enabled.

This follows Codex's documented
[Plan-mode contract](https://github.com/openai/codex/blob/main/codex-rs/collaboration-mode-templates/templates/plan.md),
where `request_user_input` is the native structured control.

## Validation

- focused resolver, preamble, host-config, generation, and golden tests
- Codex and Claude generation freshness checks
- generated Codex markdown scan for legacy or bogus tool names
- post-rebase targeted changed-area suite: 529 passed, 0 failed
- full free-suite attempted with an isolated home; the changed-area tests passed,
  while unrelated local baseline/environment checks failed because XCTest was
  unavailable, an installed-home fixture was absent under the isolated home,
  and one existing path-safety test timed out; upstream CI is the clean-room gate

### DIFF
--- a/hosts/codex.ts
+++ b/hosts/codex.ts
@@ -30,6 +30,13 @@ const codex: HostConfig = {
     { from: '.claude/skills/review', to: '.agents/skills/gstack/review' },
     { from: '.claude/skills', to: '.agents/skills' },
   ],
+  toolRewrites: {
+    'Do NOT use AskUserQuestion for interactive prompts. Auto-choose the recommended option.':
+      'If request_user_input is unavailable, report BLOCKED and stop; never auto-choose an answer.',
+    'AskUserQuestion with 4 options': 'request_user_input using the sequential 2-3-option split rule',
+    AskUserQuestions: 'request_user_input calls',
+    AskUserQuestion: 'request_user_input',
+  },
 
   suppressedResolvers: [
     'DESIGN_OUTSIDE_VOICES',  // design.ts:485 — Codex can't invoke itself
--- a/scripts/resolvers/preamble/generate-ask-user-format.ts
+++ b/scripts/resolvers/preamble/generate-ask-user-format.ts
@@ -1,6 +1,94 @@
 import type { TemplateContext } from '../types';
 
-export function generateAskUserFormat(_ctx: TemplateContext): string {
+function generateCodexRequestUserInputFormat(): string {
+  return `## request_user_input Format
+
+### Availability (read first)
+
+\`request_user_input\` is available in Codex Plan mode by default. Before each
+decision, confirm that it appears in the current tool list.
+
+If the tool is unavailable — including in Default mode — do NOT guess, silently
+choose, auto-select the recommendation, or continue the workflow. Tell the user:
+\`BLOCKED — this workflow needs interactive decisions. Switch this task to Plan
+mode and rerun it.\` Then STOP. Do not enable the under-development
+\`default_mode_request_user_input\` feature as a workaround.
+
+### Tool-call shape
+
+Send one decision per call and one item in the \`questions\` array. Use:
+
+- \`header\`: at most 12 characters.
+- \`id\`: stable snake_case.
+- \`question\`: the complete decision brief below, including context, stakes,
+  recommendation, completeness, pros/cons, and the closing Net line.
+- \`options\`: 2-3 explicit choices. Keep each label to 1-5 words, suffix the
+  recommended label with \`(Recommended)\`, and give each option one concise
+  sentence describing its concrete impact or tradeoff.
+
+The detailed brief belongs in \`question\`; do not try to pack it into option
+labels or omit it because the tool has structured options.
+
+### Decision-brief format
+
+\`\`\`
+D<N> — <one-line question title>
+Project/branch/task: <1 short grounding sentence using _BRANCH>
+ELI10: <plain English a 16-year-old could follow, 2-4 sentences, name the stakes>
+Stakes if we pick wrong: <one sentence on what breaks, what user sees, what's lost>
+Recommendation: <choice> because <one-line reason>
+Completeness: A=X/10, B=Y/10   (or: Note: options differ in kind, not coverage — no completeness score)
+Pros / cons:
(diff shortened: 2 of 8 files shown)


isaiahratahi:
@garrytan Both this focused fix and companion #2268 are ready for maintainer review. The current heads are rebased on main; the focused generation suites pass 529/529 here and 485/485 there. GitHub Actions are waiting for fork-workflow approval, and Trunk correctly rejected author submission because this account lacks write permission. When convenient, please approve the workflows and queue whichever PRs you accept. The changes remain intentionally separate.

isaiahratahi:
Withdrawing this PR because the required compatibility behavior is specific to our internal Ratahi-managed Codex environment. We are moving the implementation into our private operating repository and no longer request upstream review or merge. Thank you.

PR #2275, opened 2026-07-15
Add /autobuilder-loop and /plan-deliverables skills

sneakygriff:
## What this adds

Two skills that work as a unit, plus the supporting fixes they depend on.

### `/autobuilder-loop`

Plan-driven build loop. Resolves an approved plan, freezes the milestone set, and drives it to completion with model-routed subagents (the orchestrator never reads source, diffs, or logs — subagents return compact envelopes). Every milestone must pass a **mandatory gate** before it counts as done:

- `/plan-eng-review` + `/review`, each in a subagent that reads its own skill file and returns findings-only (fixes stay routed dispatches).
- An **adversarial cross-model forum** — Codex `gpt-5.6-sol` at `ultra` effort and Grok `grok-build`, read-only. A `[P1]` from **either** model fails the gate. If both are unavailable it degrades to a single Claude adversarial subagent tagged `[single-model]` rather than silently passing.
- Local Docker verification (a crashed app returning 500 is a FAIL, never a pass; `NO_SERVER` is a FAIL of "functional").

A bounded deferral valve allows batching **at most one** small, non-sensitive milestone's gate forward — never security/auth/migrations/public interfaces, and never the final milestone.

### `/plan-deliverables`

Authors **per-milestone acceptance criteria** — each measurable and paired with the check that validates it (the deliverable) — into the design doc.

This closes a real gap in the pipeline: `/autobuilder-loop` re-reads per-milestone `{next_milestone, acceptance_criteria}` from the plan on every iteration, but **nothing upstream authored that field**. `/office-hours` writes a single prose `## Success Criteria` blurb, `/plan-eng-review` audits proposed test coverage, and `/spec`'s Definition of Done is issue-level (epics only).

Runs after `/office-hours`, before the plan reviews. Per-piece loop:

> draft criteria → default-on cross-model gap-check (Codex + Grok, honoring the existing `codex_reviews` / `grok_reviews` kill-switches) → probe the user for judgment calls → pair each criterion with a **repo-native** validating check (or an explicit `manual-verification` when no automated check can honestly prove it) → write the block.

Idempotent and resume-safe: re-running reads existing progress and processes only what's missing.

### How the two connect

`<!-- status: pending | built-gate-pending | complete -->` milestone markers. `/plan-deliverables` writes `pending`; `/autobuilder-loop` step 1.0 selects the first `pending` milestone, and step e / the deferral valve write status back. That's an explicit handoff contract instead of inferring "next" from unchecked checkboxes at both ends.

## Hardening

`/plan-deliverables` was put through an adversarial audit (Codex `gpt-5.6-sol` ultra + Grok `grok-build` + independent verification agents) before this PR. Fixed:

- A gap-check block that **never wrote its prompt file** — both external models caught it independently; it would have sent an empty prompt to both CLIs on every milestone.
- `mktemp "…-XXXXXXXX.txt"` — the `.txt` after the X-template means **zero randomization on BSD/macOS** (collisions between concurrent runs).
- `BRANCH=$(git … | tr … || echo no-branch)` — without `pipefail` the pipeline exits 0 on `tr`, so the fallback never fires and `BRANCH` ends up empty.
- Missing idempotency/resume guards; section insert landing below the doc's closing section; prompt-injection fencing for doc-derived text; probe model availability once and run the two CLIs in parallel instead of sequentially.

Three findings were verified as false-positive / convention-consistent and deliberately **not** "fixed" (e.g. the markdown format is the shape autobuilder-loop's LLM parse subagent was co-designed for).

## Validation

- `bun run gen:skill-docs --host all` → clean; `--dry-run` freshness gate → **0 stale**.
- `bun run skill:check` → all hosts fresh; `gstack-plan-deliverables` OK across hosts.
- ⚠️ `skill:check` reports `❌ claude/SKILL.md — generated file missing`. This **pre-exists on `main`** (upstream `claude/` contains only `SKILL.md.tmpl`) and is unrelated to this branch.

## Notes for review

- Commit 1 also carries the `gen-skill-docs` and test-suite fixes that `/autobuilder-loop` depends on to generate and validate.
- Happy to split this into two PRs, trim scope, or take the skills separately — whatever's easiest to review.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,6 +20,7 @@ Invoke them by name (e.g., `/office-hours`).
 | `/plan-devex-review` | DX-mode review: TTHW, magical moments, friction points, persona traces. |
 | `/plan-tune` | Self-tune AskUserQuestion sensitivity per question. |
 | `/autoplan` | One command runs CEO → design → eng → DX review. |
+| `/autobuilder-loop` | Drive an approved plan to completion via model-routed subagents, review gates, and Docker verification. |
 | `/design-consultation` | Build a complete design system from scratch. |
 | `/spec` | Turn vague intent into a precise, executable spec in five phases. Files a GitHub issue, optionally spawns a Claude Code agent in a fresh worktree, and lets `/ship` close the source issue on merge. |
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,6 +119,7 @@ gstack/
 ├── review/          # PR review skill
 ├── plan-ceo-review/ # /plan-ceo-review skill
 ├── plan-eng-review/ # /plan-eng-review skill
+├── plan-deliverables/ # /plan-deliverables skill
 ├── autoplan/        # /autoplan skill (auto-review pipeline: CEO → design → eng)
 ├── benchmark/       # /benchmark skill (performance regression detection)
 ├── canary/          # /canary skill (post-deploy monitoring loop)
@@ -942,6 +943,7 @@ When the user's request matches an available skill, invoke it via the Skill tool
 
 Key routing rules:
 - Product ideas/brainstorming → invoke /office-hours
+- Acceptance criteria / deliverables / definition of done for a plan → invoke /plan-deliverables
 - Strategy/scope → invoke /plan-ceo-review
 - Architecture → invoke /plan-eng-review
 - Design system/plan review → invoke /design-consultation or /plan-design-review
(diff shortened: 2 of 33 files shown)


sneakygriff:
Closing in favour of a fresh PR: upstream main advanced 31 commits while this was in flight, leaving this branch conflicting. Superseded by a rebased branch with the conflict resolved and docs regenerated against current main.

Issue #2137, opened 2026-06-30
Add a standard AISP/AISOP contract for /autoplan phase gates, evidence labels, and task aggregation


optimization2026:
## Summary

`/autoplan` is one of the most important gstack skills because it coordinates multiple review phases and turns a rough plan into a reviewed implementation plan.

It already behaves like a structured workflow:

```text
input plan
  -> Phase 0 scope detection
  -> CEO review
  -> Design review if UI scope
  -> Eng review
  -> DX review if DX scope
  -> cross-phase synthesis
  -> task aggregation
  -> final approval gate
```

I propose adding a **standard AISP V1.0.0 package** as a machine-readable contract for `/autoplan`.

This would not replace the current `/autoplan` skill.

This would not require all gstack skills to use AISP.

The goal is to make `/autoplan`'s phase gates, evidence labels, task aggregation, and approval contract explicit enough to test and audit.

AISP protocol reference:

https://github.com/AIXP-Labs/AISP

---

## Problem

Several `/autoplan` issues point to the same underlying pattern: important workflow rules are currently encoded in prose, shell snippets, templates, and output conventions, but not in one explicit machine-checkable contract.

Examples:

```text
Phase selection:
  UI/DX phases can be skipped silently when Phase 0 scope detection misses terms.

Task aggregation:
  Phase 4 can return zero implementation tasks because an aggregation error is swallowed.

Consensus labeling:
  Single-voice mode can be labeled as CONFIRMED, even when Codex is unavailable.

Workspace safety:
  /autoplan can run for a long time and edit plan artifacts in-place without a strong workspace preflight.

Codex invocation:
  /autoplan may call Codex without a per-project auth/profile gate or explicit informed consent.
```

The problem is not that `/autoplan` lacks instructions.

The problem is that `/autoplan` has many non-negotiable workflow invariants that are hard to validate as one system:

```text
- phase decisions must be visible;
- skipped phases must have diagnostics;
- consensus labels must reflect actual reviewer evidence;
- task aggregation must not silently produce zero tasks;
- workspace mutations need safeguards;
- cross-model review must be transparent;
- final approval must show unresolved taste decisions and degraded evidence.
```

A standard AISP package could make these invariants explicit.

---

## Proposed solution

Add an optional AISP package for `/autoplan`.

Possible layout:

```text
autoplan/
├── SKILL.md.tmpl
├── SKILL.md
└── aisp/
    └── autoplan_review_contract_aisp/
[...]

Issue #1542, opened 2026-05-16
Document Codex refresh_token_reused recovery when Hermes auth still works


shandutta:
## Problem
When Codex CLI reports `refresh_token_reused`, the current recovery guidance can send users straight to `codex login` even if Hermes' `openai-codex` provider still has fresh working tokens. That creates a split-brain auth situation: Hermes works, standalone `codex exec` fails.

## Expected behavior
The Codex skill should distinguish a Hermes provider outage from a stale standalone Codex CLI credential store, and suggest repairing the standalone auth file only after a Hermes smoke test proves the provider route still works.

## Current downstream workaround
A local downstream skill patch adds guidance to verify Hermes `openai-codex`, repair the standalone CLI credential store from Hermes tokens via a helper script, chmod the auth file, then re-test `codex exec`.

## Notes
This is documentation/runbook guidance only; it does not change Codex or Hermes auth behavior.

0xDevNinja:
taking this one — adding a Hermes-aware sub-bullet under Error Handling > Auth so `refresh_token_reused` doesnt funnel users straight to `codex login` when the Hermes `openai-codex` provider is still serving valid tokens. doc-only patch, no auth behavior change.

PR #2278, opened 2026-07-15
fix: require per-project Codex auth approval

### LINKED ISSUE #965: `/codex` and `/autoplan` shell out to `codex` with no auth-profile gate or consent prompt
## Problem

The `/codex` skill and `/autoplan` (plan-review phases) shell out to the `codex` CLI with no check for **which account is currently authenticated** in `~/.codex/`. Whatever `codex login` was last run with is what gets used — silently, with no prompt.

Two things compound:

1. **`codex/SKILL.md.tmpl:40-47`** — the only pre-flight is `which codex`. No auth inspection, no per-project config, no `CODEX_HOME` awareness.
2. **`autoplan/SKILL.md.tmpl:70`** — "run codex" is classified as a **mechanical** decision ("one clearly right answer, auto-decide silently"). So during `/autoplan`, `codex exec` fires across multiple plan-review phases with zero user consent.

And the gap is wider than just `/codex` and `/autoplan`. Other skills chain into codex — `/ship` may suggest a codex review, `/plan-eng-review` may reach for it, and nothing stops a future skill from doing the same. "Don't run `/codex`" is not a mitigation a user can reasonably apply, because they'd have to know the full call graph of every gstack skill and intercept every chained prompt. The gate has to live at the invocation layer, not the skill-entry layer, so that every codex call is covered regardless of which skill triggered it.

The net effect: if a user has more than one codex account (common — personal vs. employer-provided, separate orgs, compliance-isolated environments), gstack will happily use whichever one happens to be in `~/.codex/` against any repo, with no way for the user to approve or even know ahead of time.

## Repro

1. `codex login` as account A.
2. `cd` into any repo — including ones where account A's credentials are the *wrong* choice.
3. Run `/autoplan` (or `/codex review`, or any skill that may chain into codex).
4. `codex exec` runs against account A. No prompt, no confirmation, no record of approval. The first the user knows is when they see the output (or a bill).

## Expected

Before the first codex invocation in a project, gstack should prompt the user for which codex auth to use, remember the answer per-project, and route subsequent invocations through `CODEX_HOME` accordingly. `/autoplan` should not treat codex as a silent auto-yes decision. Any skill that chains into codex should surface *which account* the call will use at the point of prompt, so "yes" is informed consent rather than reflex.

## Actual

- `codex/SKILL.md.tmpl:40-47` — binary-presence check only.
- `autoplan/SKILL.md.tmpl:70` — codex classified as mechanical auto-yes.
- No per-project config, no `CODEX_HOME` plumbing anywhere in the codex-calling skills.
- Chained-invocation prompts in other skills (e.g. `/ship`'s "get a codex review?") don't disclose which account will be used.

## Suggested fix (minimal)

The `codex` CLI already supports `CODEX_HOME` as an env var pointing at an alternate auth/config directory. This makes isolation trivial — users who maintain separate codex homes (e.g., `~/.codex-personal/`, `~/.codex-work/`) can have gstack pick the right one per project.

Concretely:

1. **Per-project config file:** `~/.gstack/projects/{slug}/codex.yaml` with one field — `codex_home: <path>`.
2. **Invocation-layer gate (not skill-entry layer):** wherever gstack shells out to `codex`, read the per-project config first. If missing, `AskUserQuestion` listing the `~/.codex*` directories found on the machine (glob), plus "Cancel." Save the chosen path. The prompt *is* the setup flow — no separate `/setup-codex` skill needed. Because the gate is at the invocation layer, every codex call is protected regardless of which skill triggered it.
3. **Before every `codex` invocation** (in `/codex`, `/autoplan` phases 1–4, and anywhere else), export `CODEX_HOME=<configured path>`.
4. **Transparency at the chained prompt:** any prompt that *may* trigger a codex call (e.g. `/ship`'s "get a codex review?", `/autoplan`'s CEO-review step if re-introduced as opt-in) should name the account that will be used — e.g. *"Run codex using the account approved for this project (`<path>`)?"* — so the user's "yes" is informed.
5. **In `autoplan/SKILL.md.tmpl:70`:** remove "run codex (always yes)" from the mechanical examples. Codex invocation should go through the same per-project gate as `/codex` itself, never auto-yes.
6. **Escape hatch for the eval harness:** `GSTACK_CODEX_SKIP_GATE=1` env var to bypass the gate when `test/skill-e2e-*.test.ts` drives codex directly. Test runner sets it; skill prose documents it.

Scope: ~20 lines of template changes, one config file, no new helper binary, no fingerprinting, no global flag. The same pattern likely applies to any other authenticated AI CLI gstack shells out to — worth a follow-up audit.

## Why this matters

Consent and credential isolation are separate issues, and gstack currently has neither for codex:
- **Consent:** a user should know when their codex account is about to be billed and see what prompt is being sent.
- **Isolation:** users with multiple codex accounts need a way to tell gstack which one is approved for which project.
- **Informed "yes":** even when a user does approve codex at a prompt, "yes" should mean yes to a *specific* account, not whichever one happens to be in `~/.codex/` at that moment.

All three are table-stakes for a tool that runs inside user repos and spends money on their behalf.

time-attack:
## What it does

Gstack now requires each project to approve a specific Codex profile before Codex can run.

## How it works

The shared launcher resolves the current project, reads its saved Codex home, validates it, exports it as `CODEX_HOME`, and blocks invocation when approval is missing. Skills prompt the user to select a discovered profile or cancel, then disclose the selected path.

## What was broken

Gstack could silently start Codex using whichever account happened to be authenticated globally, even when that account was inappropriate for the current project.

## How it was fixed

I added a hard invocation-layer profile gate, per-project configuration, profile disclosure, and an eval-only bypass. I routed chained review and design calls through the same launcher and removed Codex from autoplan's automatic mechanical decisions.

## Verification

The identical production-path probe failed on `origin/main` and passed after the fix in OneManCompany, gbrain, and zotero-arxiv-daily. Explicit bypass controls passed before and after in Plane and Twenty.

The focused regression and generated-skill suites passed: 434 tests, 0 failures.

Live evidence gate:

| Environment | Baseline | Candidate |
|---|---:|---:|
| OneManCompany | 1 (bug reproduced) | 0 |
| gbrain | 1 (bug reproduced) | 0 |
| zotero-arxiv-daily | 1 (bug reproduced) | 0 |
| Plane bypass control | 0 | 0 |
| Twenty bypass control | 0 | 0 |

## Issue coverage

Closes #965.

Related follow-ups that remain open because this PR does not satisfy their full scope:

- #349 — custom `CLAUDE_CONFIG_DIR` path support
- #1066 — interactive-question support in Codex Default mode
- #1542 — `refresh_token_reused` recovery guidance
- #1628 — ChatGPT-account model entitlement selection
- #2137 — the broader machine-readable `/autoplan` workflow contract


time-attack:
@16francej merge pls

PR #2279, opened 2026-07-16
fix: preserve generated skill links

creatornader:
## Summary

- Preserve generated skill paths when the frontmatter name differs from the source directory.
- Verify each generated index link points to an existing generated skill file.

## Tests

- bun test test/llms-txt-shape.test.ts


creatornader:
This PR is the generated-link repair split out from #2083 in response to review. It keeps the generator tied to each skill's generated output path and verifies every indexed link exists. #2083 will rebase onto main after this lands, so its command rename no longer carries this unrelated repair.

creatornader:
Trunk rejected my submission because this fork account does not have write access to the repository. This isolated link repair is ready for a maintainer to submit. Once it lands, I will immediately rebase #2083 onto main and remove the duplicated repair from that PR.

PR #2302, opened 2026-07-19
fix: scope benchmark routing to web performance

creatornader:
## Summary

- Narrow `/benchmark` routing to web page performance requests.
- Keep the current `/benchmark` command name while excluding model, product, business, eval, and research benchmarks.
- Add a regression test for the routing boundary and regenerate the skill index.

## Verification

- `bun run gen:skill-docs`
- `bun test test/gen-skill-docs.test.ts test/llms-txt-shape.test.ts` (411 passed)

### DIFF
--- a/benchmark/SKILL.md
+++ b/benchmark/SKILL.md
@@ -2,11 +2,11 @@
 name: benchmark
 preamble-tier: 1
 version: 1.0.0
-description: Performance regression detection using the browse daemon. (gstack)
+description: Web page performance regression detection using the browse daemon. (gstack)
 triggers:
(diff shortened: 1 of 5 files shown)


creatornader:
Can we please bump this fix? My agents are making this mistake almost a dozen times a day where they call the "benchmark" skill thinking it will help with model benchmarks etc only to find out it's completely irrelevant. A skill this narrow should not simply be treated as a generic "benchmark" skill, because it's not. I will have to uninstall this skill and/or gstack if not fixed soon, it's extremely disruptive.

Issue #2310, opened 2026-07-20
question-preference-hook emits invalid permissionDecision "defer" → aborts every AskUserQuestion call


johnmwhitman:
# Bug: `question-preference-hook` emits `permissionDecision: "defer"` → aborts every AskUserQuestion call

**Repo:** garrytan/gstack · **Version:** v1.60.1.0 · **Host:** Claude Code (current)

## Summary

The PreToolUse hook `hosts/claude/hooks/question-preference-hook.ts` returns
`permissionDecision: "defer"` on its no-op path (the `defer()` helper). The
current Claude Code harness does not honor `"defer"` as a permission decision —
it treats the hook output as invalid and **aborts the tool call**. The user sees
`Tool execution was interrupted` / `[Tool result missing due to internal error]`.

Because `defer()` is the overwhelmingly common return path (every question that
isn't a marker-tagged, never-ask, two-way door with an unambiguous recommended
option), **AskUserQuestion fails on essentially every call.** The `deny()` path
is unaffected because `"deny"` is a valid value.

## Reproduction

1. On Claude Code with gstack hooks enabled, trigger any `AskUserQuestion` call
   without a matching never-ask preference.
2. Hook returns `{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"defer"}}`.
3. Tool aborts every time.

Direct hook check:
```
echo '{"tool_name":"AskUserQuestion","tool_input":{"questions":[{"question":"Q?","options":[{"label":"A"},{"label":"B"}]}]}}' \
  | hosts/claude/hooks/question-preference-hook
# → {"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"defer"}}
```

## Root cause

Valid `permissionDecision` values are `"allow" | "deny" | "ask"`. `"defer"` is
not honored. The canonical "no opinion, proceed with the normal permission flow"
signal is to **omit `permissionDecision` entirely** (exit 0, empty stdout, or
`hookSpecificOutput` carrying only `additionalContext`).

## Fix

In the `defer()` helper, stop emitting `permissionDecision: "defer"`. Emit empty
stdout + exit 0 when there's no `additionalContext`; otherwise emit
`hookSpecificOutput` with `hookEventName` + `additionalContext` only:

```ts
function defer(additionalContext?: string): void {
  if (!additionalContext) {
    process.exit(0);
  }
  process.stdout.write(
    JSON.stringify({
      hookSpecificOutput: {
[...]

Issue #2292, opened 2026-07-17
AskUserQuestion PreToolUse hook's defer() hits claude-code#64389 race — fix: use allow instead


lotofiyan:
## Bug

`hosts/claude/hooks/question-preference-hook.ts`'s pass-through path (the common case: a plain `AskUserQuestion` call with no `<gstack-qid:...>` marker, or one that doesn't match a never-ask preference) resolves via:

```js
function defer(additionalContext?: string): void {
  const out = { hookEventName: 'PreToolUse', permissionDecision: 'defer', ... };
  process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
  process.exit(0);
}
```

This hits a confirmed Claude Code SDK race: [anthropics/claude-code#64389](https://github.com/anthropics/claude-code/issues/64389) — a `PreToolUse` hook resolving `permissionDecision: "defer"` can lose a race in the SDK and get fed a fabricated `is_error: true`, `"[Tool result missing due to internal error]"` tool_result instead of suspending correctly. A commenter on that thread confirms it reproduces against `AskUserQuestion` specifically, not just Bash.

Since this hook matches every `AskUserQuestion` / `mcp__*__AskUserQuestion` call, and the pass-through `defer()` branch is what fires for any plain question without gstack's internal marker (i.e. most real usage, any time a skill or the base agent asks a free-form question), this made `AskUserQuestion` unreliable-to-broken in the Claude Code VS Code extension host in my testing — deterministically failing across repeated calls in one session.

## Root cause is upstream, but the workaround is local

The claude-code#64389 thread has two purported fixes:
1. Delay the `defer()` write until after the assistant message stream closes (~200ms, reported as "deterministic across dozens of runs" by one reporter).
2. Use `allow`/`deny`/`ask` instead of `defer` — the thread states plainly these are not affected by the race.

I tried (1) first; it did **not** reproduce reliably in my environment — still failed after adding the delay. Switched to (2): since `AskUserQuestion` has no real "normal permission system" step for `defer` to fall through to (unlike a gated `Bash` command), swapping the pass-through case to `permissionDecision: "allow"` is behaviorally equivalent here, and it resolved the issue — confirmed working across repeated live calls afterward.

## Suggested fix

In `defer()` in `question-preference-hook.ts` (and check `auq-error-fallback-hook.ts` / any other PreToolUse hook using `defer` for AskUserQuestion for the same pattern), change:

```js
permissionDecision: 'defer',
```
to
```js
[...]

PR #2318, opened 2026-07-22
fix(hooks): question-preference-hook must not emit permissionDecision 'defer' (merges #2165 + #1924)

Akagilnc:
Picking up the handoff from #2165. @time-attack asked there on 2026-07-15:

> Please rewrite this version to include the accurate documentation and strict output coverage from PR 1924. Its implementation is otherwise the best current foundation.

Neither PR has moved since, so this branch does that merge. Credit belongs to @dpnascimento (#2165) and @jacksondc (#1924); every commit here carries both as co-authors. If either of them would rather carry it forward, I am happy to close this in favour of theirs.

## The bug

`hosts/claude/hooks/question-preference-hook.ts` emitted `permissionDecision: "defer"` on its no-opinion path. `defer` is a real Claude Code value meaning "pause this tool call and hand control back", and it is honored in print/non-interactive mode. So in any non-interactive session, every `AskUserQuestion` the hook did not auto-decide got parked and never executed. On my machine one such call was replayed several times across resumes, and no `tool_result` was ever written for it.

## What changed

- `defer()` becomes `passThrough()`: silent `exit 0`, or `hookSpecificOutput.additionalContext` alone when plan-tune has context to inject. No `permissionDecision` on that path.
- The `deny` paths (never-ask auto-decide, Conductor prose redirect) are untouched.
- Spike doc corrections, going a little beyond #1924: it restores the `deny > ask > allow` precedence fact that #1924 deleted outright, moves "No output" out of the `permissionDecision values:` list where it is not a value, and rewrites the Auto-decide example from the never-implemented `allow` + `updatedInput` to the `deny` + reason shape the hook actually emits.
- Tests: `expectNoDecision` (from #1924) now covers about ten pass-through cases, plus two cases #1924 lacked. One is `mcp__conductor__AskUserQuestion` with no preference, which is the exact shape the bug reports fire on. The other asserts `hookEventName === 'PreToolUse'` on the `additionalContext`-only output.

## Scope

Fixes the root cause behind #2006, #2035, #2292 and #2310.

#2207 is not fixed here. That one is the Conductor auto-install of the hook, owned by #2208. An early commit body on this branch listed #2207 among the root-cause issues; that wording was too broad, and this PR body is authoritative on claim surface.

## Verification

```
bun test test/question-preference-hook.test.ts test/memory-cache-injection.test.ts
  28 pass / 0 fail / 67 expect()      (upstream/main: 27 pass / 52 expect)

bun run test
  2 failures: spec-template-sync, session-runner observability
  Both also fail on untouched upstream/main. Zero new failures.

bun run slop:diff
  no new findings in 5 changed files
```

## Two things I could not verify

The cathedral e2e assertion never executed locally. `test/skill-e2e-plan-tune-cathedral.test.ts` is `EVALS=1`-gated. Forcing it with `EVALS_ALL=1 EVALS=1` gives 0 pass / 5 fail, because the fixture scaffold does not copy `lib/`, so the hook dies with `Cannot find module '../../../lib/is-conductor'` at `expect(res.status).toBe(0)`, before the pass-through assertion runs. That fixture gap predates this branch and is not fixed here. Gate coverage for this behaviour is the two unit test files above.

The Conductor deny reason carries a premise I could not confirm. In Conductor the hook still denies with a prose redirect whose reason cites the MCP variant being flaky. That premise was tied to the same `defer` path this PR removes, and I have not re-proven it under current Claude Code. The behaviour is deliberately unchanged, since it is out of scope, but that wording is now a historical claim this PR does not confirm.

I can rebase or reshape this however suits the review.

### DIFF
--- a/docs/spikes/claude-code-hook-mutation.md
+++ b/docs/spikes/claude-code-hook-mutation.md
@@ -51,7 +51,8 @@ Optional in subagent context: `agent_id`, `agent_type`.
 - `"deny"` — block (feedback to Claude, NOT a synthetic answer per Codex
   correction in D-prefixed decisions)
 - `"ask"` — escalate to user
-- `"defer"` — let permission flow continue
+- No output — let permission flow continue
+- `"defer"` — pause a non-interactive SDK/tool caller so it can resume later
 
 **`updatedInput` semantics:** shallow merge of fields present in the returned
 object onto the original `tool_input`. Only valid with
@@ -88,7 +89,7 @@ required for our hook to fire there.
   accepting.
 
 **`permissionDecision` precedence (when multiple hooks decide):**
-`deny > ask > allow > defer` — most restrictive wins.
+Claude Code treats `defer` as an explicit pause/resume decision, not as the normal pass-through path. Hooks that do not need to decide should exit 0 with no output.
 
 ## Implementation hookSpecificOutput examples
 
@@ -107,11 +108,12 @@ required for our hook to fire there.
 ```
 
 **Pass-through (no preference, or one-way safety override):**
+Exit 0 with no stdout. If the hook has context to add, omit `permissionDecision`:
 ```json
 {
   "hookSpecificOutput": {
     "hookEventName": "PreToolUse",
-    "permissionDecision": "defer"
+    "additionalContext": "optional context for Claude"
   }
 }
 ```
--- a/docs/spikes/claude-code-hook-mutation.md
+++ b/docs/spikes/claude-code-hook-mutation.md
@@ -13,8 +13,9 @@ answer via `updatedInput`? If yes, what's the exact protocol?
 
 ## Answer
 
-**Yes.** `updatedInput` is the supported mechanism. Source:
-https://code.claude.com/docs/en/hooks (confirmed 2026-04 reference).
+**Yes.** `updatedInput` is the supported mechanism (platform capability;
+this plan-tune hook does not take that path — see Implementation examples).
+Source: https://code.claude.com/docs/en/hooks (confirmed 2026-04 reference).
 
 ## Hook stdin schema (PreToolUse + PostToolUse)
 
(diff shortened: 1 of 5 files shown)


Akagilnc:
Twelve days open, no review yet. Not trying to jump the queue — just putting the confirmations that have come in since in one place, since they all land on the same change.

**Independent confirmations since 2026-07-22**

- #2320 (manhit96, Jul 22) — same bug; closed as a duplicate after applying the same change as this PR (silent `exit 0` pass-through, `additionalContext` only). Verified end-to-end on Claude Code 2.1.217 / VSCode extension.
- #2310 (devty, Jul 26) — corrects the mechanism several of the reports state. `defer` is not rejected as invalid; Claude Code **honors** it and parks the tool. The hook exits 0 and is logged as a success, `hook_deferred_tool` follows, and no `tool_result` is ever produced — which is also why `auq-error-fallback-hook` never helps: it's a `PostToolUse` hook, and the tool never completes.
- #2310 (topcoder1, Jul 31) — blast radius: Claude Code desktop 2.1.217–2.1.219, gstack v1.60.1.0, macOS — **57/57 AskUserQuestion calls wedged over 8 days**, none ever answered. Same fix, re-applied at SessionStart because upgrades rematerialize the file.

**Current state**

The same bug now has five open reports — #2006, #2035, #2292, #2310, #2340 — plus #2320 closed as a duplicate. Three open PRs: #1924, #2165, and this one. `main` is unchanged since a3259400 (Jul 15); `hosts/claude/hooks/question-preference-hook.ts:98` still emits `permissionDecision: 'defer'`.

Impact: the no-preference path runs on every unmarked question, so anyone with no plan-tune preferences loses `AskUserQuestion` entirely — silently, while the hook still reports success.

**What this PR is**

#1924 (@jacksondc) and #2165 (@dpnascimento) carried forward on current `main`, both authors credited as co-authors on the commits. The behavior change is small: pass-through becomes a silent `exit 0`, and `permissionDecision` is reserved for the documented `allow` / `deny` / `ask` values. Deny and auto-answer paths are unchanged. The rest of the diff is regression tests and a doc update.

Happy to rebase, split, or reshape it however you want. Leaving this here so the later confirmations are easy to find — and so I'm not just watching more duplicates arrive for something a few of us have already checked.

garrytan:

PR #2321, opened 2026-07-22
fix(preamble): probe gbrain engine liveness before recommending it

syong:
## Problem

The gbrain block in the skill preamble decides whether to tell the agent "prefer `gbrain search` over Grep". Today it checks three things:

1. `~/.gbrain/config.json` exists
2. `gbrain --version` responds
3. the worktree has a `.gbrain-source` pin

All three can pass while the engine is completely unreachable — **none of them opens a connection**. They prove gbrain is *installed*, not that it *works*.

So when the engine goes down (disk full, container stopped, pooler unreachable, autopilot wedged), every skill preamble keeps printing:

> GBrain configured. Prefer `gbrain search`/`gbrain query` over Grep for semantic questions…

The agent then reaches for a tool whose every query fails, instead of falling back to Grep. Nothing errors, nothing is surfaced — the guidance actively asserts health it never verified, so the outage is silent and can persist indefinitely.

I hit exactly this: the local Postgres engine was crash-looping on a full Docker disk for well over a day. The preamble recommended gbrain the entire time.

## Fix

Add a bounded liveness probe and branch on the result:

```bash
_GBRAIN_ALIVE=0
if command -v timeout >/dev/null 2>&1; then
  timeout 5 gbrain sources list >/dev/null 2>&1 && _GBRAIN_ALIVE=1
else
  gbrain sources list >/dev/null 2>&1 && _GBRAIN_ALIVE=1
fi
```

When it fails, the preamble now says so and redirects to Grep:

```
GBrain DOWN — this worktree is pinned, but a probe query failed, so
the engine is unreachable. Do NOT rely on `gbrain search`/`code-def`
this session; use Grep instead. Diagnose with `gbrain doctor --fast`.
```

Design notes:

- **Bounded.** `timeout 5` means a hung engine can't stall the preamble of every skill. Falls back to an unbounded call only where `timeout` isn't available (rare on macOS/Linux; the alternative is skipping the check entirely).
- **Engine-agnostic.** Points at `gbrain doctor --fast` rather than assuming any particular backend (PGLite / Postgres / Supabase).
- **Zero cost when gbrain isn't in play.** The probe runs only inside the existing "pin is present" branch, so non-gbrain users and unpinned worktrees are untouched.
- **Cheap.** One `sources list` per skill invocation.

## Changes

- `scripts/resolvers/preamble/generate-brain-sync-block.ts` — the actual change
- 49 generated `SKILL.md` files — regenerated via `bun run gen:skill-docs`

Insertions only; no existing lines modified.

## Verification

Tested against a genuinely stopped engine, from a pinned worktree:

| engine state | preamble output |
|---|---|
| up | normal "prefer `gbrain search`" guidance |
| **stopped** | **`GBrain DOWN` + use Grep + diagnose hint`** |
| restarted | normal guidance again |

Also confirmed the unpinned-worktree branch still prints its existing "isn't pinned yet" message and never runs the probe.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -356,10 +356,29 @@ if [ -f "$_GBRAIN_CONFIG" ] && command -v gbrain >/dev/null 2>&1; then
       _GBRAIN_PIN_PATH="$_REPO_TOP/.gbrain-source"
     fi
     if [ -n "$_GBRAIN_PIN_PATH" ]; then
+      # Liveness probe. A config file + `--version` + a pin only prove gbrain is
+      # INSTALLED; none of them prove its engine is reachable. When the engine is
+      # down (disk full, container stopped, pooler unreachable, autopilot wedged)
+      # this block would still tell the agent to "prefer gbrain search" — so the
+      # agent keeps reaching for a tool whose every query fails, instead of
+      # falling back to Grep. One bounded probe closes that gap; the timeout
+      # keeps a hung engine from stalling the preamble of every skill.
+      _GBRAIN_ALIVE=0
+      if command -v timeout >/dev/null 2>&1; then
+        timeout 5 gbrain sources list >/dev/null 2>&1 && _GBRAIN_ALIVE=1
+      else
+        gbrain sources list >/dev/null 2>&1 && _GBRAIN_ALIVE=1
+      fi
+      if [ "$_GBRAIN_ALIVE" -eq 1 ]; then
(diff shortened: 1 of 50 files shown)


syong:
Closing this — it diverged from main over the past few weeks and now conflicts with the MCP scope-resolution refactor (#2499) that reworked the same generator block. The liveness-probe idea still stands (upstream doesn't have it), but I'm keeping the change on my local install rather than maintaining a long-lived rebase. Happy to reopen a fresh PR against current main if there's interest.

PR #2344, opened 2026-07-25
fix: keep ship review retries inside one invocation

yolo-jared:
## Summary

- treat agent-applied pre-landing fixes as an internal ship retry loop instead of requiring a second user invocation
- generate and install the Codex payload with the GPT overlay, including global `~/.agents` discovery
- migrate source checkouts out of Codex skill-discovery roots to prevent recursive or Claude-source skill loading
- add regression coverage for same-invocation continuation, GPT generation, global installs, and discovery migration

## Verification

- focused host and generator suites: 484 pass, 0 fail
- new release-autonomy regressions: 5 pass, 0 fail
- official non-eval suite under isolated global Git config: 4,440 pass, 2 unrelated failures, 488 skip
  - benchmark auth-remediation wording
  - session-runner observability comment count
  - both failing test/source surfaces are unchanged from `origin/main`

The non-eval command excludes the repository's LLM and periodic E2E groups through its documented ignore list.

### DIFF
--- a/setup
+++ b/setup
@@ -18,8 +18,8 @@ INSTALL_GSTACK_DIR="$(cd "$(dirname "$0")" && pwd)"
 SOURCE_GSTACK_DIR="$(cd "$(dirname "$0")" && pwd -P)"
 INSTALL_SKILLS_DIR="$(dirname "$INSTALL_GSTACK_DIR")"
 BROWSE_BIN="$SOURCE_GSTACK_DIR/browse/dist/browse"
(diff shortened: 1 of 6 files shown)


yolo-jared:
Closing this — it has drifted into conflict against main since July and I'd rather not leave a stale PR in the queue. The underlying issue (ship review retries escaping their invocation) may or may not still be live on current main; if I confirm it still reproduces I'll open a fresh PR against an up-to-date base rather than rebasing six months of drift. Thanks, and sorry for the noise.

PR #2323, opened 2026-07-22
feat: add GitHub Copilot CLI and app support

andrey-esipov:
> **TL;DR**
>
> **What:** One declarative `copilot` host so gstack skills install and run on the GitHub Copilot CLI and the Copilot app. Closes #393. Fresh implementation against the current host architecture, not a rebase of the stale, conflicting #396.
>
> **Why the eval checks are red:** GitHub withholds repository secrets from a `pull_request` opened from a fork, so `ANTHROPIC_API_KEY` is empty and every API-calling shard dies at SDK auth before a model runs. All 57 retry attempts spent **0 tokens / $0.00**. Rerunning the same tests on clean `main` with credentials scrubbed reproduces the failures byte-for-byte, as does an unrelated fork PR from another author. No code defect involved.
>
> **Everything that does not need a secret passes:** `actionlint`, `check-freshness`, `build-image`, `windows-free-tests`, `windows-setup`, plus the new unit tests in `test/copilot-host.test.ts` and `test/host-config.test.ts`.
>
> **To get real eval signal** (maintainer, ~1 minute). Run the suite from a short-lived branch in this repo, where secrets exist:
>
> ```bash
> git fetch origin pull/2323/head
> git push origin FETCH_HEAD:refs/heads/ci-verify/copilot-host
> gh workflow run evals.yml --ref ci-verify/copilot-host
> # once green:
> git push origin --delete ci-verify/copilot-host
> ```
>
> `evals.yml` already declares `workflow_dispatch`, and test selection comes from `git diff origin/main...HEAD` (`test/helpers/touchfiles.ts:841`) rather than the event payload, so that run selects the identical 42/181 tests this PR does. The branch is only a vehicle for the run. Delete it after and merge this PR.
>
> Per-shard breakdown: [root-cause comment](https://github.com/garrytan/gstack/pull/2323#issuecomment-5051936109).

---

## Problem

gstack has no official GitHub Copilot host, so generated skills, installation paths, setup detection, and runtime assets do not work as one supported flow in either the standalone GitHub Copilot CLI or GitHub Copilot desktop app project sessions.

GitHub documents one Agent Skills contract for both surfaces. This change adds one declarative `copilot` host instead of duplicating path roots or maintaining separate CLI and app variants.

## Verified Copilot contract

| Contract | Verified behavior | Source |
|---|---|---|
[...]

PR #2345, opened 2026-07-25
ci: skip eval jobs deterministically on fork PRs

andrey-esipov:
## Problem

Fork PRs cannot pass this workflow, and today whether they *appear* to pass is decided by Docker layer cache state rather than by the code under review.

GitHub does not pass repository secrets to a `pull_request` triggered from a fork, and issues a read-only `GITHUB_TOKEN` ([fork workflow security](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflows-in-forked-repositories)). Three consequences follow, none of which any contributor can fix from a branch:

- **`evals`** — `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, and `GEMINI_API_KEY` are all empty. Agent shards fail with `error_api`; `llm-judge` throws `Could not resolve authentication method` at `new Anthropic()` before a request is made; `e2e-pty-plan-smoke` shows `Not logged in` and then times out at 300s. No model executes, so these burn **0 tokens and $0.00** across every retry.
- **`report`** — cannot post its comment: `Resource not accessible by integration (addComment)`.
- **`build-image`** — cannot push to ghcr.io with `packages: read`.

## The nondeterminism

`evals` declares `needs: build-image`, and the image tag is content-hashed over `Dockerfile.ci`, `package.json`, and `bun.lock`. So:

| Cache state on a fork PR | `build-image` | `evals` | PR appears |
|---|---|---|---|
| Image already cached | passes (no push needed) | **runs, fails 7 shards + report** | ❌ red |
| Image not cached | fails (push denied) | **skipped** | mostly green |

Identical code, opposite verdicts. #2264 merged in the second state with `build-image` red and `evals` showing `skipping`; #2323 sits in the first state with 8 red checks. Neither outcome reflects the branch.

This also means **no fork PR can reach green while these jobs run**. `e2e-pty-plan-smoke` does not use diff-based test selection (`shouldRun = EVALS && EVALS_TIER === 'gate'`), so it executes on every PR regardless of the diff, and it needs an authenticated `claude`. Confirmed across unrelated fork PRs from different authors, which fail with byte-identical errors.

## Change

Gate `build-image`, `evals`, and `report` on the PR head living in the base repository:

```yaml
if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```

Fork PRs now skip these jobs deterministically — the same outcome the cold-cache path already produces by accident, but for a stated reason and every time. **Same-repo PRs, pushes, and `workflow_dispatch` are untouched and keep full eval coverage.** Maintainers get real coverage on fork contributions the way they do today: push the commits to a base-repo branch (`git fetch origin pull/N/head` then push that ref) and let it run with secrets.

## What this does not do

- **Does not weaken coverage where coverage was real.** These jobs currently produce no signal about fork branch code — they fail before any model or push is attempted.
- **Does not add skip-on-missing-key to the test harness.** That would silently weaken same-repo runs, and it contradicts existing policy (`test/skill-e2e-first-task-scaffold.test.ts` throws `requires ANTHROPIC_API_KEY ... refusing to skip`).
- **Does not use `pull_request_target`.** Running untrusted fork code with secrets is unsafe.
- **Does not fix `build-image` push behavior generally.** Making the build run with `push: false` on forks so `Dockerfile.ci` still gets compile-checked is a reasonable follow-up, deliberately out of scope here.

## Verification

- `actionlint` (v1.7.11, matching `.github/workflows/actionlint.yml`) reports the same 2 pre-existing `runner-label` warnings before and after; **no new findings**.
- Workflow parses cleanly; job graph and `needs` edges unchanged.
- Diff is 3 `if:` conditions plus a comment. No step, matrix, secret, or permission changes.

## Note for maintainers

If Trunk requires the `evals (...)` or `report` contexts for merge, please confirm skipped jobs satisfy them as expected. I could not inspect the Trunk configuration from outside the org. Happy to adjust the approach.

Found while investigating #2323; filed separately so that PR stays free of CI edits.

### DIFF
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -14,8 +14,16 @@ env:
 
 jobs:
   # Build Docker image with pre-baked toolchain (cached — only rebuilds on Dockerfile/lockfile change)
+  # Fork PRs never receive repository secrets (ANTHROPIC_API_KEY et al) and get a
+  # read-only GITHUB_TOKEN, so nothing below can succeed from a fork: the eval
+  # shards fail at SDK auth, `report` cannot post its comment, and this job cannot
+  # push to ghcr.io. Skip deterministically instead of leaving the outcome to
+  # Docker-cache luck (a warm cache made these run and fail; a cold one made them
+  # fail to push and skip). Same-repo PRs, pushes, and workflow_dispatch are
+  # unaffected. Maintainers get real coverage via a trusted base-repo branch.
   build-image:
     runs-on: ubicloud-standard-8
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     permissions:
       contents: read
       packages: write
@@ -58,6 +66,7 @@ jobs:
   evals:
     runs-on: ${{ matrix.suite.runner || 'ubicloud-standard-8' }}
     needs: build-image
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     container:
       image: ${{ needs.build-image.outputs.image-tag }}
       credentials:
@@ -263,7 +272,7 @@ jobs:
   report:
     runs-on: ubicloud-standard-8
     needs: evals
-    if: always() && github.event_name == 'pull_request'
+    if: always() && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
     timeout-minutes: 5
     permissions:
       contents: read
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -47,8 +47,8 @@ jobs:
 
       # A fork PR's GITHUB_TOKEN only has `packages: read`, so pushing fails.
       # Still BUILD (validates Dockerfile.ci changes), just don't publish. This
-      # job intentionally has no `if:` — it must always start, because a run in
-      # which every job is skipped registers as a startup failure.
+      # job intentionally keeps no `if:` so fork PRs still get one real, honest
+      # green check here instead of a run where every job is grey.
       - if: steps.check.outputs.exists == 'false'
         uses: docker/build-push-action@v6
         with:
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -14,16 +14,8 @@ env:
 
 jobs:
   # Build Docker image with pre-baked toolchain (cached — only rebuilds on Dockerfile/lockfile change)
-  # Fork PRs never receive repository secrets (ANTHROPIC_API_KEY et al) and get a
-  # read-only GITHUB_TOKEN, so nothing below can succeed from a fork: the eval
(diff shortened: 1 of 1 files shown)


andrey-esipov:
### Heads up: this PR's own checks cannot run, and that is expected

`E2E Evals` and `Windows Free Tests` on this branch both completed with **0 jobs** and `startup_failure` ("This run likely failed because of a workflow file issue"). I chased that for a while assuming I'd broken the YAML. I hadn't.

The tell is `Windows Free Tests`. That workflow is not touched by this PR, and it passes on my other branch. It fails here anyway. So the failure is not about the content of any one file.

The actual mechanism: **GitHub refuses to start workflow runs for a pull request from a fork when that pull request modifies anything under `.github/workflows/`.** The fork's `GITHUB_TOKEN` has no `workflows` permission, so the run set cannot be constructed and every workflow in it fails at startup. It is a security control, not a bug, and it applies to the whole run set rather than just the edited file.

Corroborating evidence from this repo: I checked the last 60 merged PRs. Every merged PR that touched `.github/workflows/` came from a branch in `garrytan/gstack` itself. **Zero fork PRs have ever modified a workflow file here.** There is no counterexample, because there cannot be one.

So this PR is in the somewhat funny position of being blocked by a stricter cousin of the exact problem it is trying to fix.

### What I did verify locally

- `actionlint` v1.7.11 against the modified file: same 2 pre-existing `runner-label` warnings for `ubicloud-standard-8` as baseline `main`, no new findings.
- YAML parses cleanly (`js-yaml`).
- Job graph and every `needs:` edge unchanged from `main`.

### What a maintainer needs to do

Because of the restriction above, this change can only be exercised from a branch in the base repo:

```
git fetch https://github.com/andrey-esipov/astack.git ci/fork-pr-eval-skip
git push origin FETCH_HEAD:refs/heads/ci/fork-pr-eval-skip
```

Then open a PR from that branch, or just cherry-pick the single commit. It is one file and roughly ten lines.

### What the change does

On a fork PR:
- `build-image` still runs and still builds `Dockerfile.ci`, so Dockerfile changes stay validated. Only the `push:` to ghcr is turned off, since a fork token only has `packages: read`.
- `evals` and `report` skip.

On same-repo PRs, pushes, and `workflow_dispatch`, nothing changes. Full coverage.

### Why it is worth doing

Right now the outcome for a fork PR is decided by Docker layer cache state, not by the code:

- **Warm cache** → `build-image` finds the tag, skips the push, passes → `evals` runs → every API-calling shard fails at SDK auth because fork PRs get no secrets → 8 red checks.
- **Cold cache** → `build-image` tries to push, is denied, fails → `evals` is skipped → the PR looks clean.

Same commit, opposite verdict, depending on whether someone else happened to warm the cache first. #2264 merged with `build-image` failing and `evals: skipping`. #2323 got 8 red checks for the same underlying reason. Neither result says anything about the code.

This just makes the skip explicit instead of accidental.

PR #2349, opened 2026-07-26
fix(plan-tune): stop deferring AskUserQuestion in non-interactive sessions

mitch-schneider-1:
## Summary

The `AskUserQuestion` `PreToolUse` hook emits `permissionDecision: "defer"` on its pass-through path — no preference set, no `<gstack-qid:…>` marker, i.e. the common case. This makes `AskUserQuestion` **silently fail in every non-interactive host** while working perfectly in the terminal.

## Root cause

`defer` is an undocumented, **print-mode-only** value. The documented set is `allow | deny | ask`. From the Claude Code 2.1.216 binary:

```
- `permissionDecision` - "allow", "deny", or "ask" (PreToolUse only)
Hook … returned permissionDecision=defer in interactive mode; ignoring (defer is print-mode only)
Unknown hook permissionDecision type: … Valid types are: allow, deny, ask, defer
```

So:

| Host | Behavior |
|---|---|
| Interactive terminal | Value parsed, warning logged, **ignored** → tool works |
| Non-interactive (Cowork, Agent SDK, `claude -p`) | Value **honored** → tool_use left unresolved, no result |

In the non-interactive case the agent receives `[Tool result missing due to internal error]` and the question never reaches the user.

The failure mode is nasty precisely because it's mode-split: 100% broken in one host, 100% fine in another. That reads as flaky transport, so it gets misattributed to the platform rather than the hook. It also defeats the existing `auq-error-fallback-hook` mitigation, since a deferred tool never produces the error response that hook keys on.

## Fix

Emit **no** `permissionDecision` on pass-through. Absence is the correct "no opinion" signal and the only form that is safe in both modes.

Unchanged:
- `additionalContext` (plan-tune memory injection) still emitted when present.
- The `deny` paths — `never-ask` auto-decide and the Conductor prose redirect — are untouched.

Also corrects `docs/spikes/claude-code-hook-mutation.md`, which documented `defer` as *the* pass-through shape. That's where the bug originated, so leaving it would re-seed the regression.

## Testing

- **89 pass / 0 fail** across the 7 `AskUserQuestion` hook test files.
- Updated 11 assertions across 3 test files that pinned the old `defer` contract; they now assert absence.
- Verified live: `AskUserQuestion` confirmed working in a Cowork session where it had consistently returned no result.
- The 5 `skill-e2e-plan-tune-cathedral` eval scenarios fail identically on `main` and on this branch (pre-existing, unrelated — per the repo's E2E blame protocol).

Health check — empty stdout is correct:

```bash
echo '{"hook_event_name":"PreToolUse","tool_name":"AskUserQuestion","tool_input":{"questions":[]}}' \
  | hosts/claude/hooks/question-preference-hook
```

### DIFF
--- a/docs/spikes/claude-code-hook-mutation.md
+++ b/docs/spikes/claude-code-hook-mutation.md
@@ -107,14 +107,25 @@ required for our hook to fire there.
 ```
 
 **Pass-through (no preference, or one-way safety override):**
+
+Emit **no** `permissionDecision` — absence is "no opinion". Either write nothing to
+stdout, or emit only `additionalContext`:
 ```json
 {
   "hookSpecificOutput": {
     "hookEventName": "PreToolUse",
-    "permissionDecision": "defer"
+    "additionalContext": "…optional…"
   }
 }
 ```
+
+> ⚠️ **Do not use `permissionDecision: "defer"` here.** (This spike originally
+> recommended it; that was wrong and shipped a real bug — see the AskUserQuestion
+> note in `~/.claude/CLAUDE.md`.) The documented set is `allow | deny | ask`.
+> `defer` is an undocumented print-mode-only value: interactive Claude Code logs
+> "defer is print-mode only" and ignores it, but non-interactive sessions (Cowork,
+> Agent SDK, `claude -p`) honor it and leave the tool_use **unresolved with no
+> result**, which the agent sees as `[Tool result missing due to internal error]`.
 
 **PostToolUse capture (always):**
 ```json
--- a/hosts/claude/hooks/question-preference-hook.ts
+++ b/hosts/claude/hooks/question-preference-hook.ts
@@ -92,13 +92,29 @@ function readStdin(): Promise<string> {
   });
 }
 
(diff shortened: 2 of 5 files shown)


mitch-schneider-1:
Closing as a duplicate — #2325 already covers this fix and touches the same files. I missed it before opening; apologies for the noise.

Leaving one note here in case it's useful to whoever picks #2325 up: the reason this reads as flaky transport rather than a hook bug is that `defer` is honored in print mode but ignored in interactive mode, so the same hook is 100% broken in Cowork / Agent SDK / `claude -p` and 100% fine in a terminal. From the Claude Code 2.1.216 binary:

```
- `permissionDecision` - "allow", "deny", or "ask" (PreToolUse only)
Hook … returned permissionDecision=defer in interactive mode; ignoring (defer is print-mode only)
```

Worth also fixing `docs/spikes/claude-code-hook-mutation.md`, which documents `defer` as the pass-through shape — that's where the bug originated, so it can re-seed.

PR #2371, opened 2026-07-28
test(memory-ingest): cover the --include-unattributed skip counter (#2353)

### LINKED ISSUE #2353: --include-unattributed is inert: the skip guard compares "" against "_unattributed", so unattributable sessions always import
### Environment
gstack 1.60.1.0 · gbrain 0.42.66.0 (Postgres) · Linux

### What happens
`buildTranscriptPage` stores the **raw** git remote on the page but writes the **fallback** into the frontmatter, and the later filter tests the fallback string:

```ts
// bin/gstack-memory-ingest.ts:704
const remote = resolveGitRemote(session.cwd);   // returns "" on failure (:681)
// :726 — frontmatter gets the fallback
`git_remote: ${remote || "_unattributed"}`,
// :747 — the page field gets the RAW value
git_remote: remote,
```
```ts
// bin/gstack-memory-ingest.ts:1170
if (!args.includeUnattributed && page.git_remote === "_unattributed") {
```

`page.git_remote` is `""` whenever the remote could not be resolved, so `"" === "_unattributed"` is always false and the branch is dead. `repoSlug` (`:687`) *does* apply the fallback, so the slug and frontmatter both read `_unattributed` — only the guard's input doesn't.

`buildArtifactPage` gets this right: `:784` stores `git_remote: slug_repo`, the already-defaulted value.

The earlier guard at `:1165` (`!session.cwd`) doesn't cover this — these sessions **do** have a `cwd`, it just no longer resolves.

### What I expected
With `--include-unattributed` off (the default), a session whose repo cannot be determined is skipped and counted in `skipped_unattributed`.

### Minimal repro
```bash
mkdir -p /tmp/not-a-repo && cd /tmp/not-a-repo
# run any agent session here, then:
gstack-memory-ingest.ts            # no --include-unattributed
# → page transcripts/<agent>/_unattributed/<date>-<id> is imported
# → summary reports "skipped (unattrib): 0"
```

### Why it matters
On one brain, **354 of 650** transcript pages carry `git_remote: _unattributed` — 54% of the corpus imported against the default policy.

The compounding factor is that the remote is resolved at **ingest** time from a `cwd` captured at **session** time. Of those 354 pages, **333 (94%)** point at a directory that no longer exists, and 84 are under `worktree` paths. Agent sessions that run in ephemeral git worktrees are therefore *systematically* unattributable: `git -C <deleted dir> remote get-url origin` throws at `:673` and returns `""` every time.

The user-facing counter also reads `0` (`:1887`), so nothing signals that the policy isn't being applied.

### Suggested fix
Normalise once, at the source — one line at `:747`:

```ts
git_remote: remote || "_unattributed",
```

Optionally, recover the attribution instead of discarding it: Claude Code encodes the project directory in the transcript path (`~/.claude/projects/<encoded-cwd>/…`), which survives worktree deletion and would correctly attribute most of these.

noron12234:
Fixes #2353.

## Root cause

`buildTranscriptPage` applies the `_unattributed` fallback in two of the three
places it computes it, and the one it skips is the one the guard reads:

```ts
const remote = resolveGitRemote(session.cwd);   // "" when the remote can't be resolved
const slug_repo = repoSlug(remote);             // → "_unattributed"
…
`git_remote: ${remote || "_unattributed"}`,     // :729 frontmatter — defaulted
…
git_remote: remote,                             // :750 page record — RAW
```

```ts
if (!args.includeUnattributed && page.git_remote === "_unattributed") {   // :1173
```

`page.git_remote` is `""`, so the comparison is never true, the branch is dead, and
`--include-unattributed` has no effect in either direction. The earlier `!session.cwd`
guard doesn't cover it — these sessions have a `cwd`, it just no longer resolves to a
remote (deleted directories, ephemeral worktrees).

`buildArtifactPage` already stores the defaulted value (`git_remote: slug_repo`, :787),
so this is the transcript builder catching up.

## Fix

One line, plus a comment noting the guard depends on it:

```ts
git_remote: remote || "_unattributed",
```

Taking the fix at the source rather than loosening the guard to `!page.git_remote ||
… === "_unattributed"` keeps the field consistent with the frontmatter and the slug,
which is what `buildArtifactPage` already does.

## Verification

Reproduced with a real session JSONL whose `cwd` is a directory that isn't a git repo,
run through `--bulk --no-write`:

| | written | skipped (unattrib) |
|---|---|---|
| before | **1** | **0** |
| after | 0 | 1 |
| after, `--include-unattributed` | 1 | 0 |
| after, cwd *is* a repo with a remote | 1 | 0 |

Three regression tests added to `test/gstack-memory-ingest.test.ts`, in the existing
`runScript` / `writeClaudeCodeSession` style. Confirmed they actually bite — reverting
just the one-line change makes the first one fail:

```
(fail) unattributed skip > skips a session whose cwd is not a git repo
Expected to contain: "written:               0"
Received: "… written:               1 … skipped (unattrib):    0 …"
```

Suites run clean with the fix in place:

```
bun test test/gstack-memory-ingest.test.ts                  26 pass  0 fail
bun test test/memory-ingest-no-put_page.test.ts \
         test/memory-ingest-timeout.test.ts \
         test/gstack-memory-helpers.test.ts                 34 pass  0 fail
```

## Note on scope

The issue also suggests recovering attribution from the Claude Code project-directory
encoding rather than discarding it. That's a behaviour change with its own design
questions, so it's left out — this PR only makes the existing flag do what it says.

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -744,7 +744,10 @@ function buildTranscriptPage(path: string, session: ParsedSession): PageRecord {
     source_path: path,
     session_id: session.session_id,
     cwd: session.cwd,
-    git_remote: remote,
+    // Apply the same "_unattributed" fallback the frontmatter and repoSlug use.
+    // The unattributed skip guard reads this field, so leaving the raw "" here
+    // makes `--include-unattributed` inert (`"" === "_unattributed"` is never true).
+    git_remote: remote || "_unattributed",
     start_time: session.start_time,
     end_time: session.end_time,
     partial: session.partial,
--- a/test/gstack-memory-ingest.test.ts
+++ b/test/gstack-memory-ingest.test.ts
@@ -296,6 +296,64 @@ describe("internal: parseTranscriptJsonl + buildTranscriptPage shape", () => {
   });
 });
 
+// ── unattributed skip policy ───────────────────────────────────────────────
+
+describe("gstack-memory-ingest unattributed skip", () => {
+  function sessionIn(cwd: string): string {
+    return (
+      `{"type":"user","message":{"role":"user","content":"hi"},"timestamp":"2026-05-01T00:00:00Z","cwd":"${cwd}"}\n` +
+      `{"type":"assistant","message":{"role":"assistant","content":"hello"},"timestamp":"2026-05-01T00:00:01Z"}\n`
+    );
+  }
+
+  it("skips a session whose cwd is not a git repo", () => {
+    const home = makeTestHome();
+    const work = join(home, "work");
+    mkdirSync(work, { recursive: true });
+    writeClaudeCodeSession(home, "not-a-repo", "sess1", sessionIn(work));
+
+    const r = runScript(["--bulk", "--no-write"], { HOME: home, GSTACK_HOME: join(home, ".gstack") });
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("written:               0");
+    expect(r.stdout).toContain("skipped (unattrib):    1");
+
+    rmSync(home, { recursive: true, force: true });
+  });
+
+  it("imports that same session when --include-unattributed is passed", () => {
+    const home = makeTestHome();
+    const work = join(home, "work");
+    mkdirSync(work, { recursive: true });
+    writeClaudeCodeSession(home, "not-a-repo", "sess1", sessionIn(work));
+
+    const r = runScript(["--bulk", "--no-write", "--include-unattributed"], {
+      HOME: home,
+      GSTACK_HOME: join(home, ".gstack"),
+    });
+    expect(r.exitCode).toBe(0);
+    expect(r.stdout).toContain("written:               1");
+    expect(r.stdout).toContain("skipped (unattrib):    0");
+
+    rmSync(home, { recursive: true, force: true });
+  });
+
+  it("still imports a session whose cwd does resolve to a git remote", () => {
+    const home = makeTestHome();
+    const repo = join(home, "repo");
+    mkdirSync(repo, { recursive: true });
+    spawnSync("git", ["-C", repo, "init", "-q"]);
+    spawnSync("git", ["-C", repo, "remote", "add", "origin", "https://github.com/foo/bar.git"]);
+    writeClaudeCodeSession(home, "real", "sess2", sessionIn(repo));
+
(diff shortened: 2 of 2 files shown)


noron12234:
Closing this for now — I'm stepping back to re-review my own work before asking for anyone's time. Thanks, and sorry for the noise.

noron12234:
Reopening after re-checking the one thing that could have made this wrong — whether normalising `page.git_remote` breaks any other reader.

It does not. `git_remote` has exactly four sites in `bin/gstack-memory-ingest.ts` on current `main`:

| line | what it does |
|---|---|
| 726 | frontmatter — already writes `${remote \|\| "_unattributed"}` |
| 747 | `buildTranscriptPage` — writes the raw `remote` (this PR) |
| 784 | `buildArtifactPage` — writes `slug_repo`, already defaulted to `"_unattributed"` |
| 1170 | the skip guard — `page.git_remote === "_unattributed"` |

So 747 is the only site that skips the fallback, and 1170 is the only consumer of the field. The guard therefore compares `""` against `"_unattributed"` and never fires, which is #2353. Aligning 747 with 726/784 is the minimal fix and there is no other reader to regress.

The three added tests cover both directions plus the attributed case, and they fail if the one-line change is reverted.

noron12234:
Rebased onto current `main`. The code half of this PR is now **obsolete** — `main` already carries `git_remote: remote || "_unattributed"` at `bin/gstack-memory-ingest.ts:795`, plus a regression test for the deny-policy path (`a deny policy keyed _unattributed applies to unattributable transcripts (#2353)`). I confirmed it is the same fix by reverting that one line locally and watching your test go red:

```
git_remote: remote,      # reverted
44 pass, 1 fail
(fail) a deny policy keyed _unattributed applies to unattributable transcripts (#2353)
```

So I dropped my version of the code change and the diff is now **test-only**: 58 lines covering the skip *counter* rather than the policy filter — that a non-repo session reports `skipped (unattrib): 1` and `written: 0`, that `--include-unattributed` flips it to `0`/`1`, and that a session with a real remote is unaffected.

`bun test test/gstack-memory-ingest.test.ts` → **45 pass, 0 fail** on the rebased branch.

That said: `main` already references `--include-unattributed` in 16 places in that file, so this may well be coverage you already have. Your call — happy for you to close it if it is redundant, or I can trim it to whichever of the three cases is actually missing. No hard feelings either way, the underlying bug is fixed and that was the point.

noron12234:
Closing this. As I noted on 08-27 the code half is already on `main`, so there is nothing left here worth your review time.

Thanks for the patience on it.

PR #2372, opened 2026-07-28
fix(mktemp): X's must be trailing — breaks on macOS (2nd run) and Alpine (1st run) (#2370)

### LINKED ISSUE #2370: /codex: mktemp templates put X's before a .txt suffix — breaks on macOS after the first run, and the failure looks like a model stall
## Summary

Every `mktemp` template in `codex/SKILL.md.tmpl` places the `XXXXXX` placeholders **before** a `.txt` suffix. BSD `mktemp` (macOS) only substitutes X's when they are the **trailing** characters of the template. With a suffix it treats the whole string as a literal filename.

The result is not an immediate error, which is what makes it nasty:

- **Run 1** creates a file literally named `codex-err-XXXXXX.txt` and exits 0. Everything works.
- **Run 2 and every run after** fail with `mkstemp failed: ... File exists`.

Because the literal file persists in `$TMPDIR`, the skill is broken from the second invocation onward and **stays** broken until `$TMPDIR` is cleared.

## Why it's worse than a broken temp file

The failure mode defeats the non-zero-exit surfacing added in #1327.

`$TMPERR` is assigned from the failed `mktemp`, so it is empty. The redirect `2>"$TMPERR"` then fails, `codex` never runs, and the skill's own handler prints:

```
[codex exit 1]
```

with nothing after it, because `head -1 "$TMPERR"` has no file to read. That is indistinguishable from "the model stalled and returned nothing" — precisely the misdiagnosis #1327's comment says the exit-code surfacing exists to prevent:

> Surface non-zero exits (parse errors, arg-shape breaks, etc.) so the calling agent doesn't read "no output" as a silent model/API stall and burn 30-60min misdiagnosing it. See #1327.

I lost a run to exactly this before spotting it.

## Reproduction

macOS, BSD `mktemp`:

```console
$ rm -f "${TMPDIR}demo-XXXXXX.txt"

$ mktemp "${TMPDIR}demo-XXXXXX.txt"; echo "exit=$?"
/var/folders/.../T/demo-XXXXXX.txt
exit=0

$ mktemp "${TMPDIR}demo-XXXXXX.txt"; echo "exit=$?"
mktemp: mkstemp failed on /var/folders/.../T/demo-XXXXXX.txt: File exists
exit=1
```

Note run 1 created a file whose name contains six literal `X` characters — no substitution happened.

Correct form:

```console
$ mktemp "${TMPDIR%/}/demo-XXXXXX"
/var/folders/.../T/demo-0v7hIP
```

## Evidence this is happening in the wild

Literal artifacts accumulated in my `$TMPDIR` from real `/codex` runs across multiple sessions:

```
/var/folders/.../T/codex-err-XXXXXX.txt
/var/folders/.../T/codex-err-review-XXXXXX.txt
/var/folders/.../T/codex-out-XXXXXX.txt
/var/folders/.../T/codex-prompt-XXXXXX.txt
```

`codex-err-review-XXXXXX.txt` does not match any template in the current skill, so it is left over from an older revision — these have been quietly piling up for a while. They are also never cleaned: `rm -f "$TMPERR"` removes the literal file, but only on the paths that reach it.

## Affected sites

All in `codex/SKILL.md.tmpl` (the generated `codex/SKILL.md` mirrors them):

| Line | Template |
|---|---|
| 161 | `TMPERR=$(mktemp "$TMP_ROOT/codex-err-XXXXXX.txt")` |
| 209 | `_PROMPT_FILE=$(mktemp "$TMP_ROOT/codex-prompt-XXXXXX.txt")` |
| 338 | `TMPERR=${TMPERR:-$(mktemp "$TMP_ROOT/codex-err-XXXXXX.txt")}` |
| 437 | `TMPRESP=$(mktemp "$TMP_ROOT/codex-resp-XXXXXX.txt")` |
| 438 | `TMPERR=$(mktemp "$TMP_ROOT/codex-err-XXXXXX.txt")` |

`git log -S` dates the pattern to d8523301 (#197, the original `/codex` skill), so it predates the `$TMP_ROOT` refactor in #1056 — the CHANGELOG for that wave shows the prior form was `mktemp /tmp/codex-*-XXXXXX.txt`, i.e. the suffix came along for the ride rather than being introduced by it.

## Secondary: `$TMP_ROOT` already ends with a slash

`bin/gstack-paths` emits:

```
TMP_ROOT=/var/folders/dy/_4txks415gx7mmdct58hmn_m0000gn/T/
```

so `"$TMP_ROOT/codex-err-..."` expands to `...T//codex-err-...`. Harmless on POSIX, but worth normalizing while touching these lines.

## Suggested fix

Move the X's to the end and strip the trailing slash at each of the 5 sites:

```diff
-TMPERR=$(mktemp "$TMP_ROOT/codex-err-XXXXXX.txt")
+TMPERR=$(mktemp "${TMP_ROOT%/}/codex-err-XXXXXX")
```

If the `.txt` extension matters to anything downstream, the portable pattern is to create then rename:

```bash
TMPERR=$(mktemp "${TMP_ROOT%/}/codex-err-XXXXXX") && mv "$TMPERR" "$TMPERR.txt" && TMPERR="$TMPERR.txt"
```

Nothing in the skill appears to depend on the extension, so dropping it looks like the simpler correct move.

Worth a grep across other skills for the same shape — I only checked `codex/`, where it is the sole offender among `*.md` templates.

Hardening thought, separate from the fix: the `[codex exit N]` handler could fall back to a message when `$TMPERR` is empty or unset, so a temp-file failure can never again present as silent model output.

## Environment

- gstack `v1.60.1.0` (`a3259400`)
- macOS, `Darwin 25.5.0 arm64`, BSD `mktemp` (no `--version`)
- zsh
- Global git install at `~/.claude/skills/gstack`, team mode on

Happy to send a PR for the 5-line change if useful.

noron12234:
Fixes #2370.

## Scope

The issue reports the five `mktemp` templates in `codex/SKILL.md.tmpl`. The same
mistake exists in three more places it doesn't mention, so this PR fixes all of them:

| file | sites | template |
|---|---|---|
| `codex/SKILL.md.tmpl` | 5 | `codex-{err,prompt,resp}-XXXXXX.txt` ← #2370 |
| `claude/SKILL.md.tmpl` | 3 | `…-response-XXXXXX.json`, `…-error-XXXXXX.txt`, `…-diff-XXXXXX.patch` |
| `scripts/resolvers/review.ts` | 1 | `gstack-codex-oh-XXXXXXXX.txt` → `office-hours/SKILL.md` |
| `bin/gstack-developer-profile` | 2 | `developer-profile.json.XXXXXX.tmp` |

`claude/SKILL.md.tmpl` is the clearest tell — line 97 is already correct and the two
lines directly under it are not:

```bash
PROMPT_FILE=$(mktemp /tmp/gstack-claude-prompt-XXXXXX)        # correct
RESP_FILE=$(mktemp /tmp/gstack-claude-response-XXXXXX.json)   # broken
ERR_FILE=$(mktemp /tmp/gstack-claude-error-XXXXXX.txt)        # broken
```

## Measured behaviour

Three platforms, same two templates:

| | `codex-err-XXXXXX.txt` (old) | `codex-err-XXXXXX` (new) |
|---|---|---|
| **macOS** (BSD mktemp) | run 1 creates a file named literally `codex-err-XXXXXX.txt`; run 2+ → `mkstemp failed … File exists` | fresh name every run |
| **Alpine** (busybox mktemp) | **`mktemp: : Invalid argument` on every run**, including the first | fresh name every run |
| **Debian** (GNU coreutils 9.7) | works — substitutes and keeps the suffix | works |

```console
# macOS
run 1: /var/…/codex-err-XXXXXX.txt
run 2: mktemp: mkstemp failed on /var/…/codex-err-XXXXXX.txt: File exists
run 3: mktemp: mkstemp failed on /var/…/codex-err-XXXXXX.txt: File exists

# Alpine
run 1: mktemp: : Invalid argument
run 2: mktemp: : Invalid argument
```

Only GNU tolerates a suffix, which is why this survived — Linux CI is green while
macOS and Alpine users hit it. Note Alpine fails on the *first* run, so it's not even
the delayed failure mode the issue describes.

`bin/gstack-developer-profile` fails differently: its `trap 'rm -f "$TMPOUT"' EXIT`
usually cleans the literal file up, so single runs look fine. But the name is constant,
so two concurrent invocations collide on the same path and one process's trap deletes
the other's temp file — exactly the guarantee `mktemp` exists to provide:

```console
old: A → /…/developer-profile.json.XXXXXX.tmp
     B → mktemp: mkstemp failed … File exists
new: A → developer-profile.json.tmp.hpkoP3
     B → developer-profile.json.tmp.JHnL0U
```

## Fix

Move the X's to the end. The suffixes aren't load-bearing — nothing globs on them and
every cleanup is `rm -f "$VAR"` — so they're dropped. `developer-profile` keeps its
`.tmp` marker, moved ahead of the X's.

## Verification

- Generated docs rebuilt with `bun run gen:skill-docs`; the tracked diff is exactly
  6 files / 17 lines and **every changed line is a `mktemp` line** (`git diff -U0`
  filtered on non-`mktemp` content comes back empty).
- `bun test test/gen-skill-docs.test.ts test/gen-skill-docs-idempotency.test.ts
  test/gen-skill-docs-out-dir.test.ts test/codex-hardening.test.ts` → **436 pass, 0 fail**.

## Not included

The issue also notes the temp files are never cleaned up (`rm -f "$TMPERR"` missing on
some paths) and that stale `codex-err-review-XXXXXX.txt` artifacts from older revisions
linger. That's a separate change to the skill's control flow — happy to follow up if
you want it in the same pass.

### DIFF
--- a/bin/gstack-developer-profile
+++ b/bin/gstack-developer-profile
@@ -55,7 +55,7 @@ do_migrate() {
 
   # Run migration in a temp file, then atomic rename.
   local TMPOUT
-  TMPOUT=$(mktemp "$GSTACK_HOME/developer-profile.json.XXXXXX.tmp")
+  TMPOUT=$(mktemp "$GSTACK_HOME/developer-profile.json.tmp.XXXXXX")
   trap 'rm -f "$TMPOUT"' EXIT
 
   cat "$LEGACY_FILE" | bun -e "
@@ -182,7 +182,7 @@ do_log_session() {
   ensure_profile
 
   local TMPOUT
-  TMPOUT=$(mktemp "$GSTACK_HOME/developer-profile.json.XXXXXX.tmp")
+  TMPOUT=$(mktemp "$GSTACK_HOME/developer-profile.json.tmp.XXXXXX")
   trap 'rm -f "$TMPOUT"' EXIT
 
   PROFILE_FILE_PATH="$PROFILE_FILE" RECORD_INPUT="$INPUT" TMPOUT_PATH="$TMPOUT" bun -e "
--- a/claude/SKILL.md.tmpl
+++ b/claude/SKILL.md.tmpl
@@ -95,8 +95,8 @@ Create temp files:
 
 ```bash
 PROMPT_FILE=$(mktemp /tmp/gstack-claude-prompt-XXXXXX)
-RESP_FILE=$(mktemp /tmp/gstack-claude-response-XXXXXX.json)
-ERR_FILE=$(mktemp /tmp/gstack-claude-error-XXXXXX.txt)
+RESP_FILE=$(mktemp /tmp/gstack-claude-response-XXXXXX)
+ERR_FILE=$(mktemp /tmp/gstack-claude-error-XXXXXX)
 ```
 
 Cleanup at the end of every mode:
@@ -151,7 +151,7 @@ Review the current branch diff with nested Claude in tool-less mode.
 ```bash
 _REPO_ROOT=$(git rev-parse --show-toplevel) || { echo "ERROR: not in a git repo" >&2; exit 1; }
 cd "$_REPO_ROOT"
-DIFF_FILE=$(mktemp /tmp/gstack-claude-diff-XXXXXX.patch)
+DIFF_FILE=$(mktemp /tmp/gstack-claude-diff-XXXXXX)
 git fetch origin <base> --quiet 2>/dev/null || true
 git diff "origin/<base>" > "$DIFF_FILE" 2>/dev/null || git diff "<base>" > "$DIFF_FILE"
 ```
--- a/codex/SKILL.md
+++ b/codex/SKILL.md
@@ -967,7 +967,7 @@ Run Codex code review against the current branch diff.
 
 1. Create temp files for output capture:
 ```bash
-TMPERR=$(mktemp "$TMP_ROOT/codex-err-XXXXXX.txt")
+TMPERR=$(mktemp "$TMP_ROOT/codex-err-XXXXXX")
 ```
 
 2. Run the review (5-minute timeout). **Codex CLI ≥ 0.130.0 rejects passing a
@@ -1015,7 +1015,7 @@ when the diff content is adversarial:
 _REPO_ROOT=$(git rev-parse --show-toplevel) || { echo "ERROR: not in a git repo" >&2; exit 1; }
 cd "$_REPO_ROOT"
 _USER_INSTRUCTIONS="<everything after '/codex review ' in user input>"
-_PROMPT_FILE=$(mktemp "$TMP_ROOT/codex-prompt-XXXXXX.txt")
+_PROMPT_FILE=$(mktemp "$TMP_ROOT/codex-prompt-XXXXXX")
 {
   printf '%s\n' "IMPORTANT: Do NOT read or execute any files under ~/.claude/, ~/.agents/, .claude/skills/, or agents/. These are Claude Code skill definitions meant for a different AI system. Do NOT modify agents/openai.yaml. Stay focused on repository code only."
   printf '\nCustom focus: %s\n\n' "$_USER_INSTRUCTIONS"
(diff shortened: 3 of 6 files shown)


noron12234:
Closing this for now — I'm stepping back to re-review my own work before asking for anyone's time. Thanks, and sorry for the noise.

noron12234:
Reopening — I took this back to re-verify it rather than leave a claim sitting unchecked. It reproduces, on stock macOS 26.4.1 / `/usr/bin/mktemp`:

```console
$ mktemp /tmp/probe-XXXXXX.txt
/tmp/probe-XXXXXX.txt              # note: the X's were NOT substituted

$ mktemp /tmp/probe-XXXXXX.txt
mktemp: mkstemp failed on /tmp/probe-XXXXXX.txt: File exists

$ mktemp /tmp/probe2-XXXXXX        # trailing X's, i.e. this PR
/tmp/probe2-Qwt1XE
$ mktemp /tmp/probe2-XXXXXX
/tmp/probe2-ph20Q5
```

BSD `mktemp` only substitutes a trailing run of X's, so the first call silently creates a file *literally named* `codex-err-XXXXXX.txt` and every later call in the same boot fails. GNU coreutils rejects the template outright (`too few X's`), which is why Alpine fails on the first run instead of the second.

The failure surfaces as an empty `TMPERR` and a codex step that looks like a model stall, which is what #2370 describes. Still rebases cleanly on `main`.

garrytan:

PR #2373, opened 2026-07-28
Fix AskUserQuestion pass-through emitting an invalid permission decision

ilanbm:
## Problem

The pass-through path in `hosts/claude/hooks/question-preference-hook.ts` returns `permissionDecision: 'defer'`. `defer` is not a valid PreToolUse decision (`allow | deny | ask`), so Claude Code rejects the hook output and aborts the AskUserQuestion call itself. With the hook active, every question fails with "Stopped — Tool execution was interrupted", while the hook exits 0 and logs nothing — the failure presents as a harness bug rather than a hook bug. Introduced in v1.60.1.0; still present on main.

## Fix

A hook with no opinion emits nothing: pass-through now exits silently instead of inventing a decision value. The Conductor transport-avoidance `deny` path is unchanged.

## Test

`test/question-preference-hook-contract.test.ts` pins the output contract mechanically: pass-through, non-target tools, and malformed stdin are silent with exit 0, and any emitted decision must come from the valid set. Hermetic via a temp `GSTACK_STATE_ROOT`. `bun test test/question-preference-hook-contract.test.ts` — 6 pass.

### DIFF
--- a/hosts/claude/hooks/question-preference-hook.ts
+++ b/hosts/claude/hooks/question-preference-hook.ts
@@ -93,12 +93,21 @@ function readStdin(): Promise<string> {
 }
 
 function defer(additionalContext?: string): void {
-  const out: Record<string, unknown> = {
-    hookEventName: 'PreToolUse',
-    permissionDecision: 'defer',
-  };
-  if (additionalContext) out.additionalContext = additionalContext;
-  process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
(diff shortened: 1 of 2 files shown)


ilanbm:
Superseded by #2472 (v1.61.0.0) — same root cause and the same fix (pass-through is exit 0 with exactly-empty stdout, never a `permissionDecision`), and `test/question-preference-hook.test.ts` there covers strictly more than the contract test in this PR. Closing in favor of yours. Thanks for the fix wave.

Issue #2091, opened 2026-06-22
/codex temp-file creation fails on macOS (BSD mktemp): .txt suffix after XXXXXX + trailing-slash TMP_ROOT


scotthardin:
### Summary
On macOS, `/codex` (all three modes) fails before Codex ever runs: the `mktemp`
calls that create the stderr/prompt/response temp files error out, the bash block
exits 1, and the user sees no review. Two compounding bugs, both in gstack.

### Environment
- gstack: 1.58.4.0
- macOS: 26.5.1 (Apple silicon)
- mktemp: /usr/bin/mktemp (BSD)
- $TMPDIR: /var/folders/.../T/  (note the trailing slash — macOS default)

### Symptom
mktemp: mkstemp failed on /var/folders/.../T//codex-err-XXXXXX.txt: File exists
mktemp: mkstemp failed on /var/folders/.../T//codex-prompt-XXXXXX.txt: File exists
/opt/homebrew/bin/bash: line 13: : No such file or directory
=== CODEX_EXIT=1 ===
(The empty `TMPERR`/`_PROMPT_FILE` vars then cascade into "No such file or directory"
for every later `"$TMPERR"` / `cat "$_PROMPT_FILE"`.)

### Root cause
1. **Suffix after the placeholder.** `skills/codex/SKILL.md` uses templates like
   `mktemp "$TMP_ROOT/codex-err-XXXXXX.txt"` (also `codex-prompt-XXXXXX.txt`,
   `codex-resp-XXXXXX.txt`). GNU `mktemp` tolerates a suffix after `XXXXXX`; **BSD
   `mktemp` (macOS) requires the placeholder at the END of the template** and
   mishandles the `.txt` suffix.
2. **Trailing slash in `TMP_ROOT`.** `bin/gstack-paths` emits `TMP_ROOT` straight
   from `$TMPDIR`, which on macOS ends in `/`. The template then expands to a
   double-slash path (`…/T//codex-err-…`), compounding the failure.

### Affected code
- `skills/codex/SKILL.md` (and `SKILL.md.tmpl`) — Steps 2A / 2B / 2C, every
  `mktemp "$TMP_ROOT/codex-*-XXXXXX.txt"`.
- `bin/gstack-paths` — `TMP_ROOT` carries `$TMPDIR`'s trailing slash.

### Suggested fix
Drop the `.txt` suffix (X's at the end) and strip the trailing slash:
```bash
TR="${TMP_ROOT%/}"
[...]

PR #2376, opened 2026-07-28
fix(paths): shell-quote gstack-paths output so eval round-trips values

### LINKED ISSUE #2374: Windows: eval "$(gstack-paths)" strips backslashes from TMP_ROOT — mktemp fails and skills exit before Codex runs
### Summary

On Windows (git-bash/MSYS), `eval "$(gstack-paths)"` corrupts `TMP_ROOT` before any skill
gets to use it. `$TMP` on Windows is a backslash path like `C:\Users\me\AppData\Local\Temp`,
and `bin/gstack-paths` emits it as a bare, unquoted assignment. `eval` then consumes the
backslashes as escape sequences, so the caller ends up with `C:UsersmeAppDataLocalTemp`.

Every skill that does `mktemp "$TMP_ROOT/..."` after Step 0.6 then fails, and the failure
cascades into exactly the same "empty `TMPERR` → `: No such file or directory`" symptom
already documented for macOS in #2091 — but from a different root cause, so the fix there
does not help here.

### Environment

- gstack: 1.60.1.0 (`a3259400`)
- OS: Windows 11 Pro 26200, `MINGW64_NT-10.0-26200`
- bash: GNU bash 5.2.37(1)-release (x86_64-pc-msys)
- `TMPDIR`: unset (normal on Windows)
- `TMP`: `C:\Users\me\AppData\Local\Temp`

### Reproduction

```bash
P=~/.claude/skills/gstack/bin/gstack-paths

# The script's own output is correct:
env -u TMPDIR TMP='C:\Users\test\AppData\Local\Temp' bash "$P" | grep TMP_ROOT
# TMP_ROOT=C:\Users\test\AppData\Local\Temp

# ...but the documented usage, eval, corrupts it:
env -u TMPDIR TMP='C:\Users\test\AppData\Local\Temp' bash -c 'eval "$('"$P"')"; echo "[$TMP_ROOT]"'
# [C:UserstestAppDataLocalTemp]        <-- every separator gone

# A path with a space is worse — eval word-splits it and TMP_ROOT ends up empty:
env -u TMPDIR TMP='C:\Program Files\Temp' bash -c 'eval "$('"$P"')"; echo "[$TMP_ROOT]"'
# bash: line 3: FilesTemp: command not found
# []
```

### Downstream impact

`skills/codex/SKILL.md` Step 2A (and every other skill using `$TMP_ROOT`) does:

```bash
TMPERR=$(mktemp "$TMP_ROOT/codex-err-XXXXXX.txt")
... codex review "..." 2>"$TMPERR"
```

With the corrupted value:

```
mktemp: failed to create file via template 'C:UserstestAppDataLocalTemp/codex-err-XXXXXX.txt': No such file or directory
bash: line 1: : No such file or directory
exit=1
```

`mktemp` fails, `TMPERR` is empty, and `2>"$TMPERR"` is a redirect to an empty filename, so
**Codex never runs at all**. The user sees an immediate exit 1 with no review and no
obvious cause. In the `C:\Program Files` case `TMP_ROOT` is empty instead, so `mktemp`
targets the filesystem root and fails with `Permission denied`.

This is the same user-visible failure as #2091 / #2370, which is worth noting because
those are both filed as macOS/BSD-mktemp problems — a Windows user hitting this will find
them, try the BSD fix, and still be broken.

### Root cause

`bin/gstack-paths` ends with:

```bash
echo "GSTACK_STATE_ROOT=$_state_root"
echo "PLAN_ROOT=$_plan_root"
echo "TMP_ROOT=$_tmp_root"
```

The documented contract is `eval "$(gstack-paths)"`, so these lines are re-parsed by the
shell. An unquoted RHS means backslashes are escapes and spaces are word separators.

The header comment does anticipate hostile values:

> Security: output values are not sanitized — callers may receive paths with shell-special
> characters if env vars contain them. Skills should always quote expansions
> (`"$GSTACK_STATE_ROOT"`, not `$GSTACK_STATE_ROOT`).

That mitigation does not apply to this failure. The corruption happens *during* `eval`,
before the caller has any variable to quote. Callers already quote correctly and are still
broken. Note this is not only a Windows concern: any POSIX `TMPDIR` containing a space or
backslash corrupts the same way.

### Proposed fix

Emit shell-quoted assignments so `eval` round-trips the exact value. `printf %q` is a bash
builtin and the script is already `#!/usr/bin/env bash`:

```diff
-echo "GSTACK_STATE_ROOT=$_state_root"
-echo "PLAN_ROOT=$_plan_root"
-echo "TMP_ROOT=$_tmp_root"
+printf 'GSTACK_STATE_ROOT=%q\n' "$_state_root"
+printf 'PLAN_ROOT=%q\n' "$_plan_root"
+printf 'TMP_ROOT=%q\n' "$_tmp_root"
```

Verified round-trip through `eval` on all four cases:

| Input | Result |
|---|---|
| `C:\Users\test\AppData\Local\Temp` | PASS |
| `C:\Program Files\Temp` | PASS |
| `/tmp/o'brien` | PASS |
| `/plain/posix/path` | PASS |

One tradeoff worth a maintainer decision: the header also documents a plain
non-eval usage (`gstack-paths` → prints `GSTACK_STATE_ROOT=...`). With `%q`, POSIX paths
print unchanged, but a Windows path prints escaped (`C:\\Users\\...`). If that
human-readable form matters, an alternative is keeping `echo` and changing the contract to
`set -a; . <(gstack-paths)`, though that is a wider change across every skill's Step 0.6.

### Notes

Found while dogfooding `/codex review` on Windows. Working around it with a literal
`TMP_ROOT=/tmp` before the `mktemp` makes `/codex review` work normally on 1.60.1.0,
which isolates this as the only blocker on that path.

Related but distinct: #2091, #2370 (same cascade, BSD `mktemp` root cause, macOS),
#1782 (`gstack-config` truncating paths at the first space — same unquoted-path family).

fangearhq-boop:
Fixes #2374.

> Re-filed. The original PR (#2375) was closed automatically when its head fork was
> deleted by mistake — GitHub closes any PR whose head repository goes away, and such a PR
> can't be reopened. Same commit (`0134f29e`), same tests, no content changes. Apologies for
> the duplicate notification.

### The bug

`bin/gstack-paths` is consumed as `eval "$(gstack-paths)"` (the documented contract, used
at Step 0.6 of every skill), but emits bare `KEY=VALUE` lines with an unquoted right-hand
side. `eval` re-parses those: backslashes become escape sequences, spaces become word
separators.

On Windows `$TMP` is always a backslash path, so this fires every time:

```bash
TMP='C:\Users\me\AppData\Local\Temp'
eval "$(gstack-paths)"
echo "$TMP_ROOT"
# C:UsersmeAppDataLocalTemp      <- every separator gone
```

The `mktemp "$TMP_ROOT/codex-err-XXXXXX.txt"` that follows then fails, `TMPERR` is empty,
and `2>"$TMPERR"` is a redirect to an empty filename, so the bash block exits 1 **before
Codex ever runs**:

```
mktemp: failed to create file via template 'C:Users...Temp/codex-err-XXXXXX.txt': No such file or directory
bash: line 1: : No such file or directory
exit=1
```

A value containing a space is worse — `eval` word-splits it, prints
`FilesTemp: command not found`, and leaves the variable empty, so `mktemp` targets the
filesystem root and fails with `Permission denied`.

This is the same user-visible cascade as #2091 and #2370, both filed as macOS/BSD-`mktemp`
bugs. The root cause here is different, so those fixes don't help: a Windows user searching
the symptom finds them, applies them, and is still broken.

Not Windows-only — any POSIX `TMPDIR` containing a space or a backslash corrupts the same way.

### The fix

Emit with `printf %q` so `eval` reproduces values byte-for-byte:

```diff
-echo "GSTACK_STATE_ROOT=$_state_root"
-echo "PLAN_ROOT=$_plan_root"
-echo "TMP_ROOT=$_tmp_root"
+printf 'GSTACK_STATE_ROOT=%q\n' "$_state_root"
+printf 'PLAN_ROOT=%q\n' "$_plan_root"
+printf 'TMP_ROOT=%q\n' "$_tmp_root"
```

The script is already `#!/usr/bin/env bash`, so `%q` is available. Plain POSIX paths are
unchanged by `%q`, so existing callers and the existing assertions see identical output.

I also reworded the header comment. The old note said callers should quote expansions,
which doesn't cover this failure — the corruption happens *during* `eval`, before the
caller has a variable to quote. Callers here already quote correctly and were still broken.

### Tests

Three `eval`-round-trip cases added to `test/gstack-paths.test.ts`: backslash, space, and
embedded single quote. They assert the full documented calling convention rather than the
raw stdout, which is what the existing tests check.

Verified they are real regression tests — against the old `echo` they fail, with `%q` they pass:

| | old `echo` | `printf %q` |
|---|---|---|
| `test/gstack-paths.test.ts` | 2 fail, 10 pass | **12 pass, 0 fail** |
| `test/gstack-paths.test.ts` + `test/spec-template-invariants.test.ts` | — | **60 pass, 0 fail** |

(2 rather than 3 failures on the old code because the backslash case is skipped on this
Windows runner — see below. On Linux CI all three fail without the fix.)

Test-environment note worth flagging for review: the backslash case is skipped on `win32`.
MSYS/Git Bash rewrites both `C:\...` → `/c/...` and backslashes → forward slashes in env
values, so a literal backslash cannot be injected through the environment on a Git Bash
runner. What it guards is pure `eval` semantics, so exercising it on Linux/macOS CI is
sufficient. This follows the same reasoning as the existing `HOME`-unset skips in that
file. The space and quote cases do run on Windows, and both catch the bug there.

The three test files that reference `gstack-paths` or `TMP_ROOT` are
`test/gstack-paths.test.ts`, `test/spec-template-invariants.test.ts`, and
`test/skill-e2e-plan-tune-cathedral.test.ts` (evals tier). The first two pass in full.
No templates are touched, so generated `SKILL.md` output is unaffected.

### Alternative considered

Keeping `echo` and changing the contract to `set -a; . <(gstack-paths)` also fixes it and
preserves the plain human-readable output documented on line 4. I didn't take it because
it changes Step 0.6 in every skill template plus each host variant, where this is three
lines in one file. Happy to switch if you'd rather have the unescaped display form.

Found while dogfooding `/codex review` on Windows: with a literal `TMP_ROOT=/tmp`
substituted before the `mktemp`, `/codex review` works normally on 1.60.1.0, which isolates
this as the only blocker on that path.

### DIFF
--- a/bin/gstack-paths
+++ b/bin/gstack-paths
@@ -13,9 +13,15 @@
 #   PLAN_ROOT:         GSTACK_PLAN_DIR -> CLAUDE_PLANS_DIR -> $HOME/.claude/plans -> .claude/plans
 #   TMP_ROOT:          TMPDIR -> TMP -> .gstack/tmp (and mkdir -p, best-effort)
 #
-# Security: output values are not sanitized — callers may receive paths with
-# shell-special characters if env vars contain them. Skills should always quote
-# expansions ("$GSTACK_STATE_ROOT", not $GSTACK_STATE_ROOT).
+# Output: values are emitted shell-quoted (printf %q) so `eval` round-trips them
+# byte-for-byte. This matters on Windows, where $TMP is a backslash path like
+# C:\Users\me\AppData\Local\Temp — with a bare `echo`, eval consumes the
+# backslashes as escapes and the caller gets C:UsersmeAppDataLocalTemp. A value
+# containing a space (C:\Program Files\Temp) is worse: eval word-splits it and
+# the variable ends up empty. Quoting here is the only fix that works, because
+# the corruption happens during eval, before the caller has anything to quote.
+# Callers should still quote expansions ("$GSTACK_STATE_ROOT") for the same
+# reason any path variable needs quoting.
 set -u
 
 # State root: where gstack writes projects/, sessions/, analytics/.
@@ -60,6 +66,6 @@ fi
 # will discover that on their own write attempt. Don't fail the eval here.
 mkdir -p "$_tmp_root" 2>/dev/null || true
 
-echo "GSTACK_STATE_ROOT=$_state_root"
-echo "PLAN_ROOT=$_plan_root"
-echo "TMP_ROOT=$_tmp_root"
+printf 'GSTACK_STATE_ROOT=%q\n' "$_state_root"
+printf 'PLAN_ROOT=%q\n' "$_plan_root"
+printf 'TMP_ROOT=%q\n' "$_tmp_root"
--- a/test/gstack-paths.test.ts
+++ b/test/gstack-paths.test.ts
@@ -104,6 +104,57 @@ describe('gstack-paths', () => {
     expect(got).toHaveProperty('TMP_ROOT');
   });
 
+  // Regression: values must survive `eval "$(gstack-paths)"`, which is the
+  // documented calling convention. A bare `echo` emits an unquoted RHS, so eval
+  // re-parses it: backslashes become escapes and spaces become word separators.
+  // On Windows $TMP is always a backslash path, so every skill that then runs
+  // mktemp "$TMP_ROOT/..." fails and the bash block dies before doing any work.
+  // These run identically on POSIX — the values are just strings.
+  function evalRoundTrip(env: Record<string, string | undefined>, varName: string): string {
+    const result = spawnSync(
+      'bash',
+      ['-c', `eval "$(bash "$1")"; printf '%s' "\${${varName}}"`, 'sh', BIN],
+      {
+        env: { PATH: process.env.PATH, USERPROFILE: '', ...env } as Record<string, string>,
+        encoding: 'utf-8',
+      },
+    );
+    if (result.status !== 0) {
+      throw new Error(`eval round-trip failed (status ${result.status}): ${result.stderr}`);
+    }
+    return result.stdout;
+  }
+
+  // Values are POSIX-shaped on purpose: MSYS/Git Bash rewrites `C:\...` env
+  // values to `/c/...` before bash sees them, so a literal Windows path would
+  // assert the translation layer rather than the quoting. A backslash is a
+  // backslash to eval either way, which is the behavior under test.
+  test('eval round-trip preserves backslashes (#2374)', () => {
+    // Skip on Windows: MSYS also rewrites backslashes to forward slashes in
+    // env values, so a literal backslash cannot be injected through the
+    // environment on a Git Bash runner. The escape-eating this guards against
+    // is pure eval semantics, so exercising it on Linux/macOS CI is sufficient
+    // — same reasoning as the HOME-unset skips above.
+    if (process.platform === 'win32') return;
+    const backslashed = '/tmp/back\\slash/dir';
+    expect(evalRoundTrip({ TMPDIR: backslashed, HOME: '/h' }, 'TMP_ROOT')).toBe(backslashed);
+  });
+
+  test('eval round-trip preserves spaces (#2374)', () => {
+    // Bare echo made eval word-split this, leaving the variable empty and
+    // emitting `<second-word>: command not found`.
+    const spaced = '/tmp/two words/dir';
+    expect(evalRoundTrip({ TMPDIR: spaced, HOME: '/h' }, 'TMP_ROOT')).toBe(spaced);
+  });
+
+  test('eval round-trip preserves quotes, and leaves plain paths alone (#2374)', () => {
+    expect(evalRoundTrip({ TMPDIR: "/tmp/o'brien", HOME: '/h' }, 'TMP_ROOT')).toBe("/tmp/o'brien");
+    expect(evalRoundTrip({ GSTACK_HOME: '/tmp/state root' }, 'GSTACK_STATE_ROOT')).toBe(
+      '/tmp/state root',
+    );
+    expect(evalRoundTrip({ HOME: '/tmp/myhome' }, 'PLAN_ROOT')).toBe('/tmp/myhome/.claude/plans');
+  });
+
   test('output is shell-evalable: only KEY=VALUE lines, no extra prose', () => {
     const result = spawnSync('bash', [BIN], {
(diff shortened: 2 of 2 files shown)


fangearhq-boop:
Flagging a duplicate so this can be closed out one way or the other: #1580 fixes the same bug, and has been open since May. Two PRs for one bug is probably why neither has moved, so here's the tradeoff laid out to make it a single call.

**#1580 (@yannickspiess)** wraps values in single quotes via a hand-rolled `_shell_quote` helper using `sed`. Portable to POSIX `sh`/dash, so it keeps working if the shebang ever changes. One file, no tests.

**This PR** uses `printf %q`. That's a bash builtin, which is fine today since `bin/gstack-paths` is already `#!/usr/bin/env bash`, but it is the less portable of the two if that ever changes.

So on the mechanism itself, #1580 is arguably the safer choice and I don't want to pretend otherwise.

Where this PR adds things #1580 doesn't:

- **Regression tests.** Three eval-round-trip cases in `test/gstack-paths.test.ts` (backslash, space, embedded quote) that fail against the old `echo` and pass after. Without a test, this reverts the next time someone touches the emitter.
- **The Windows case specifically.** `$TMP` is a backslash path there, so `eval` eats the separators and `TMP_ROOT` comes back as `C:UsersmeAppDataLocalTemp`. Spaces are worse: `eval` word-splits and the variable ends up empty. #1580's framing is spaces-only, which is the narrower half of the bug.
- **The link to why this looks like a different bug.** The downstream symptom is `mktemp "$TMP_ROOT/..."` failing, so `TMPERR` is empty, so `2>"$TMPERR"` is a redirect to an empty filename and the block exits 1 before Codex runs. The user sees no review and no error. That is the same visible cascade as #2091 and #2370, both filed as macOS/BSD-`mktemp` issues. A Windows user finds those, applies them, and is still broken, because the root cause is different. Worth having written down wherever this lands.

Happy either way, including closing this in favor of #1580 — I'd just ask that the tests come across with it, since that's the part that keeps it fixed. If you'd prefer the POSIX helper with these tests on top, say so and I'll rework this PR to that shape.

#2374 is the underlying report and is still open.

garrytan:

Issue #1036, opened 2026-04-17
/codex skill ignores Codex MCP, uses fragile `codex exec` pipe that stalls or truncates (~50% failure rate)


henrysouchien:
The `/codex` skill invokes Codex via `Bash` → `codex exec "<prompt>" --json 2>/dev/null | python3 -u -c "..."` instead of using the `mcp__codex__codex` / `mcp__codex__codex-reply` MCP tools that Claude Code already has available.

This approach is unreliable — roughly half my invocations fail, in one of two ways:

1. **Stalls with no output.** Bash call hangs until the 5-minute timeout, no response ever reaches Claude. Codex may or may not have actually run; impossible to tell because stderr is discarded via `2>/dev/null`.
2. **Output truncated mid-response.** Claude receives a partial answer that looks complete but is cut off, sometimes mid-sentence.

Likely causes:
- `codex exec --json` stream + Python stdin parser can deadlock on full pipe buffers.
- `2>/dev/null` hides the actual failure mode, so there's no signal when something breaks.
- Bash heredoc/quoting of the prompt can mangle large inputs.

The skill preamble explicitly steers Claude away from the MCP tool, which seems backwards — `mcp__codex__codex` handles streaming + structured output + session continuity cleanly and wouldn't hit any of these issues.

**Expected:** prefer `mcp__codex__codex` when available; fall back to `codex exec` only if MCP isn't installed. At minimum, don't swallow stderr.

**Repro:** `/codex review` on any non-trivial plan file, 2–3 times. Expect at least one stall or truncation.

**Environment:** Claude Code on macOS, gstack 0.16.2.0, codex CLI installed, `mcp__codex__*` MCP tools available.

**Workaround:** invoke Codex manually via the MCP tools, skip the skill.

Necmttn:
I would treat `codex exec` as a transport fallback with receipts, not as the default path. For each `/codex` invocation: chosen transport (`mcp` or `exec`), Codex session/thread id when available, stderr capture status, stream close reason, byte count, timeout, and truncation marker.

The dangerous case is partial output that looks complete. A required `complete=true/false` marker or terminal event from the transport would make stalls and mid-response truncation visible instead of turning into bad review input.

Generated with ax - https://github.com/Necmttn/ax

time-attack:
@16francej Please close this issue as resolved. The Codex runner now detaches stdin, enforces timeouts, captures errors, checks authentication, and verifies that a turn completed through #1056 and the streaming work in #536.

[...]

Issue #971, opened 2026-04-11
codex exec hangs in skill bash blocks: missing </dev/null causes 'Reading additional input from stdin' deadlock (codex CLI 0.120.0+)


loning:
## Summary

Every `codex exec` invocation in gstack skill files hangs forever on `codex CLI 0.120.0+` (current released version). Symptom: process at 0% CPU, stdout empty, stderr shows `Reading additional input from stdin...`. The skill workflow then times out or sits indefinitely.

Reproduced today against gstack v0.16.3.0 while running `/plan-eng-review` with codex CLI 0.120.0. First call (with `--enable web_search_cached`) sat at 0% CPU for 13+ minutes, no output. Second call (without web_search) hit timeout 124 with `Reading additional input from stdin...` in stderr. Third call adding `</dev/null` succeeded in 2:25.

## Root cause

`codex exec --help` (0.120.0):

> [PROMPT]
> Initial instructions for the agent. **If not provided as an argument (or if `-` is used), instructions are read from stdin. If stdin is piped and a prompt is also provided, stdin is appended as a `<stdin>` block.**

When skill bash blocks invoke `codex exec "\$PROMPT" ...`, the parent shell's stdin is inherited. In Claude Code's Bash tool (and any background or non-TTY context), stdin is **not** a TTY but also **not** closed — codex detects this as "stdin is piped, append it" and blocks waiting for EOF that never comes. The positional `\$PROMPT` arg is silently followed by an indefinite wait for stdin.

Independent contributing issue: `--enable web_search_cached` is deprecated in 0.120.0 and seems to make hangs more likely / more silent (no clear error).

## Fix

Add `< /dev/null` to every `codex exec` invocation. The prompt is already passed as an arg, so no real stdin input is needed.

Diff (one-liner per call site):

\`\`\`diff
- codex exec "<prompt>" -C "\$_REPO_ROOT" -s read-only -c 'model_reasoning_effort="high"' --enable web_search_cached 2>"\$TMPERR_PV"
+ codex exec "<prompt>" -C "\$_REPO_ROOT" -s read-only -c 'model_reasoning_effort="high"' < /dev/null 2>"\$TMPERR_PV"
\`\`\`

Recommend dropping `--enable web_search_cached` at the same time (deprecated and the prompt template doesn't need it for outside-voice review).

## Affected files in v0.16.3.0

- \`plan-eng-review/SKILL.md:1057\` — outside voice plan review
[...]

PR #2379, opened 2026-07-28
fix(codex): wire the timeout wrapper into the /review and /ship diff passes

aegixx:
Refs #1036.

#1056 added `_gstack_codex_timeout_wrapper` to `bin/gstack-codex-probe` and wired it into `codex/SKILL.md`. The `/review` and `/ship` Codex passes never adopted it. Its entire change to `review/SKILL.md` was appending `< /dev/null`, the stdin fix from #971/#1045, so those passes still run on a bare 5-minute Bash gate and the stalls #1036 describes remain reachable there.

Before this PR, on v1.60.1.0:

| file | wrapper uses | bare `300000` |
| --- | --- | --- |
| `codex/SKILL.md` (+ `.tmpl`) | 5 | 2 |
| `review/SKILL.md` | **0** | 2 |
| `scripts/resolvers/review.ts` | **0** | 5 |
| `ship/sections/adversarial.md` | **0** | 2 |

### The cap is the cause, not the pipe

#1036 attributes the stalls to pipe deadlock and swallowed stderr. Measuring one truncated `/review` adversarial pass (codex-cli 0.145.0, macOS) points somewhere simpler. Its rollout log shows `task_started`, no completion event, 12 shell commands and 6 web searches, last event at **287s** of a 300s budget, cut mid-tool-call. The same prompt under a longer cap: **rc=0 at 336s**, `turn.completed` present. It was 36 seconds short, with no stall and no deadlock.

That also explains the roughly 50% failure rate #1036 reports: cost straddles the gate, so survival tracks diff size. `--enable web_search_cached` is on at both sites and accounted for about 32s of it.

### Changes

For the adversarial and structured-review passes in both skills:

1. Re-source `gstack-codex-probe` inside each Codex block. Shell functions do not survive between Bash blocks, so the wrapper is not in scope at the call site even though `/review` already sources this same file for its auth probe.
2. Wrap both calls with `_gstack_codex_timeout_wrapper 540`.
3. Raise the Bash gate to `600000`, deliberately **above** the wrapper so the wrapper fires first and a stall surfaces as a diagnosable exit 124 instead of a harness kill returning nothing.
4. Drop the "`timeout` doesn't exist on macOS" instruction. The wrapper resolves `gtimeout`, then `timeout`, then runs unwrapped, so that case is already handled; the claim is what steered these sites away from the wrapper.
5. Treat a timeout as missing coverage rather than an informational note. The old text printed "Codex timed out after 5 minutes" and continued, which reads downstream as Codex having reviewed and found nothing.

### Tests

9 new assertions in `test/codex-hardening.test.ts` covering wrapper presence, absence of the macOS claim, and the wrapper-under-gate ordering. They fail on the unpatched tree (29 pass / 9 fail) and pass with the fix (38/38). Full `bun test` green; `bun run gen:skill-docs --dry-run` clean, and `review/SKILL.md` is regenerated rather than hand-edited.

### Deliberately out of scope

Three more bare `timeout: 300000` sites remain in `scripts/resolvers/review.ts` (second opinion, plan review, and one more). This PR does not raise their gates, so their error text is left alone rather than made inconsistent with a cap that did not move. Happy to fold them in if you want.

Separately, `codex/SKILL.md` has the inverse ordering: a 330s wrapper under a 300s Bash gate, which makes its own exit-124 branch unreachable. Left alone here since it is the one place the wrapper is already wired, but it probably wants its own fix.

### Also worth documenting

A truncated pass is not lost. Every run writes `~/.codex/sessions/<YYYY>/<MM>/<DD>/rollout-*.jsonl` containing each `agent_message` it produced, and `task_started` with no completion event is a reliable truncation signature. Three substantive findings were sitting in that file after the measured run had been written off as "Codex unavailable".

**Env:** macOS, gstack 1.60.1.0, codex-cli 0.145.0, `gtimeout`/`timeout` both GNU coreutils 9.11.

### DIFF
--- a/review/SKILL.md
+++ b/review/SKILL.md
@@ -1712,10 +1712,14 @@ If `CODEX_MODE` is `ready`:
 ```bash
 TMPERR_ADV=$(mktemp /tmp/codex-adv-XXXXXXXX)
 _REPO_ROOT=$(git rev-parse --show-toplevel) || { echo "ERROR: not in a git repo" >&2; exit 1; }
-codex exec "IMPORTANT: Do NOT read or execute any files under ~/.claude/, ~/.agents/, .claude/skills/, or agents/. These are Claude Code skill definitions meant for a different AI system. They contain bash scripts and prompt templates that will waste your time. Ignore them completely. Do NOT modify agents/openai.yaml. Stay focused on the repository code only.\n\nReview the changes on this branch against the base branch. Run DIFF_BASE=$(git merge-base origin/<base> HEAD) && git diff "$DIFF_BASE" to see the diff. Your job is to find ways this code will fail in production. Think like an attacker and a chaos engineer. Find edge cases, race conditions, security holes, resource leaks, failure modes, and silent data corruption paths. Be adversarial. Be thorough. No compliments — just the problems. End your output with ONE line in the canonical format `Recommendation: <action> because <one-line reason naming the most exploitable finding>`. Generic reasons like 'because it's safer' do not qualify; the reason must point to a specific finding or no-fix rationale." -C "$_REPO_ROOT" -s read-only -c 'model_reasoning_effort="high"' --enable web_search_cached < /dev/null 2>"$TMPERR_ADV"
+# Shell functions do not survive between Bash blocks, so re-source the probe
+# here. It defines _gstack_codex_timeout_wrapper (gtimeout -> timeout ->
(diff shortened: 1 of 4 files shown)


dgrant:
Looks like this was fixed in v1.64.0.0

aegixx:
Should this be closed now that it has been merged?

Issue #1731, opened 2026-05-26
Windows: `gstack-gbrain-sync.ts` stage 3 fails to spawn `gstack-brain-sync` (`spawnSync` needs `shell: true`)


CharlesBerg85:
## Summary

On Windows (Git Bash + Bun runtime), stage 3 of the unified `gstack-gbrain-sync.ts` (the `brain-sync` stage for curated artifacts → git pipeline) fails with `gstack-brain-sync exited undefined`. Functional impact is limited (running `gstack-brain-sync` directly works fine), but the noise pollutes the sync summary on every gstack skill start and makes real errors harder to spot.

## Environment

- **OS:** Windows 11
- **Shell:** Git Bash (MSYS2)
- **gstack version:** `1.45.0.0`
- **Bun version:** (Bun-bundled with gstack install)
- **gbrain version:** `0.18.2`

## Root cause

`bin/gstack-gbrain-sync.ts:964-971` — the `runBrainSyncPush` function uses:

```ts
spawnSync(brainSyncPath, ["--once"], { stdio: [...], timeout: 60000 })
```

— without `shell: true`.

`gstack-brain-sync` is a bash-shebang script:

```bash
#!/usr/bin/env bash
```

Windows can't `exec` a shebang file directly without going through a shell. So `spawnSync` fails to spawn (returns `result.status === undefined`), and the wrapper interprets that as a stage failure.

## Reproduction

On any Windows machine with gstack installed + gbrain configured:

```bash
# In Git Bash
gstack-gbrain-sync --incremental
```

Expected: stage 3 completes cleanly.
Actual: stage 3 reports `gstack-brain-sync exited undefined`, even though running `gstack-brain-sync --once` directly works (exits 0, empty queue).

## Proposed fix

Add `shell: true` to both `spawnSync` calls in `runBrainSyncPush`, OR wrap with explicit `bash` invocation on Windows:

```ts
// Option A — simple:
spawnSync(brainSyncPath, ["--once"], { stdio: [...], timeout: 60000, shell: true })

// Option B — explicit cross-platform:
const isWindows = process.platform === "win32";
const cmd = isWindows ? "bash" : brainSyncPath;
const args = isWindows ? [brainSyncPath, "--once"] : ["--once"];
spawnSync(cmd, args, { stdio: [...], timeout: 60000 });
```

## Workaround currently in use

`gstack-gbrain-sync --no-brain-sync` to skip stage 3 explicitly. Functional but noisy in the sync summary.

[...]

PR #2386, opened 2026-07-29
fix(gbrain): stop a path case difference from destroying the source index

ShahriarLak:
## The bug

On Windows, `/sync-gbrain` fails on every run, in every mode:

```
ERR   code   source registration failed: gbrain sources remove <id> failed:
⚠️  This will permanently delete: 129 pages, 627 chunks, 627 embeddings
To proceed, pass --confirm-destructive (or use soft-delete: gbrain sources archive <id>)
```

The flag demand is the symptom, not the cause. In `ensureSourceRegistered`, gbrain held the source at `C:/Users/me/projects/quill` while the caller derived `C:/Users/me/Projects/Quill`. Those were compared with `!==`, so the same directory spelled differently was classified as drift, and every run took the remove-then-add path.

**Passing `--confirm-destructive` would be worse than the bug.** It really deletes the source's pages, chunks and embeddings and re-walks from scratch, over a mismatch that was never real. Anyone hitting this and following the error message's advice loses their index.

## The fix

`lib/gbrain-sources.ts` compares paths case-folded with separators normalised. Windows and macOS both default to case-insensitive filesystems, so this is correct there; on Linux a case-distinct pair is possible in principle but gstack never creates one, and re-using an existing index is a much cheaper failure than deleting it.

Two related fixes in the same area:

**`bin/gstack-gbrain-source-wireup`** carries the identical defect, plus a second trigger. Its multi-machine guard parses `sources list --json` with `jq`, so on a machine without jq `existing_path` comes back empty, the guard can never fire, and *every* mismatch falls through to remove-then-add. It now falls back to `sed`, compares normalised, and on a genuine mismatch warns and continues instead of deleting. Its own comment already states that gbrain stores pages by content and `local_path` is informational metadata, which is the argument for never deleting on a path change.

**`bin/gstack-gbrain-sync.ts`** spawned `gstack-brain-sync` with `shell: true` on Windows. The existing comment (#1731) notes Windows can't spawn a shebang script without a shell, but `shell: true` means `cmd.exe`, which can't run an extensionless bash-shebang script either. It reported `'...gstack-brain-sync' is not recognized as an internal or external command` and exited 1 on every run. It now passes the script to `bash`.

## Verification

Windows 11, gbrain 0.42.52.0, gstack 1.60.1.0.

Before: `1 ok, 2 error, 0 skipped` (the two errors above), on every run.

After:

```
  OK    code         synced gstack-code-<repo> (page_count=129) (18.8s)
  OK    memory       gbrain import: 0 imported, 0 unchanged, 0 failed (2.4s)
  OK    brain-sync   curated artifacts pushed (1.7s)

  3 ok, 0 error, 0 skipped
```

No destructive operation attempted, index intact at 129 pages.

## Note on the dry-run

Worth flagging separately: `--dry-run` previews this as an idempotent `gbrain sources add ...` and gives no hint that a real run would attempt a `sources remove` first. That made the dry-run actively misleading while diagnosing this. Not addressed here, since it's a reporting change rather than a correctness one.

### DIFF
--- a/bin/gstack-gbrain-source-wireup
+++ b/bin/gstack-gbrain-source-wireup
@@ -294,18 +294,40 @@ do_wireup() {
       # brain-worktree (same basename, different parent), don't ping-pong the
       # registration. Just sync from our local worktree — gbrain stores pages
       # by content, not by local_path. The metadata is informational only.
-      local existing_path
+      # A path mismatch must never cost the index. This branch used to
+      # remove-then-re-add, which on gbrain >= 0.42 fails outright (the remove
+      # demands --confirm-destructive) and, if "fixed" by passing that flag,
+      # would permanently delete every page/chunk/embedding for the source and
+      # re-walk from scratch. Both outcomes are wrong, because the comment above
+      # is right: gbrain stores pages by CONTENT and local_path is informational
+      # metadata. A changed path does not justify dropping the index.
+      #
+      # Two things made this fire spuriously:
+      #   1. Case. On Windows gbrain may hold C:/Users/me/projects/quill while
+      #      the caller derives C:/Users/me/Projects/Quill — the same directory,
+      #      compared case-sensitively, so "the path changed".
+      #   2. jq. The multi-machine guard below parses `sources list --json` with
+      #      jq. Where jq isn't installed, existing_path comes back EMPTY, the
+      #      guard can never fire, and every mismatch falls through to the
+      #      destructive branch.
+      #
+      # So: normalise before comparing (case-fold + forward slashes), fall back
+      # to sed when jq is missing, and on a genuine mismatch warn and carry on
+      # rather than delete.
+      local existing_path norm_existing norm_worktree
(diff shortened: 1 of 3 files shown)


ShahriarLak:
Closing — upstream fixed this independently and better.

`samePath()` in `lib/gbrain-sources.ts` now realpath-normalises both sides:

```ts
const real = (p: string): string => {
  try { return realpathSync(p) } catch { return p }
}
return real(registered) === real(requested)
```

On Windows `realpathSync` returns the canonical on-disk casing, so
`projects/quill` and `Projects/Quill` compare equal and the drift branch never
fires. That covers the case difference this PR targeted, and handles symlink
aliases too, which my case-folding comparison did not. It is also why the branch
now shows as conflicting — it edits the same function.

Verified on the original repro (Windows 11, gstack 1.61.0.0): a full
`/sync-gbrain --full` completed with the source synced at 142 pages and no
destructive `sources remove`.

Thanks for sorting it. Nothing needed here.

PR #2397, opened 2026-07-30
fix(ios-qa): search front-most content first in the debug bridges

IDSTUK:
## The defect

`ElementsBridgeImpl.snapshot()` walked from `activeKeyWindow`. A presented sheet
sits **after** the screen it covers in `window.subviews`, so `window.subviews`
order is not front-most order — the covered screen is emitted first and the sheet
last. A client taking the first match for a label therefore activates a control
the user cannot reach.

Measured against a real app with a sheet presented:

| Label | Layer | First index |
|---|---|---|
| `Get Ready To Race` | covered screen | 8 |
| `Add` | covered screen | 35 |
| `Create` | **the sheet** | 210 |

This is easy to misread as "sheets are missing from `/elements`" — they are not,
they are just last. It shows up in practice as a label collision resolved by the
wrong layer: a tap aimed at a button in the sheet activates a same-named button
on the screen behind it, and the agent sees a success.

Menus, alerts and action sheets are a second case: each gets its **own
`UIWindow`**, so keying off `isKeyWindow` missed them entirely — absent from
`/elements`, missing from `/screenshot`, and untappable via `/tap`.

## The fix

`ScreenshotBridgeImpl` gains three helpers:

- `orderedWindows(in:)` — visible windows front-most first (`windowLevel`, then
  insertion order)
- `frontmostWindow()` — what the user is actually touching, not merely the key window
- `searchRoots()` — per window, the top-most presented view controller's view
  before the window itself

Then:

- `/elements` walks those roots in order, with an `ObjectIdentifier` seen-set
- `/tap` targets `frontmostWindow()`
- `/type` and `/swipe` search the roots in order
- `/screenshot` composites the windows back-to-front, so a menu is captured
  rather than dropped

The seen-set also fixes long-standing duplication: `collect` visits each view
through both the accessibility-children pass and the subviews pass, so nodes were
emitted several times over. On the app this was measured against, **315 entries
became 69 with nothing lost** — the covered screen is still fully present, just
after the sheet.

Also removes the two now-dead private copies of
`activeScene`/`activeKeyWindow` in `ElementsBridgeImpl` and `MutationBridgeImpl`.

## Verification

- `test/fixtures/ios-qa/FixtureApp` builds for iOS Simulator (`xcodebuild build
  -scheme DebugBridgeUI -destination 'generic/platform=iOS Simulator'`)
- `bun test test/skill-e2e-ios.test.ts` — 7 pass, 1 skip, 0 fail
- The equivalent change runs live in a two-app project on iOS 26.3. After it, the
  same sheet reports `Cancel` at index 0 and `Create` at 2, with the covered
  screen from 34; taps into presented content land on the right layer.

The template and the test fixture are kept byte-identical, as they were before.

## Note on scope

`handleTap` here is coordinate-only via `DebugBridgeTouch`, so upstream has no
label-collision bug in `/tap` itself — the collision surfaces through whatever
the client resolves from `/elements`. The window-scoping half of the fix
(`frontmostWindow`, the screenshot composite) applies to `/tap` and
`/screenshot` regardless.

### DIFF
--- a/ios-qa/templates/Bridges.swift.template
+++ b/ios-qa/templates/Bridges.swift.template
@@ -41,27 +41,73 @@ enum ScreenshotBridgeImpl {
     /// (modern API, replaces UIGraphicsBeginImageContext). Returns nil if
     /// no key window is available (e.g., app backgrounded).
     static func capturePNG() -> Data? {
-        guard let scene = activeScene(), let window = activeKeyWindow(in: scene) else { return nil }
-        let bounds = window.bounds
+        guard let scene = activeScene() else { return nil }
+        let windows = orderedWindows(in: scene)
+        guard let bounds = windows.first?.bounds else { return nil }
         let renderer = UIGraphicsImageRenderer(bounds: bounds)
         let image = renderer.image { _ in
             // drawHierarchy is the documented way to snapshot real UIKit
             // layers including layer-backed views. afterScreenUpdates: false
             // because we want the CURRENT visible state, not a forced layout.
-            window.drawHierarchy(in: bounds, afterScreenUpdates: false)
+            //
+            // Back-to-front across every window: a UIMenu, alert or action
+            // sheet lives in its OWN window, so drawing only the key window
+            // silently drops it from the screenshot.
+            for window in windows.reversed() {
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2406, opened 2026-07-31
fix(gbrain): make --full do a full code walk, not a delta one

ShahriarLak:
`/sync-gbrain --full` does not do a full walk. `runCodeImport()` walks with a bare `gbrain sync --strategy code --source X` — right strategy, but an **incremental** walk. It only revisits files changed since the source's checkpoint, so a file missed at the *original* import is never revisited and stays out of the index indefinitely.

The `reindex-code` pass below cannot rescue it: it re-chunks pages that already exist and never walks the filesystem. That is the same property the comment directly above the call already relies on when explaining why the walk must run first — that fix made a fresh source get pages at all, but left `--full` unable to discover a file the first walk skipped. This PR is the missing half.

### Why it is expensive

The failure is **silent**. Nothing errors, nothing warns, and the verdict block still reports `OK` while `gbrain search` and `gbrain code-def` answer out of a partial index. It presents as "gbrain is weak at code questions" rather than "the index is incomplete", so it survives indefinitely.

### Measured

Two local code sources, before and after, counting exported functions resolvable via `gbrain code-def`:

| source | before | after |
|---|---|---|
| A | 61/201 (30%) | 180/201 (89%) — **79 files had no page at all** |
| B | whole source files missing | 93% |

Both had been serving search from a partial index for weeks.

### Reproduce on your own brain

```
gbrain sync --source <id> --strategy code --full --dry-run
```

Compare `N file(s) would be imported` against that source's `page_count`. To check one file: `gbrain get <file-slug>` returning `page_not_found` means it was never imported.

Worth knowing while testing: the default strategy is **markdown** and `--strategy` is per-invocation, never persisted on the source — drop the flag and the same command reports `strategy=markdown` and a handful of files.

### Scope

Two lines of behaviour change, gated on `args.mode === "full"`, so incremental runs are untouched and stay fast. `--yes` because this is spawned non-interactively and a full walk otherwise prompts to confirm import cost. The failure `summary` now echoes the real argv instead of a hardcoded string that could disagree with what ran.

Verified: file parses under `bun build`; an incremental `--dry-run` confirms the walk does **not** gain `--full`; a `--full` run performs the discovery pass and imports the previously-missing files.

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -895,7 +895,25 @@ async function runCodeImport(args: CliArgs): Promise<StageResult> {
     };
   }
 
-  const walkResult = spawnGbrain(["sync", "--strategy", "code", "--source", sourceId], {
+  // `--full` must do a FULL walk, not a delta one.
+  //
+  // A bare `sync --strategy code` is incremental: it only revisits files that
+  // changed since the source's checkpoint. So a file missed at the ORIGINAL
+  // import is never revisited and stays invisible indefinitely — and the
+  // reindex-code pass below cannot rescue it, because it re-chunks pages that
+  // already exist and never walks the filesystem (the same property the comment
+  // above already relies on).
+  //
+  // The failure is silent: no error, no warning, and the verdict block still
+  // reports OK while `gbrain search` and `gbrain code-def` answer out of a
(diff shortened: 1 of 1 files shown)


garrytan:

Issue #2250, opened 2026-07-13
setup-gbrain calls bin/gstack-memory-ingest & bin/gstack-gbrain-sync without .ts extension → "script missing at install root" on every host


inside-ziwu:
## Summary

`/setup-gbrain`'s transcript-ingest steps call `bin/gstack-memory-ingest` and `bin/gstack-gbrain-sync` **without the `.ts` extension**. Neither bare name exists in `bin/` — only `gstack-memory-ingest.ts` and `gstack-gbrain-sync.ts` ship. The agent dutifully reports the script as missing from the gstack install root and the ingest/full-sync steps dead-end. Both files are also mode `644` (no `+x`), so even a correctly-spelled direct exec would fail without `bun run`.

Hit this live running `/gstack-setup-gbrain` under Codex; the Claude render has the same references, so every host is affected.

## Affected call sites

`setup-gbrain/SKILL.md.tmpl` (canonical source, confirmed present on `main` today):

- Line 690: `~/.claude/skills/gstack/bin/gstack-memory-ingest --probe`
- Line 735: `~/.claude/skills/gstack/bin/gstack-gbrain-sync --full --no-brain-sync`

Rendered copies inherit it: Claude render `setup-gbrain/SKILL.md:1444,1489`; Codex render `gstack-setup-gbrain/SKILL.md:1432,1477` (as `$GSTACK_ROOT/bin/...`).

## Repro

```bash
ls bin/ | grep -E 'memory-ingest|gbrain-sync'
# gstack-gbrain-sync.ts
# gstack-memory-ingest.ts        ← no bare names

grep -n 'bin/gstack-memory-ingest[^.]\|bin/gstack-gbrain-sync[^.]' setup-gbrain/SKILL.md.tmpl
# 690: ... bin/gstack-memory-ingest --probe
# 735: ... bin/gstack-gbrain-sync --full --no-brain-sync

~/.claude/skills/gstack/bin/gstack-memory-ingest --probe
# no such file or directory
```

## Expected

The repo already has the correct pattern one skill over — `sync-gbrain/SKILL.md` invokes the same orchestrator as:

```bash
bun run ~/.claude/skills/gstack/bin/gstack-gbrain-sync.ts <user-args>
```

## Suggested fix (either works)

1. Change both call sites to `bun run .../gstack-memory-ingest.ts --probe` / `bun run .../gstack-gbrain-sync.ts --full --no-brain-sync`, matching `sync-gbrain`.
[...]

PR #2409, opened 2026-07-31
fix(setup-gbrain): correct stale gstack-memory-ingest/gstack-gbrain-sync invocation paths

### LINKED ISSUE #2393: setup-gbrain SKILL.md:1445 gives a gstack-memory-ingest path that does not exist
**Component:** `setup-gbrain/SKILL.md` (gstack 1.60.1.0)

`setup-gbrain/SKILL.md:1445` instructs the agent to run:

```bash
~/.claude/skills/gstack/bin/gstack-memory-ingest --probe
```

That path does not exist on a stock install. `bin/` contains `gstack-memory-ingest.ts`, not
an executable of that name, so the command fails with
`timeout: failed to execute process: No such file or directory`.

`setup-gbrain/memory.md:64` gives the working form:

```bash
bun run bin/gstack-memory-ingest.ts --probe
```

Two documents in the same skill directory disagree, and the one the agent is told to follow
during Step 7.5 is the broken one. Step 7.5 branches on the probe's output, so the failure
lands in the middle of an unattended decision.

SomSamantray:
## What

`setup-gbrain`'s docs told the agent to run `gstack-memory-ingest` and `gstack-gbrain-sync` as bare command names in 8 places (4 in `SKILL.md.tmpl`, 4 in `memory.md`).

## Why

Both are `.ts` files with no `package.json` bin alias stripping the extension, so the bare form fails with "No such file or directory" the moment an agent follows the doc literally — Step 7.5 branches on the probe command's output, so this breaks mid-flow during an unattended decision.

## How

- Corrected all 8 sites to `bun run <path>/<name>.ts <flags>`, matching how `bin/gstack-gbrain-sync.ts` already invokes the ingest script internally, and matching `memory.md`'s own pre-existing correct line.
- Regenerated `setup-gbrain/SKILL.md` from the fixed template (separate commit, per this repo's own bisection convention).
- Added `test/setup-gbrain-bin-invocation-paths.test.ts`, a grep-based regression test (same pattern as `test/setup-gbrain-path4-structure.test.ts`) that fails if a bare invocation reappears, including across a backslash line-continuation.

Fixes #2393, #2250.

### DIFF
--- a/setup-gbrain/SKILL.md
+++ b/setup-gbrain/SKILL.md
@@ -1441,7 +1441,7 @@ curated `~/.gstack/` artifacts into gbrain so the retrieval surface
 
 Run the probe to size the operation:
 ```bash
-~/.claude/skills/gstack/bin/gstack-memory-ingest --probe
+bun run ~/.claude/skills/gstack/bin/gstack-memory-ingest.ts --probe
 ```
 
 Read the output. If `Total files in window: 0`, skip — there's nothing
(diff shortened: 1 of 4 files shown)


garrytan:

PR #2411, opened 2026-07-31
fix(claude): use host invocation for auth detection

xing-qnex:
## Summary

Fix the generated `gstack-claude` skill so Codex and other sandboxed hosts use the real host Claude Code invocation as the authentication check.

## Root cause

The skill previously inferred authentication from `$HOME/.claude/.credentials.json` or `ANTHROPIC_API_KEY`. That is unreliable when Claude Code uses the macOS Keychain and the calling agent's sandbox cannot see host credentials. The temp-file templates also appended suffixes after `XXXXXX`, which is not accepted by macOS `mktemp`.

## Changes

- Resolve `claude` and invoke the same binary in the host execution context.
- Do not report an auth blocker until the actual `claude -p` invocation returns an auth/login error.
- Keep review/challenge calls tool-less with `--disable-slash-commands --tools ""`.
- Use macOS-compatible `mktemp` templates with `XXXXXX` at the end.
- Add deterministic generated-skill and real host `mktemp` regression tests.

## Validation

- `bun test test/gen-skill-docs.test.ts`: 403 passed, 0 failed
- macOS smoke for prompt/response/error/diff temp templates: passed
- `git diff --check`: passed

The full repository suite was also sampled, but it contains unrelated environment-dependent failures involving gbrain fixtures, local ports, `~/.gstack` sandbox writes, and SwiftPM. This PR's targeted generation suite is green.

### DIFF
--- a/claude/SKILL.md.tmpl
+++ b/claude/SKILL.md.tmpl
@@ -32,7 +32,7 @@ The generated external invocation name is `gstack-claude`.
 
 ---
 
-## Step 0: Check Claude CLI
+## Step 0: Resolve Claude CLI
 
 ```bash
 CLAUDE_BIN=$(command -v claude 2>/dev/null || echo "")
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2414, opened 2026-07-31
Fix Windows terminal-agent leak: one orphaned process + console window per watchdog tick

SYKhayyat:
## Summary

On Windows, the browse daemon leaks one `bun run terminal-agent.ts` process and one **visible console window** every 60 seconds, without bound, until the host runs out of committable memory.

Found on a 48GB Windows 11 machine: 7 orphaned agents and climbing, a terminal stealing foreground focus once a minute, and 0.5GB free RAM against a 69.4GB/70.4GB commit charge. The daemon's parent session had already exited, so nothing owned the orphan.

Three bugs had to line up. Fixing any one alone leaves the others latent, so this fixes all three.

### 1. `isProcessAlive()` reported live processes as dead

`browse/src/error-handling.ts` shelled out to `tasklist /FI "PID eq <pid>"` with a 3000ms timeout.

**A `Bun.spawnSync` that hits its `timeout` still returns**, carrying partial stdout. The `.includes()` PID match then comes back `false` — a live process reported dead. Demonstrated:

```
timed-out spawnSync stdout length: 93 -> includes pid match: false
```

Measured `tasklist` latency was **685–1665ms on an idle box**, and considerably worse under memory pressure. The 3s timeout was reachable in ordinary use, not just pathological cases.

### 2. The false negative orphaned the agent instead of killing it

`killAgentByRecord()` validates liveness before signalling, so a false negative made it skip the kill and return `false` — while the watchdog respawned regardless. The old agent survived, unreferenced.

This is self-reinforcing: every orphan adds memory pressure, which slows the next `tasklist`, which produces the next false negative.

### 3. The crash-loop guard could not fire

```js
const AGENT_WATCHDOG_TICK_MS = 60000;   // tick
const RESPAWN_GUARD_WINDOW_MS = 60_000; // window
const RESPAWN_GUARD_MAX = 3;            // trip threshold
```

The window equals the tick, so **at most one respawn can ever sit inside it** and `respawnHistory.length >= 3` is unreachable. The guard was structurally incapable of firing at the default tick rate, which is why a steady one-per-tick leak ran forever instead of stopping after 3.

## Fixes

| Bug | Fix |
|---|---|
| 1 | `isProcessAlive()` uses signal 0 on every platform. Node and Bun both map `process.kill(pid, 0)` to an `OpenProcess` existence check on Windows, so the POSIX idiom is portable. `EPERM` is treated as alive (process exists, we just can't signal it) — returning `false` there would reintroduce the original failure. |
| 2 | Fixed by 1 — the kill now actually happens. |
| 3 | `RESPAWN_GUARD_WINDOW_MS` scales with the tick, so "3 respawns in quick succession stops the loop" holds at any tick value. |
| — | `spawnTerminalAgent()` passes `windowsHide: true`. The agent is a background daemon with all three stdio streams already ignored, so it has nothing to display. No-op on macOS/Linux. |

Signal 0 measured against the old probe on the affected machine:

```
signal-0 avg: 0.0036 ms
tasklist    : 270.1 ms   (~74,000x slower)
```

It also spawns nothing and cannot time out, which removes the failure mode rather than widening the timeout.

## Second commit: the static tripwires never ran on Windows

I only found bug 3 because `terminal-agent-watchdog.test.ts` asserted the literal `RESPAWN_GUARD_WINDOW_MS = 60_000` — pinning the bug as an invariant. Going to update it surfaced a larger problem.

14 test files locate themselves via:

```js
path.resolve(new URL(import.meta.url).pathname, '..', '..', 'src')
```

On Windows `new URL(...).pathname` is `/C:/Users/...` (leading slash), so `path.resolve` prepends the cwd's drive and yields `C:\C:\Users\...`. Every one of those tripwires threw `ENOENT` instead of asserting anything.

```
before: 12 pass / 80 fail
after:  92 pass /  0 fail
```

The inert tripwires included the `pkill`-teardown regression guard, CDP session cleanup, SSE subscriber wiring, and the PTY lease routes — precisely the invariants meant to fail CI on regression.

CI stayed green because `windows-free-tests.yml` runs a curated 13-file allowlist that excludes all of them; the workflow's own comments already track full Windows parity as a follow-up. This isn't a surprise so much as a gap that had grown teeth.

Fix is `import.meta.path`, Bun's absolute native path for the current file — a drop-in with identical arity, so every `'..'` segment count is unchanged.

## Testing

- New `browse/test/process-liveness-windows.test.ts` (6 tests) pins all three bugs, including an assertion that the liveness probe **spawns no subprocess at all** — the property that made the original implementation timeout-prone in the first place.
- `terminal-agent-watchdog.test.ts` test 4 now pins the window's *relationship to the tick* rather than the literal that let the bug ship.
- 100 pass / 0 fail across the 16 affected files on Windows 11 + Bun 1.3.14.
- Both commits verified independently green (bisectable).
- `browse/dist/` untouched (gitignored locally); rebuilt `server-node.mjs` only to confirm the fixes reach the bundle.

Full-directory `bun test browse/test/` still aborts early here on `batch.test.ts`, which needs a live Chromium — unrelated to these changes and present at clean `HEAD` too.

## Notes

- **No `VERSION` / `CHANGELOG` bump.** Per `CLAUDE.md` those happen at `/ship` time, and an external branch guessing a version is a guaranteed conflict. Happy to add an entry in whatever voice/version you want.
- Bug 1's fix changes `isProcessAlive` on **all** platforms, not just Windows. The macOS/Linux behaviour was already `process.kill(pid, 0)`; the only semantic change there is that `EPERM` now reports alive instead of dead, which is the correct answer and matters for cross-user PID checks.
- Reported and fixed on Windows 11 26200, Bun 1.3.14, gstack v1.58.4.0.

### DIFF
--- a/browse/src/error-handling.ts
+++ b/browse/src/error-handling.ts
@@ -7,8 +7,6 @@
 
 import * as fs from 'fs';
 
-const IS_WINDOWS = process.platform === 'win32';
-
 // ─── Filesystem ────────────────────────────────────────────────
 
 /** Remove a file, ignoring ENOENT (already gone). Rethrows other errors. */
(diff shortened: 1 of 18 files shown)


garrytan:

PR #2417, opened 2026-07-31
fix(gbrain): respect existing source pins

exGeni:
## Problem

`/sync-gbrain` always derives a new worktree-scoped source ID, even when the
repository already has a valid `.gbrain-source` pin created through the native
GBrain source workflow. This silently bypasses the selected source boundary,
registers a second federated source, and routes later dream/cycle checks to the
wrong source.

## Fix

- Reuse a local pin only when its ID is valid, it is registered, and its
  registered path resolves to the current checkout.
- Treat a confirmed pin as user-managed: sync and attach it without add/remove,
  legacy migration, or federation changes.
- Keep dry-run spawn-free by reading only the local marker for previews.
- Route explicit dream and full-cycle checks through the same resolver and the
  config-backed GBrain environment.
- Fall back to the existing generated source ID for missing, invalid, stale, or
  unreadable pins.

This is complementary to #2386. That PR protects path-spelling drift inside
source registration; this PR prevents a valid pre-existing pin from reaching
registration or generated-ID migration in the first place.

## Verification

- Red test against the old implementation: 37 passed, 1 failed; the old code
  planned a generated federated source instead of the pinned source.
- Targeted suites: 75 passed, 0 failed.
- Full free suite: passed outside the filesystem sandbox. The first sandboxed
  run had four unrelated EROFS failures because worktree tests write under
  `~/.gstack-dev`; the unrestricted rerun passed.
- Live disposable-repository dry-run selected the existing non-federated pin
  and planned no source add or federation change.

Regression coverage includes spawn-free dry-run, symlink-equivalent registered
paths, non-dry-run sync/attach with no add/remove, dream routing, unreadable
markers, and config-backed environment use.

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -29,7 +29,7 @@
  * than building a gstack-side daemon.
  */
 
-import { existsSync, statSync, mkdirSync, writeFileSync, readFileSync, unlinkSync, renameSync } from "fs";
+import { existsSync, statSync, mkdirSync, writeFileSync, readFileSync, unlinkSync, renameSync, realpathSync } from "fs";
 import { join, dirname } from "path";
 import { execSync, spawnSync } from "child_process";
 import { homedir, hostname } from "os";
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2419, opened 2026-07-31
v1.60.1.1 feat: allow Emulation.setEmulatedMedia on the CDP allowlist

meshailabs:
## The gap

browse cannot test any CSS media-feature branch today. Under the deny-default CDP allowlist, `Emulation.setEmulatedMedia` 403s:

```
$ browse cdp Emulation.setEmulatedMedia '{"features":[{"name":"prefers-reduced-motion","value":"reduce"}]}'
Cause: deny-default posture; method has not been audited and added to cdp-allowlist.ts.
```

That makes `prefers-reduced-motion`, `prefers-contrast`, `prefers-color-scheme`, and `forced-colors` all unreachable. `/qa` and `/design-review` can see that a reduced-motion rule exists in the stylesheet, but cannot prove it fires. For skills whose stated job is live visual and accessibility QA, that is a real hole.

I hit it QA'ing a site whose CRT scanline overlay is suppressed under `prefers-reduced-motion`. There was no way to verify the suppression short of adding the method by hand and rebuilding.

## The change

One allowlist entry, scoped `tab` / `trusted`, matching the three `Emulation` entries already present (`setDeviceMetricsOverride`, `clearDeviceMetricsOverride`, `setUserAgentOverride`).

Why `trusted` is right here: the method returns an empty result. It emits no page content, so there is no marker-injection or exfil vector to wrap. Why `tab` is right: it mutates only the active tab's render state, exactly like `setDeviceMetricsOverride`.

The justification notes the one sharp edge, mirroring the existing `setUserAgentOverride` note: the override persists on the tab until cleared with an empty `features` array.

## Verification

| Gate | Result |
|---|---|
| `browse/test/cdp-allowlist.test.ts` (the 4-field CI lint) | 6 pass, 192 assertions |
| `browse/test/adversarial-security.test.ts` | 2 pass |
| `bun run gen:skill-docs --dry-run` (skill-docs CI) | 0 stale |
| `bun run build` | no generated drift; only this file changes |

Working end to end after a rebuild, against a live site:

| Emulated state | `body::after` display |
|---|---|
| `prefers-reduced-motion: reduce` | `none` |
| `prefers-contrast: more` | `none` |
| reset (empty `features`) | `block` |

Every gate above was run per-file on purpose. A bare `bun test` exits 0 after roughly 23 of 409 files regardless of failures (#2421), so a full-suite run is not evidence of anything here. Per-file runs return a real exit code and a summary line.

## Pre-existing failures, unrelated to this PR

`main` is currently red: **87 test failures across four `browse/test/sidebar-*.test.ts` files**, from the v1.14.0.0 sidebar rip (#1216) removing endpoints and `browse/src/sidebar-agent.ts` without removing their tests. Filed as #2420 with the full root cause.

None of it is from this PR. I verified by running the affected files with this change applied and on a pristine `main` checkout: identical results either way (`sidebar-integration.test.ts` = 2 pass / 11 fail in both).

Correction to an earlier revision of this description: I first wrote "11 failures", having only seen the tail of the output. The true figure is 87 across four files.

## Notes

- Branched from `a325940` (current `main`), not from my fork's stale default branch.
- VERSION bumped to `1.60.1.1` via `bin/gstack-next-version --bump micro`, which reported `no collision; clean bump from base` against the live queue.
- CHANGELOG entry follows the short `### Added` form used by other micro releases (e.g. `0.18.0.1`).


garrytan:

PR #2423, opened 2026-08-01
fix(bin): PreToolUse hook schema in gstack-team-init required mode (#2413)

### LINKED ISSUE #2413: gstack-team-init required writes a PreToolUse hook that never blocks
**Version:** gstack 1.60.1.0
**File:** `bin/gstack-team-init:130`

## Summary

The hook generated by `gstack-team-init required` emits a permission payload
Claude Code does not recognise, then exits 0. Unrecognised JSON on exit 0 is a
non-blocking error, so the `Skill` call proceeds and `required` mode enforces
nothing. Every repo bootstrapped this way has a gate that only *looks* like a
gate: the user sees the BLOCKED text on stderr and is waved straight through.

## Generated code

`.claude/hooks/check-gstack.sh`, tail:

```bash
echo '{"permissionDecision":"deny","message":"gstack is required but not installed. See stderr for install instructions."}'
exit 0
```

## Why it fails

Per the [hooks reference](https://code.claude.com/docs/en/hooks), a PreToolUse
hook must nest the decision inside `hookSpecificOutput`:

```json
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "deny",
    "permissionDecisionReason": "..."
  }
}
```

Two mismatches: `permissionDecision` is top-level rather than nested, and
`message` is not a field (the key is `permissionDecisionReason`). Neither the
current nested schema nor the deprecated top-level `{"decision":"block"}` form
matches what is emitted, so nothing is parsed. JSON is only processed on exit
0, and an exit 0 whose JSON does not validate is treated as a non-blocking
error — the tool call runs.

## Reproduce

```bash
# in a repo bootstrapped with: gstack-team-init required
HOME=/tmp/no-gstack bash .claude/hooks/check-gstack.sh; echo "exit: $?"
```

Observed: the BLOCKED text on stderr, the payload above on stdout, `exit: 0`.
Then invoke any skill on a machine without `~/.claude/skills/gstack/bin` — it
runs.

## Suggested fix

Nest the payload and exit 2. Exit 2 blocks a PreToolUse call even when the JSON
fails schema validation, so the gate holds on older and newer Claude Code
alike:

```bash
  cat <<'JSON'
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"gstack is required but not installed. See stderr for install instructions."}}
JSON
  exit 2
```

Verified locally: absent gstack now exits 2 with schema-valid JSON and stderr
intact; present gstack still emits `{}` and exits 0.

## Two things worth deciding alongside the fix

1. **Blast radius.** The matcher is `"Skill"` with no name filter, so once the
   gate genuinely blocks it blocks *every* skill, including Claude Code's own
   built-ins, not just `gstack-*`. That is a stronger claim than "gstack is
   required" and may warrant filtering on the skill name.
2. **Windows.** The hook is a `.sh` invoked as a bare command. Claude Code
   falls back to PowerShell when Git Bash is absent, so on those machines the
   hook cannot execute and fails open regardless of this fix. Worth either
   documenting Git Bash as a hard requirement of `--team` or shipping a `.ps1`
   alongside.

Masashi-Ono0611:
## What was broken (plain English)
`gstack-team-init required` writes a Claude Code hook meant to block AI-assisted work when gstack isn't installed globally. The hook prints the right warning to the terminal, but the machine-readable part of its answer was in a shape Claude Code doesn't recognize, and it said "everything's fine" (exit code 0) even on the deny path. So the block never actually happened — the warning showed up, then the tool call ran anyway.

## How it was fixed (plain English)
The hook's answer is now wrapped in the shape Claude Code actually expects, and it exits with the code that means "actually stop this." The allow path (when gstack is installed) is untouched.

Fixes #2413

## Technical detail
`.claude/hooks/check-gstack.sh` (generated by `bin/gstack-team-init required`) emitted:
```
{"permissionDecision":"deny","message":"..."}
exit 0
```
Per the [hooks reference](https://code.claude.com/docs/en/hooks), a PreToolUse hook must nest the decision inside `hookSpecificOutput`, and JSON is only parsed on exit 0 — an exit-0 JSON payload that doesn't match the expected schema is treated as a non-blocking error, so the tool call proceeds regardless. Fix:
```
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"..."}}
exit 2
```
Exit code 2 blocks the call even if a future schema change makes the JSON invalid again, so the gate holds defensively.

## Verification
- Manual repro matching the issue's exact steps: `HOME=/tmp/no-gstack bash .claude/hooks/check-gstack.sh; echo "exit: $?"` now prints the nested JSON and `exit: 2` (was `exit: 0` before this fix, confirmed against `main`).
- gstack-present path still returns `{}` / `exit: 0` (unchanged, verified).
- Added `test/gstack-team-init-hook-schema.test.ts`: generates the hook via `gstack-team-init required` in a temp repo and asserts both the absent-gstack (exit 2, schema-valid deny) and present-gstack (exit 0, `{}`) branches.
- `bun test` full suite run on this branch and compared against `main`: no new failures introduced by this change (pre-existing failures — Playwright browser binary not installed, gbrain/Xcode/auth-dependent tests, and sidebar tests referencing already-ripped functionality — are identical/superset on `main`).

## Scope note
The same flat, non-nested `permissionDecision` pattern this issue reports also exists in `freeze/bin/check-freeze.sh` and `careful/bin/check-careful.sh` (both currently rely on implicit exit 0 via `set -euo pipefail`). Left out of this PR to keep the diff scoped to the exact file/line #2413 reports — happy to send a follow-up if that's wanted.

### DIFF
--- a/bin/gstack-team-init
+++ b/bin/gstack-team-init
@@ -127,8 +127,8 @@ Install it:
 
 Then restart your AI coding tool.
 MSG
-  echo '{"permissionDecision":"deny","message":"gstack is required but not installed. See stderr for install instructions."}'
-  exit 0
+  echo '{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"gstack is required but not installed. See stderr for install instructions."}}'
+  exit 2
 fi
 
 echo '{}'
--- a/test/gstack-team-init-hook-schema.test.ts
+++ b/test/gstack-team-init-hook-schema.test.ts
@@ -1,0 +1,81 @@
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { execSync, execFileSync } from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+
+// Regression guard for #2413: gstack-team-init required generated a
+// PreToolUse hook that emitted a flat {"permissionDecision":...} payload and
+// exited 0. Claude Code only nests decisions under hookSpecificOutput, and
+// only exit code 2 reliably blocks a PreToolUse call. The old shape was
+// silently ignored as a non-blocking error, so `required` mode enforced
+// nothing — the BLOCKED text printed to stderr but the tool call proceeded.
+
+const ROOT = path.resolve(import.meta.dir, '..');
+const TEAM_INIT = path.join(ROOT, 'bin', 'gstack-team-init');
+
+let repoDir: string;
+let fakeHomeAbsent: string;
+let fakeHomePresent: string;
+
+describe('gstack-team-init required: PreToolUse hook schema (#2413)', () => {
+  beforeEach(() => {
+    repoDir = fs.mkdtempSync(path.join(os.tmpdir(), 'gstack-team-init-repo-'));
+    execFileSync('git', ['init', '-q'], { cwd: repoDir });
+    execFileSync(TEAM_INIT, ['required'], { cwd: repoDir, encoding: 'utf-8' });
+
+    fakeHomeAbsent = fs.mkdtempSync(path.join(os.tmpdir(), 'gstack-team-init-home-absent-'));
+
+    fakeHomePresent = fs.mkdtempSync(path.join(os.tmpdir(), 'gstack-team-init-home-present-'));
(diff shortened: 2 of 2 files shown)


garrytan:

PR #2426, opened 2026-08-01
fix(careful): three bypasses in check-careful.sh (quoted args, rm -R, ${IFS})

wtamminga:
Three ways a destructive command currently reaches the user unwarned through `/careful`'s PreToolUse hook. Each is one commit with tests.

I found these while auditing my own pinned copy of gstack, then confirmed all three reproduce on `main` at `a325940`.

## 1. A quoted argument hides everything after it

`check-careful.sh` extracted the command with

```
grep -o '"command"[[:space:]]*:[[:space:]]*"[^"]*"'
```

`[^"]*` stops at the first escaped quote inside the JSON string value, so anything after a quoted argument was truncated away before the pattern checks ran:

| command | reaches the checks as | result |
|---|---|---|
| `git commit -m "wip" && rm -rf /` | `git commit -m \` | allowed |
| `bash -c "rm -rf /"` | `bash -c \` | allowed |
| `echo "x"; rm -rf ~` | `echo \` | allowed |

The `python3` fallback couldn't rescue these — it only ran when `CMD` was empty, and truncation leaves `CMD` non-empty.

This one is already visible in the repo. `test/hook-scripts.test.ts` carried a note on the SQL block explaining that `psql -c "DROP TABLE users"` couldn't be used in a test because the extractor truncated it, so those cases were written without quotes. The workaround was documented; the security consequence was that the SQL checks were also being defeated by any realistic quoted invocation.

Fixed by parsing the payload with `python3` (falling back to `node`) and **failing closed** when it can't be parsed — a hook whose job is gating destructive commands shouldn't allow-by-default on unreadable input. A well-formed payload with no `command`, or a non-string `command`, still allows.

## 2. `rm -R` is not matched

`rm` accepts `-R` as an exact synonym for `-r`. The destructive check matched `-[a-zA-Z]*r`, lowercase only, so `rm -R /`, `rm -Rf ~` and `rm -fR /` all classified as non-recursive.

Widened to `[rR]` in both the destructive check and the build-artifact safe exception — doing both keeps `rm -Rf node_modules` allowed instead of turning it into a false positive.

## 3. `${IFS}` sidesteps every whitespace pattern

Every check inspects the command as a string, but bash executes what the string *means* after expansion. `${IFS}` contains no literal whitespace, so

```
rm${IFS}-rf${IFS}/
```

matches none of the `rm\s+` patterns while executing as a full recursive delete.

Rather than try to out-parse bash, this adds a small tripwire that asks when it sees `${IFS}`/`$IFS` word-splitting or a base64 decode piped into a shell. Those are rare in commands anyone means to run unattended, so the false-positive cost is low — and a decode that *doesn't* feed a shell (`cat file.b64 | base64 -d > out.bin`) still allows.

## Testing

Added cases to `test/hook-scripts.test.ts` alongside the existing ones, and converted the SQL tests to the realistic quoted forms now that they work.

I verified the script directly by piping hook payloads into `bash careful/bin/check-careful.sh` — 15 must-ask and 8 must-allow cases, all passing, including the pre-existing safe-exception and compound-command cases so nothing regressed. **I did not run the full `bun test` suite**, since I was working from a scratch clone under a policy of not executing a third-party repo wholesale; the new tests are written to the file's existing `runHook`/`runHookRaw` helpers and conventions but have not been executed by the bun runner. Worth a CI run before merge.

## Notes

- Two files touched: `careful/bin/check-careful.sh` and `test/hook-scripts.test.ts`. No generated `SKILL.md` or templates involved.
- Three commits, bisected one-logical-change-each per CONTRIBUTING/CLAUDE.md.
- `#2` and `#3` are independent of `#1` and can be taken separately if you'd rather split them.
- None of this touches ETHOS.md, promotional material, or voice.

### DIFF
--- a/careful/bin/check-careful.sh
+++ b/careful/bin/check-careful.sh
@@ -56,6 +56,23 @@ fi
 # Normalize: lowercase for case-insensitive SQL matching
 CMD_LOWER=$(printf '%s' "$CMD" | tr '[:upper:]' '[:lower:]')
 
+# --- Shell-obfuscation tripwire ---
+# Every check below inspects the command as a STRING, but bash executes what the
+# string MEANS after expansion. ${IFS} holds the default field separator and
+# contains no literal whitespace, so
+#
+#   rm${IFS}-rf${IFS}/
+#
+# matches none of the `rm\s+` patterns while executing as a full recursive
+# delete. The same holds for a command assembled by a base64 decode piped to a
+# shell. Rather than try to out-parse bash, treat these splitting/decoding
+# primitives as a reason to ask: they are vanishingly rare in commands a human
+# actually means to run unattended.
+if printf '%s' "$CMD" | grep -qE '\$\{IFS\}|\$IFS|\$\(echo[^)]*base64[^)]*\)|base64[[:space:]]+(-d|--decode)[^|]*\|[[:space:]]*(sh|bash)' 2>/dev/null; then
+  printf '{"permissionDecision":"ask","message":"[careful] Shell obfuscation detected (IFS word-splitting or base64-to-shell). Read the command carefully before approving."}\n'
+  exit 0
+fi
+
 # --- Check for safe exceptions (one standalone rm of build artifacts) ---
 # Match the complete command. Parsing only the last rm is unsafe because shell
 # syntax or comments can hide an earlier destructive command, for example:
--- a/careful/bin/check-careful.sh
+++ b/careful/bin/check-careful.sh
@@ -61,7 +61,7 @@ CMD_LOWER=$(printf '%s' "$CMD" | tr '[:upper:]' '[:lower:]')
 # syntax or comments can hide an earlier destructive command, for example:
 #   rm -rf / # rm -rf node_modules
 # Unknown syntax fails closed and falls through to the destructive checks.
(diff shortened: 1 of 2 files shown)


wtamminga:
Follow-up datapoint, no rush on review.

I've now moved my own install from `dc6252d` to `a325940` (this PR's base) and re-applied these three fixes on top. Two things worth reporting back:

1. **The fixes hold at HEAD.** Against `a325940` + this patch, a 25-case corpus passes: 16 must-ask (including `git commit -m "wip" && rm -rf /`, `bash -c "rm -rf /"`, `rm -R /`, `rm -fR /`, `rm${IFS}-rf${IFS}/`, `rm -rf / # rm -rf node_modules`) and 8 must-allow still returning `{}` (`rm -rf node_modules`, `rm -rf ./dist`, `rm -rf packages/web/node_modules`, `ls -la`, `npm install`, `git status`). No false positives. An unparseable payload returns `ask` rather than allow.

2. **Unrelated find while repinning — `investigate/SKILL.md` still uses `${CLAUDE_SKILL_DIR}`.** `9cc41b7` re-anchored `careful`, `freeze`, and `guard` to `$HOME`, but `investigate` was missed. It has two `PreToolUse` hook commands (lines 26 and 31) plus a freeze-availability probe (lines 913-914) still on the old variable. Because those hooks are written as `[ -x "$S" ] && bash "$S" || exit 0`, an empty variable means the path never resolves and the hook silently exits 0 — so `investigate`'s freeze boundary is inert on Claude Code 2.1.162+ rather than failing loudly. Happy to send that as a separate one-line PR if you'd like it split out; I didn't want to widen the scope of this one.

Not blocking on either — just closing the loop since I'm now running this patch in anger.

garrytan:

PR #2429, opened 2026-08-01
fix(safety): unknown source exits 1 instead of documented 2 (#2390)

### LINKED ISSUE #2390: gstack-question-preference: user-origin rejection exits 1 (not the documented 2) for any source outside the hardcoded REJECTED_SOURCES list
**Repo:** garrytan/gstack · **Version:** v1.60.1.0 (`a325940`) · **Host:** Claude Code 2.1.186, macOS

## Summary

~30 `SKILL.md` files carry this instruction in the Question Tuning section:

> Exit code 2 = rejected as not user-originated; do not retry.

and `CHANGELOG.md` states the guarantee explicitly:

> The binary enforces this with exit code 2 for rejected writes.

`bin/gstack-question-preference` only exits `2` for **four literal source
strings**. Any other value that isn't user-originated exits `1` — the same code
as malformed JSON, a missing field, or an invalid `preference`. The caller
therefore cannot distinguish *"poisoning attempt blocked"* from *"I sent a bad
payload"*, which is exactly the distinction the documented contract exists to
make.

The write is refused either way, so this is a **contract/signal bug, not a
security hole** — it fails closed.

## Reproduction

```console
$ export GSTACK_HOME=$(mktemp -d)   # sandbox, don't touch real prefs

# in the denylist → documented behaviour
$ gstack-question-preference --write '{"question_id":"x","preference":"never-ask","source":"inline-tool-output"}'
gstack-question-preference: rejected — source "inline-tool-output" is not user-originated (profile poisoning defense)
$ echo $?
2

# NOT in the denylist, equally not user-originated → undocumented behaviour
$ gstack-question-preference --write '{"question_id":"x","preference":"never-ask","source":"tool-output"}'
gstack-question-preference: invalid source "tool-output"; allowed: plan-tune, inline-user
$ echo $?
1
```

Same `1` for every structural error, so `1` carries no information:

| payload | exit |
|---|---|
| `source: "inline-tool-output"` (in denylist) | **2** |
| `source: "tool-output"` | 1 |
| `source: "web"` / `"pr-body"` / any other non-allowed value | 1 |
| `source` field missing | 1 |
| malformed JSON | 1 |
| `preference: "talvez"` | 1 |
| valid `inline-user` write | 0 |

## Root cause

`bin/gstack-question-preference`, lines 147-160 — a denylist decides the exit
code, but the actual gate two lines below is an allowlist:

```js
const ALLOWED_SOURCES = ['plan-tune', 'inline-user'];
const REJECTED_SOURCES = ['inline-tool-output', 'inline-file', 'inline-file-content', 'inline-unknown'];
if (!j.source) {
  ...
  process.exit(1);
}
if (REJECTED_SOURCES.includes(j.source)) {
  ...
  process.exit(2);          // ← only these four strings
}
if (!ALLOWED_SOURCES.includes(j.source)) {
  ...
  process.exit(1);          // ← everything else that is also not user-originated
}
```

Anything the caller labels with a source name not on the four-item denylist —
a new label added by a future skill, a host-specific one, or a typo of an
intended value — is silently demoted from "blocked injection" to "generic
error".

The bash wrapper is fine: it propagates the node exit code verbatim
(`exit $RC`, line 209).

## Impact

- An agent following the `SKILL.md` instruction keys "do not retry" to exit 2.
  On exit 1 it may retry the write, or surface it to the user as a tooling bug
  rather than as a blocked injection attempt.
- Any monitoring or test keyed on exit 2 undercounts rejected writes.
- The denylist has to be kept in sync with every source label anyone ever
  introduces, and forgetting one degrades the signal silently.

## Suggested fix

`ALLOWED_SOURCES` is already the security gate, so the denylist is redundant
for the decision — only for the wording of the message. Exit 2 whenever
`source` is present but not allowed, and keep 1 for structural errors:

```js
if (!j.source) {
  process.stderr.write('gstack-question-preference: source field required (one of: ' + ALLOWED_SOURCES.join(', ') + ')\n');
  process.exit(1);                                    // caller bug → 1
}
if (!ALLOWED_SOURCES.includes(j.source)) {
  const known = REJECTED_SOURCES.includes(j.source);
  process.stderr.write(
    'gstack-question-preference: rejected — source "' + j.source + '" is not user-originated' +
    (known ? ' (profile poisoning defense)' : '; allowed: ' + ALLOWED_SOURCES.join(', ')) + '\n',
  );
  process.exit(2);                                    // not user-originated → 2
}
```

That makes the code match the ~30 SKILL.md files and the CHANGELOG without
weakening anything: the set of accepted writes is unchanged.

Happy to send a PR if the shape above is what you'd want.

gregario:
## Summary

Fixes #2390 — the user-origin gate in `gstack-question-preference --write` exits with code 1 (validation error) instead of code 2 (poisoning rejection) when a source value is not in ALLOWED_SOURCES but also not in the hardcoded REJECTED_SOURCES list.

The gate has three branches: missing source (exit 1), explicitly rejected source like `inline-tool-output` (exit 2), and anything else not in ALLOWED_SOURCES (was exit 1, should be exit 2). The third branch is semantically a user-origin rejection — the source is not user-originated — but was incorrectly using the validation-error exit code. This matters because callers distinguish exit 1 (retry with corrected input) from exit 2 (do not retry, this is a security rejection). The fix aligns the exit code and error message with the REJECTED_SOURCES branch.

## Verification

- `bun test` green (60 tests across 2 related files)
- Regression test `test/gstack-question-preference.test.ts` fails on main with:
  ```
  error: expect(received).toBe(expected)
  Expected: 2
  Received: 1
  ```
- Passes after fix applied
- Discrimination verified: reverting only `bin/gstack-question-preference` while keeping the test causes failure

### DIFF
--- a/bin/gstack-question-preference
+++ b/bin/gstack-question-preference
@@ -155,8 +155,8 @@ do_write() {
       process.exit(2);
     }
     if (!ALLOWED_SOURCES.includes(j.source)) {
-      process.stderr.write('gstack-question-preference: invalid source \"' + j.source + '\"; allowed: ' + ALLOWED_SOURCES.join(', ') + '\n');
-      process.exit(1);
+      process.stderr.write('gstack-question-preference: rejected — source \"' + j.source + '\" is not user-originated (profile poisoning defense)\n');
+      process.exit(2);
     }
 
     // Optional free_text — sanitize (no injection patterns, no newlines, <=300 chars)
--- a/test/gstack-question-preference.test.ts
+++ b/test/gstack-question-preference.test.ts
@@ -281,8 +281,16 @@ describe('--write user-origin gate (profile-poisoning defense)', () => {
       '--write',
       JSON.stringify({ question_id: 'q1', preference: 'never-ask', source: 'anonymous' }),
     );
-    expect(r.status).not.toBe(0);
-    expect(r.stderr).toContain('invalid source');
+    expect(r.status).toBe(2);
+    expect(r.stderr).toContain('profile poisoning defense');
+  });
+
+  test('unknown source exits 2 (user-origin rejection), not 1 (validation error)', () => {
+    const r = run(
+      '--write',
+      JSON.stringify({ question_id: 'q1', preference: 'never-ask', source: 'tool-output' }),
+    );
+    expect(r.status).toBe(2);
   });
 });
 


garrytan:

PR #2430, opened 2026-08-01
fix(browse): skip .gitignore append when git already ignores .gstack/ (#2385)

### LINKED ISSUE #2385: browse: ensureStateDir appends .gstack/ to a tracked .gitignore even when git already ignores it globally
## What happens

`ensureStateDir()` appends `.gstack/` to the git root's `.gitignore` on every browse-server start where that file doesn't already contain the literal line. It decides whether to append by reading the file's text, so if you already ignore `.gstack/` through your global `core.excludesFile`, you still get the line written into the repo's `.gitignore`. Revert it and the next daemon start puts it back.

The damage shows up on shared repos. Mine is a Jetpack checkout where `.gitignore` is tracked and team-owned, so every `/browse`, `/qa`, `/design-review`, `/canary`, or `/scrape` leaves a modified tracked file sitting in `git status` — one `git add -A` away from riding along in a PR that has nothing to do with gstack.

## Where

`browse/src/config.ts`, in `ensureStateDir()`. Same on main as of 1.60.1.0:

```js
const content = fs.readFileSync(gitignorePath, 'utf-8');
if (!content.match(/^\.gstack\/?$/m)) {
  const separator = content.endsWith('\n') ? '' : '\n';
  fs.appendFileSync(gitignorePath, `${separator}.gstack/\n`);
}
```

That text match is the entire test. Git is never asked whether the path is already ignored, so `core.excludesFile`, `.git/info/exclude`, and any parent `.gitignore` are all invisible to it.

Callers: `cli.ts` (two sites) and `server.ts` (two sites), so it fires on any daemon start.

## Repro

```sh
# .gstack/ is already in ~/.config/git/ignore
mkdir repro && cd repro && git init -q
printf 'node_modules/\n' > .gitignore
git add .gitignore && git commit -qm init

git check-ignore -v .gstack/
# ~/.config/git/ignore:8:.gstack/   .gstack/     <- already ignored

bun -e 'import {resolveConfig,ensureStateDir} from "<gstack>/browse/src/config.ts"; ensureStateDir(resolveConfig(process.env));'

cat .gitignore
# node_modules/
# .gstack/     <- appended anyway

git status --porcelain
#  M .gitignore
```

Reverting `.gitignore` and re-running reproduces it every time. It is idempotent within a checkout, so it doesn't stack duplicates; it just comes back after any revert.

## Suggested fix

Ask git instead of reading the file, and bail early when the answer is yes:

```js
export function isIgnoredByGit(projectDir: string, relPath: string): boolean {
  try {
    const proc = Bun.spawnSync(['git', 'check-ignore', '-q', '--', relPath], {
      cwd: projectDir, stdout: 'pipe', stderr: 'pipe', timeout: 2_000,
    });
    // 0 = ignored, 1 = not ignored, 128 = error (not a repo, bad pathspec)
    return proc.exitCode === 0;
  } catch {
    return false;
  }
}
```

then in `ensureStateDir()`, after the state dir is created:

```js
if (isIgnoredByGit(config.projectDir, '.gstack/')) return;
```

Returning `false` on exit 128 keeps the current behavior for non-repos and for a broken or missing git, so the only change is "already ignored, so don't touch the file."

Regression test that fails before and passes after — it sets `core.excludesFile` inside a scratch repo rather than depending on the developer's real global config:

```js
test('leaves .gitignore alone when git already ignores .gstack/ globally', () => {
  // git init tmpDir; write .gstack/ to an excludes file; git config core.excludesFile <that file>
  // .gitignore starts as 'node_modules/\n'
  ensureStateDir(config);
  expect(fs.readFileSync(gitignorePath, 'utf-8')).toBe('node_modules/\n');
  expect(fs.existsSync(path.join(tmpDir, '.gstack'))).toBe(true);
});
```

Running this locally: 42 pass / 0 fail in `browse/test/config.test.ts` with the guard, 41 pass / 1 fail without it. The four existing `ensureStateDir` gitignore tests use non-repo temp dirs, so `check-ignore` returns 128 there and they're unaffected.

Happy to open a PR if the approach looks right.

gregario:
## Summary

Fixes #2385 — `ensureStateDir()` appends `.gstack/` to the project's `.gitignore` even when git already ignores the path via `core.excludesFile`, `.git/info/exclude`, or a parent `.gitignore`.

The function only checked the local `.gitignore` file text for the pattern. If the line wasn't there, it appended — regardless of whether git's full ignore machinery already covered it. This corrupts tracked `.gitignore` files on shared repos, creating spurious diffs that ride along in unrelated PRs.

The fix adds a `git check-ignore -q` call before the text check. If git already ignores `.gstack/` through any mechanism, the function returns early without touching the file. On error or non-repo environments (exit 128), it falls back to the existing behavior.

## Verification

- `bun test browse/test/config.test.ts` green (42/42)
- Regression test `browse/test/config.test.ts` ("leaves .gitignore alone when git already ignores .gstack/ globally") fails on main with:
  ```
  error: expect(received).toBe(expected)
  "node_modules/
  + .gstack/
  "
  Expected  - 0
  Received  + 1
  ```
- Passes after fix applied
- Discrimination verified: reverting only `browse/src/config.ts` causes failure

### DIFF
--- a/browse/src/config.ts
+++ b/browse/src/config.ts
@@ -82,9 +82,12 @@ function isIgnoredByGit(projectDir: string, relPath: string): boolean {
   try {
     const proc = Bun.spawnSync(['git', 'check-ignore', '-q', '--', relPath], {
       cwd: projectDir, stdout: 'pipe', stderr: 'pipe',
+      timeout: 2_000,
     });
     return proc.exitCode === 0;
   } catch {
+    // git not found, timed out, or not a repo (exit 128). Fall through to
+    // the text-check path — appending is the safe default when unsure.
     return false;
   }
 }
--- a/browse/src/config.ts
+++ b/browse/src/config.ts
@@ -78,6 +78,17 @@ export function resolveConfig(
   };
 }
 
+function isIgnoredByGit(projectDir: string, relPath: string): boolean {
+  try {
+    const proc = Bun.spawnSync(['git', 'check-ignore', '-q', '--', relPath], {
+      cwd: projectDir, stdout: 'pipe', stderr: 'pipe',
+    });
+    return proc.exitCode === 0;
+  } catch {
+    return false;
+  }
+}
+
 /**
  * Create the .gstack/ state directory if it doesn't exist.
  * Throws with a clear message on permission errors.
@@ -96,6 +107,9 @@ export function ensureStateDir(config: BrowseConfig): void {
   }
 
   // Ensure .gstack/ is in the project's .gitignore
+  // First, check if git already ignores .gstack/ (via global excludes, .git/info/exclude, or parent .gitignore)
+  if (isIgnoredByGit(config.projectDir, '.gstack/')) return;
+
   const gitignorePath = path.join(config.projectDir, '.gitignore');
   try {
     const content = fs.readFileSync(gitignorePath, 'utf-8');
--- a/browse/test/config.test.ts
+++ b/browse/test/config.test.ts
@@ -124,6 +124,41 @@ describe('config', () => {
       expect(fs.existsSync(path.join(tmpDir, '.gitignore'))).toBe(false);
       fs.rmSync(tmpDir, { recursive: true, force: true });
     });
+
+    test('leaves .gitignore alone when git already ignores .gstack/ globally', () => {
+      const { spawnSync } = require('child_process');
+      const tmpDir = path.join(os.tmpdir(), `browse-gitignore-global-${Date.now()}`);
+      fs.mkdirSync(tmpDir, { recursive: true });
+
+      // Set up a real git repo
+      spawnSync('git', ['init', '-q'], { cwd: tmpDir });
+      spawnSync('git', ['config', 'user.email', 'test@test.com'], { cwd: tmpDir });
+      spawnSync('git', ['config', 'user.name', 'Test'], { cwd: tmpDir });
+
+      // Write a global excludes file that ignores .gstack/
+      const excludesFile = path.join(tmpDir, 'global-gitignore');
+      fs.writeFileSync(excludesFile, '.gstack/\n');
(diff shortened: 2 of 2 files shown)


kraftbj:
Hit this same bug a few days back and had a fix sitting on a local branch, same approach: `git check-ignore -q` before the append. No sense duplicating it, but two things from mine are worth folding in.

**Bound the spawn with a timeout.** `Bun.spawnSync` with no `timeout` blocks indefinitely if git hangs:

```ts
const proc = Bun.spawnSync(['git', 'check-ignore', '-q', '--', relPath], {
  cwd: projectDir,
  stdout: 'pipe',
  stderr: 'pipe',
  timeout: 2_000,
});
```

Not hypothetical on a network filesystem or with a slow credential helper. `ensureStateDir` runs on every browse start, so a stalled git means the browser never comes up at all, which is a worse failure than the bug being fixed here.

**Document what the `false` return means.** `isIgnoredByGit` returns `false` for two different states: git says "not ignored" (exit 1), and git couldn't answer at all (exit 128, not a repo, binary missing). Collapsing them is the correct call, since falling back to the text check is the safe direction. Assuming "ignored" when git failed would silently skip the append in a repo that actually needs it. That reasoning isn't obvious from the code though, and the next person to touch this will have to re-derive it. Two lines of comment above the return covers it.

FWIW your test is stronger than the one I wrote. Committing the `.gitignore` first and asserting `check-ignore` returns 0 before exercising the code means it can't pass for the wrong reason. Mine did neither.

gregario:
Both good catches, incorporated.

Added `timeout: 2_000` to the spawn and a comment above the `return false` explaining why both cases collapse to the same path. You're right that a stalled git on every browse start is worse than the original bug.

Re the failing CI: those are the LLM-judge and e2e-browse evals (Tier 2/3, non-deterministic `claude -p` subprocess tests). Green on main, unrelated to this change.

Pushed the update.

garrytan:

PR #2432, opened 2026-08-01
fix(setup): accept --host cursor and --host slate (#2361)

### LINKED ISSUE #2361: setup --host rejects cursor and slate, but both are registered in ALL_HOST_CONFIGS
`hosts/index.ts` line 21 registers ten hosts:

```
export const ALL_HOST_CONFIGS: HostConfig[] = [claude, codex, factory, kiro, opencode, slate, cursor, openclaw, hermes, gbrain];
```

`setup` accepts only eight of them. Lines 88 and 139 both enumerate `claude, codex, kiro, factory, opencode, openclaw, hermes, gbrain, or auto`. Passing `--host cursor` or `--host slate` exits with `Unknown --host value`.

Both have full host configs (`hosts/cursor.ts`, `hosts/slate.ts`) and `bun run gen:skill-docs --host cursor` works, so the generator supports them and only the installer doesn't.

Practical effect: there is no documented way to install gstack for Cursor, even though Cursor is a supported host. Since Cursor discovers skills by scanning for `.claude/skills`, `.cursor/skills`, `.agents/skills`, and `.codex/skills` anywhere on disk, it finds the generated per-host bundles inside the gstack checkout and lists the same ~55 skills many times over. On my machine that showed as 664 `SKILL.md` files under `~/.claude/skills`, where roughly 80 are real. Deleting the nine gitignored bundles brought it to 167.

Two things that would help, either one on its own:

1. Add `cursor` and `slate` to the `--host` list in `setup` so the installer matches the host registry.
2. Have `setup` clear stale generated bundles for hosts other than the one being installed, so a single checkout does not multiply into every scanner's view.

gregario:
## Summary

Fixes #2361 — `./setup --host cursor` and `./setup --host slate` exit with "Unknown --host value" even though both hosts have full configs in `hosts/cursor.ts` and `hosts/slate.ts` and are registered in `ALL_HOST_CONFIGS`.

The case statement at setup line 104 listed only `claude|codex|kiro|factory|opencode|auto` as pass-through hosts. Cursor and Slate are local IDE hosts with the same installation pattern (directory-based skill discovery) but were never added to the allowlist. The fix adds them alongside the other local hosts.

## Verification

- `bun test test/setup-host-validation.test.ts` green (3/3)
- Regression tests fail on main:
  - "--host cursor is accepted" → fails because stderr contains "Unknown --host value"
  - "--host slate is accepted" → same
  ```
  Expected to not contain: "Unknown --host value"
  Received: "Unknown --host value: cursor (...)"
  ```
- Both pass after fix applied
- Discrimination verified: reverting only `setup` causes both failures

### DIFF
--- a/setup
+++ b/setup
@@ -85,7 +85,7 @@ NO_TEAM_MODE=0
 PLAN_TUNE_HOOKS_MODE=""   # "" = resolve from env/config/prompt; "yes"/"no" = explicit
 while [ $# -gt 0 ]; do
   case "$1" in
-    --host) [ -z "$2" ] && echo "Missing value for --host (expected claude, codex, kiro, factory, opencode, openclaw, hermes, gbrain, or auto)" >&2 && exit 1; HOST="$2"; shift 2 ;;
+    --host) [ -z "$2" ] && echo "Missing value for --host (expected claude, codex, kiro, factory, opencode, cursor, slate, openclaw, hermes, gbrain, or auto)" >&2 && exit 1; HOST="$2"; shift 2 ;;
     --host=*) HOST="${1#--host=}"; shift ;;
     --local) LOCAL_INSTALL=1; shift ;;
     --prefix)    SKILL_PREFIX=1; SKILL_PREFIX_FLAG=1; shift ;;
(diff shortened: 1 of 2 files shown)


garrytan:

Issue #2435, opened 2026-08-02
bun run test exits 0 when tests fail — batch.test.ts kills the runner with process.exit(0)


vaskockorovski:
## What happens

`bun run test` can exit **0** with a large number of failing tests, and print no final tally.

## Cause

`browse/test/batch.test.ts:47` schedules:

```js
setTimeout(() => process.exit(0), 500);
```

When that timer fires it terminates the whole `bun test` process with status 0 — the run is truncated and every failure recorded up to that point is discarded. Because the `test` script is `bun test … && (bun run slop:diff …)`, the `&&` tail then runs, which is itself a signal that the exit code was 0.

## Environment

- `bun` 1.3.13
- Ubuntu 24.04.4 LTS (kernel 6.8.0-136)
- `main` @ `a325940`

## Repro

```
bun install
bun run test > out.log 2>&1; echo "EXIT=$?"
grep -c '^(fail)' out.log            # non-zero
grep -E 'Ran [0-9]+ tests' out.log   # no tally — the run did not finish
```

Observed here: `EXIT=0` with 82 `(fail)` lines and no tally block. `browse/test/batch.test.ts` is the last file to emit output in every truncated run.

## Why it's worth fixing

The suite currently cannot fail CI, so a regression landing anywhere before that timer fires goes green.

The most portable symptom, if it helps anyone else hitting this: **a runner that prints no final tally did not finish** — the missing summary is a more reliable tell than the exit code, because the exit code is the thing being falsified.

## Note on scoping while investigating

Worth flagging because it cost a couple of runs to notice: `bun test test/` is a substring filter rather than a directory scope, so it still pulls in `browse/test/**`, and `--ignore 'browse/**'` did not exclude them either. Two consecutive "isolated" runs reproduced an identical signature, which briefly read as evidence against the diagnosis — it was the same unscoped run twice. An explicit file list was the only reliable way to isolate the non-browse suite:

```
bun test $(ls test/*.test.ts | grep -vE 'skill-e2e-|skill-llm-eval|skill-routing-e2e|codex-e2e|gemini-e2e')
```

That form gives a truthful `exit 1` and a real tally.

## No fix attached

Happy to send a PR if you'd like a particular approach — either letting the test await shutdown rather than exiting, or moving the `process.exit(0)` behind an env guard so it only fires in the standalone context it was written for.

tranthanhnhatkhoa:
[...]

PR #2448, opened 2026-08-04
feat: tiered outside-voice review — cheap loop + frontier gate

vaskockorovski:
## Tiered outside-voice review: cheap loop + frontier gate (toward #619)

`outside_voice` config family that runs the **iterative review loop** on a cheap OpenAI-compatible backend and keeps the **final gate** on codex. Built because the loop's value is fix-verify on an evolving artefact — a fix is new code — so the long, repetitive half is exactly the half that does not need frontier pricing, while the last look still does.

**Measured on this branch's own build**, from the ledger and usage log this PR ships, both re-derived at time of writing rather than quoted from notes:

| | rounds | P1 | P2 | P3 | new-work | false-positive |
|---|---|---|---|---|---|---|
| loop (MiniMax M3) | 27 | 68 | 139 | 128 | 141 | 71 |
| gate (codex) | 17 | 4 | 26 | 5 | 20 | 8 |

**Loop token spend US$2.30** across 50 calls (1,215,790 in / 1,609,873 out at $0.30/$1.20). The same volume at frontier rates is US$46.33 — **20.2×**. Corroborated twice, from an 8-call sample early in the build (20.0×) and the full 50-call lane here.

The number #619 will care about more than the price: **the cheap tier's false-positive rate is not worse than the frontier tier's** — 71 of 335 findings (21%) against codex's 8 of 35 (23%). Cheap did not mean noisy. What differed was *kind*, not quality: the gate found repo-wide, cross-file contracts; the loop found within-diff defects. They behaved like different instruments rather than the same instrument at two prices, which is the actual argument for tiering rather than simply switching.

### What this adds
- `outside_voice` config family (flat keys, validated): `loop` phase on a cheap backend, `final_gate` on codex. **An unconfigured install behaves exactly as today** — `outside_voice_*` unset resolves to `codex`, `codexPreflight` is untouched, and every unmigrated skill still dispatches `codex exec`. (Not byte-identical: `codex/SKILL.md` itself changed substantially. The unchanged thing is behaviour, and a test pins the unset→codex resolution.)
- One adapter (`bin/gstack-outside-voice`) replacing per-skill hardcoding; backend flag translation lives in the adapter, no new dependencies.
- Structured findings contract (`--findings-out`, per-request nonce, fail-loud on unparseable) replacing free-form parsing — a count parsed from a format the reviewer may vary is not a measurement.
- Per-voice, per-round append-only ledger (`--voice`, finding classes, reasoning tokens, cost) — corrections are events, never edits. It is what produced the table above.

### Scope — what is and is not wired
`/codex` is the **only** skill routed through the adapter, across all three of its review paths (loop, default final gate, custom-focus). **Thirteen other skills still call `codex exec` directly** and are unaffected by `outside_voice_*` config. So this is a first step toward #619 rather than a completion of it — happy to migrate the rest in follow-ups, or in this PR if you would prefer it landed whole.

### Honest caveats, from the same ledger

- **The bar this branch was held to was re-defined partway, deliberately, and it is worth saying so plainly.** The original bar was a review round returning no P1 and no P2. On self-modifying review tooling that turned out to be unreachable rather than merely distant: every fix is new code that seeds the next round's findings, so the bar measured "the artefact stopped changing", which contradicts fixing anything. It was replaced by a **pure-verification round** — no fixes landed after the reviewed commit, read-only review, requiring **zero P1 and zero new-work P2**. That was reached: one P2 was raised and verified false (below). P1s had been extinct for seven rounds before it.
- **The one open finding, and why it is not fixed.** The final round flagged `bin/gstack-config` for not stripping inline comments from the two backend keys, claiming `outside_voice_loop: openrouter # cheap loop` parses as `openrouter#cheaploop`. Probed rather than accepted: it resolves to `openrouter`, as does the gate key. The parser takes `awk '{print $2}'`, so a space-separated trailing comment lands in fields 3+ and is discarded. The only shape that does fail is `openrouter#nospace` — which YAML does not treat as a comment either, so refusing it is correct. Left unrewritten because any edit would have destroyed the verification round's whole property.
- **Self-modifying material makes review loops compound rather than taper.** A meaningful share of gate findings were contradictions the run's own fixes created in adjacent prose or code, not defects in the original work. Worth knowing before anyone budgets a round count for a change of this shape. The ledger records structural sweeps (6 here) as the counter-measure: when findings start coming from your own fixes, the answer is a consistency sweep, not another round.
- **A runaway-detector fired once, at round 21**, and is recorded in the ledger rather than quietly retried past. The decision to continue was that the round in question had just performed a structural sweep, the findings were reaching real users rather than internal tooling, and the branch was one round from its verification pass.
- **This branch also fixes the `batch.test.ts` early-exit reported in #2435**, so a red suite can no longer report green by truncation. The scoped, explicit-file-list form documented on that issue yields a truthful exit 1 and a real tally: **3785 pass / 8 fail across 207 files**. All 8 are pre-existing and none is attributable to this branch — verified by diffing failure *name sets*, not counts.
  **The fix is necessary and not sufficient:** it was 8 files, not 1 (the issue named `batch.test.ts` because it happened to die first), and even with all 8 guarded, `bun run test` *still* exits 0 without a tally — browse's production code calls `process.exit` in `cli.ts` and `server.ts`, and a test driving a real shutdown takes the runner down with it. A separate defect this PR does not claim to fix.
- **`test/fixtures/golden-ship-claude.md` is an orphan.** No test or script reads it, and its contents predate the current generated shape. This branch edited it as part of a repo-wide sweep; it was deliberately not hand-patched further, since fabricating a "golden" no generator produces is worse than an unused stale file. Flagged as a deletion candidate rather than silently touched.

### Cost note — the honest shape of the 20× figure
The **US$2.30** is a **floor**: computed from list rates rather than reconciled against a billing dashboard, and 16 token fields across those 50 calls are an `unknown` sentinel (timeouts and errors where usage was never returned), so real spend sits slightly above it. The frontier side is **hypothetical by necessity** — codex reports neither prompt nor completion counts, so the gate's own calls carry zero token data and the US$46.33 prices the *loop's* volume at frontier rates rather than measuring what codex actually consumed. Codex has repository access and different reasoning behaviour, so it would not consume the same tokens. What makes ~20× trustworthy despite that is corroboration, not precision: two independent samples, taken weeks and 42 calls apart, landed on 20.0× and 20.2×.

Full per-round dataset — false positives, self-inflicted counts and every correction event — in the ledger. Happy to reshape config naming or scope to taste.


vaskockorovski:
Upstream CI on this PR shows 8 failures (7 eval suites + `report`). I looked before
flagging: all 8 are one root cause — **fork PRs don't receive repo secrets** — not a
regression in this branch.

**Mechanism.** `evals.yml` is `on: pull_request` and reads keys from `secrets.*`
(L248–250). GitHub withholds repo secrets from `pull_request` runs from a fork, so
every job in run 30876990754 logs:

    ANTHROPIC_API_KEY:
    OPENAI_API_KEY:
    GEMINI_API_KEY:

(empty in the *passing* jobs too).

**How each failure follows.**

- `llm-judge` — the SDK error verbatim:
  `error: Could not resolve authentication method. Expected either apiKey or authToken to be set.`
- `e2e-workflow` / `e2e-plan` / `e2e-design` / `e2e-review` / `e2e-qa-workflow` —
  `exitReason === "error_api"`, preceded by
  `/document-release: $0.00 (1 turns, 0.0k tokens, 1s)`. Zero tokens, zero cost, one
  second: the API call never happened. The content guardrail immediately above
  (`expect(hasOriginalEntries).toBe(true)`, L106) passes; only the exit-reason
  assertion on L109 fails.
- `e2e-pty-plan-smoke` (31m44s) — same cause, different shape. The seed step's
  `if (key) seed.customApiKeyResponses = …` no-ops on an empty key, so the TUI wedges
  on the "use detected ANTHROPIC_API_KEY?" dialog your own comment at L151–156
  anticipates. Six attempts × ~308.8s ≈ 31 min.
- `report` — downstream only: `[ "$FAILED" -gt 0 ] && STATUS="❌ FAIL"`.

**Counterfactual.** The scheduled run with secrets present
(`evals-periodic`, run 30792498501, 2026-08-03) passes the same suites on the same
files one day before this PR: e2e-workflow ✓ e2e-plan ✓ e2e-design ✓ e2e-review ✓
e2e-qa-workflow ✓.

**Not us.** This branch modifies none of the failing test files. The only shared files
it touches are a comment in `test/helpers/tool-map.ts` and a codex flag in
`document-release/sections/release-body.md` — neither can yield `error_api` at 0 tokens.
Its own suite is green locally: `bun test test/outside-voice-config.test.ts` →
75 pass / 0 fail.

**Two notes that may save you time.**

1. Annotation counts aren't a failure signal on this workflow — `--retry 2` annotates
   attempts that later pass, so the *passing* jobs in run 30792498501 also carry
   annotations (`skill-e2e-design.test.ts:155`, `skill-e2e-plan.test.ts:181`).
2. `e2e-qa-bugs` going green here doesn't indicate health: it logged
   `0 pass / 0 fail` across 5 tests — diff-selection skipped all of them.

Happy to rerun once approved with secrets, or to split anything out that would make
review easier.

PR #2451, opened 2026-08-04
fix: exec gstack-brain-sync through bash on Windows

vaston-viji:
## Problem

The #1731 fix gated the orchestrator's brain-sync spawns behind `shell: NEEDS_SHELL_ON_WINDOWS`, but that routes through cmd.exe, which cannot execute an extensionless bash shebang script. Every `/sync-gbrain` run on Windows errors:

```
'C:\Users\...\gstack\bin\gstack-brain-sync' is not recognized as an internal or external command
```

and the stage reports `ERR brain-sync — gstack-brain-sync exited 1`.

## Fix

In `runBrainSyncPush` (bin/gstack-gbrain-sync.ts), exec the script through bash explicitly on Windows: `spawnSync("bash", [brainSyncPath, flag])`. Git Bash ships with git, which gstack already requires, so bash is a safe dependency. POSIX keeps the direct shebang exec with no shell. Spawn errors (e.g. bash missing from PATH) now surface in the stage summary instead of a bare exit code.

The #1731 tripwire test pinned the old shape (`spawnSync(brainSyncPath, ...)` with the shell flag), so it is updated to pin the new invariant: brain-sync spawns must exec bash on Windows and must not rely on `shell: true`.

## Verification

On Windows 11 (bun 1.3.14):
- Before: `ERR brain-sync — gstack-brain-sync exited 1` with the "not recognized" cmd error.
- After: `OK brain-sync — curated artifacts pushed (0.2s)`.
- `bun test test/gbrain-spawn-windows-shell.test.ts`: 5 pass, 0 fail.

Related: #2450 fixes the separate Windows probe-timeout false-negative in the same sync pipeline.

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -1145,25 +1145,35 @@ function runBrainSyncPush(args: CliArgs): StageResult {
     return { name: "brain-sync", ran: false, ok: true, duration_ms: 0, summary: "skipped (gstack-brain-sync not installed)" };
   }
 
-  // #1731: gstack-brain-sync is a bash shebang script; Windows can't spawn it
-  // without a shell, which surfaced as "brain-sync exited undefined".
-  spawnSync(brainSyncPath, ["--discover-new"], {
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2452, opened 2026-08-04
fix: Windows build and ACL issues

chiragborse1:
Fixes two Windows-specific issues that prevented gstack from working out-of-the-box:

1. **Build failure for browse/server-node.mjs**
   - On Windows (MSYS/Git Bash), Bun cannot resolve MSYS-style `/c/...` absolute paths.
   - The build-node-server.sh script now converts the GSTACK_DIR to a Windows-style path via `cygpath -m` when running under MINGW/MSYS/CYGWIN.

2. **ACL lockout on .gstack directory**
   - gstack's file-permissions.ts used bare username (`chirag`) for icacls grants.
   - When the hostname equals the username (e.g., host `CHIRAG`, user `chirag`), the bare name resolves to the *machine SID* instead of the user SID.
   - This left the real user without access to the .gstack directory, causing EPERM on reads/writes and the "Another instance is starting the server, waiting..." loop.
   - Changed to use the fully-qualified `%USERDOMAIN%\%USERNAME%` (with fallback to the SID via `whoami /user`) so the correct user gets access.

After applying these patches:
- A fresh `.gstack` directory is created with the correct `CHIRAG\chirag:(OI)(CI)(F)` ACL.
- The browse server starts successfully (tested with `browse goto https://example.com` → text extraction).
- Both the Bun-compiled `browse.exe` and the Node.js compatibility bundle (`server-node.mjs`) are built and functional.

Tested on:
- Windows 10 (MSYS2/Git Bash from git-bash)
- Bun 1.3.14
- Node 24.18.0

Fixes:
- "FileNot able to build browse/server-node.mjs on Windows (FileNotFound opening root directory)"
- Browse server fails with EPERM on .gstack directory due to mistaken ACL granting machine SID

Related to: https://github.com/garrytan/gstack/issues/997 (the codesign note is unrelated, but same class of Windows-specific issues)

### DIFF
--- a/browse/scripts/build-node-server.sh
+++ b/browse/scripts/build-node-server.sh
@@ -8,6 +8,11 @@
 set -e
 
 GSTACK_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+# Windows (MSYS/Git Bash): convert to a Windows-style path — Bun cannot open
+# MSYS /c/... absolute paths ("FileNotFound opening root directory").
+case "$(uname -s)" in
+  MINGW*|MSYS*|CYGWIN*) GSTACK_DIR="$(cygpath -m "$GSTACK_DIR")" ;;
+esac
 SRC_DIR="$GSTACK_DIR/browse/src"
 DIST_DIR="$GSTACK_DIR/browse/dist"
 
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2457, opened 2026-08-04
Fix: wire GSTACK_CHROMIUM_PATH into headless launch, forward proxy config, and trust custom CA

erangini-ai:
## Summary

On sandboxed hosts (e.g., Claude Code web/remote) where Playwright's Chromium download is blocked, gstack bundles a pre-installed Chromium and sets `GSTACK_CHROMIUM_PATH`. However, only headed `/browse` (`launchHeaded()`) actually used that variable — headless `/browse` (`launch()`) ignored it entirely, still hitting the "Executable doesn't exist" error.

This PR fixes that, plus two related proxy/Certificate Authority gaps that prevent the browser from functioning inside MITM-proxied environments.

## Changes

1. **Wire `GSTACK_CHROMIUM_PATH` into headless launch**
   `launch()` now reads `GSTACK_CHROMIUM_PATH` and passes it as `executablePath` when set, matching the existing `launchHeaded()` behavior. Falls back to Playwright's default when the variable is absent.

2. **Forward `HTTPS_PROXY` to Chromium**
   The browser process now passes `--proxy-server=<value>` (derived from `HTTPS_PROXY`) so outbound traffic routes through the session's proxy instead of attempting a direct connection.

3. **Trust the proxy's custom Certificate Authority**
   Added `ignoreHTTPSErrors: true` (Playwright‑level) so the browser accepts certificates injected by the MITM‑style proxy. Also sets `NODE_EXTRA_CA_CERTS` for the Node process's own TLS stack (has no effect on Chromium itself, but useful for other network calls in the same context).

## Testing

Verified in a proxy‑restricted Firecracker VM using a `file://` URL (zero external network dependency):

- **Headless** `/browse` navigates and renders successfully — no "Executable doesn't exist" error.
- **Headed** `/browse` (xvfb‑run) passes the same test, no Chromium download attempted.
- The `--proxy-server` flag is confirmed present in the launched Chromium command line.

Real‑internet navigation remains subject to the host's outbound network policy; this fix is focused solely on eliminating the browser‑launch failure.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -369,6 +369,29 @@ export class BrowserManager {
       console.log(`[browse] Extensions loaded from: ${extensionsDir}`);
     }
 
+    // Support custom Chromium binary via GSTACK_CHROMIUM_PATH env var, same as
+    // launchHeaded(). Falls back to Playwright's bundled/resolved Chromium
+    // (undefined → chromium.launch() uses its own default resolution) when unset.
+    const executablePath = process.env.GSTACK_CHROMIUM_PATH || undefined;
+
+    // Forward the session's outbound proxy to Chromium. Host sandboxes (e.g.
+    // Claude Code web/remote) route all HTTPS through a local MITM-style
+    // agent proxy via HTTPS_PROXY; without --proxy-server Chromium tries a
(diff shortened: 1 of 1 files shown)


garrytan:

PR #2460, opened 2026-08-05
fix(diff-scope): detect Drizzle migrations, root-level migrations/, and singular migration/

cjones6489:
## What

`SCOPE_MIGRATIONS` returned `false` on a PR whose diff contained a migration, so
`/ship` never dispatched the **data-migration specialist** — on a diff that dropped a
column `DEFAULT`. Found while shipping a Drizzle + Postgres app; I force-included the
specialist manually, and it turned out to be the pass that mattered most on that diff.

The pattern list matched `db/migrate/`, `*/migrations/`, `alembic/` and `prisma/migrations/`.
Three gaps:

1. **Drizzle** writes bare `.sql` files plus a `meta/` journal into its `out` dir (default
   `./drizzle`), matching none of them. This is the case that surfaced it.
2. **`*/migrations/*` requires a parent segment**, so a **root-level** `migrations/` dir —
   Knex, Sequelize, golang-migrate, Atlas — fell through.
3. **Flyway** and some Java/Ecto layouts use the singular `migration/`.

Matching is deliberately broad, on the asymmetry that a false positive costs one extra
specialist pass while a false negative silently skips the reviewer on the diff that most
needed one.

## Also: the test file could not run on Windows at all

Every one of `test/diff-scope.test.ts`'s 16 tests failed before this change — not from a
regression, it had simply never worked there. `createRepo` derived the parent directory with:

```js
const dirPath = fullPath.substring(0, fullPath.lastIndexOf('/'));
```

`join()` emits **backslashes** on Windows, so `lastIndexOf('/')` returned `-1`,
`substring(0, -1)` gave `''`, and `mkdirSync('')` threw `ENOENT`. Now uses `dirname()`.

I verified this is pre-existing rather than mine by stashing both files and re-running:
**0 pass / 16 fail** on a clean checkout.

## Verification

`bun test test/diff-scope.test.ts` on Windows:

| | before | after |
|---|---|---|
| pass | 0 | **21** |
| fail | 16 | **0** |

Six new cases cover Drizzle output (bare `.sql`, `meta/` journal, monorepo package),
root-level `migrations/`, singular `migration/`, and a **negative** case asserting the
widened patterns do *not* fire on `drizzle.config.ts` or on prose like
`docs/migration-guide.md` — only on a real path segment.

## Note

Branched from `v1.60.1.0`, which is what my install was pinned at. `bin/gstack-diff-scope`
and `test/diff-scope.test.ts` are both untouched upstream since then, so it should apply
cleanly — happy to rebase if you'd prefer.

### DIFF
--- a/bin/gstack-diff-scope
+++ b/bin/gstack-diff-scope
@@ -63,8 +63,16 @@ while IFS= read -r f; do
     .github/*) CONFIG=true ;;
     requirements.txt|pyproject.toml|go.mod|Cargo.toml|composer.json) CONFIG=true ;;
 
-    # Migrations: database migration files
-    db/migrate/*|*/migrations/*|alembic/*|prisma/migrations/*) MIGRATIONS=true ;;
+    # Migrations: database migration files.
+    # Match BROADLY on purpose: a false positive costs one extra specialist pass, a false
+    # negative silently skips the data-migration reviewer on the diff that needed it most.
+    # (Observed: a Drizzle repo dropping a column DEFAULT scored SCOPE_MIGRATIONS=false.)
+    db/migrate/*|db/migration/*) MIGRATIONS=true ;;
+    migrations/*|migration/*|*/migrations/*|*/migration/*) MIGRATIONS=true ;;
+    alembic/*|prisma/migrations/*) MIGRATIONS=true ;;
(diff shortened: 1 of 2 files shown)


cjones6489:
Closing this — opened in error on my side, not a considered submission. Apologies for the noise.

PR #2461, opened 2026-08-05
Fix browse daemon crash-loop diagnostics + a shell-injection bug found along the way

phuttimatebenchanakatkul:
## Summary
- The detached browse daemon's stdout/stderr were wired to `'ignore'` on every platform, so every `console.error('[browse] FATAL: ...')` from a crash (Chromium disconnect, uncaughtException, unhandledRejection) went nowhere — a restart was indistinguishable from any other cause of a dropped session, with nothing on disk recording why.
- `cli.ts`: both the Windows and macOS/Linux spawn paths now redirect stdout/stderr to `<stateDir>/browse-daemon.log` (append mode, accumulates across the daemon's full lifetime and every respawn).
- That immediately surfaced two real, previously invisible bugs, both fixed here:
  1. **`spawn bun ENOENT`** — `bun-polyfill.cjs`'s `Bun.spawn`/`spawnSync` polyfill called Node's `child_process.spawn` with a bare command name (`'bun'`), which Windows can't resolve without PATHEXT lookup.
  2. **`browser?.process is not a function`** — `resolveDisconnectCause()` (the crash classifier itself) threw on some browser/connection-mode shapes, and being an unhandled fire-and-forget rejection, took the whole daemon down a second, harder-to-diagnose way.

## A note on how the fix for #1 evolved
My first attempt used `shell:true` with hand-rolled double-quote escaping to fix the PATHEXT resolution. I verified empirically (not just reasoned about it) that this does **not** neutralize cmd.exe metacharacters (`&|^%<>`), and is reachable via `browser-skill-commands.ts`'s `...opts.skillArgs` passthrough from `$B skill run <name> --arg k=v`. Spawning a resolved `.cmd` path with an arg containing `& echo INJECTED > proof.txt` actually created the file.

Replaced it with [`cross-spawn`](https://github.com/moxystudio/node-cross-spawn) — already resolved in `node_modules` transitively via `@modelcontextprotocol/sdk`, now added as a direct dependency — the established library for correct Windows PATHEXT resolution + argument escaping together. Re-verified the same injection payload no longer executes, and normal resolution (`bun --version`) still works.

## Test plan
- [x] Killed the live Chromium child under a running daemon — log now shows a single clean `FATAL: Chromium process crashed or was killed`, terminal-agent respawns successfully, daemon auto-recovers to a healthy new PID
- [x] Re-ran the same test after reverting each fix individually to confirm each is load-bearing (not just cosmetically present)
- [x] Confirmed the shell-injection payload (`& echo INJECTED > proof.txt`) no longer executes after the `cross-spawn` fix
- [x] Confirmed normal resolution (`bun --version` via both `spawn` and `spawnSync`) still works, including paths/args containing spaces
- [x] 24/24 pre-existing `browser-manager-unit` tests pass
- [x] `bun build --compile` succeeds for both `cli.ts` and the Node server bundle

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -91,7 +91,32 @@ export function shouldEnableChromiumSandbox(): boolean {
  * restarts on backoff.
  */
 export async function resolveDisconnectCause(browser: Browser | null): Promise<'clean' | 'crash'> {
-  const proc = browser?.process();
+  // browser.process() has thrown "browser?.process is not a function" in
+  // the wild (confirmed live: killing the headless-shell child process
+  // triggered it) -- some browser instances/connection modes apparently
+  // don't carry a real .process() the way a plain chromium.launch() result
+  // does. That's a synchronous throw inside this async function, which
+  // rejects the promise; the caller (handleChromiumDisconnect) awaits it
+  // without its own try/catch and is itself invoked via `void` (fire-and-
+  // forget) from the 'disconnected' event handler, so the rejection had
+  // nowhere to go but server.ts's top-level unhandledRejection handler --
+  // which exits the WHOLE daemon. The function whose entire job is
+  // classifying a disconnect was, on this path, causing a second, harder
+  // to diagnose crash of its own. Default to 'crash' (the conservative,
+  // pre-existing behavior for anything abnormal) when we can't even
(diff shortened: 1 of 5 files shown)


garrytan:

PR #2467, opened 2026-08-06
fix(make-pdf): hide empty TOC page numbers and dot leaders in print

dylansacks62:
## Problem

`make-pdf generate --toc` produces a table of contents where dot leaders run from each entry title into blank space. It reads as a broken document.

Cause: `.toc-page` is filled by a CSS `target-counter`, which only resolves under Paged.js. Paged.js is not vendored in gstack, and Chrome has no native `target-counter`, so the spans stay permanently empty while the leader dots still render. `browse/src/meta-commands.ts:545` already documents the constraint: *"make-pdf v1 ships without Paged.js; TOC renders without page numbers."*

This is not fixable in CSS alone as long as the counter has nothing to resolve against.

## Fix

Hide `.toc-dots` and `.toc-page` in print until page numbers actually exist. `screenCss()` already does exactly this for HTML output, so this brings print in line with the path that was already correct. A comment marks the two rules to restore the day Paged.js lands.

Also: `buildTocBlock` mapped every heading at H2 or deeper to `level-2`, so H3 sub-sections sat at the same indent as numbered sections and read as top-level peers. H3 now gets its own `level-3` class.

## Result

Clean, clickable, correctly nested TOC. No trailing dots. Cover page, running headers, `N of M` footers unaffected.

## Testing

- `bun test` in `make-pdf/`: **189 pass, 0 fail**, 393 `expect()` calls across 11 files.
- Verified visually on a 9-page letter-format document with a cover page: rasterized the output and confirmed zero leader-dot glyphs in the TOC region, correct H2/H3 nesting, and intact headers and footers.

## Note

If Paged.js is ever vendored, revert the single `display: none` rule and the original two declarations are restored verbatim from the comment. The `level-3` change is independent and should stay either way.

### DIFF
--- a/make-pdf/src/print-css.ts
+++ b/make-pdf/src/print-css.ts
@@ -273,9 +273,14 @@ function tocRules(enabled: boolean): string {
     `  padding: 3pt 0;`,
     `}`,
     `.toc li .toc-title { flex: 0 0 auto; }`,
-    `.toc li .toc-dots { flex: 1 1 auto; border-bottom: 1px dotted #aaa; margin: 0 6pt; transform: translateY(-4pt); }`,
-    `.toc li .toc-page { flex: 0 0 auto; color: #666; font-variant-numeric: tabular-nums; }`,
+    // v1 ships without Paged.js, so target-counter never resolves and
+    // .toc-page stays empty — leader dots would trail into blank space.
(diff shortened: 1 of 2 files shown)


garrytan:

PR #859, opened 2026-04-06
fix(design): restrict session file permissions to owner-only

garagon:
## What

Add `{ mode: 0o600 }` to both `writeFileSync` calls in `design/src/session.ts`.

## Why

Design session files are written to `/tmp` without restricted permissions. Default umask (typically 0644) makes them world-readable on shared systems. Sessions contain `originalBrief` and `feedbackHistory` with user design prompts and iteration data.

The rest of the codebase uses `mode: 0o600` for sensitive file writes (16 locations across `browse/src/server.ts`, `browse/src/sidebar-agent.ts`, `browse/src/browser-manager.ts`, `browse/src/cli.ts`, `browse/src/meta-commands.ts`, and `design/src/auth.ts`).

## Files

- `design/src/session.ts`: `createSession()` line 52 and `updateSession()` line 78

PR #2468, opened 2026-08-06
fix(design): create OpenAI key file owner-only to close write-then-chmod race

bunlongheng:
## What

`saveApiKey()` in `design/src/auth.ts` persists the OpenAI API key to `~/.gstack/openai.json` by writing at the process default umask and only tightening permissions afterwards:

```ts
fs.writeFileSync(configPath(), JSON.stringify({ api_key: key }, null, 2));
fs.chmodSync(configPath(), 0o600);
```

On a typical umask (022) the file is created `0644` (group/world-readable). Between the `writeFileSync` and the `chmodSync` there is a window where the file containing the OpenAI API key is readable by any other local user. This is CWE-377 (insecure file creation) / CWE-367 (TOCTOU) — low severity, only exploitable by a co-located local user on a shared/multi-user host, but it's a real key-disclosure window and a cheap thing to close.

## Fix

Create the file owner-only up front by passing `{ mode: 0o600 }` to `writeFileSync`, so it is never briefly world-readable. The trailing `chmodSync` is kept as a backstop that still tightens a pre-existing loose file.

```ts
fs.writeFileSync(configPath(), JSON.stringify({ api_key: key }, null, 2), { mode: 0o600 });
fs.chmodSync(configPath(), 0o600);
```

This matches the convention already used for session files in `design/src/session.ts` (`{ mode: 0o600 }` on create, landed in #859). #859 hardened the session files but the actual API key file kept the write-then-chmod pattern — this brings it in line.

## Test

Adds a `saveApiKey` regression test that writes the key under a deliberately permissive umask (`0o000`) and asserts the resulting file is exactly `0600` with no group/other bits.

```
$ bun test design/test/auth.test.ts
 7 pass
 0 fail
```

## Scope

Two files, `design/src/auth.ts` (+3/-1) and `design/test/auth.test.ts` (+21). No behavior change beyond the file-creation mode.

### DIFF
--- a/design/src/auth.ts
+++ b/design/src/auth.ts
@@ -111,7 +111,10 @@ export function describeApiKeySource(resolution: ApiKeyResolution): string {
 export function saveApiKey(key: string): void {
   const dir = path.dirname(configPath());
   fs.mkdirSync(dir, { recursive: true });
-  fs.writeFileSync(configPath(), JSON.stringify({ api_key: key }, null, 2));
+  // Create the file owner-only up front so the API key is never briefly
+  // world/group-readable in the window between write and chmod. The trailing
+  // chmodSync is kept as a backstop to tighten a pre-existing loose file.
+  fs.writeFileSync(configPath(), JSON.stringify({ api_key: key }, null, 2), { mode: 0o600 });
   fs.chmodSync(configPath(), 0o600);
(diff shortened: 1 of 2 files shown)


garrytan:

Issue #2194, opened 2026-07-07
gbrain serve (local-stdio MCP) holds PGLite exclusively; probe exit-124 misclassified as broken-config — /sync-gbrain deterministically skips code+memory inside any Claude session (lock variant of #1964, timeout raise cannot fix)


Kerumen:
## Symptom

Same output as the #1964 family (`#1792` `#1832` `#1969` `#2051`), but with a root cause none of them cover:

```
gstack-gbrain-sync (incremental):
  SKIP  code         skipped — local engine broken-config — config at ~/.gbrain/config.json is malformed; see /setup-gbrain Step 1.5 (0.9s)
  SKIP  memory       skipped — local engine broken-config — config at ~/.gbrain/config.json is malformed; see /setup-gbrain Step 1.5
  OK    brain-sync   curated artifacts pushed (2.8s)
```

Meanwhile the engine is healthy: `gbrain doctor --fast --json` answers with only warnings, and `~/.gbrain/config.json` is valid, well-formed JSON (pglite engine, voyage-code-3 embeddings).

## Root cause: `gbrain serve` holds the PGLite database exclusively

PGLite is an embedded single-process engine — whichever process opens `~/.gbrain/brain.pglite` owns it. With the recommended default setup (`/setup-gbrain` → PGLite + local-stdio MCP), **Claude Code spawns `gbrain serve` at session start, and serve opens the database eagerly at boot** — even if no `mcp__gbrain__*` tool is ever called in the session:

```
$ pgrep -fl gbrain
95427 bun /Users/.../.bun/bin/gbrain serve        # parent PID = the live `claude` process

$ lsof +D ~/.gbrain/brain.pglite | head -3
COMMAND   PID          USER   FD   TYPE   ... NAME
bun     95427 yannpringault    6u   REG   ... .../brain.pglite/pg_wal/000000010000000000000004
bun     95427 yannpringault    7u   REG   ... .../brain.pglite/global/2672
```

Every gbrain CLI command that needs the DB then times out at connect:

```
$ gbrain sources list --json
gbrain sources: connect timed out (default 10000ms; pass --timeout=Ns to override).
$ echo $?
124
```

[...]

PR #1991, opened 2026-06-12
v1.60.2.0 fix(developer-profile): exclude mode:resources entries from SESSION_COUNT, TIER, and NUDGE_ELIGIBLE

mvann:
## Summary

Reopens the fix from #1821 (closed for staleness), now **rebased onto current main (v1.57.10.0) → v1.57.11.0** as requested:

> "Closing — this targets v1.55.1.0, which is behind main (now v1.57.0.0), and has merge conflicts. If the developer-profile fix is still relevant, please rebase on current main and reopen. Thanks!"

The fix is still relevant: upstream never touched `bin/gstack-developer-profile` or its test, so both defects are still live on main today. The only conflicts on rebase were version bookkeeping (VERSION/CHANGELOG/package.json), now resolved.

## The bug (deterministic, on the normal /office-hours path)

Every `/office-hours` session writes two profile rows, not one: the real session, plus a `mode:"resources"` bookkeeping row that Phase 6 unconditionally appends (`office-hours/sections/design-and-handoff.md:510`) to dedupe which founder-resource links you've already seen. At the start of the next session the profile is read back (`design-and-handoff.md:278`) to pick the greeting tier and decide whether to fire the builder→founder nudge.

`do_read()` counted those bookkeeping rows as real sessions:

- **`SESSION_COUNT` / `TIER` inflation.** `const count = sessions.length` counted resources rows, so after one real session `SESSION_COUNT` read 2 and tier thresholds (intro=1, welcome_back=2-3, regular=4-7, inner_circle=8+) crossed ~2x too fast. A brand-new user gets greeted as a returning "regular."
- **`NUDGE_ELIGIBLE` armed off bookkeeping.** The predicate `sessions.filter(e => e.mode !== 'startup')` was doubly wrong: it counted resources rows *as builder sessions* and excluded real startup sessions. The builder→founder nudge could fire at someone who never actually came back to build.

This is not undefined behavior or malformed input — gstack's own Phase 6 writes the row every session, and the inflation compounds with each run.

## The fix (`bin/gstack-developer-profile`, `do_read()`)

- Hoist the existing `realSessions = sessions.filter(e => e.mode !== 'resources')` filter above `count`, so `SESSION_COUNT` / `TIER` use it (LAST_*/CROSS_PROJECT/DESIGN_COUNT already did).
[...]

PR #2472, opened 2026-08-07
v1.61.0.0 fix wave: guards failing open / silent failures (9 fixes, 4 community PRs absorbed)

### LINKED ISSUE #1985: /sync-gbrain code stage fails on existing sources (gbrain ≥0.42 --confirm-destructive guard); use reindex-code in place

### LINKED ISSUE #2006: question-preference-hook returns permissionDecision: "defer" on pass-through paths, interrupting AskUserQuestion cards
## Summary

The `question-preference-hook` currently uses a helper named `defer()` for normal pass-through cases, but that helper returns:

```json
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "defer"
  }
}
```

That is not equivalent to "no decision." In Claude Code's current hook contract, `defer` has a specific meaning: pause the tool call so an external caller can resume it later. Claude's docs describe it as intended for `claude -p` / SDK-style integrations, and say it is only honored in non-interactive mode.

For this hook's ordinary no-op paths, gstack should abstain instead: omit `permissionDecision` and optionally include `additionalContext`.

Relevant docs page: https://code.claude.com/docs/en/hooks

## Affected local version

- gstack version: `1.56.0.0`
- gstack commit: `cab774c` (`v1.56.0.0 Token-reduction Phase B + AUQ paranoid safety net (#1849)`)
- Branch: `main`
- Claude Code version observed locally: `2.1.177`
- Hook file: `hosts/claude/hooks/question-preference-hook.ts`
- Hook wiring: `~/.claude/settings.json`, `hooks.PreToolUse`, matcher `(AskUserQuestion|mcp__.*__AskUserQuestion)`

## User-visible symptom

Every `AskUserQuestion` card rendered and then failed with:

> Tool execution was interrupted

It happened regardless of model or conversation context.

The local patch changed only the pass-through helper so it no longer emits `permissionDecision: "defer"`. After that, a real `AskUserQuestion` card rendered and accepted a click successfully.

## Why this has broad impact

This hook is wired to all `AskUserQuestion` calls:

```json
{
  "matcher": "(AskUserQuestion|mcp__.*__AskUserQuestion)",
  "command": "~/.claude/skills/gstack/hosts/claude/hooks/question-preference-hook"
}
```

The helper is used for many normal cases, including:

- empty stdin
- JSON parse failure
- non-`AskUserQuestion` tool names
- questions with no marker
- no saved preference
- `always-ask` preference
- one-way safety override
- ambiguous recommendation parsing
- catch-all error handling

That means users can hit this even with zero saved `/plan-tune` preferences.

## Root cause

In `hosts/claude/hooks/question-preference-hook.ts`, the pass-through helper used to be:

```ts
  const out: Record<string, unknown> = {
    hookEventName: 'PreToolUse',
    permissionDecision: 'defer',
  };
  if (additionalContext) out.additionalContext = additionalContext;
  process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
  process.exit(0);
}
```

That code appears to intend "no enforcement, let the question proceed." But `permissionDecision: "defer"` means "pause this tool call for later resume," not "let it proceed normally."

Claude Code docs say:

- PreToolUse supports `allow`, `deny`, `ask`, and `defer`.
- `defer` means the tool call exits with `stop_reason: "tool_deferred"` so an external caller can collect input and resume later.
- `defer` is only honored in non-interactive mode with `claude -p`.
- For no-decision behavior, hooks should omit a decision or return no output.

## Minimal proposed fix

Change the pass-through helper so it emits no `permissionDecision`:

```ts
  const out: Record<string, unknown> = {
    hookEventName: 'PreToolUse',
  };
  if (additionalContext) out.additionalContext = additionalContext;
  process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
  process.exit(0);
}
```

The `deny()` helper should stay as-is, because `permissionDecision: "deny"` is valid and is the hook's intended enforcement path.

## Cleaner proposed fix

Rename the helper from `defer()` to something like `passThrough()` or `abstain()` to avoid confusing the hook's "no enforcement" path with Claude Code's real `permissionDecision: "defer"` behavior.

## Tests and docs that likely need updating

Search results in the installed gstack repo show no other live hook implementation still returning `permissionDecision: "defer"` after the local patch, but tests and one design note still expect/document it:

- `test/question-preference-hook.test.ts`
- `test/memory-cache-injection.test.ts`
- `test/skill-e2e-plan-tune-cathedral.test.ts`
- `docs/spikes/claude-code-hook-mutation.md`

Examples:

```ts
expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBe('defer');
```

Those expectations should become one of:

```ts
expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBeUndefined();
```

or, if the helper is changed to exit silently on paths without `additionalContext`, assert empty stdout for pure pass-through cases and assert `additionalContext` only where needed.

## Verification from local incident

Confirmed locally:

- The hook was installed and wired to `AskUserQuestion` in `~/.claude/settings.json`.
- gstack was at `1.56.0.0`, commit `cab774c`.
- Claude Code was `2.1.177`.
- The old helper returned `permissionDecision: "defer"` for normal no-op paths.
- The user had no saved preferences, so the enforcement path was not involved.
- The hook error log was empty, so this was not a hook crash.
- After changing the helper to omit `permissionDecision`, a real `AskUserQuestion` card rendered and accepted a click.

## Suggested labels

- bug
- hooks
- AskUserQuestion
- plan-tune

### LINKED ISSUE #2024: one-way-doors: revoke/reset credential patterns miss nouns rotate has — 'reset my secret' leaks through as two-way
## Observed problem

`scripts/one-way-doors.ts` is the secondary keyword safety net for AskUserQuestion calls that fire without a registry id. Its three credential patterns are meant to be parallel across `revoke`/`reset`/`rotate` (the inline comment says they "allow filler words between verb and noun", and the test header for #1839 states "the credential-rotation patterns must be parallel across revoke/reset/rotate").

They are not. The noun alternations differ:

```
/\brevoke\s+[\w\s]*\b(api key|token|credential|access key|password)\b/i,        // missing: secret
/\breset\s+[\w\s]*\b(api key|token|password|credential)\b/i,                     // missing: secret, access key
/\brotate\s+[\w\s]*\b(api key|token|secret|credential|access key|password)\b/i, // complete
```

So a destructive credential operation phrased with the "wrong" verb/noun pair classifies as **two-way** (auto-decidable) instead of one-way (always ask). This is a false-negative in a safety net — the exact failure class #1839 fixed for `rotate ... password`.

## Current behavior on upstream `main` (c7ae6320, v1.58.1.0)

```
$ bun -e 'import {classifyQuestion} from "./scripts/one-way-doors.ts"; ...'
two-way   revoke my secret      <- should be one-way
two-way   reset my secret       <- should be one-way
two-way   reset my access key   <- should be one-way
ONE-WAY   rotate my secret
```

## Expected behavior

`revoke`/`reset`/`rotate` should share an identical credential noun list, so every `<verb> my <noun>` combination across {api key, token, secret, credential, access key, password} classifies as one-way.

- open PRs and issues touching `one-way-doors` / "one-way door" — none cover this (only #1839/#1911 fixed `rotate ... password`, which is already on main)
- searched open PRs for "revoke reset rotate" and the credential-noun subsystem — no collision
- confirmed the existing `test/one-way-doors.test.ts` only checks the non-password nouns against `rotate`, never against `revoke`/`reset`

## Candidate fix shape

Unify the noun alternation across all three verbs (add `secret` to `revoke`; add `secret` and `access key` to `reset`), and extend the regression test to assert every verb × noun pair is one-way.

### LINKED ISSUE #2032: design variants: --count silently generates zero variants on NaN/0/negative input

### LINKED ISSUE #2035: AskUserQuestion silently broken: PreToolUse hook emits permissionDecision:'defer' → 'Tool result missing due to internal error'
## Summary

The PreToolUse hook for `AskUserQuestion` (`question-preference-hook`) emits `permissionDecision: "defer"` on its no-op path. Current Claude Code builds choke on that value and abort the tool call, so **every `AskUserQuestion` call fails** with:

```
[Tool result missing due to internal error]
```

This hits the common case: any AUQ call where no `never-ask` preference matches falls through to `defer()`, which is essentially all of them. The result is that the questions UI is silently broken for the user — the call is made, no answer ever comes back.

## Environment

- gstack: **v1.55.1.0**
- Claude Code: recent build (Opus 4.8 era, June 2026), VSCode extension
- OS: macOS (darwin 25.5.0)
- `bun` present and on PATH. The hook itself runs **cleanly** — exit 0, no entries in `~/.gstack/hook-errors.log`. The problem is the *output it emits*, not a crash.

## Root cause

`hosts/claude/hooks/question-preference-hook.ts` → `defer()`:

```ts
  const out: Record<string, unknown> = {
    hookEventName: 'PreToolUse',
    permissionDecision: 'defer',
  };
  if (additionalContext) out.additionalContext = additionalContext;
  process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
  process.exit(0);
}
```

`defer()` is the default/fall-through path for any AUQ call without a matching `never-ask` preference. On a normal question it emits:

```json
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"defer"}}
```

Empirically, current Claude Code builds reject this `permissionDecision: "defer"` and abort the tool call with an internal error. (The hooks docs are ambiguous on whether `"defer"` is an accepted `permissionDecision` value alongside `allow`/`deny`/`ask`; regardless, emitting it currently breaks the call, whereas the documented "no decision" form — exit 0 with no stdout — works.)

## Reproduction

1. gstack 1.55.1.0 with the PreToolUse hook registered in `~/.claude/settings.json` (`matcher: "(AskUserQuestion|mcp__.*__AskUserQuestion)"`).
2. Trigger any `AskUserQuestion` that has no matching `never-ask` preference.
3. The tool result is `[Tool result missing due to internal error]`. 100% reproducible.

## Isolation evidence

Neutralizing the shims to `exit 0` no-ops pinpoints the PreToolUse hook:

| PreToolUse | PostToolUse | Result |
|---|---|---|
| real | real | ❌ internal error |
| no-op | no-op | ✅ works |
| no-op | real | ✅ works → culprit is PreToolUse |
| fixed (below) | real | ✅ works |

## Suggested fix

The documented "no decision / proceed via the normal permission flow" output is **exit 0 with no stdout**. Only emit JSON when there is actually `additionalContext` to inject, and without a `permissionDecision`:

```ts
  // "Proceed via the normal permission flow" is exit 0 with no output.
  // Emitting permissionDecision:'defer' makes recent Claude Code builds reject
  // the tool call with an internal error, silently breaking every AUQ call.
  if (additionalContext) {
    process.stdout.write(
      JSON.stringify({
        hookSpecificOutput: {
          hookEventName: 'PreToolUse',
          additionalContext,
        },
      }),
    );
  }
  process.exit(0);
}
```

Verified locally: with this change, `AskUserQuestion` works with **both** gstack hooks active (PreToolUse + PostToolUse).

The `deny()` path is unaffected — `permissionDecision: "deny"` is a standard value and works fine; only the `defer()` no-op path needs changing.

### LINKED ISSUE #2039: /careful fails open on chained rm: 'rm -rf /; rm -rf node_modules' is silently allowed
### Summary

The `/careful` PreToolUse hook (`careful/bin/check-careful.sh`) fails open on chained `rm -r` commands: a destructive recursive delete is silently allowed when the command is chained with a second `rm` that targets a build-artifact directory.

### Observed problem

The "safe exception" block (lines 28-48) whitelists `rm -rf` of build artifacts (`node_modules`, `dist`, `.next`, etc.). It extracts the rm targets with a single greedy expression:

```sh
RM_ARGS=$(printf '%s' "$CMD" | sed -E 's/.*rm[[:space:]]+(-[a-zA-Z]+[[:space:]]+)*//;s/--recursive[[:space:]]*//')
```

`.*rm` matches greedily, so for a chained command it only ever inspects the targets of the **last** `rm` in the string. If that final rm targets a safe directory, the whole command is whitelisted and the hook returns `{}` (allow) without warning, even though an earlier rm in the chain is destructive.

### Current behavior on upstream `main` (c7ae6320)

```
$ printf '{"tool_input":{"command":"rm -rf /; rm -rf node_modules"}}' | bash careful/bin/check-careful.sh
{}
```

No warning. The `rm -rf /` is waved through because the greedy extraction only sees the trailing `node_modules`.

Reversing the order (safe rm first, dangerous rm last) correctly warns, which confirms the asymmetry is caused by the last-rm-only extraction:

```
$ printf '{"tool_input":{"command":"rm -rf node_modules; rm -rf /home/user/data"}}' | bash careful/bin/check-careful.sh
{"permissionDecision":"ask","message":"[careful] Destructive: recursive delete (rm -r). This permanently removes files."}
```

### Expected behavior

Any chained command containing a destructive recursive `rm` should warn (`permissionDecision: "ask"`), regardless of where the safe rm sits in the chain. Single-command artifact cleanups (`rm -rf node_modules`, `rm -rf .next dist`) should still be allowed.

### Candidate fix shape

The target-extraction shortcut is only reliable for a single `rm` invocation. Gate the safe-exception so it applies only when the command has no shell separator (`;`, `|`, `&`, newline). When a separator is present, fall through to the existing destructive-pattern check, which already warns on any `rm -r`. This is the fail-safe direction for a guard and preserves the single-command artifact-cleanup convenience.

### Duplicate searches performed

- Open PRs touching `careful/bin/check-careful.sh`: #1110 (text-output false positives — over-warning, the opposite direction; it splits on `&&`/`||` but turns `;` into a space and re-feeds the same greedy safe-exception, so it does not fix this) and #1509 (hook activation gate + `hookSpecificOutput` shape — does not touch the detection logic).
- Open issues: #1060 (false positives in commit messages — over-warning), #1091 (broad proposal to add structured execution judgment). Neither covers this fail-open.
- Verified the bug reproduces on current `main` (c7ae6320).

### LINKED ISSUE #2051: gstack misclassifies a gbrain thin-client (remote-HTTP MCP, no local engine) as broken-config, suppressing brain-aware planning blocks
## Summary

On a machine where gbrain runs as a **thin-client** — the engine lives on a remote host and the agent talks to it purely over a remote-HTTP MCP server, with no local Postgres/PGLite engine — `./setup` prints:

```
gbrain not detected — brain-aware blocks suppressed in planning-skill SKILL.md files.
```

and `gstack-config gbrain-refresh` reports `local-status: broken-config`. The brain is fully functional over MCP (reads/writes work in-session), but gstack decides it's absent and strips the brain-aware blocks from the planning skills (office-hours, plan-ceo-review, etc.).

## Root cause

Two independent assumptions in the detector both assume a **local** engine and a **hardcoded MCP name**:

**1. `detectMcpMode()` in `bin/gstack-gbrain-detect` hardcodes the MCP server name `gbrain`.**
All three tiers look up a server literally named `gbrain`:
- `claude mcp get gbrain --json`
- grep for `^gbrain:` in `claude mcp list`
- `mcpServers.gbrain` in `~/.claude.json`

A thin-client whose MCP server is registered under any other name returns `gbrain_mcp_mode: "none"`, even though a perfectly good remote-HTTP gbrain MCP is connected.

**2. `freshClassify()` in `lib/gbrain-local-status.ts` only probes a local engine.**
It runs `gbrain sources list --json` against a local DB. On a thin-client there is no local `database_url`, so the probe fails and the function hits its defensive default:

```ts
// Defensive default per codex #8: unrecognized failures classify as
// broken-config so the user sees the raw stderr surfaced upstream.
return "broken-config";
```

There is no state representing "no local engine, but a healthy remote brain over MCP."

Combined, a legitimate thin-client lands on `mcp_mode: none` + `local_status: broken-config` → "gbrain not detected" → brain-aware blocks suppressed, despite the brain being live.

## Impact

Thin-client / remote-engine setups (a documented gbrain topology) silently lose brain-aware planning in gstack, with a misleading "broken-config" signal that suggests the user's gbrain install is broken when it isn't.

## Suggested fix

- `detectMcpMode()`: don't hardcode the server name `gbrain`. Detect any MCP server whose endpoint/command resolves to gbrain, or allow the server name to be configured.
- `freshClassify()`: add a state for a healthy remote/thin-client brain (e.g. `remote-ok`). When a remote-HTTP gbrain MCP is present (or the config points at a remote endpoint with no local engine), treat the brain as available and keep the brain-aware blocks, instead of falling through to `broken-config`.

## Environment

- gstack v1.58.3.0
- gbrain v0.42.51.0 (thin-client; remote engine, remote-HTTP MCP)
- macOS

### LINKED ISSUE #2052: /context-restore can load the wrong worktree checkpoint (shared origin-derived slug + newest-across-all-branches restore)

### LINKED ISSUE #2067: gstack-developer-profile --read double-counts mode:"resources" rows into SESSION_COUNT, TIER, and NUDGE_ELIGIBLE
## `gstack-developer-profile --read` double-counts `mode:"resources"` bookkeeping rows into SESSION_COUNT, TIER, and NUDGE_ELIGIBLE

`bin/gstack-developer-profile` derives a builder's `SESSION_COUNT`, `TIER`, and `NUDGE_ELIGIBLE` from the `sessions[]` array in `~/.gstack/developer-profile.json`. But `/office-hours` appends **two** rows per run: the real session row (`mode: "startup"` or `"builder"`), and a separate `mode: "resources"` bookkeeping row recorded after resources are surfaced (`office-hours/SKILL.md:1620`, written via `--log-session`). The read path counts both.

This was masked while #1677 was live (the writer wrote to the legacy `builder-profile.jsonl` the reader never re-read, so `SESSION_COUNT` was stuck at 0). Now that the profile is unified onto `developer-profile.json` and sessions actually persist, the resources rows are read — so the bug is live: counts read roughly **2x** real sessions.

## Root cause (`bin/gstack-developer-profile`, current `main`)

The file already knows resources rows aren't sessions. Lines 236-238 filter them for `LAST_*`/`CROSS_PROJECT`, with a comment naming them "the Phase 6 auto-append":

```
const realSessions = sessions.filter(e => e.mode !== 'resources');
```

But that filter is applied to only 2 of the 4 consumers. The other two still read raw `sessions`:

- **Line 228:** `const count = sessions.length;` → `SESSION_COUNT` and `TIER` (thresholds `>=8` inner_circle, `>=4` regular, `>=1` welcome_back) are inflated.
- **Line 255:** `const builderSessions = sessions.filter(e => e.mode !== 'startup').length;` → doubly wrong: it **counts** `resources` rows (they're `!== 'startup'`) and **excludes** real `startup` sessions. Feeds `NUDGE_ELIGIBLE` (line 256: `builderSessions >= 3 && totalSignals >= 5`), so the builder→founder nudge can arm off bookkeeping alone.

## Repro

Fresh `$HOME`. Run `/office-hours` to completion 4 times as real design sessions.

- Each run appends 1 session row + 1 `mode:"resources"` row → `sessions.length == 8`.
- `gstack-developer-profile --read` reports `SESSION_COUNT: 8` / `TIER: inner_circle`.
- Truth: 4 real sessions, which should be `TIER: regular`.

Two real sessions already report `SESSION_COUNT: 4` / `regular` when the user is actually at `welcome_back`. The founder nudge can fire after as few as 3 runs on bookkeeping rows rather than 3 real `builder` sessions.

## Impact

Low-severity (greeting tier and the builder→founder nudge are personalization/tone, not crash or data loss), but it fires on the **default happy path** for every `/office-hours` user now that sessions persist: wrong-tier greeting and a premature "have you thought about whether this could be a company?" nudge.

## Fix

Hoist the existing `realSessions` filter above `count`, and make the nudge count real `mode:"builder"` sessions:

- `count = realSessions.length` → SESSION_COUNT / TIER / CROSS_PROJECT / designs all key off real sessions.
- `builderSessions = realSessions.filter(e => e.mode === 'builder').length` → nudge measures repeat *builders*, which is what it was meant to measure.

Write path untouched; this is a read-only aggregation change. A PR with this fix plus regression tests (count, tier both-sides-of-gate boundaries, cross-project, and nudge in the presence of resources rows) is up at #1991.

## Relationship to #1677

Distinct bug, sequential. #1677 was a write-path split-brain (`SESSION_COUNT` always 0 because writes went to a file the reader never read). Unifying the profile fixed that and **exposed** this one: now that rows persist and get read, the `mode:"resources"` rows get over-counted. #1677 = sessions don't persist; this = persisted sessions are over-counted.

### LINKED ISSUE #2473: office-hours-spec-review E2E deterministically hits error_max_turns (9 turns) on main
`office-hours-spec-review` (in `test/skill-e2e-plan.test.ts`, ~line 560) fails deterministically with `error_max_turns` at exactly 9 turns, everywhere:

- PR #2472 CI (e2e-plan job): 3/3 attempts, 9 turns, ~30-43s, ~$0.25 each
- CI eval-store comparison: FAIL → FAIL vs the previous baseline run (same 9-turn shape)
- Local run on pure `origin/main` @ a3259400: `error_max_turns`, 9 turns, 33s

The deterministic 9-turn ceiling on every run suggests the scenario's turn budget no longer fits the flow (a skill prose change or CC version shift added turns), rather than flakiness. Candidate next steps: dump the failing transcript (the harness drops reasoning/tool traces on failure — see the E2E observability TODO), check what the agent spends turns 1-9 on, and either raise `maxTurns` or trim the scenario's prompt scope.

Not caused by PR #2472 (receipts above — identical failure on main without the branch).

garrytan:
## v1.61.0.0 fix wave: guards failing open / silent failures

Fixes #2035, fixes #2006, fixes #2039, fixes #2024, fixes #2032, fixes #2067, fixes #2052, fixes #2051, fixes #1985. Also fixes #2473 (three E2E budget/starvation fixes landed here after CI diagnosis; all verified locally and green on CI run 3).

One bundled PR, 13 bisect commits. Every fix ships with a tripwire test **verified failing against the unfixed code first**, so the next recurrence of this bug class trips CI instead of shipping.

### The headline (#2035/#2006)

Every AskUserQuestion died with "Tool result missing due to internal error" on current Claude Code builds. `permissionDecision:'defer'` is a real PreToolUse value, but since CC v2.1.89 it means "pause this tool call for external resumption" (a headless feature), never "abstain" — interactive sessions have nothing to resume the paused call. The hook's pass-through is now exit 0 with exactly-empty stdout (or additionalContext-only output, which keeps plan-tune memory nuggets working). The protocol reference doc that taught the old behavior is corrected in the same commit. Deployment: `/gstack-upgrade` delivers it — the registered hook shim execs the TypeScript live, no settings change.

### Community PRs absorbed (land-and-own, authorship preserved)

| PR | Author | Fix | Our hardening on top |
|----|--------|-----|----------------------|
| #2040 | @jbetala7 | /careful chained-rm bypass | capital `-[rR]` + `$(`/backtick excluded from safe targets (both fail-opens verified by executing the script) |
| #2054 | @jbetala7 | /context-restore branch preference | — |
| #2031 | @jbetala7 | gbrain `--confirm-destructive` | drift remove routed through the #1734 data-loss guards + realpath drift check + drift logging |
| #1991 | @mvann | developer-profile resources-row exclusion + 8 regression tests | — (version-bump commit superseded by the wave's consolidated release commit) |

Thank you @jbetala7 and @mvann. The fork PRs can be closed with credit once this lands.

### Also closed as already-fixed (receipts in CHANGELOG)

#1965, #1950, #1964 — verified fixed in-tree with regression tests; close-out comments cite the exact test lines.

### Merge reconciliation with #2264 (time-attack wave)

Main landed overlapping guard fixes mid-wave. Reconciled semantically in the merge commit:
- /careful: adopted main's anchored full-command whitelist (stricter, catches comment-hiding) and re-applied this wave's two hardenings on top. Union of both waves' test batteries passes.
- one-way-doors: main's singular unification + this wave's plurals and `--summary-stdin` runtime wiring.
- gbrain-local-status: main's `engine-locked` (#2194) + this wave's `thin-client` (#2051) coexist; `--is-ok` keeps each wave's intent (engine-locked = stop, thin-client = usable).

### Verification

- Free suite: green (bun exit 0). Fail-first proof recorded per fix (4 careful, 4 gbrain-sources, 14 hook, 6 one-way-doors, 12 flag-utils rows failed pre-fix, pass post-fix). Parity suite, ship goldens (all 3 hosts), and gen-skill-docs freshness (`gen:skill-docs && git diff --exit-code`) all clean.
- Detached eval run (156 E2E + 20 judge tests selected by diff): EXIT=0. Includes the gate-tier plan-tune cathedral test exercising the hook end-to-end, and the five periodic tests this wave's touchfiles trigger.
- Slop: zero new findings (hook file scores identical pre/post-wave).

### Known residuals (flagged, with receipts)

- `plan-tune-inspect` E2E: retry-flaky. Branch 3/6 attempts failed vs main 0/3 — not statistically significant (Fisher p≈0.46); failure mode is the agent giving up after one slug-derivation stumble; the only plan-tune SKILL.md delta is one preamble line unrelated to the log-review flow. **Unverified — may or may not be related.** Passed on retry in the suite.
- `context-save-list-current-branch` E2E: FAIL→FAIL vs the previous eval run — pre-existing, not introduced here.
- `test/eval-list-cli.test.ts` (main's new infra): fails in any workspace with the `.claude/skills/gstack` dev symlink (its `getProjectEvalDir()` slug detection escapes the test's HOME sandbox via the repo-local symlink); passes in clean worktrees on this branch (2/2). Pre-existing isolation gap, not touched by this wave.
- `benchmark-cli` NOT READY + `observability` #11: fail on pure origin/main in the same environment — pre-existing.
- `design/test/variants-retry-after.test.ts` HTTP-date timing flake: reproduced on the pre-wave commit; filed into the design/test CI-wiring TODO.

### Follow-ups filed in TODOS.md

Wire `design/test/` into CI (all 8 files invisible to every runner), /context-save worktree-identity hardening (#2052 residual), conditional gbrain reindex-in-place gated on the new drift log.


garrytan:
**CI status: 22/23 green. The one red check is pre-existing, with receipts.**

`evals (e2e-plan)` fails on `office-hours-spec-review` — all 3 attempts `error_max_turns` at exactly 9 turns (~30-43s, ~$0.25 each). Attribution:

1. CI's own eval-store comparison marks it FAIL → FAIL (the previous baseline run failed with the identical 9-turn signature).
2. Ran the same test on pure `origin/main` (a3259400) locally: `error_max_turns`, 9 turns, 33s, $0.26 — byte-identical failure shape.

The scenario deterministically burns its turn budget on main and branch alike, so it broke independently of this PR (filed separately). Every other eval suite in this run is green, including both gate-tier runners and the two suites that exercise this PR's hook fix end-to-end (`plan-tune` cathedral in gate, `e2e-pty-plan-smoke`).

PR #2474, opened 2026-08-07
Fix: decision cards never appear in the Claude desktop app (AskUserQuestion parked by permissionDecision 'defer')

### LINKED ISSUE #2466: Decision cards never appear in the Claude desktop app (PreToolUse hook sends permissionDecision: 'defer')
## Summary

If you use gstack inside the **Claude desktop app**, AskUserQuestion decision cards never appear. The question is silently swallowed and the agent gets back `[Tool result missing due to internal error]`. The user just sees the agent go quiet, or ask a question in prose that they never got buttons for.

It works fine in the terminal CLI, which is probably why this hasn't surfaced before.

**Cause:** `hosts/claude/hooks/question-preference-hook.ts` emits `permissionDecision: 'defer'` on its no-enforcement path. In Claude Code, `defer` does not mean "no opinion" — it means *park this tool call so a later resume can pick it up*.

## Root cause detail

From the Claude Code 2.1.221 binary, the `defer` branch of the PreToolUse handler:

```js
case "defer": {
  if (!n.options.isNonInteractiveSession) {
    C(`Hook ${ce.hookName} returned permissionDecision=defer in interactive mode; ignoring (defer is print-mode only)`, {level:"warn"});
    break;                                   // ignored → card renders normally
  }
  ...
  return T.push({ message: al({ type: "hook_deferred_tool", ... }) })   // tool parked, never runs
}
```

So `defer` is only honored when the session is flagged `isNonInteractiveSession` — and the Claude desktop app is flagged non-interactive even with a human sitting in front of it. The AskUserQuestion call gets parked, never executes, and never returns a result.

The valid enum is `allow | deny | ask | defer`, so this passes schema validation and fails silently.

## Evidence

Bisected against a single session with all three AUQ hooks registered:

| Hooks live | Result |
|---|---|
| none | card renders ✅ |
| `question-preference-hook` only (emits `defer`) | `[Tool result missing due to internal error]` ❌ |
| all three, with fix below | card renders ✅ |

Across that machine's full transcript history: **3,700+ tool calls of every other kind resolved 100%. AskUserQuestion: 8 calls, 0 resolved.**

Reproduced on gstack `1.60.1.0` at `origin/main` (`a325940`), Claude Code 2.1.221, macOS.

## Why omitting is the correct fix (not switching to another value)

The hook's own docstring states the intent:

> `never-ask + one-way → defer (safety override; one-way always asks)`

If `defer` parked the tool, a one-way door would *never* reach the human — the exact inverse of the documented safety override. So the intent here is "no enforcement, proceed normally, let the user be asked". Claude Code encodes that by **omitting** `permissionDecision`; it is `.optional()` in the PreToolUse `hookSpecificOutput` schema:

```js
E.object({
  hookEventName: E.literal("PreToolUse"),
  permissionDecision: DM_().optional(),
  permissionDecisionReason: E.string().optional(),
  updatedInput: E.record(E.string(), E.unknown()).optional(),
  additionalContext: E.string().optional()
})
```

## Fix

In `defer()`, drop the `permissionDecision` key and keep everything else:

```ts
  const out: Record<string, unknown> = {
    hookEventName: 'PreToolUse',
  };
  if (additionalContext) out.additionalContext = additionalContext;
  process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
  process.exit(0);
}
```

This preserves everything that path was doing:

- `additionalContext` still flows, so plan-tune memory-nugget injection is unaffected
- the `deny` paths (never-ask auto-decide, and the Conductor prose redirect) are untouched — `deny` is a valid, honored value
- the PostToolUse logging hooks are unaffected

Only `allow` / `deny` / `ask` are safe to emit from a PreToolUse hook that expects the tool to actually run.

## Tests

14 assertions across `test/question-preference-hook.test.ts`, `test/memory-cache-injection.test.ts`, and `test/skill-e2e-plan-tune-cathedral.test.ts` assert `permissionDecision === 'defer'` and need updating to `toBeUndefined()`. After that, 93/93 pass across the five hook suites.

Happy to open a PR with the fix, the updated assertions, and a regression guard asserting the hook never emits `'defer'` if that's useful.

habassa5:
## Summary

Fixes #2466 — AskUserQuestion decision cards never appear when gstack runs in the **Claude desktop app**. The question is silently swallowed, the user sees nothing, and the agent gets back `[Tool result missing due to internal error]`.

`hosts/claude/hooks/question-preference-hook.ts` emitted `permissionDecision: 'defer'` on its no-enforcement path. In Claude Code, `defer` does **not** mean "no opinion" — it means *park this tool call so a later resume can pick it up*.

From the Claude Code 2.1.221 binary:

```js
case "defer": {
  if (!n.options.isNonInteractiveSession) {
    C(`Hook ${ce.hookName} returned permissionDecision=defer in interactive mode; ignoring (defer is print-mode only)`, {level:"warn"});
    break;                                   // ignored → card renders normally
  }
  ...
  return T.push({ message: al({ type: "hook_deferred_tool", ... }) })   // parked, never runs
}
```

`defer` is honored only when the session is flagged `isNonInteractiveSession` — and the Claude desktop app is flagged non-interactive even with a human present. So the AskUserQuestion call is parked, never executes, and never returns.

**This is why it went unreported:** in the terminal CLI `defer` is ignored and cards work fine, which is how most people run gstack. It reproduces only in the desktop app.

## The fix

Omit `permissionDecision` on the no-enforcement path. It is `.optional()` in the PreToolUse `hookSpecificOutput` schema:

```js
E.object({
  hookEventName: E.literal("PreToolUse"),
  permissionDecision: DM_().optional(),
  permissionDecisionReason: E.string().optional(),
  updatedInput: E.record(E.string(), E.unknown()).optional(),
  additionalContext: E.string().optional()
})
```

Omitting it is also what this path already *means*. The hook's own docstring states:

> `never-ask + one-way → defer (safety override; one-way always asks)`

If `defer` parked the tool, a one-way door would never reach the human — the exact inverse of the documented safety override. So the intent is "no enforcement, proceed normally, let the user be asked."

Only `allow` / `deny` / `ask` are safe to emit when the tool is expected to actually run.

## What is preserved

- `additionalContext` still flows → plan-tune memory-nugget injection unaffected
- `deny` paths untouched → never-ask auto-decide and the Conductor prose redirect both still work (`deny` is a valid, honored value)
- PostToolUse logging hooks unaffected

## Verification

Bisected on a real desktop-app session with all three AUQ hooks registered:

| Hooks live | Result |
|---|---|
| none | card renders ✅ |
| `question-preference-hook` only (emits `defer`) | `[Tool result missing due to internal error]` ❌ |
| all three, with this patch | card renders ✅ |

Across that machine's full transcript history before the fix: **3,700+ tool calls of every other kind resolved 100%; AskUserQuestion went 0-for-8.** After the fix, decision cards render and return normally, including multi-question and `multiSelect` cards.

## Tests

14 assertions across three suites encoded the old contract and are updated to `toBeUndefined()`. Added a regression guard (`never emits permissionDecision:defer`) covering the no-preference, no-marker, empty-stdin, and non-AUQ paths, plus the one-way-door safety override.

**93/93 passing** across the five hook suites against current `main` (v1.60.2.0):

```
bun test test/question-preference-hook.test.ts test/memory-cache-injection.test.ts \
         test/auq-error-fallback-hook.test.ts test/question-log-hook.test.ts \
         test/hook-scripts.test.ts
# 93 pass, 0 fail
```

### DIFF
--- a/hosts/claude/hooks/question-preference-hook.ts
+++ b/hosts/claude/hooks/question-preference-hook.ts
@@ -93,9 +93,24 @@ function readStdin(): Promise<string> {
 }
 
 function defer(additionalContext?: string): void {
+  // Deliberately does NOT emit permissionDecision:'defer'.
+  //
+  // In Claude Code, 'defer' does not mean "no opinion" — it means "park this
+  // tool call so a later resume can pick it up". It is ignored in interactive
+  // sessions, but HONORED whenever the session is flagged non-interactive,
+  // which includes the Claude desktop app even with a human present. A parked
+  // AskUserQuestion never runs and never returns a result, so the model sees
+  // "[Tool result missing due to internal error]" and the decision card never
+  // reaches the user.
+  //
+  // This function's contract is "no enforcement — proceed normally, let the
+  // user be asked" (see the never-ask + one-way safety override above, which
+  // requires the question to still reach the human). The correct encoding of
+  // that in Claude Code is to omit permissionDecision entirely; it is
+  // .optional() in the PreToolUse hookSpecificOutput schema. additionalContext
+  // is still delivered, so plan-tune memory injection is unaffected.
   const out: Record<string, unknown> = {
     hookEventName: 'PreToolUse',
-    permissionDecision: 'defer',
   };
   if (additionalContext) out.additionalContext = additionalContext;
   process.stdout.write(JSON.stringify({ hookSpecificOutput: out }));
--- a/test/memory-cache-injection.test.ts
+++ b/test/memory-cache-injection.test.ts
@@ -91,7 +91,7 @@ describe('memory injection', () => {
         ],
       },
     });
-    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBe('defer');
+    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBeUndefined();
     expect(r.parsed?.hookSpecificOutput?.additionalContext).toContain('verbose explanations');
   });
 
@@ -115,7 +115,7 @@ describe('memory injection', () => {
         ],
       },
     });
-    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBe('defer');
+    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBeUndefined();
     expect(r.parsed?.hookSpecificOutput?.additionalContext).toBeUndefined();
   });
 
@@ -219,7 +219,7 @@ describe('per-session memory cache', () => {
         ],
       },
     });
-    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBe('defer');
+    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBeUndefined();
     expect(r.parsed?.hookSpecificOutput?.additionalContext).toBeUndefined();
   });
 });
--- a/test/question-preference-hook.test.ts
+++ b/test/question-preference-hook.test.ts
@@ -126,7 +126,7 @@ describe('defers (no enforcement)', () => {
       },
     });
     expect(r.status).toBe(0);
-    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBe('defer');
+    expect(r.parsed?.hookSpecificOutput?.permissionDecision).toBeUndefined();
   });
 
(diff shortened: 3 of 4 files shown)


garrytan:

PR #2475, opened 2026-08-07
fix(browse): dispatch change event on fill for change-only validators

intelliot:
Playwright's `Locator.fill()` sets the value and dispatches an `input` event, but never `change`. Frameworks that validate on `change` (AngularJS `ng-change`, debounced strength/match checks — e.g. cPanel's Jupiter theme "Add FTP Account" password-match check) never see the update: the DOM value is correct but the framework's own validator still reports a mismatch, and submitting the form re-fails even after the field visually looks right.

Hit this today driving a real cPanel FTP-account form via `$B fill` — two identical password values kept failing "Passwords do not match" until switching to a driver that dispatches real per-keystroke events.

**Fix:** dispatch `change` immediately after `fill()`, without forcing `blur` (which would break `fill` → `press Enter` flows that rely on focus staying put).

**Testing:** added `browse/test/fill-change-event.test.ts` with a minimal change-only-listener fixture reproducing the bug class. Verified the new test fails without the fix (git stash confirmed) and passes with it. Full `bun run test` suite passes clean on this branch.

### DIFF
--- a/browse/src/write-commands.ts
+++ b/browse/src/write-commands.ts
@@ -377,11 +377,14 @@ export async function handleWriteCommand(
       const value = valueParts.join(' ');
       if (!selector || !value) throw new Error('Usage: browse fill <selector> <value>');
       const resolved = await session.resolveRef(selector);
-      if ('locator' in resolved) {
-        await resolved.locator.fill(value, { timeout: 5000 });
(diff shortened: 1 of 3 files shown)


garrytan:

PR #2483, opened 2026-08-08
fix(resolvers): empty find must not fall through to cwd

tranthanhnhatkhoa:
Follow-up to #2482, which fixed this defect in `bin/gstack-codex-session-import`. Same bug, three more sites, two of them user-visible.

## The bug

`find ... | xargs ls -t` runs `ls -t` with **no operands** when `find` matches nothing — and `ls -t` with no operands lists the **current directory**.

| site | effect when the dir is empty or absent |
|---|---|
| `scripts/resolvers/review.ts:859` | `PLAN` becomes a random `.md` from wherever `/review` was invoked, then gets **read as the plan** |
| `preamble/generate-context-recovery.ts:15` | "RECENT ARTIFACTS" lists unrelated cwd files |
| `preamble/generate-context-recovery.ts:24` | `LATEST_CHECKPOINT` points at a random file, which the skill is then told to read |

All three only misfire when the projects dir has no `ceo-plans` / `checkpoints` / `plans` yet — a **fresh install**, which is exactly where a wrong answer is least likely to be recognized as wrong. A new user's first `/review` can silently review against someone else's markdown.

Lines 857–858 in `review.ts` pipe to `grep -l`, which reads stdin when given no files and correctly matches nothing. Left alone.

## The fix

`-exec ls -t {} +` never runs the command when nothing matched. Same substitution as #2482, and the same reason for avoiding `xargs -r`: GNU-only, repo targets macOS, no existing use to follow.

## Regeneration

These are resolvers, so the change propagates into generated docs:

- `bun run gen:skill-docs` — 46 files
- the three ship goldens in `test/fixtures/golden/` refreshed by hand; they're byte-for-byte copies of generated output and have **no update script**, so `gen:skill-docs` leaves them stale and the golden tests fail until they're copied over

Worth knowing: `.agents/skills/gstack-ship/SKILL.md` and `.factory/skills/gstack-ship/SKILL.md` are gitignored, so `git status` shows nothing while their goldens go stale. Only the test catches it.

I checked every regenerated line — all are one of the three fixes or its comment. No unrelated drift.

## Verified

```
suite before   17 fails
suite after    17 fails      (no new failures)
golden tests   3 fail -> 0
```

The 17 are pre-existing and unrelated; #2480, #2481 and #2482 cover 15 of them, and the remaining 2 are environmental on my box (no `claude` on PATH for the test user, and `cdp-e2e` needs Chrome).

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -630,7 +630,7 @@ eval "$(~/.claude/skills/gstack/bin/gstack-slug 2>/dev/null)"
 _PROJ="${GSTACK_HOME:-$HOME/.gstack}/projects/${SLUG:-unknown}"
 if [ -d "$_PROJ" ]; then
   echo "--- RECENT ARTIFACTS ---"
-  find "$_PROJ/ceo-plans" "$_PROJ/checkpoints" -type f -name "*.md" 2>/dev/null | xargs ls -t 2>/dev/null | head -3
+  find "$_PROJ/ceo-plans" "$_PROJ/checkpoints" -type f -name "*.md" -exec ls -t {} + 2>/dev/null | head -3
   [ -f "$_PROJ/${_BRANCH}-reviews.jsonl" ] && echo "REVIEWS: $(wc -l < "$_PROJ/${_BRANCH}-reviews.jsonl" | tr -d ' ') entries"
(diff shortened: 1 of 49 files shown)


garrytan:

PR #2491, opened 2026-08-08
fix(codex): use flags that codex exec resume actually accepts

cappellaadrian:
## Problem

`/codex` consult mode saves a thread id and offers "Continue the conversation" on the next run. That resume path has never been able to work: `codex exec resume` accepts a much smaller flag set than `codex exec`, and the skill passes it two flags it rejects.

Verified against codex-cli 0.147.0:

```
$ codex exec --help | grep -E '^\s+-(s|C)'
  -s, --sandbox <SANDBOX_MODE>
  -C, --cd <DIR>

$ codex exec resume --help | grep -E '^\s+-(s|C)'
(no matches)
```

Result, using the exact command from the skill:

```
$ codex exec resume $TID "..." -C "$_REPO_ROOT" -s read-only -c 'model_reasoning_effort="medium"' --json
error: unexpected argument '-C' found
exit=2
```

## Safety angle

The `-s read-only` removal is not just a flag rename. Without an explicit sandbox override, a resumed session inherits the sandbox policy from the Codex CLI's own config file. So a conversation that starts read-only on round 1 could gain write access from round 2 onward. `-c 'sandbox_mode="read-only"'` pins it on every round, which matches the skill's stated "Never modify files. Codex runs in read-only sandbox mode" rule.

## Changes

- `-s read-only` becomes `-c 'sandbox_mode="read-only"'` on the resume call
- `-C "$_REPO_ROOT"` becomes a `cd "$_REPO_ROOT" &&` prefix on the resume call
- `< /dev/null` added to all Codex invocations in `/codex` and `/autoplan`. `codex exec` reads stdin and prints `Reading additional input from stdin...`, which can hang a non-interactive run
- Inline note documenting the flag asymmetry so the `-s` form does not get reintroduced

Templates edited, then `bun run gen:skill-docs --host all`. Split into two commits per the bisection convention in CLAUDE.md: template change, then regenerated output.

## Verification

End to end against a real Codex thread in a throwaway repo:

```
OLD (-s + -C):              exit=2, error: unexpected argument '-C' found
NEW (cd + -c sandbox_mode): exit=0, agent_message: OK2
thread continuity:          same thread_id returned
context retention:          asked what it replied in round 1, answered correctly
```

`bun test`: 5 failures before this change, the same 5 after, zero regressions. For transparency those 5 pre-existing failures on a clean `main` checkout are the 3 golden-file ship baselines, `gstack-uninstall > clean system outputs nothing to remove`, and `gen-skill-docs > package.json version matches VERSION file`. None are touched by this PR.

One test did legitimately catch a mistake in an earlier draft: my explanatory note originally spelled out the Codex home path, which tripped the `zero Codex paths` guard in `gen-skill-docs.test.ts`. Reworded to describe the file without the path.

## Not included

No VERSION bump or CHANGELOG entry, to avoid conflicting with your release flow. Happy to add both if you would like them in the PR.

### DIFF
--- a/autoplan/SKILL.md
+++ b/autoplan/SKILL.md
@@ -883,7 +883,7 @@ Override: every AskUserQuestion → auto-decide using the 6 principles.
   What alternatives were dismissed too quickly? What competitive or market risks are
   unaddressed? What scope decisions will look foolish in 6 months? Be adversarial.
   No compliments. Just the strategic blind spots.
-  File: <plan_path>" -C "$_REPO_ROOT" -s read-only --enable web_search_cached
+  File: <plan_path>" -C "$_REPO_ROOT" -s read-only --enable web_search_cached < /dev/null
   ```
   Timeout: 10 minutes
 
@@ -1000,7 +1000,7 @@ Override: every AskUserQuestion → auto-decide using the 6 principles.
   accessibility requirements (keyboard nav, contrast, touch targets) specified or
   aspirational? Does the plan describe specific UI decisions or generic patterns?
   What design decisions will haunt the implementer if left ambiguous?
-  Be opinionated. No hedging." -C "$_REPO_ROOT" -s read-only --enable web_search_cached
+  Be opinionated. No hedging." -C "$_REPO_ROOT" -s read-only --enable web_search_cached < /dev/null
   ```
   Timeout: 10 minutes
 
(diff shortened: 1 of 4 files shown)


cappellaadrian:
Closing: this is already fixed upstream in #1391 (v1.30.0.0). My local install was pinned at v0.16.1.0 from April and I diagnosed against that instead of current main. Sorry for the noise. The independent verification against codex-cli 0.147.0 does confirm the flag semantics your fix landed on: `codex exec resume` rejects both `-s/--sandbox` and `-C/--cd`, and the resumed thread keeps context correctly with `-c 'sandbox_mode="read-only"'`.

PR #1, opened 2026-03-11
docs: add README and CLAUDE.md

garrytan:
## Summary
- Add open source README with install instructions (project-level and user-level), copy-paste CLAUDE.md snippet for users, command reference, and architecture overview
- Add CLAUDE.md with dev workflow for contributors (install, test, build, deploy)

## Test plan
- [ ] Verify README renders correctly on GitHub
- [ ] Confirm install instructions work end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2492, opened 2026-08-08
fix: scope headed-Chromium profile per project, not machine-wide

himiaocc:
## Summary

Two separate bugs, both causing the same visible symptom (headed browser opens then vanishes within seconds — reported as a persistent crash by a user running two sibling project windows concurrently, both driving a captcha-gated login).

**1. Chromium profile shared machine-wide, not per-project**
- `resolveChromiumProfile()` defaulted to a single `$HOME/.gstack/chromium-profile`, even though each browse daemon is already isolated per git repo root (`resolveConfig().stateDir`).
- Every daemon restart runs `killOrphanChromium()` (cli.ts), which reads the profile's `SingletonLock` and kills whatever PID holds it — no way to distinguish a genuinely dead orphan from another project's live headed window.
- Fix: default now resolves to `<project stateDir>/chromium-profile`. Also fixed two call sites (`browser-manager.ts` `handoff()`, `cli.ts` `chromiumProfileDir()`) that had their own hardcoded copies of the old global path instead of going through the shared resolver.

**2. Parent-process watchdog was killing `handoff()` sessions within ~15s (found during verification of fix #1 — turned out to be the dominant cause, reproduces with zero concurrency)**
- `handoff()` transitions an already-running headless daemon to headed mode in-place. `BROWSE_PARENT_PID` on that daemon was set at its original headless spawn — some short-lived CLI process that's expected to exit almost immediately (headless mode's watchdog branch already tolerates this).
- Once `handoff()` flips `connectionMode` to `'headed'`, the watchdog's next 15s tick took the OTHER branch: "headed + parent gone → shutdown", intended for `/pair-agent`/`/open-gstack-browser` where `BROWSE_PARENT_PID` is deliberately a real long-lived owner. Applied to a handoff-transitioned daemon, the already-dead CLI parent triggered a shutdown almost immediately after every handoff succeeded.
- Fix: track whether headed mode was reached via `handoff()` specifically (`BrowserManager.handoffTransition` / `didHandoffTransition()`) and have the watchdog treat "headed via handoff, parent gone" the same as plain headless (stay alive).

## Test plan
- [x] `bun test browse/test/config.test.ts browse/test/handoff.test.ts browse/test/browser-manager-unit.test.ts browse/test/watchdog.test.ts browse/test/server-factory.test.ts` — all pass (config 42, handoff+browser-manager+config combined 81, watchdog 4 incl. new regression test, server-factory 33)
- [x] Verified via `git stash` that a full `bun test browse/test/` run truncates identically on unmodified `main` — pre-existing test-runner quirk, unrelated to this change
- [x] New subprocess-based regression test in `watchdog.test.ts` (`BROWSE_TEST_FORCE_HANDOFF_HEADED` env hook) proving the daemon stays alive when its parent dies while in handoff-headed state
- [x] Real-world verification against a live captcha-gated site: ran `handoff`, then waited 75s (5 watchdog ticks) checking `status` + actually driving the browser with `text` — same PID throughout, still headed, still responsive. (First verification pass only checked the immediate `handoff()` return value and missed bug #2 entirely — this is the corrected, patient verification.)

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -186,6 +186,15 @@ export class BrowserManager {
   // ─── Handoff State ─────────────────────────────────────────
   private isHeaded: boolean = false;
   private consecutiveFailures: number = 0;
+  // Set true only by handoff() (headless daemon transitioning to headed
+  // in-place via the CLI `handoff` command). Distinguishes that case from
+  // launchHeaded()/connect (daemon spawned headed from the start, e.g. by
+  // pair-agent with a real long-lived parent to track). The server's
+  // parent-process watchdog uses this to tell "the CLI process that
+  // happened to send this HTTP request is gone (expected — every CLI
+  // invocation is short-lived)" from "the process that's supposed to own
+  // this headed browser's lifecycle is gone (should shut down)".
+  private handoffTransition: boolean = false;
 
   // ─── Watch Mode ─────────────────────────────────────────
   private watching = false;
@@ -253,6 +262,16 @@ export class BrowserManager {
   public onDisconnect: ((exitCode?: number) => void | Promise<void>) | null = null;
 
   getConnectionMode(): 'launched' | 'headed' { return this.connectionMode; }
+  didHandoffTransition(): boolean { return this.handoffTransition; }
+
+  /** TEST-ONLY: simulate the state handoff() leaves behind, without
+   * launching a real browser — lets watchdog.test.ts exercise the
+   * "headed via handoff, parent gone" branch as a fast subprocess test.
+   * Never called by any real code path. */
+  __forceHandoffHeadedForTest(): void {
+    this.connectionMode = 'headed';
+    this.handoffTransition = true;
+  }
 
   // ─── Watch Mode Methods ─────────────────────────────────
   isWatching(): boolean { return this.watching; }
@@ -1608,6 +1627,7 @@ export class BrowserManager {
       this.pages.clear();
       this.tabSessions.clear();
       this.connectionMode = 'headed';
+      this.handoffTransition = true;
 
       // Same Layer C stealth as launch()/launchHeaded(). Must run BEFORE
       // restoreState() navigates so the init scripts apply to the restored
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -1568,7 +1568,13 @@ export class BrowserManager {
         console.log('[browse] Handoff: extension not found — headed mode without side panel');
       }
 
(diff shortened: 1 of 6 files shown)


frederik-kaster-noygear:
Verified this reproduces, and the `killOrphanChromium()` diagnosis in your summary matches what I landed on from a different direction. Some measurements in case they are useful for review.

**The ProcessSingleton guard is doing real work, and removing the lock removes it.** Reproduced on Linux with throwaway profiles and Playwright's bundled Chromium:

| Scenario | Result |
|---|---|
| Second `launchPersistentContext()` on an in-use profile, locks intact | Refused. `Failed to create .../SingletonLock: File exists (17)`, then `Failed to create a ProcessSingleton for your profile directory ... Aborting now to avoid profile corruption` |
| Same, after deleting `Singleton{Lock,Socket,Cookie}` first | Both instances run. 8 processes on the profile becomes 16. The second logs `database is locked` (UKM), `Failed to initialize database` (top_sites), `Unable to open the password store database`, `Could not create/open login database`, `GCM Store/LOCK` |

One incidental thing, since you are editing that docstring: a clean Playwright `context.close()` removes the three files itself, so their absence on an idle profile is normal and not evidence that something deleted them. Only a hard kill leaves them behind. That tripped me up for a while.

**A hole this PR narrows but does not close.** Per-project profiles fix the cross-project case, which is the reported symptom. Within a single project, `emergencyCleanup()` and `shutdown()` in `server.ts` still call `cleanSingletonLocks(resolveChromiumProfile())` unconditionally, with no check that this process ever launched a browser:

* a daemon started with `BROWSE_HEADLESS_SKIP=1` (four test files use it) reaches shutdown having never opened a browser, and deletes the locks anyway
* so does one whose launch threw, because the shutdown cleanup sits outside the `if (!skipBrowser)` block that guards the launch

Confirmed by observation rather than reading: point such a daemon at a profile held by a live headed browser, send SIGINT, and the profile is left with zero of its three lock files. The next launch then succeeds instead of being refused.

Resolution steps 1 and 2 also still permit deliberate sharing, since `CHROMIUM_PROFILE` is how gbrowser's gbd points per-workspace daemons at a chosen profile.

**Offer.** I have a follow-up that gates the shutdown-path cleanup on ownership: `BrowserManager` records the profile it actually launched into at both `launchPersistentContext()` sites and releases that record only on a clean `close()`, so a close that timed out still clears locks that really are stale. Both shutdown paths then call `cleanOwnedSingletonLocks()`, which no-ops without ownership. It touches the same four files as this PR, so it is strictly a follow-up. Happy to rebase it once this lands, or to fold it in here if you would rather review one change. Not opening a competing PR either way.

PR #2498, opened 2026-08-09
fix(redact-prepush): six paths where the hook exits 0 on a real credential

lubosxyz:
The pre-push credential scanner can be made to allow a push that carries a live key. Six separate paths, same outcome: `exit 0`, nothing scanned, secret on the remote.

| # | trigger | shipped | fixed |
|---|---|---|---|
| 1 | credential on a line whose **content** starts with `++` | ALLOW | BLOCK |
| 2 | malformed stdin (unparseable ref update) | ALLOW | BLOCK(unscannable) |
| 3 | `diff.external` set | ALLOW | BLOCK |
| 4 | push to a remote other than `origin` | ALLOW | BLOCK |
| 5 | junk remote sha | ALLOW | BLOCK(unscannable) |
| 6 | `localSha="0"` read as a branch delete | ALLOW | BLOCK(unscannable) |

**#1** is the one I'd fix first even alone: added lines are filtered with `!line.startsWith("+++")`, which is meant to drop the `+++ b/file` header — but a genuine added line whose content begins with `++` renders as `+++…` too. A committed `.patch`/`.diff` carrying a key sails through unscanned.

**#3** needs no attacker: `diff.external` is ordinary user config. With it set, `git diff` emits no `+` lines, the scanner reads an empty diff, and everything is allowed. Silently.

## The false positive that teaches the bypass

Over the engine's 1 MiB cap the hook reports `engine.input_too_large` as HIGH and tells the user to **"Rotate the credential (a pushed secret is compromised)"** — for content it never scanned. On one machine that produced five entries in `~/.gstack/security/prepush-skip.jsonl`, including *"0 HIGH at 1.33 MB"* and *"synthetic credential-scrubbing fixture"*. None held a credential.

That matters beyond ergonomics: a gate that cries wolf trains its user to reach for `GSTACK_REDACT_PREPUSH=skip`, and the habit is then applied to the push where the gate was right. The fix scans in windows so large-but-clean input passes, and reserves blocking for content that genuinely could not be read.

## Fail-closed is preserved

An unreadable range still blocks (#1946 stands). What changed is the message: instead of `died of signal 9` with no explanation, it names the range, the reason, how many MB were read, and the concrete next step.

## Verification

`test/redact-prepush-fail-open-verify.sh` — 15 scenarios, each run against **both** the shipped and the fixed scanner, asserting an expected verdict per row and exiting non-zero on mismatch. Rows 2, 4, 5, 6, 9, 14 must agree between the two columns: that is the no-detection-regression check, including credentials straddling the window cut.

```
GATE: PASS (15/15)
```

I verified this independently of the change's author, against `bin/gstack-redact-prepush` and `lib/redact-engine.ts` that are **byte-identical** (sha256 `5c06f7de…` / `c1f5df32…`) to current `main` at v1.61.0.0, so the result applies to this base.

## Relationship to open PRs

Disjoint from #2398 (pushed range vs whole repo), #2328 (path-ignore) and #2358 (chained `pre-push.local` stdin newline) — I diffed all three and none touches these six paths. #2328 is motivated by the same false positive from the other direction; #2358 is a seventh fail-open path in the wrapper that this PR does not address.

Happy to rebase behind any of them if you'd rather land those first.

### DIFF
--- a/bin/gstack-redact-prepush
+++ b/bin/gstack-redact-prepush
@@ -22,46 +22,141 @@
  *   - MEDIUM → warn (non-blocking). LOW/WARN → silent.
  *   - GSTACK_REDACT_PREPUSH=skip → log + exit 0 (escape valve).
  *
- * Installed/uninstalled via `gstack-redact install-prepush-hook` (see the
- * gstack-redact CLI), which chains any pre-existing hook.
+ * STREAMING (2026-08): the diff is consumed incrementally and scanned in bounded
+ * windows instead of being buffered whole. This fixes three defects that made the
+ * hook an unreliable gate on large repos — see
+ * docs/prepush-redact-hook-2026-08/README.md in rejstrik-ai for the measurements:
+ *
+ *   1. `scan()` refuses input over its own 1 MiB cap (DEFAULT_MAX_BYTES) by
+ *      returning a synthetic HIGH `engine.input_too_large`. The hook reported that
+ *      as a credential hit, so ANY push adding >1 MiB of text was blocked with
+ *      "Rotate the credential (a pushed secret is compromised)" — a false positive
+ *      on innocuous data, and the added lines were never actually scanned.
+ *      Windowing keeps every scan() call under the cap, so the content is really
+ *      scanned. This scans MORE than before, never less.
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2500, opened 2026-08-10
Probe AGENTS.md for skill routing; resolve GSTACK_ROOT in team-init

gamerey43:
## What

Two small fixes to how gstack locates itself and detects routing, plus a regression test:

1. **The preamble's `HAS_ROUTING` probe now checks `AGENTS.md` as well as `CLAUDE.md`.** Every skill's preamble runs:

   ```bash
   _HAS_ROUTING="no"
   for _RF in CLAUDE.md AGENTS.md; do
     if [ -f "$_RF" ] && grep -q "## Skill routing" "$_RF" 2>/dev/null; then
       _HAS_ROUTING="yes"
     fi
   done
   ```

   Non-Claude hosts (Codex, Cursor, generic agent harnesses) route skills via `AGENTS.md` — the de-facto cross-harness convention file. Today a repo with `AGENTS.md` routing but no `CLAUDE.md` reports `HAS_ROUTING: no` and gets nagged to create `CLAUDE.md`. This fixes that, and the routing-injection prompt stops firing for repos that already route via `AGENTS.md`.

2. **`gstack-team-init` resolves `GSTACK_ROOT` instead of hardcoding `~/.claude/skills/gstack`.** Required-mode enforcement (the CLAUDE.md verification line and the generated `.claude/hooks/check-gstack.sh`) now resolves the install the same way skill preambles do: `$GSTACK_ROOT` env → `~/.codex/skills/gstack` → `~/.gstack/repos/gstack` → legacy `~/.claude/skills/gstack`. Install instructions point at the canonical global location throughout. This stops required mode from false-blocking on installs that keep the repo at `~/.gstack/repos/gstack` (the layout the setup produces on non-Claude hosts).

3. **Regression test for the probe** (`test/routing-probe.test.ts`): extracts the *live* generated block from `review/SKILL.md` and executes it across the three scenarios — both files routed, `AGENTS.md`-only, neither — plus a structural guard that the probe references `AGENTS.md`. The `AGENTS.md`-only case fails against the old probe, so a generator edit that breaks `AGENTS.md` detection is caught even when regeneration is clean.

## Why

- A repo routing via `AGENTS.md` should be treated as routed, not prompted to add `CLAUDE.md`.
- Required mode should enforce against the real install layout, not a path that only exists for Claude Code installs.
- The probe's behavior is now pinned by a test, so the AGENTS.md support can't silently regress.

## Testing

- `bun test test/routing-probe.test.ts` — 4 pass.
- Mutation check: the old CLAUDE.md-only probe returns `no` on an AGENTS-only repo, which the new test fails on.
- `gstack-team-init required` exercised in a throwaway repo: hook emits `{}` (allow) with the patched resolution on a global install, `deny` with a fake empty `HOME`.
- All skill docs regenerated via `bun run gen:skill-docs` (default + `--host codex`).


garrytan:

PR #2506, opened 2026-08-10
fix(config): make key validation locale-independent

### LINKED ISSUE #2494: gstack-config rejects every key containing the letter "i" under Turkish locale (a-z range is collation-dependent)
## Summary

`bin/gstack-config` validates keys with `[a-zA-Z0-9_]`. Under a Turkish locale (`LANG=tr_TR.UTF-8`) the `a-z` range is resolved by **collation order**, and the letter `i` falls outside it (Turkish treats dotless `ı` and dotted `i` as separate letters). Every config key containing an `i` is rejected — which is most of them.

Worse, the failure is silent: skill preambles read config with `2>/dev/null || echo <default>`, so the error is swallowed and **every stored preference falls back to its default**.

## Environment

- gstack `1.61.0.0` (commit `94993f7`)
- `LANG=tr_TR.UTF-8`, `LC_ALL` and `LC_COLLATE` unset
- GNU grep 3.12, bash 5.x, Linux

## Reproduction

```bash
$ LANG=tr_TR.UTF-8 gstack-config get proactive
Error: key must contain only alphanumeric characters, underscores, and an optional @<endpoint-id> suffix

$ LANG=tr_TR.UTF-8 gstack-config get telemetry
anonymous
```

Minimal repro of the underlying cause — which ASCII lowercase letters fail `[a-z]` under `tr_TR`:

```bash
$ LANG=tr_TR.UTF-8 bash -c 'for ch in a b c d e f g h i j k l m n o p q r s t u v w x y z; do
    printf "%s" "$ch" | grep -qE "^[a-z]+$" || printf "%s " "$ch"; done; echo'
i
```

Exactly one letter: `i`.

## Which keys break

| Works | Rejected |
|---|---|
| `telemetry`, `auto_upgrade` — neither contains `i` | `proactive`, `question_tuning`, `explain_level`, `checkpoint_mode`, `codex_reviews`, `cross_project_learnings`, `skip_eng_review`, `artifacts_sync_mode`, `routing_declined`, `gstack_contributor`, `update_check`, `plan_tune_hooks`, `checkpoint_push`, `skill_prefix`, `workspace_root`, … |

Both the `get` and `set` paths are affected (`bin/gstack-config:267` and `:283`).

## Why this is worse than it looks

Skill preambles all use this pattern:

```bash
_PROACTIVE=$(~/.claude/skills/gstack/bin/gstack-config get proactive 2>/dev/null || echo "true")
_QUESTION_TUNING=$(~/.claude/skills/gstack/bin/gstack-config get question_tuning 2>/dev/null || echo "false")
```

stderr is discarded and the non-zero exit hits the `||` fallback, so a Turkish-locale user gets defaults everywhere and **no indication anything is wrong**. In a session on this machine the preamble reported `PROACTIVE: true`, `EXPLAIN_LEVEL: default`, `CHECKPOINT_MODE: explicit`, `QUESTION_TUNING: false` — none of which were the stored values; all were fallbacks. `gstack-config set` also fails, so the user cannot change them either. Silent both ways.

This likely affects any locale whose collation reorders ASCII letters, not just `tr_TR`.

## Root cause

`bin/gstack-config:267` and `:283`:

```bash
if ! printf '%s' "$KEY" | grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'; then
```

Bracket **ranges** (`a-z`) are locale-collation dependent in POSIX. Bracket **classes** (`[:alnum:]`) are classification dependent and behave correctly.

## Proposed fix

```diff
-    if ! printf '%s' "$KEY" | grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'; then
+    if ! printf '%s' "$KEY" | grep -qE '^[[:alnum:]_]+(@[[:alnum:]]+)?$'; then
```

Applied to both sites. Verified on this machine: all keys read and write correctly under `tr_TR.UTF-8` with no `LC_*` override.

One caveat worth a maintainer decision: `[[:alnum:]]` under a UTF-8 locale also accepts non-ASCII letters (e.g. `ü`, `ç`), so it is slightly more permissive than the original intent. If strict ASCII is required, `LC_ALL=C grep` or a `case` glob would be tighter:

```bash
case "$KEY" in
  *[!a-zA-Z0-9_@]*) ...reject... ;;
esac
```

(bash `case` globs are also collation-sensitive, so `LC_ALL=C` is the safer explicit form.)

## Suggested follow-up beyond this file

The silent-fallback pattern is the real amplifier. Even with the regex fixed, `2>/dev/null || echo <default>` means any future `gstack-config` failure disappears. This release notes a "silent failure" theme — this looks like the same class: a guard that reports success while doing nothing.

A quick `rg 'gstack-config get' --glob '*/SKILL.md'` will show how widespread the pattern is.

Until this lands, either apply the one-line patch above, or prefix calls with `LC_COLLATE=C`. Note that a shell-wide `LC_COLLATE=C` also changes sorting for every other tool, so the patch is the narrower fix.

Math1987:
## Summary

- pin both gstack-config key validators to the C locale
- preserve the strict ASCII key contract instead of widening it with a locale-aware character class
- add a cross-platform regression tripwire plus get/set behavior coverage

Fixes #2494

## Why

POSIX bracket ranges follow locale collation. With GNU grep under tr_TR.UTF-8, the ASCII letter i can fall outside [a-z], so valid config keys are rejected and skill preambles silently fall back to their defaults.

## Tests

- fail-first: the new regression test failed on main because neither validator pinned LC_ALL
- `bun test test/gstack-config-key-locale.test.ts test/user-slug-fallback.test.ts test/docs-config-keys.test.ts test/explain-level-config.test.ts test/gstack-config-redact-keys.test.ts` — 35 pass, 0 fail
- `bash -n bin/gstack-config`
- `git diff --check`

## Full-suite note

I also attempted `bun test`, but it did not complete cleanly in my multi-repository macOS workspace: unrelated existing suites timed out or depended on unavailable local services and fixture assumptions (including global discovery, gbrain fixtures, and local server binding). The focused config suite above is green.

### DIFF
--- a/bin/gstack-config
+++ b/bin/gstack-config
@@ -264,7 +264,7 @@ case "${1:-}" in
     # endpoint-namespaced keys introduced by the brain-aware planning layer).
     # Endpoint ids are sha8/sha16 hex for remote MCP URLs, or the literal
     # "local" for stdio/PGLite engines (see endpoint_hash).
-    if ! printf '%s' "$KEY" | grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'; then
+    if ! printf '%s' "$KEY" | LC_ALL=C grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'; then
       echo "Error: key must contain only alphanumeric characters, underscores, and an optional @<endpoint-id> suffix" >&2
       exit 1
     fi
@@ -280,7 +280,7 @@ case "${1:-}" in
     VALUE="${3:?Usage: gstack-config set <key> <value>}"
     # Validate key (alphanumeric + underscore + optional @<endpoint-id> suffix).
     # Accepts hex hashes and the literal "local" from endpoint_hash.
-    if ! printf '%s' "$KEY" | grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'; then
+    if ! printf '%s' "$KEY" | LC_ALL=C grep -qE '^[a-zA-Z0-9_]+(@[a-zA-Z0-9]+)?$'; then
       echo "Error: key must contain only alphanumeric characters, underscores, and an optional @<endpoint-id> suffix" >&2
       exit 1
     fi
--- a/test/gstack-config-key-locale.test.ts
+++ b/test/gstack-config-key-locale.test.ts
@@ -1,0 +1,76 @@
+/**
+ * Locale-independent key validation tests for bin/gstack-config.
+ *
+ * POSIX bracket ranges such as a-z follow the active collation order. Under
+ * GNU grep with tr_TR.UTF-8, that excludes the ASCII letter i and silently
+ * breaks most stored preferences. macOS BSD grep does not reproduce the bug,
+ * so the source-level tripwire pins the C-locale boundary on every platform.
+ */
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { spawnSync } from "child_process";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+
+const ROOT = path.resolve(import.meta.dir, "..");
+const CONFIG = path.join(ROOT, "bin", "gstack-config");
+
+let stateRoot: string;
+
+function run(args: string[]) {
+  const result = spawnSync(CONFIG, args, {
+    encoding: "utf8",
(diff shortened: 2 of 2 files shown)


Math1987:
Thanks — this fix and its regression test were absorbed in #2571 (commit 008dd65), with attribution to @Math1987 / #2506. Closing this PR as superseded.

PR #2510, opened 2026-08-10
fix(sync): run gstack-brain-sync through bash, not cmd.exe, on Windows

ShahriarLak:
## The bug

On Windows the `brain-sync` stage of `/sync-gbrain` fails on **every** run:

```
'C:\Users\...\.claude\skills\gstack\bin\gstack-brain-sync' is not recognized as an internal or external command,
operable program or batch file.

  ERR   brain-sync   gstack-brain-sync exited 1 (0.0s)
```

## Why the #1731 fix didn't cover it

#1731 added `shell: NEEDS_SHELL_ON_WINDOWS` to these spawns. That is correct for the `gbrain.cmd` shim and does nothing here: `shell: true` routes through **cmd.exe**, which resolves `.cmd`/`.bat` via PATHEXT but has no concept of a shebang, so an extension-less bash script is rejected outright.

A `.cmd` shim needs a **shell**. A shebang script needs an **interpreter**. The two cases look identical and are not, which is why one fix appeared to cover both.

## Why it matters more than one red line

`artifacts_sync_mode` defaults to pushing curated artifacts to git. With this stage dead, a Windows user's learnings accumulate in `~/.gstack` and are never committed, while the sync report shows one red line among otherwise green stages. It reads as a minor stage failure and is actually silent data stagnation.

## The change

New `bashScriptInvocation()` in `lib/gbrain-exec.ts`:

- Resolves Git for Windows' `bash.exe` and passes the script as `argv[0]`.
- Prefers Git bash over a bare `bash` on PATH: **WindowsApps ships a `bash.exe` that is the WSL launcher**, and if it wins PATH order it reads `C:\...` as a Linux path and the script never sees the repo.
- Forward-slashes the script path, since bash treats backslashes as escapes.
- `GSTACK_BASH` overrides for unusual installs.
- Returns `null` when no bash exists, so the stage reports `skipped (no bash found; set GSTACK_BASH to your Git bash.exe)` rather than an unactionable spawn error.

POSIX is unchanged: the script is exec'd directly, no interpreter, no shell.

## Tests

The #1731 tripwire asserted the shape that does not work (`spawnSync(brainSyncPath, ...)` + `shell: NEEDS_SHELL_ON_WINDOWS`), so it now asserts the opposite and fails if a raw spawn returns. Six unit tests cover the resolver: POSIX passthrough, Git-bash resolution, path slashing, `shell: false`, `GSTACK_BASH` override, and the no-bash `null`.

The `gbrain.cmd` assertions in that file are untouched — that part of #1731 is right and still enforced.

## Verification (Windows 11, gstack 1.61.0.0, bun 1.3.11)

Before: `ERR brain-sync   gstack-brain-sync exited 1 (0.0s)`
After: `OK    brain-sync   curated artifacts pushed (4.2s)` — and the artifacts repo committed and pushed on its own.

The suite is nondeterministic on Windows (the same tree gave 743 and 741 failures on back-to-back runs), so I compared the affected files rather than the whole run: **14 failures before, 14 after**, with 6 new passing tests. No behaviour change on POSIX, where the invocation is byte-identical to today's.

### DIFF
--- a/bin/gstack-gbrain-sync.ts
+++ b/bin/gstack-gbrain-sync.ts
@@ -40,7 +40,7 @@ import { detectEngineTier, withErrorContext, canonicalizeRemote } from "../lib/g
 import { ensureSourceRegistered, sourcePageCount, parseSourcesList, cycleCompleted, type CycleStatus } from "../lib/gbrain-sources";
 import { detectAutopilot, decideSourceRemove, decideCodeSync } from "../lib/gbrain-guards";
 import { localEngineStatus, type LocalEngineStatus } from "../lib/gbrain-local-status";
-import { buildGbrainEnv, spawnGbrain, execGbrainJson, NEEDS_SHELL_ON_WINDOWS } from "../lib/gbrain-exec";
+import { buildGbrainEnv, spawnGbrain, execGbrainJson, NEEDS_SHELL_ON_WINDOWS, bashScriptInvocation } from "../lib/gbrain-exec";
 import { checkOwnedStagingDir } from "../lib/staging-guard";
 
 // ── Types ──────────────────────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


garrytan:

PR #2515, opened 2026-08-11
fix(browse): sidebar Terminal — duplicate WS subprotocol header + doubled CJK IME input

mindsurf0176:
Two independent bugs kept the sidebar Terminal pane unusable. Both verified end-to-end against a live agent in Chrome for Testing 145 / Bun 1.3.6.

## 1. Duplicate Sec-WebSocket-Protocol header breaks the WS handshake

**`/pty-session` mints fine, but `new WebSocket(url, ['gstack-pty.' + token])` never completes** — the pane sits at an empty cursor and typing does nothing.

Bun >= 1.3 auto-echoes the first offered subprotocol in the `server.upgrade()` 101 response. The agent *also* sets an explicit `Sec-WebSocket-Protocol` header (needed back when Bun didn't auto-echo), producing a duplicate response header:

```
HTTP/1.1 101 Switching Protocols
Sec-WebSocket-Protocol: gstack-pty.<token>   ← Bun auto-echo
Sec-WebSocket-Protocol: gstack-pty.<token>   ← manual echo
```

RFC 6455 allows at most one selected subprotocol, so strict clients reject the handshake — reproduced with python `websockets` (`InvalidHeader: multiple values`).

**Fix:** drop the manual echo; Bun's auto-echo selects the right protocol (the sidebar offers exactly one).

## 2. Korean/CJK IME input is sent twice

The local workaround for #1272 sent `e.data` manually on `compositionend`. But this xterm build's `_finalizeComposition` emits the same composed string through `onData` one macrotask later (`setTimeout(0)`), so every composed syllable went out twice (`안녕` → `안녕안녕`). ASCII was unaffected because it never goes through composition.

**Fix:** remove the custom composition listeners — this xterm version already suppresses partial jamo while composing and emits the final string exactly once through `onData`. Supersedes the #1272 workaround.

## Verification

- Raw-socket handshake: exactly one `Sec-WebSocket-Protocol` header in the 101 (new regression test — `Headers.get()` normalizes duplicates away, so fetch-level assertions can't see this bug class).
- Full E2E: mint → WS upgrade with `gstack-pty.<token>` subprotocol → `{type:"start"}` → claude PTY boots and answers in the sidebar pane.
- `bun test` free suite green. A third commit fixes two stale static guards in `terminal-agent.test.ts` that were already failing on main (lazy-spawn guard still grepped `spawnClaude(` after the v1.44 `maybeSpawnPty(` refactor; the echo guard pinned the removed `acceptedProtocol`).

Note: fork PR, so paid eval/E2E workflows won't have secrets — the free suite covers these changes.

### DIFF
--- a/browse/src/terminal-agent.ts
+++ b/browse/src/terminal-agent.ts
@@ -598,12 +598,10 @@ function buildServer() {
         // first that matches a known token.
         const protoHeader = req.headers.get('sec-websocket-protocol') || '';
         let token: string | null = null;
-        let acceptedProtocol: string | null = null;
         for (const raw of protoHeader.split(',').map(s => s.trim()).filter(Boolean)) {
           const candidate = raw.startsWith('gstack-pty.') ? raw.slice('gstack-pty.'.length) : raw;
           if (validTokens.has(candidate)) {
             token = candidate;
-            acceptedProtocol = raw;
             break;
           }
         }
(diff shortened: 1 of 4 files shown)


garrytan:

PR #2531, opened 2026-08-12
fix(ship): version-bump skipped subdirectory manifests and wrote npm-invalid versions

CarringtonCreative:
Two ways `gstack-version-bump` left a release half-done, both silent. Found while shipping two versions of a real project through `/ship`.

## 1. Subdirectory manifests were skipped entirely

The tool only ever looked at `./package.json`. A repo whose only Node package lives in `web/`, `app/`, or `frontend/` has no root manifest, so `classify` reported `pkgExists: false` and every bump wrote `VERSION` alone — leaving the manifest to be edited by hand. That's the exact drift this tool exists to prevent, in the one layout where it silently did nothing.

The path now resolves `--package-json-path` → `.gstack/package-json-path` → `./package.json`, mirroring the existing `resolveVersionPath`. A subdirectory package is covered by a one-line pin.

## 2. It wrote versions npm rejects

`VERSION` is 4-digit `MAJOR.MINOR.PATCH.MICRO`. npm's semver is 3-component and rejects the fourth, so mirroring 1:1 into an npm-managed manifest writes a version npm will not install.

When a `package-lock.json` sits beside the manifest — proof npm actually manages it — the MICRO is now dropped (`0.1.26.0` → `0.1.26`), and the lockfile's **two** version fields are mirrored as well: the document root and `packages[""]`, the entry describing the root package itself. npm keeps both in step and nothing else in a release does, so a lockfile left behind drifts one field per bump until someone runs npm. Pure JSON edit — no npm spawn, no dependency-tree churn.

**Without a lockfile nothing validates the field, so the historical 1:1 mirror is preserved.** gstack's own `package.json` carries `1.60.1.0` and keeps doing so.

This required an `expectedPkg` parameter on `classifyState`. It defaults to `VERSION` — but for an npm-managed manifest it's the 3-component truncation. Without that, a correctly-synced `0.1.27` reads as drift against `0.1.27.0` forever and every `classify` returns `DRIFT`.

## Tests

7 new cases in two blocks, covering the negatives as well as the happy paths:

```
subdirectory manifest (no root package.json)
  write finds a pinned manifest and bumps it
  --package-json-path overrides the pin
npm-managed manifest (lockfile present)
  write drops the MICRO and syncs BOTH lockfile version fields
  a correctly-synced 3-component manifest is NOT read as drift
  repair syncs manifest + lockfile to the 3-component form
  no lockfile keeps the historical 4-digit mirror (gstack itself)
```

- `bun test test/gstack-version-bump.test.ts` — 21/21
- Full Tier 1 free suite (`bun test browse/test/ test/ make-pdf/test/` with the `test` script's ignores) — exit 0, zero failures on a clean fork clone at `94993f7`
- **Not run:** Tier 2/3 (`test:e2e`, `test:evals`, ~$4). Per CONTRIBUTING they need `EVALS=1` and can't nest inside Claude Code. Flagging rather than letting silence imply coverage.

## Note on the golden fixtures

Step 12 of the ship skill documents the new behaviour, so all three ship-skill goldens are updated.

Heads-up for whoever reviews: on a **first** run in a fresh clone, `host-config.test.ts` reports the Codex and Factory goldens as failing. That's a test-ordering artifact, not a real failure — `.agents/` and `.factory/` are untracked generated files, and `gen-skill-docs.test.ts` regenerates them *after* `host-config.test.ts` has already compared them. A second run is green. Happy to file that separately if it's worth fixing.

I verified the updated fixtures are byte-identical to what a clean clone generates, so no local-environment rendering rode along.

### DIFF
--- a/bin/gstack-version-bump
+++ b/bin/gstack-version-bump
@@ -30,13 +30,28 @@
 //       DRIFT_STALE_PKG path: sync package.json.version to the current VERSION
 //       file. No bump. Validates the VERSION pattern first.
 //
-// Contract: classify NEVER writes. write/repair mutate VERSION + package.json
-// only. No git mutation, no network. Mirrors gstack-next-version's reader/writer
-// split so /ship composes them.
+// Contract: classify NEVER writes. write/repair mutate VERSION + the manifest
+// (+ its lockfile) only. No git mutation, no network. Mirrors
+// gstack-next-version's reader/writer split so /ship composes them.
+//
+// Manifest resolution (all three subcommands accept --package-json-path):
+//   --package-json-path <p>  →  .gstack/package-json-path  →  ./package.json
+// A repo whose only Node package lives in a subdirectory (web/, app/,
+// frontend/) has no ROOT package.json. The tool used to report
+// pkgExists:false there and write VERSION alone, leaving the manifest to be
+// bumped by hand — the drift this tool exists to prevent, in the one layout
+// where it silently did nothing.
+//
+// npm semver: VERSION is 4-digit MAJOR.MINOR.PATCH.MICRO; npm rejects a
+// fourth component. When a package-lock.json sits beside the manifest — proof
+// npm actually manages it — the MICRO is dropped and the lockfile's two
+// version fields are mirrored too. Without a lockfile nothing validates the
+// field and the historical 1:1 mirror is preserved, so gstack's own
+// package.json keeps carrying 1.60.1.0.
 
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
 import { execFileSync } from "node:child_process";
-import { join } from "node:path";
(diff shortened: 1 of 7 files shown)


CarringtonCreative:
Filed the golden-fixture first-run failure I mentioned above as #2532 — it reproduces on pristine `v1.61.0.0` with no changes, so it's independent of this PR.

Correcting one thing I wrote in the description: I attributed it to `gen-skill-docs.test.ts` regenerating the files *after* `host-config.test.ts` compares them. The real cause is that `.agents/` and `.factory/` are gitignored and **only** `gen-skill-docs.test.ts` creates them, so `host-config.test.ts` can't run standalone in a fresh clone at all. Details and repro in #2532.

garrytan:

PR #2539, opened 2026-08-12
fix(browse): forward windowsHide through the Bun polyfill spawn shims

jerrynicholsai:
## The bug

On Windows, a `bun.exe` console window reopens by itself every so often, with no scheduled task, Run key, or startup shortcut behind it. I chased it down to the browse daemon and it turned out to be three things stacked, only the last of which is a defect:

1. A browse daemon left running from a headed session cannot self-terminate. The parent-process watchdog is disabled by design in headed mode (`browse/src/server.ts:686`), which is correct, but it means an orphaned daemon lives until killed.
2. That daemon runs a 60s ticker that respawns the terminal-agent whenever the recorded child PID is dead (`browse/src/server.ts:1493-1548`).
3. Each respawn opens a **visible console window**, because `browse/src/bun-polyfill.cjs` accepts a `Bun.spawn` options object and forwards only `stdio`, `env`, and `cwd` to `child_process.spawn`.

`windowsHide` is dropped in translation. Node defaults it to `false`; `Bun.spawn` hides the window. So the shim silently inverts the behavior on the one platform it exists to serve, per its own header comment ("Bun API polyfill for Node.js — Windows compatibility layer").

`stdio: 'ignore'` silences the child's output but does not suppress its console window, so the existing spawn options don't cover this.

## The fix

Both shims forward the option and default it to `true`, matching the Bun API being emulated. An explicit `windowsHide: false` still passes through, so nothing loses the escape hatch. `spawnTerminalAgent` also sets it explicitly at the call site, which is where the symptom surfaced.

## Tests

Three cases added to `browse/test/bun-polyfill.test.ts`, in the file's existing subprocess style: the default for `spawn`, the default for `spawnSync`, and that an explicit `false` is honored. They patch `child_process` before the polyfill is required, so they assert on what actually reaches Node rather than on the source text.

I verified each one goes red against the unpatched shim (3 fail) and green with it (7 pass), so they are known to detect the defect rather than merely passing.

## One drive-by, and why it was unavoidable

The suite could not run at all on Windows before this. The tests interpolate an absolute path into a JS string literal, so `C:\Users\...` had its backslashes consumed as escapes and every `require()` failed with MODULE_NOT_FOUND. All four tests failed for that reason alone, not on their assertions.

The path is now normalized to forward slashes, which Node accepts on Windows. This file goes from **0/4 to 7/7 on Windows**. I kept the change to the one line that builds the path plus the interpolations, so the existing tests are otherwise untouched.

## Verification

- `bun test browse/test/bun-polyfill.test.ts`: 7 pass, 0 fail (was 0 pass, 4 fail on Windows).
- `bun test browse/test/terminal-agent-pid-identity.test.ts browse/test/terminal-agent-integration.test.ts`: 9 pass, 6 fail, **identical to the clean baseline** on this machine. Those 6 are pre-existing Windows failures in that suite and are not touched by this change.
- I did not complete a full `bun test` run. It far exceeds the "<5s Tier 1" figure in CONTRIBUTING on Windows and timed out locally at 10 minutes. Happy to run anything specific you want covered.

Environment: Windows 10 Pro 19045, bun 1.3.14, branched from d078622 (v1.62.0.0).

### DIFF
--- a/browse/src/bun-polyfill.cjs
+++ b/browse/src/bun-polyfill.cjs
@@ -75,6 +75,10 @@ globalThis.Bun = {
       timeout: options.timeout,
       env: options.env,
       cwd: options.cwd,
+      // Node defaults windowsHide to false; Bun.spawn hides the console
+      // window. Without this the shim silently inverts the behavior on the
+      // one platform it exists to serve. See the spawn() note below.
+      windowsHide: options.windowsHide !== false,
     });
 
     return {
@@ -91,6 +95,11 @@ globalThis.Bun = {
       stdio,
       env: options.env,
       cwd: options.cwd,
+      // stdio:'ignore' silences a child's output but does not suppress its
+      // console window on Windows. The terminal-agent respawn (server.ts
+      // watchdog, 60s ticker) therefore popped a visible bun.exe window on
+      // every respawn until this was forwarded.
+      windowsHide: options.windowsHide !== false,
     });
 
     return {
--- a/browse/src/terminal-agent-control.ts
+++ b/browse/src/terminal-agent-control.ts
@@ -77,6 +77,10 @@ export function spawnTerminalAgent(opts: {
       ...(opts.extraEnv || {}),
     },
     stdio: ['ignore', 'ignore', 'ignore'],
+    // Explicit for the Node fallback path (dist/bun-polyfill.cjs), where the
+    // host default is the opposite of Bun's. A visible console window on every
+    // watchdog respawn is the symptom when this is missing.
+    windowsHide: true,
   });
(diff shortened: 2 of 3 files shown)


garrytan:
Thank you @jerrynicholsai — spot on. The Bun polyfill's spawn shims were dropping `windowsHide` on the floor, so even callers that passed it still popped console windows; your forward-through fix closes that gap for every spawn routed through the polyfill, covered in `browse/test/bun-polyfill.test.ts`.

Your commit lands with your authorship in #2577 (`01adcf8e`). Closing this PR as superseded by #2577, which carries it.

Issue #2536, opened 2026-08-12
bun test is non-deterministic on a clean clone: 3 failures then 0, same command


CarringtonCreative:
`bun test` — the documented Tier 1 gate — produces different results run to run on an unmodified clone. Same command, same tree, no changes in between: 3 failures, then 0.

## Reproduce (pristine `v1.61.0.0`, no modifications)

```bash
git clone --depth 1 https://github.com/garrytan/gstack.git && cd gstack
bun install
bun test browse/test/ test/ make-pdf/test/ \
  --ignore 'test/skill-e2e-*.test.ts' --ignore test/skill-llm-eval.test.ts \
  --ignore test/skill-routing-e2e.test.ts --ignore test/codex-e2e.test.ts \
  --ignore test/gemini-e2e.test.ts
```

(That's the `test` script from `package.json`, minus the trailing `slop:diff`.)

**Run 1 — 3 failures:**

```
(fail) --list > handles missing proposals file                                        [5006.61ms]
(fail) gstack-community-dashboard > connects to Supabase when config exists           [6887.86ms]
(fail) gstack-gbrain-install D5 detect-first > --dry-run falls through to fresh clone
       when no valid clone detected                                                   [7388.56ms]
```

**Run 2, identical command, nothing changed — 0 failures.**

A third run of just those three files together produced yet another failing set — only `gstack-gbrain-install D5 detect-first > rejects a pre-existing path that lacks a valid gbrain package.json`, with the other two green.

## Each file is deterministic in isolation

Two runs each, same clone:

```
test/distill-apply.test.ts          11 pass  0 fail   (x2)
test/telemetry.test.ts              40 pass  0 fail   (x2)
test/gbrain-detect-install.test.ts  16 pass  0 fail   (x2)
```

So the tests themselves are fine. Something about running the suite together isn't.

## What it isn't

**Not shared-state pollution** — at least not for `distill-apply`, which is hermetic:

```ts
beforeEach(() => {
  stateRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'gstack-apply-'));
  ...
});
afterEach(() => { fs.rmSync(stateRoot, { recursive: true, force: true }); });

// and in run():
env.GSTACK_STATE_ROOT = stateRoot;
delete env.GSTACK_HOME;
```

Fresh temp dir per test, state root redirected into it, `GSTACK_HOME` stripped from the child env. Another test writing to `~/.gstack` can't reach it.

[...]

PR #2540, opened 2026-08-12
fix(gbrain-install): --dry-run no longer requires the network (fixes flaky bun test)

CarringtonCreative:
Fixes the `bun test` non-determinism in #2536. One-line guard, plus a comment explaining why it's there.

## The bug

`gstack-gbrain-install --dry-run` prints a plan and exits without cloning anything — but it makes a live request to github.com first:

```sh
if ! $VALIDATE_ONLY; then
  check_prereq bun ...
  check_prereq git ...
  if ! curl -s --head --max-time 10 https://github.com >/dev/null 2>&1; then
    fail "cannot reach https://github.com. Check your network and try again."
  fi
fi
```

The check is gated on `VALIDATE_ONLY`. `--dry-run` sets `DRY_RUN`. So the dry-run path hits the network for a result nothing downstream uses — both dry-run branches `exit 0` before any clone.

When that curl loses a race for sockets or DNS, `fail()` exits **3** and a dry run reports "cannot reach https://github.com" on a machine that is online.

## Direct proof, no test runner involved

60 concurrent `--dry-run` invocations against temp `HOME`/`GSTACK_HOME`:

| | Failures |
|---|---|
| before | **9 / 60** (~15%), all `cannot reach https://github.com` |
| after | **0 / 60** |

## Why this made the suite flaky

The three `gstack-gbrain-install D5 detect-first` tests each call `run(INSTALL, ['--dry-run'])` and assert `expect(r.status).toBe(0)`. A full `bun test` spawns many concurrent processes, the curl occasionally loses, and whichever D5 test was running goes red.

Every symptom lines up: green in isolation (16/16, repeatedly), a *different* one of the three failing each run, sub-second failures rather than timeouts, and the observed exit code of exactly 3.

## Validation

The documented Tier 1 gate, on a clean clone:

| | Runs | Failed |
|---|---|---|
| before (#2536) | 7 | **3 (43%)** |
| after | 10 | **0** |

Ten consecutive clean runs against a 43% per-run failure rate is roughly a 0.4% coincidence.

## Scope

Real installs are unchanged — they still get the fail-fast offline check rather than a hanging `git clone`. Only the path that never clones skips it.

I first blamed a `PATH` leak and proposed defaulting `SAFE_PATH` in the test helper. That was wrong, and I've retracted it in #2536 — it would also have broken these tests on any machine where bun lives at `~/.bun/bin`, which `SAFE_PATH` excludes.

## Not addressed here

`telemetry > connects to Supabase when config exists` appeared once in my first sample and looks like the same shape — a real network call inside the free suite. I haven't investigated it and make no claim about it; noting it in case it's worth a look.

Environment: macOS (Darwin 25.5.0), bun 1.3.13, clean clone at `94993f7`.

### DIFF
--- a/bin/gstack-gbrain-install
+++ b/bin/gstack-gbrain-install
@@ -84,7 +84,14 @@ if ! $VALIDATE_ONLY; then
   # GitHub reachability — fail fast if offline rather than hanging `git clone`.
   # --max-time 10, --head (no body), quiet. Status code 200-4xx means we reached
   # the server (even 404 is reachability proof).
-  if ! curl -s --head --max-time 10 https://github.com >/dev/null 2>&1; then
+  #
+  # Skipped under --dry-run: a dry run prints a plan and exits without ever
+  # cloning, so requiring the network buys nothing and costs a real failure mode.
+  # It made `--dry-run` fail (exit 3, "cannot reach https://github.com") whenever
+  # the curl lost a race for sockets/DNS — reproducible at ~15% by running 60
+  # dry-runs concurrently, and the cause of intermittent red in the D5 tests,
+  # which call this exact path.
+  if ! $DRY_RUN && ! curl -s --head --max-time 10 https://github.com >/dev/null 2>&1; then
     fail "cannot reach https://github.com. Check your network and try again."
   fi
 fi


garrytan:

PR #2542, opened 2026-08-13
fix(browse): create node bundle output directory

guyua9:
## Summary
- create `browse/dist` before Bun writes the Node-compatible server bundle
- keep the build helper self-contained on clean checkouts

## Validation
- `bash -n browse/scripts/build-node-server.sh`
- `git diff --check`

This is intentionally limited to the missing output-directory precondition.


garrytan:

PR #2543, opened 2026-08-13
fix(redact): scan large diffs in slices; stop digit-only UUIDs matching as cards/phones

luckywenapere:
Two fixes to the redact guardrail, both found while using gstack on a real
monorepo — the workflow CONTRIBUTING.md recommends.

They're independent; happy to split into two PRs if you'd prefer.

## 1. `engine.input_too_large` blocks ordinary pushes

`gstack-redact-prepush` calls `scan()` once on the whole pushed diff, so any
diff over `DEFAULT_MAX_BYTES` (1 MiB) comes back as a single
`engine.input_too_large` HIGH finding and blocks the push.

That threshold is reached by ordinary work. A feature branch catching up to a
busy `main` produced **1,146,782 bytes** of added lines — only 9% over, and
only ~7% of it was the lockfile; the rest was genuine source.

The failure mode is worse than a noisy block:

- The message names no file, no line and no credential, so it reads as a bug.
- The documented remedy — "Rotate the credential" — doesn't apply.
- The documented escape hatch is `--no-verify`, which disables scanning
  **entirely**.

So a size cap that fires on normal branches actively trains people to bypass
the guard. It also silently discarded real findings: once actually scanned,
that same diff reported **21 MEDIUM** (PII/internal) that had never been
surfaced while the scan was aborting on size.

**Fix:** scan added lines in 768 KiB line-aligned slices and union the
findings.

No detection coverage is lost, because every pattern is single-line: none in
`redact-patterns.ts` carries the `m` or `s` flag, the `BEGIN PRIVATE KEY`
patterns capture only the header line rather than the key body, and the engine
already iterates line by line. A line boundary therefore cannot bisect a
detectable secret, so no inter-slice overlap is needed.

**Fail-closed is preserved.** A *single line* larger than the budget is still
handed to the engine intact, so a genuinely unscannable blob (minified bundle,
embedded base64) trips `input_too_large` and blocks exactly as before.

Kept in the hook rather than the engine on purpose: slice-relative `line`/`col`
are fine there (only `severity`, `id` and `preview` are read), but would break
engine callers that rely on absolute line numbers.

Verified by invoking the hook with real pre-push stdin:

| Case | Result |
|---|---|
| 1.79 MiB benign diff | exit 0, scanned — previously blocked |
| secret in the **last** slice of 1.8 MiB | exit 1, `HIGH pem.private_key` |
| single 1.5 MiB line | exit 1, `input_too_large` (fail-closed intact) |
| the real diff that needed `--no-verify` | exit 0 |

## 2. Digit-only UUIDs match as credit cards and phone numbers

Placeholder UUIDs — `00000000-0000-0000-0000-000000000000`,
`11111111-1111-…`, `22222222-2222-…` — are the standard fixture shape in test
suites, and their digit runs collide with two MEDIUM patterns. A 13–19 digit
slice of one passes Luhn often enough to matter, so `pii.cc` fires; the hyphen
groups read as national phone formatting, so `pii.phone.e164` fires.

Of the 21 MEDIUM findings above, **14 were exactly this** — every one from a
test file, reported as "Credit-card number (Luhn-valid)" and "Phone number".
That volume is what stops people reading MEDIUM output at all.

**Fix:** add `insideUuid(match)` and gate both patterns on it. The same branch
now reports 7 findings, all genuine (a published contact address, a documented
support link, upstream lockfile maintainer emails, one migration filename), and
the three test files responsible for all 14 drop to zero.

**Containment must be TOTAL, deliberately.** A span merely adjacent to or
overlapping a UUID still reports, so suppression can only fire when the whole
match is demonstrably UUID interior. A real card sitting next to a fixture UUID
is still caught — pinned by a test, since that's how this change could quietly
become a hole.

It takes the match rather than the span because the decision needs surrounding
context; the span offset is derived exactly as `redact-engine.ts` derives it
(group 1 when present, else `match[0]`), so the two can't disagree about where a
span starts. The UUID scan is windowed to 40 chars either side to stay cheap on
a multi-megabyte buffer.

Scoped to cards and phones only. `pii.ssn` can't collide (its 3-2-4 grouping
never aligns with 8-4-4-4-12) and `pii.ip_public` needs dots, so widening the
gate there would add risk for no benefit.

## Verification

- `test/redact-engine.test.ts` — **78 pass** (2 new tests added)
- engine + pattern-lint + autoredact + CLI suites — **138 pass, 0 fail**
- `bun run build` — clean
- The new UUID test **fails against the pre-change patterns file** and passes
  after, so it's a real regression guard rather than a tautology.
- `redact-pattern-lint.test.ts` passes, so the added regex is linear-time.

One caveat on my environment, in the interest of not overclaiming: I'm on
Windows, and a full `bun test` here has failures I did **not** introduce —
`browse/test/batch.test.ts` (a `beforeEach`/`afterEach` hook timeout), an
`appendSemanticReview` POSIX file-mode `0600` assertion, and three
`gstack-config-redact-keys` round-trips. I confirmed each is identical against a
pristine checkout of the files I touched. Worth a CI run on Linux to be sure.

Separately, and not fixed here: `test/redact-prepush-hook.test.ts` builds `PATH`
with `:` as the separator, so on Windows its `git` stub never applies and the
"diff killed by a signal BLOCKS" case fails (14 pass / 1 fail, before and after
this PR). Happy to fix that in a follow-up if useful — it seemed out of scope
for these two.

---

Disclosing the tooling in case it affects how you'd like to review. The
verification in each section above was actually executed rather than asserted:
the hook was invoked with real pre-push stdin for all four cases, and the new
UUID test was confirmed to fail against the pre-change patterns file before it
was kept.

### DIFF
--- a/bin/gstack-redact-prepush
+++ b/bin/gstack-redact-prepush
@@ -114,6 +114,64 @@ function addedLinesFor(localSha: string, remoteSha: string): string {
   return added.join("\n");
 }
 
+/**
+ * Byte budget per scan() call. Kept comfortably under redact-engine's
+ * DEFAULT_MAX_BYTES (1 MiB) so a slice never trips its oversize guard.
+ */
+const SCAN_CHUNK_BYTES = 768 * 1024;
+
+/**
+ * Scan added lines in line-aligned slices, unioning the findings.
+ *
+ * Why: the engine refuses input over its byte cap and fails closed, which is
+ * right for one scan() call but wrong as a push policy — a feature branch
+ * catching up to a busy main legitimately produces more added lines than the
+ * cap (1,146,782 bytes against the 1 MiB default in the push that prompted
+ * this, and only ~7% of that was the lockfile). The push then blocked on
+ * `engine.input_too_large` — a size error naming no credential — which trains
+ * people to reach for --no-verify, defeating the guardrail far more thoroughly
+ * than a large diff does.
+ *
+ * Slicing loses NO detection coverage, because every pattern is single-line:
+ * none in redact-patterns.ts carries the `m` or `s` flag, the
+ * BEGIN-PRIVATE-KEY patterns capture only the header line rather than the key
+ * body, and the engine itself iterates line by line. A line boundary therefore
+ * cannot bisect a detectable secret, so no inter-slice overlap is needed.
+ *
+ * Fail-closed is preserved: a SINGLE line over the budget is still passed to
+ * the engine intact, so a genuinely unscannable blob (minified bundle,
+ * embedded base64) trips input_too_large and blocks exactly as before.
+ *
+ * Findings' line/col are slice-relative, which is fine here — this hook only
+ * reads severity, id and preview. Do not lift this into the engine, where
+ * callers rely on absolute line numbers.
+ */
+function scanAddedLines(added: string, opts: Parameters<typeof scan>[1]): Finding[] {
+  const findings: Finding[] = [];
+  let slice: string[] = [];
+  let sliceBytes = 0;
+
+  const flush = () => {
+    if (slice.length === 0) return;
+    findings.push(...scan(slice.join("\n"), opts).findings);
+    slice = [];
+    sliceBytes = 0;
+  };
+
+  for (const line of added.split("\n")) {
+    // +1 for the newline that rejoins it.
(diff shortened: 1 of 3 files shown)


garrytan:

PR #2545, opened 2026-08-13
v1.63.1.0 fix(ship): allocate from git when the PR queue is unreachable

CarringtonCreative:
## Summary

`gstack-next-version` set `offline: true` whenever `gh pr list` failed and returned an **empty claim set**. `/ship`'s Step 12 then said *"fall back to local `BUMP_LEVEL` arithmetic"* — and local arithmetic cannot see a sibling's claim, so the fallback handed out a version another open PR already held.

Git already knows what the API was being asked for, so `offline` now degrades the queue **view**, not the **allocation**.

## Before / after

Fixture: base `0.1.66.0`, a sibling branch holding `0.1.67.0`, `gh` failing.

| | Allocated |
|---|---|
| old allocator | `0.1.67.0` — collides with the sibling |
| new allocator | `0.1.68.0` — bumps past it |

## What it reads when the host is down

- **Every remote-tracking branch's `VERSION` file** — those *are* the open PRs' branches, whether or not the API can enumerate them.
- **Versions already shipped in the base's commit subjects** — catches a number that merged and was then re-picked, which the `VERSION` file cannot see because it only holds the newest value. Bounded at 400 commits, and it warns when it truncates rather than implying it read all history.

New output field `fallback: "git" | null` so `/ship` can distinguish a git-derived pick from a host-derived one. **Additive** — the fallback only runs when the host query already returned nothing, so the online path is untouched (verified: `fallback: null`, same version as before).

## Why this mattered

Found while auditing a downstream repo that had **four duplicate version pairs** in `main` — `v0.1.6.0`, `v0.1.9.0`, `v0.1.32.0`, `v0.1.57.0` — spanning three weeks. The most recent was reproduced exactly: `gh pr list` failed mid-ship, the fallback allocated a number an open PR already held, and both merged. Two commits now claim one version there, so a version read off a running container no longer maps to a single commit.

## Test coverage

`test/gstack-next-version.test.ts`: **34 pass** (28 pre-existing + 6 new).

New tests cover: a sibling branch's claim found via remote-tracking refs; the pick landing past that sibling (`0.1.68.0`, not `0.1.67.0`); the shipped-subject scan; graceful degradation outside a git repo (warning, never a throw); and the `fallback` output contract in both directions (`"git"` when `gh` fails via a stub, `null` when online).

Adjacent suites unaffected: `gstack-version-bump` + `ship-version-sync` **35 pass**, `skill-validation` + `parity-suite` + `parity-sectioned` **340 pass** (13/13 skills parity).

## Self-review finding, fixed before commit

A comment claimed the 400-commit scan "says so" when it truncates. It didn't — the only warning in that block was the `git log` failure path, so a capped scan was silent. Made the comment true (warns when the log comes back exactly full), then verified by running a copy with the cap lowered to 5 against this repo's 118 commits: warning fires. At the real cap it correctly stays silent here.

## Not included, deliberately

`/ship`'s Step 12 instruction still tells the agent to use local arithmetic. That edit lives in `ship/SKILL.md.tmpl`, which currently also carries **unrelated in-flight work** in my working tree (the `--package-json-path` manifest resolution), and its regenerated artifacts collide with `d078622`. Rather than bundle someone else's unreviewed change, this PR ships the allocator only. **The allocator is safe on its own** — a git-derived pick is collision-safe whether or not the prose has caught up; the instruction just understates it. Happy to follow up with the prose change as a separate PR.

## Test plan

- `bun test gstack-next-version` — 34 pass
- `bun run bin/gstack-next-version --base main --bump patch --current-version <v>` — online: `fallback: null`
- Same command with a failing `gh` on `PATH` — `offline: true`, `fallback: "git"`, version still allocated, warning explains the degraded view


garrytan:

PR #2557, opened 2026-08-13
fix(browse): remove the dead security shield and its unfed /health status

frederik-kaster-noygear:
## Problem

The sidebar's SEC shield has been dead UI since the PTY terminal rewrite (ed1e4be2, #1216). That commit deleted `updateSecurityShield()` and both of its call sites along with the chat poll loop, but left the markup and CSS in place. Nothing has set the element's `data-status` since, and nothing unhides it — it ships `style="display:none"` and the only code that ever set `display: inline-flex` was the deleted function.

**Re-wiring it to `/health.security` is not viable, because that field is not live data either.** `getStatus()` reads `classifierStatus` out of `~/.gstack/security/session-state.json`, and the only writer of that file was `sidebar-agent.ts`, which the same rewrite deleted. `writeSessionState()` now has no production caller anywhere in the tree — its sole remaining caller is a unit test:

```
$ grep -rn "writeSessionState" browse/src browse/test
browse/src/security.ts:548:export function writeSessionState(...)   # definition
browse/test/security.test.ts:294:    writeSessionState(state);       # only caller
```

So `/health` reports either a permanent `inactive` (no file), or a stale `protected` wherever an old state file survived on disk. A single test run is enough to leave one behind — that unit test writes real fixture data (`sessionId: "test-session-123"`, classifiers `ok`) into `~/.gstack`, after which `/health` reports **`protected`** indefinitely.

That makes the shield worse than absent if it is ever unhidden: a green shield sourced from leftover state reads as *"no threats detected"* when the real state is *"not measured"*. Same fail-open class as #2026 (dashboard reporting "0 attacks" on backend error) and the project's own `9cc41b71` guards-failing-open wave.

## Fix

Remove the indicator and the status feed behind it, rather than wiring a needle to a gauge that measures nothing:

- shield markup in `sidepanel.html`, its CSS, and the stale comment in `sidepanel.js` that documented the indicator as "not driven yet"
- the `security` field from `/health` and the `getStatus` import
- `getStatus`, `SecurityStatus`, `StatusDetail`, and the now-unreachable `SessionState` / `readSessionState` / `writeSessionState`
- the tests covering them, including a `getStatus` shape assertion that passed on leftover fixture data and so never caught the false green

**The live defenses are untouched.** The canary, content-security L1-L3, and the L4 sidecar on `/pty-inject-scan` all report through their own call sites and never through `/health`. Per-tab decision files are unaffected.

## `browse/test/security-sidepanel-dom.test.ts`

Deleted outright, because every test in it is already dead:

- 2 assert the shield's `data-status` flips off `/health.security`
- 4 drive the security banner, whose markup `ed1e4be2` removed
- the whole suite stubs `/sidebar-chat`, an endpoint that no longer exists

The file looks healthy only because every case is `test.skipIf(!CHROMIUM_AVAILABLE)` and silently skips wherever Playwright is absent.

#2230 independently deletes this same file. If that lands first, this hunk drops out as a delete/delete; the change is otherwise standalone.

## On CI

`main` is currently red in these files — 75 failures across the security/sidebar suites, all of them testing the chat path `ed1e4be2` removed (`pollChat`, `stopAgent`, `/sidebar-agent/stop`, the browser tab bar). This branch has **the identical 75 failures, name-for-name**, with no regressions in either direction:

```
baseline (main):  298 pass, 75 fail
this branch:      296 pass, 75 fail    # -2 = the two tests removed above
comm -13 baseline mine  ->  (empty)
comm -23 baseline mine  ->  (empty)
```

#2230 is the PR that repairs that suite.

## Note

No regression test asserting the absence of the `security` key: it would fail on a correct future change that re-feeds the status from live signals (`isSidecarAvailable`, the content filters), punishing the fix rather than the regression. The reasoning is recorded in a comment where the field used to be.


garrytan:

PR #2560, opened 2026-08-13
fix(memory-ingest): pass --include-gitignored to gbrain import

gawievanblerk:
## The bug

`gstack-artifacts-init` writes an ignore-everything `.gitignore` at the root of `~/.gstack`:

```
# gstack-artifacts sync: ignore-everything base. Paths are included explicitly via
# .brain-allowlist and `git add -f` from gstack-brain-sync. Do not edit.
*
```

`gstack-memory-ingest` stages pages into `~/.gstack/.staging-ingest-<pid>-<ts>/`, which is inside that repo, and gbrain's markdown collector honours `.gitignore`. So the collector walks the staging directory, matches every staged file against `*`, and collects nothing.

**The failure is silent.** `gbrain import` exits 0 having imported nothing, while the ingest reports `written: N` from the *staged* count rather than the imported count. A run that indexes nothing is indistinguishable from a healthy one, and the memory corpus quietly stops growing.

## Reproduction

No gbrain or network needed. `git ls-files --others --exclude-standard` is the same view a gitignore-honouring collector takes:

```bash
git init .
mkdir -p .staging-ingest-12345/learnings
echo x > .staging-ingest-12345/learnings/page.md
printf '*\n' > .gitignore
git ls-files --others --exclude-standard   # -> empty
```

Observed on gstack 1.61.0.0 / 1.62.0.0 with gbrain 0.45.9.0 against a Supabase engine, but nothing here is engine or version specific.

## The fix

Pass `--include-gitignored` on the `gbrain import` call, so the import is independent of whatever `.gitignore` sits above the staging directory.

The alternative is adding a negation to the generated `.gitignore`:

```
!.staging-ingest-*/
!.staging-ingest-*/**
```

That does work (verified), but the file is gstack-owned and explicitly marked "Do not edit", so any regeneration silently reintroduces the bug. The flag fixes it at the call site instead, where the staging directory's location is actually known.

Staging dirs are created and deleted within a single ingest run and are never committed, so the `.brain-allowlist` plus `git add -f` path used by `gstack-brain-sync` is unaffected.

## Tests

`test/memory-ingest-include-gitignored.test.ts`, following the shape of the existing `memory-ingest-no-put_page.test.ts` source pin:

1. `--include-gitignored` is present in active code (comments stripped).
2. The flag sits on the same `spawnGbrainAsync` call as the `import` subcommand, so it cannot drift onto another call.
3. A behavioural test for the collision itself, so the reason for the flag survives even if the call site is refactored.

Both source pins fail against the unpatched file; the behavioural test documents the collision and passes either way. The three sibling memory-ingest suites (`gstack-memory-ingest`, `memory-ingest-no-put_page`, `memory-ingest-timeout`, 28 tests) still pass.

### DIFF
--- a/bin/gstack-memory-ingest.ts
+++ b/bin/gstack-memory-ingest.ts
@@ -1416,7 +1416,20 @@ function runGbrainImport(
     // inside Next.js / Prisma / Rails projects with their own
     // .env.local (codex review #7 — defense in depth on top of the
     // parent gstack-gbrain-sync seeding the bun grandchild's env).
-    const child = spawnGbrainAsync(["import", stagingDir, "--no-embed", "--json"]);
+    // --include-gitignored is load-bearing, not a convenience. Pages are
+    // staged into ~/.gstack/.staging-ingest-<pid>-<ts>/, and ~/.gstack is a
+    // git repo whose .gitignore is `*`. `gbrain import` honours .gitignore,
+    // so without this flag it collects files=0 and imports NOTHING, while
+    // still reporting `written: N` from the staged count. Silent data loss
+    // on every run. A working run logs `import.collect_files done ... files=N`
+    // with N > 0 and takes minutes, not seconds.
+    const child = spawnGbrainAsync([
+      "import",
+      stagingDir,
+      "--no-embed",
+      "--include-gitignored",
+      "--json",
+    ]);
     _activeImportChild = child;
     let stdout = "";
     let stderr = "";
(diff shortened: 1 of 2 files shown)


garrytan:
Thank you @gawievanblerk — this was exactly the right fix. gbrain >=0.42's git-ignore fast path was skipping the ignore-all `~/.gstack` staging dir, so memory-ingest silently ingested 0 pages (#2144); your `--include-gitignored` flag on the import child is what makes ingest work again, and the wave stacks defense-in-depth on top of it (`GIT_CEILING_DIRECTORIES` on the import child, `7531ae2b`).

Your commit lands with your authorship in #2577 (`b1ee32ad`). Closing this PR as superseded by #2577, which carries it.

PR #2561, opened 2026-08-14
fix(windows): decisions are filed under a project named "unknown", and never sync

source-utsho:
On Windows, every decision logged by `gstack-decision-log` is filed under a project literally named
`unknown`, and the Context Recovery block that should read it back never fires. Separately — and on
every platform — the decision store is not in the sync allowlist, so it never leaves the machine
even with `artifacts_sync_mode` on.

Measured on one machine before the fix: **62 decisions accumulated across 10 days and 170 gstack
skill runs, and were surfaced zero times.** Two of them were owner decisions that existed in no
other store, in a directory outside git and outside backup.

---

## 1. `resolveSlug` returns the literal string `"unknown"` on Windows

`bin/gstack-slug` is a `#!/usr/bin/env bash` script with no file extension. Windows honors neither
the shebang nor PATHEXT for an explicit path, so `CreateProcess` fails **ENOENT** and `r.stdout` is
`undefined`. `lib/bin-context.ts:12` never inspected `r.error`:

```js
const r = spawnSync(slugBinPath, { encoding: "utf-8" });
const m = (r.stdout || "").match(/^SLUG=(.+)$/m);
return m ? m[1].trim() : "unknown";   // ← every Windows call lands here
```

Reproduced directly:

```
spawn error : ENOENT / spawnSync C:\...\bin/gstack-slug ENOENT
resolveSlug : unknown
```

`decisionPaths()` then writes to `~/.gstack/projects/unknown/` — **one bucket shared by every
project on the machine**. Mine mixed three.

### Why it is silent

Three things had to line up, and they did:

1. **Both** decision bins (`gstack-decision-log` and `gstack-decision-search`) call the same
   `resolveSlug`, so they miss *identically*. Writes and searches agree with each other and the
   store is internally consistent — just in the wrong place.
2. The **read** path is bash (`generate-context-recovery.ts` → `eval "$(gstack-slug)"`), which
   resolves the slug **correctly** to the real project directory.
3. That block then guards on `[ -f "$_PROJ/decisions.active.json" ]` with **no `else`**. The file
   isn't there, so the whole ACTIVE DECISIONS section is skipped and prints nothing.

A missing file and a working store produce identical output. Nothing logs, nothing exits non-zero.

### Why `shell: true` is not the fix here

Unlike #1731, this is not a PATHEXT/shim-resolution problem — `cmd.exe` cannot run a bash script
either. And re-spawning through `bash` is worse than it looks: on Windows `bash` frequently resolves
to `C:\WINDOWS\system32\bash.exe`, i.e. **WSL**, whose `$HOME` and `/mnt/c/...` paths produce a
*different* slug **and** a *different* `slug-cache` directory. That trades one split store for
another, silently.

So this ports `gstack-slug`'s own three steps natively — cache → git remote → basename — keeping its
`[a-zA-Z0-9._-]` alphabet (`tr -cd` deletes, it does not substitute) and, critically, its **MSYS-form
cache key** (`C:\Users\j\foo` → `/c/Users/j/foo` → `_c_Users_j_foo`). Reproducing that key is what
keeps the native path and the shell path on the same answer; getting it wrong would re-derive instead
of reusing, and could drift.

**The fallback is win32-gated, so POSIX behaviour is byte-identical.**

## 2. No allowlist glob matches the decision store

`gstack-decision-log:40` enqueues `projects/<slug>/decisions.jsonl` after every write. None of the
16 managed globs in `.brain-allowlist` match it, so `compute_paths_to_stage` rejects every one at its
*"must match at least one allowlist glob"* check.

The writer and the syncer disagree, silently: enabling artifacts sync backs up learnings, plans,
designs and timelines — everything **except** the durable decision ledger. A dropped path prints
exactly what a synced one does when the queue is otherwise empty.

Fixed by adding the three `decisions.*` globs and classing them `artifact` so they also sync under
`artifacts-only`.

## Tests

`test/bin-context-windows-slug.test.ts` (9) exercises the fallback **on every platform** — only the
*gating* is win32-specific — so POSIX CI catches a regression that would otherwise surface only on a
Windows user's disk. Covers the MSYS cache-key derivation, all three resolution steps, https/ssh
remote forms, `tr -cd` deletion semantics, cache write-back with no trailing newline, plus a static
gate pinning the platform check in the style of `gbrain-spawn-windows-shell.test.ts`.

`test/artifacts-allowlist-decisions.test.ts` (3) reads the heredocs out of `gstack-artifacts-init`
rather than executing it, because `gstack-artifacts-init.test.ts` drives the real script through
`#!/bin/bash` shims and a colon-separated `PATH` and therefore cannot run on Windows — the platform
where the companion bug bites. (I confirmed that file already fails on Windows unmodified, so I left
it alone.) It applies `fnmatch.fnmatchcase` semantics, matching what `compute_paths_to_stage` uses.

Both files added to `windows-free-tests.yml`.

**Verified by planted negative, not just by passing:**

| Change reverted | Result |
|---|---|
| `resolveSlug` body only (exports kept) | 2 fail, incl. *"a spawn that cannot run resolves to a real slug, not 'unknown'"* |
| `gstack-artifacts-init` allowlist | 2 fail, marker test correctly still passes |
| Nothing | 12 pass |

Windows CI subset, run locally on Windows 11 / bun 1.3.14: **39 pass, 6 skip, 0 fail**.

## Not changed here

`docs/gbrain-sync.md:34` still says `gstack-brain-init`, renamed to `gstack-artifacts-init` in
v1.27.0.0 — `test/no-stale-gstack-brain-refs.test.ts` exists but doesn't catch this one. Left out to
keep the diff focused; happy to send it separately.

Also worth a look, though out of scope: `gstack-artifacts-init` creates the repo with `gh`/`glab`
and then verifies it via `git ls-remote` over **SSH**. On a machine where the CLI and the SSH key
authenticate as *different* accounts, that aborts at line 188 before writing anything, even though
`gstack-brain-sync` only ever does `git push origin HEAD` and works fine over HTTPS. And `git init`
there could use `core.autocrlf=false` — it stores JSONL that must round-trip byte-exactly.

### DIFF
--- a/.github/workflows/windows-free-tests.yml
+++ b/.github/workflows/windows-free-tests.yml
@@ -115,6 +115,8 @@ jobs:
             browse/test/server-sanitize-surrogates.test.ts \
             test/setup-windows-fallback.test.ts \
             test/bin-windows-bun-import-paths.test.ts \
+            test/bin-context-windows-slug.test.ts \
+            test/artifacts-allowlist-decisions.test.ts \
             test/build-script-shell-compat.test.ts \
             test/docs-config-keys.test.ts \
             test/brain-sync-windows-paths.test.ts \
--- a/bin/gstack-artifacts-init
+++ b/bin/gstack-artifacts-init
@@ -240,6 +240,14 @@ projects/*/*-design-*.md
 projects/*/*-test-plan-*.md
 projects/*/*-eng-review-test-plan-*.md
 projects/*/timeline.jsonl
+# The decision store. gstack-decision-log enqueues projects/<slug>/decisions.jsonl
+# after EVERY write, but no glob above matched it, so compute_paths_to_stage rejected
+# all of them at its "must match at least one allowlist glob" check -- a writer
+# enqueueing a path the syncer is guaranteed to drop. Without these the durable
+# decision ledger never leaves the machine, on any platform.
+projects/*/decisions.jsonl
+projects/*/decisions.active.json
+projects/*/decisions.archive.jsonl
 retros/*.md
 developer-profile.json
 builder-journey.md
@@ -267,6 +275,9 @@ cat > "$GSTACK_HOME/.brain-privacy-map.json" <<'EOF'
   {"pattern": "projects/*/*-design-*.md", "class": "artifact"},
   {"pattern": "projects/*/*-test-plan-*.md", "class": "artifact"},
   {"pattern": "projects/*/*-eng-review-test-plan-*.md", "class": "artifact"},
+  {"pattern": "projects/*/decisions.jsonl", "class": "artifact"},
+  {"pattern": "projects/*/decisions.active.json", "class": "artifact"},
+  {"pattern": "projects/*/decisions.archive.jsonl", "class": "artifact"},
   {"pattern": "retros/*.md", "class": "artifact"},
   {"pattern": "builder-journey.md", "class": "artifact"},
   {"pattern": "projects/*/timeline.jsonl", "class": "behavioral"},
--- a/lib/bin-context.ts
+++ b/lib/bin-context.ts
@@ -6,12 +6,102 @@
  */
 
 import { spawnSync } from "child_process";
+import { existsSync, mkdirSync, readFileSync, renameSync, writeFileSync } from "fs";
+import { homedir } from "os";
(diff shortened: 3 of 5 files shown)


garrytan:

Issue #2242, opened 2026-07-11
Headed mode broken on macOS: rebrand invalidates the Chrome-for-Testing bundle signature (GPU exit_code=5), and the parent-PID watchdog kills handoff sessions


abhij9750:
## Summary

Two independent bugs make headed mode unusable on macOS (arm64, gstack v1.58.5.0). Both have a clear root cause and a small fix. This is *not* #1781 (busy-vs-dead daemon / SingletonLock) — that fix is in the tree and working; these are separate.

Symptoms, in the order a user hits them:

1. `$B handoff` never opens a window: `GPU process exited unexpectedly: exit_code=5`, preceded by a bogus crashpad warning (`in_range_cast.h:38 value -634136515 out of range`).
2. `$B stop` prints `[browse] Server crashed twice in a row — aborting` and leaves a *healthy* daemon behind.
3. Retrying handoff crash-loops the daemon.

---

## Bug 1 — the rebrand corrupts the Chrome-for-Testing bundle (headed dead)

`browse/src/browser-manager.ts` (~L503-540, in `launchHeaded()`) rebrands the browser for the macOS menu bar:

```ts
const patched = plistContent.replace(/Google Chrome for Testing/g, 'GStack Browser');
fs.writeFileSync(chromePlist, patched);
```

That regex is unscoped, so it also rewrites **`CFBundleExecutable`** — which then names a binary that does not exist (the file in `Contents/MacOS/` is still `Google Chrome for Testing`). `fs.copyFileSync(iconSrc, destIcon)` a few lines later overwrites the signed `Contents/Resources/app.icns` as well.

Result on disk:

```
Contents/Info.plist:  CFBundleExecutable  = GStack Browser   ← no such file
Contents/MacOS/    :  "Google Chrome for Testing"            ← the only binary
$ spctl -a -vv ".../Google Chrome for Testing.app"
  → rejected (source=no usable signature)
```

Chromium's macOS launcher needs the outer bundle to validate before it can spawn the GPU/renderer helper apps, so the GPU process dies at launch (`exit_code=5`) and crashpad can't attach (hence the garbage-handle warning — there is no real crash report; `~/Library/Logs/DiagnosticReports/` is empty).

Headless is unaffected because `chrome-headless-shell` is a bare Mach-O: no `.app`, no plist, no helper processes. That's why the failure looks so mode-specific.

[...]

PR #2565, opened 2026-08-14
fix(browse): cancel the parent watchdog when handoff promotes a daemon to headed

Screddyice:
## The bug

`handoff` destroys the thing it just created.

It exists so a human can take over a headless session — log in, solve a CAPTCHA, clear an MFA prompt. In a Claude Code session the browser disappears roughly fifteen seconds later and takes the session with it. Driving two registrar control panels last week that cost **five daemon deaths and three logins**, each one discarding the authenticated session.

## Why

The parent-process watchdog (`server.ts`) assumes connection mode is fixed at boot: headless daemons outlive their parent, headed ones do not. Its comment says so, and it is right for the boot case — the env guards (`BROWSE_PARENT_PID=0`, `BROWSE_HEADED=1`) cover daemons that *start* headed.

`handoff` breaks the assumption. `browser-manager.ts` swaps in a headed context on a **running** daemon and sets `connectionMode = 'headed'` with no restart, so a daemon that legitimately registered a watchdog is suddenly on the fatal side of the branch:

```js
const headed = activeBrowserManager.getConnectionMode() === 'headed';
if (headed || tunnelActive) { activeShutdown?.(); }
```

The parent is typically a short-lived shell, and Claude Code's Bash tool kills one after every invocation — a case the watchdog's own comment already calls out. So the next poll fires and shuts the daemon down.

`browse connect` is unaffected: it cold-restarts with the watchdog disabled and detaches into its own process group.

## The fix

`BrowserManager` exposes `onHeadedPromotion`, fired **only** on runtime promotion, not on a headed boot. The server binds it to a canceller for the interval it already owned but previously discarded.

Bound on both the module-level manager and any embedder-supplied one, since the watchdog reads `activeBrowserManager` — binding only the default would let embedders (gbrowser) promote silently.

One placement note worth keeping: the binding sits beside the `browserManager` declaration rather than beside `clearParentWatchdog`. Putting it with the function reads better, since the function lives with the watchdog it cancels, but that touches `browserManager` in its temporal dead zone, which aborts module evaluation and leaves every later `const` uninitialized. `findport.test.ts` catches it instantly, which is how I found it.

## Tests

`watchdog.test.ts` already noted in its header that all three of its cases fix mode via env at spawn time, so none reaches the headed branch — the gap this PR closes.

Driving a real `handoff` needs a headed Chromium, which does not belong in the free tier, so the wiring is pinned with static tripwires instead, matching `cdp-session-cleanup.test.ts` and `server-auth.test.ts`. Verified in both directions: removing the notification call fails them, restoring it passes.

`cli.ts:1093` is untouched, so the `BROWSE_PARENT_PID: '0'` literal that `server-auth.test.ts` pins is intact.

## Suite status

Full `bun test`: the same **6 pre-existing failures on this branch and on `main`** (`gstack-gbrain-detect`, `gstack-artifacts-init`). They pass in isolation on both branches, so it is test-order pollution rather than anything from this change. `findport.test.ts` — the one my first attempt broke — is 9/9 on both.

## Note

There is a second, separate path that shares the same predicate: the SIGTERM handler also quits when runtime mode is headed, gated by neither env var. It does not bite `connect`-launched daemons because they are detached, and I have left it alone rather than widen this PR. Happy to follow up if you want it addressed.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -195,6 +195,15 @@ export class BrowserManager {
 
   // ─── Headed State ────────────────────────────────────────
   private connectionMode: 'launched' | 'headed' = 'launched';
+
+  /**
+   * Fired when a RUNNING daemon is promoted to headed mode (see handoff()),
+   * as opposed to starting headed. The server uses it to cancel the
+   * parent-process watchdog, which was registered on the assumption that mode
+   * is fixed at boot and would otherwise kill the freshly handed-off browser
+   * the next time the spawning shell exits.
+   */
+  onHeadedPromotion?: () => void;
   private intentionalDisconnect = false;
 
   // ─── Tab Count Guardrail (D5 + Codex single-tab flag) ───────
@@ -1602,6 +1611,14 @@ export class BrowserManager {
       this.pages.clear();
       this.tabSessions.clear();
       this.connectionMode = 'headed';
+
+      // Promotion, not a headed boot. The server registered a parent-process
+      // watchdog because this daemon started headless, and that watchdog kills
+      // headed daemons when their parent exits — which for a CLI-spawned daemon
+      // is immediately. Without this the handed-off browser dies ~15s later,
+      // taking whatever the user was mid-way through (a login, an MFA prompt)
+      // with it.
+      this.onHeadedPromotion?.();
 
       // Same Layer C stealth as launch()/launchHeaded(). Must run BEFORE
       // restoreState() navigates so the init scripts apply to the restored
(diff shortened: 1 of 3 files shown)


garrytan:
Thank you @Screddyice — sharp catch. The parent-PID watchdog was killing perfectly healthy sessions the moment handoff promoted a daemon to headed; your cancel-on-promotion fix is the piece that makes the macOS 26 headed-mode repair (#2242) hold together, and it's pinned by `browse/test/watchdog.test.ts`.

Your commit lands with your authorship in #2577 (`f3ea49ff`). Closing this PR as superseded by #2577, which carries it.

PR #2568, opened 2026-08-14
fix(version-bump): write/repair sync package-lock.json alongside package.json

### LINKED ISSUE #2567: gstack-version-bump write/repair leave package-lock.json stale — every /ship bump dirties the tree in npm repos
## Symptom

In any repo with a committed `package-lock.json`, every `/ship` version bump leaves the lockfile carrying the previous version (top-level `version` and `packages[""].version`). The tree goes dirty on the next `npm install`, and downstream clean-tree asserts (e.g. a gate/measure step) reject the work several steps later — nowhere near the cause.

## Root cause

`bin/gstack-version-bump` is explicit about it — the header contract reads: *"write/repair mutate VERSION + package.json only."* The lockfile also carries the version (twice, in lockfileVersion >= 2), so the contract itself omits a file it logically must own. The binary contains zero references to `package-lock`.

## Evidence

One consumer repo patched this at the symptom **six times** before tracing it upstream (commits `ea47287`, `1668025`, `b1afdb0`, `b9c462b`, `eda8276`, `15ac158`, each a `chore(lock): sync package-lock`). Repeated identical chore commits after every ship are the fingerprint.

## Fix

PR incoming: `write` and `repair` also sync both lockfile fields when the file exists, under the same exit-3 half-write semantics as the package.json write. `classify` stays read-only and unchanged (a hand-edited-lockfile drift state is out of scope, same philosophy as DRIFT_UNEXPECTED for package.json).

ortonom:
Fixes #2567.

## What

`gstack-version-bump write` and `repair` now also sync `package-lock.json` when it exists — both its top-level `version` and `packages[""].version` (lockfileVersion >= 2). Dependency entries are untouched; lockfileVersion 1 syncs top-level only; no lockfile is a no-op. Same exit-3 half-write semantics as the existing package.json write, so `classify` → `repair` still converge after a partial failure.

## Why

The lockfile carries the version too, so the current contract ("VERSION + package.json only") guarantees a stale lockfile after every bump in npm repos. The tree goes dirty on the next `npm install` and downstream clean-tree asserts fail far from the cause. One consumer repo accumulated six identical `chore(lock): sync package-lock` symptom commits before tracing it here (details in #2567).

## Tests

Extended `test/gstack-version-bump.test.ts`: lockfile v3 write (both fields synced, deps untouched), repair heals a stale lock, lockfile v1 no-crash, plus updated the two `write` output assertions for the new `packageLock` field. `bun test`: 18 pass, 0 fail.

`classify` is deliberately unchanged — read-only, and a hand-edited lockfile drift state is out of scope under the same philosophy as `DRIFT_UNEXPECTED`.

### DIFF
--- a/bin/gstack-version-bump
+++ b/bin/gstack-version-bump
@@ -31,8 +31,8 @@
 //       file. No bump. Validates the VERSION pattern first.
 //
 // Contract: classify NEVER writes. write/repair mutate VERSION + package.json
-// only. No git mutation, no network. Mirrors gstack-next-version's reader/writer
-// split so /ship composes them.
+// + package-lock.json (when present) only. No git mutation, no network. Mirrors
+// gstack-next-version's reader/writer split so /ship composes them.
 
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
 import { execFileSync } from "node:child_process";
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2572, opened 2026-08-15
v1.64.1.0 v1.64.1.0: the code-smell fix wave — every pipeline guard now provably fires (net −24,943 lines)

garrytan:
## The code-smell fix wave — v1.64.0.0

One PR, 36 bisected commits, net **−24,943 lines** across 183 files. Born from a three-agent audit of ~181K lines of accumulated AI-written code, reviewed through /plan-ceo-review + /plan-eng-review with two outside-voice passes (15 user decisions + 21 review findings absorbed into the plan before the first commit).

### What this repairs

| Class | What was wrong | What's true now |
|---|---|---|
| CI integrity | The free suite (358 test files) ran in NO CI job; two files had crashed at import for 48 versions | New Free Tests workflow, one bun process per file (truncation impossible by construction) |
| CI integrity | Codex/Factory freshness gates diffed gitignored paths (structurally could not fail); 7 of 10 hosts ungated | One `--host all` pass + tracked-drift diff + untracked-strays check; passed in 31s on this branch |
| Security surface | Deleted `/sidebar-chat` endpoint still on TUNNEL_PATHS; a shadow dispatcher bypassing every security gate sat one boolean from live | Allowlist is exactly `/connect` + `/command` (pinned); dispatcher deleted, JS-origin assertion unconditional |
| Security honesty | CLAUDE.md documented an ML ensemble (721MB DeBERTa + Haiku transcript classifier) with zero production callers | Dead layers deleted; docs describe the live testsavant sidecar path; do-not-re-document note added |
| Test selection | touchfiles pointed at a file deleted 48 versions ago — 3 E2E tests unreachable by any diff | Paths validated on disk every run (the guard caught a 4th rotted entry on its first execution) |
| Duplication w/ drift | 10 copy-paste host files, ~40 drifted eval gates, byte-identical cookie twins + a diverged 4th parse, ~20 atomic-write copies (one with a real race) | defineHost() (byte-identical JSON dump-diff proof), one describeE2ETier, one session-store factory, one lib/fs-atomic |

### Proof spine (per-commit, in the messages)

- **Zero-diff regens** after every claimed-safe generator/host change (commits: codex-helpers deletion, registry cleanup, boundary wiring, defineHost, dead-field deletion).
- **JSON dump-diff empty** for the defineHost conversion (10 configs byte-identical; verified no function-valued fields, so the JSON proof is complete).
- **Reviewed behavior diffs** where output intentionally changes: spec/SKILL.md 127,462 → 80,924 bytes (double-preamble bug), tier corrections (−271 lines per tier-1 skill), learningsMode (confined to gitignored host trees; goldens byte-identical).
- **Pin tests updated in the same commit** as every source change they pin, intent preserved.

### Verification

- Free suite per-file: **352/358 green.** The 6 non-green: 4 environmental on this machine, reproduced identically at clean HEAD pre-wave (compare-board, gstack-update-check, snapshot, handoff — the last needs the headed Chrome for Testing build `npx playwright install` provides; CI's image has it). 2 were wave misses caught by this very verification and fixed (an orphaned import of a deleted generator; package.json version sync).
- `gen:skill-docs --host all`: exit 0, zero tracked drift, zero strays (and the CI gate confirms on the pushed branch).
- Gate-tier evals: 39/41 shards green; the 2 non-green (cso, skillify) were load-contention artifacts of the 15-way sharded run competing with a concurrent local suite — both files re-run SOLO on this exact commit: cso 0 fail, skillify 5/5 incl. the happy path. Periodic tier runs detached now; result will be posted as a PR comment.
- `slop:diff`: net −7 findings.
- Honest gaps: the 8 tier-corrected skills have ~no eval coverage; verified by regen-diff review + live smoke (each loads and renders the intended preamble sections at its new tier). First CI run of the free-tests job on Linux is its maiden voyage — shakeout budgeted.

### Notes for the reviewer

- The openclaw-adapter deletion rode in the jsonl-store commit (concurrent staging during parallel prep) — contents unaffected, both pure deletions.
- Pre-existing red pins repaired with attribution where the wave touched their files: canDispatchOverTunnel two-arg (broken on main since v1.63), spawnClaude→maybeSpawnPty (since v1.44), a cross-file test-isolation bug in content-security filters.
- The ranked follow-up queue (12 items: pin-test conversion, server.ts route split, resolver markdown-partials, typecheck beachhead, knip ratchet, …) lives in the plan file and lands in TODOS.md in a follow-up.


garrytan:
## Periodic-tier eval result (informational — periodic does not gate merge)

`eval:bg:periodic` (sharded, detached): **62 shards — 45 passed, 9 failed, 8 timed out at the 1800s shard cap, 0 never started.**

**What this proves about the wave:** every failed/timed-out file is from the set migrated to `describeE2ETier` in this PR, and all of them RAN full paid sessions (31s–1800s durations, real turns). Migration breakage would look like the opposite — instant all-skip passes. The gate semantics work.

**Attribution status, stated per the blame protocol:**
- Gate tier (the merge gate): green with receipts — 39/41 shards + the 2 stragglers (cso, skillify) re-run solo on this exact commit, 0 fail.
- The periodic failures/timeouts are consistent with the tier's known character (opus-heavy finding-count suites, external Codex CLI, long PTY chains) under an overnight sharded run, and several show the 0-turn/$0.00 load-starvation signature seen and cleared in the gate tier. They have NOT been individually re-run — **unverified: may or may not be related to this branch.** Recommended check: compare against the next weekly periodic cron on main post-merge; any test red here and green there gets a targeted solo run on this branch.
- One environment note: the detached run resolved bun v1.3.13 vs the interactive session's v1.3.10.

Raw log: `~/.gstack-dev/eval-runs/evals-periodic-montreal-v1-garrytan-gbrain-code-smell-audit-20260815-051727-28169.log`; per-shard JSONs under `~/.gstack/projects/garrytan-gstack/evals/shards/`.

garrytan:
Caught up to origin/main (merged the parallel v1.64.0.0 tracker wave, #2571) and queue-advanced this wave to **v1.64.1.0**.

Merge notes — the two waves converged independently on several repairs, which made resolution mostly a matter of picking equivalent fixes:
- Both waves fixed the same stale pins (canDispatchOverTunnel two-arg, spawnClaude rename) and both pruned the dead chat-pipeline describes from sidebar-ux.test.ts (main's header even documents the same silent-truncation discovery, their PR #2172).
- **One wave finding reversed with receipts:** the audit called `claude/SKILL.md.tmpl` a half-landed dead skill; main's wave shipped real fixes to it — it's the deliberate outside-voice skill for non-Claude hosts, skipped on the claude host by design. Deletion reverted, `skipSkills: ['claude']` restored, CHANGELOG entry corrected.
- **One fresh catch by this wave's CI coverage:** #2571 landed a generated-file assertion that is red on every clean checkout (expects the claude-host-skipped `claude/SKILL.md`) — invisible until now because nothing ran that suite in CI. Fixed by deriving the exclusion from the host config.
- `spec/SKILL.md` conflict resolved the only correct way: regenerated from merged sources.

Post-merge: gen-skill-docs/host-config/skill-validation/size-budget/catalog-budget/touchfiles/tier-alignment suites all 0 fail; the PR's per-file free-tests job and gate-tier E2E matrix re-arbitrate on the merged code.

garrytan:
## free-tests: all 19 failures root-caused, 4 commits

Every failure class was reproduced or traced to a receipt — none were guessed at.

| Class | CI failures | Root cause | Fix |
|---|---|---|---|
| **Working-tree mutation** | hermetic-skills-seeding, host-config (claude golden), skill-census, skill-validation (codex frontmatter), spec-template-sync | `gstack-config-key-locale.test.ts` runs `set skill_prefix true` against the repo's own `bin/gstack-config`; the key's auto-relink side effect resolves the install dir from its own path — the repo — and `gstack-patch-names` rewrote all 52 tracked SKILL.md names to `gstack-*`, poisoning every later suite that reads the live tree. Reproduced locally with a fresh HOME; the culprit was isolated by running the per-file loop with a tree-dirt check after each file. | `GSTACK_SETUP_RUNNING=1` in the test env (the documented relink suppression; relink itself stays covered by `relink.test.ts`'s mock install) — `c8bf0474`. Plus a post-loop tree-mutation tripwire in the workflow so any future mutator is named instead of blamed on collateral — `96423aff`. |
| **Missing container tools** | jsonl-merge, brain-sync, gstack-detach, gbrain-source-wireup, skill-validation (`file`), make-pdf e2e ×3 | `Dockerfile.ci` had no python3 (several bash bins shell out to it), no `file` (binary-tracking check), no poppler-utils (the make-pdf gates literally print "install poppler-utils"), no color-emoji font | added to the image, mirroring make-pdf-gate.yml — `96423aff` |
| **Git identity** | session-update-autostash, team-mode, gstack-artifacts-init, gstack-decision-bins | commit-making tests fail with "Author identity unknown" in the container | `git config --global user.*` + `safe.directory` step — `96423aff` |
| **Linux xargs semantics** | gstack-codex-session-import | real bin bug: GNU xargs runs `ls -t` once on EMPTY input, listing the cwd and producing a bogus `LATEST` from the repo root; BSD xargs (macOS) skips the run, which masked it locally | `xargs -r` pins BSD behavior on both — `45bc0295` |
| **Real ratchet failure** | parity-suite | the two parallel v1.64 waves each added shared-preamble prose: 3 skills over their size ratios on the v1.57.7.0 anchor, 4 carved skeletons 22–280 B over absolute caps | rebaseline to v1.64.1.0 (union-normalized, old fixture retained for the audit trail) + ~1 KB cap headroom with per-cap rationale — `8f048370` |
| **Documented flake** | design/variants-retry-after | TODOS.md-tracked timing flake (HTTP-date Retry-After rounding) | curated skip with reason — `96423aff` |

**Bonus finding:** the image's bun pin never worked — `bun.sh/install` ignores a `BUN_VERSION` env var, so every rebuild silently installed latest (the failing run was on 1.3.14 vs the 1.3.10 devs run locally). Now passed as the positional arg the installer actually reads.

Local verification: every previously-failing suite that can run on macOS is green (`parity-suite` + all four carve-guard consumers, `gstack-config-key-locale` under a fresh HOME with zero tree dirt, `codex-session-import`, `hermetic-skills-seeding`, `host-config`, `skill-census`, `spec-template-sync`). The container-only classes (tools, identity) are verified by the CI run on this push — image rebuilds first, so allow the extra ~15 min.

garrytan:
**Follow-up: three more real bugs surfaced by the loop after the first fix wave, all fixed — board is now fully green.**

- `6c0b2e93` — `gstack-session-update`'s backgrounded update subshell inherited the session hook's stdio pipes; once the caller closed them, `git pull`'s autostash notice hit a closed pipe and the pull died of SIGPIPE (`PULL_FAILED exit=141`, empty stderr). Reachable by any production hook runner that closes stdio promptly, not just CI. The detached fork now owns its stdio; observability stays on the log file.
- `c394e483` — `gstack-decision-bins`'s scope-filter test relied on implicit branch detection, which returns undefined on CI's detached-HEAD checkout on both the log and search sides, so a branch-scoped decision could never surface. Branch is now explicit on both sides; the filter logic is what's under test.
- `970d68a7` — lease-expiry assertion demanded same-millisecond mints; back-to-back `mintLease()` calls straddled a ms boundary (`...525` vs `...526`). Now asserts a shared TTL window.

Final run: 394 files enumerated, 9 curated skips with reasons, everything else green, tree-mutation tripwire silent.

PR #2580, opened 2026-08-15
fix: human smoke confirmation gate (land-and-deploy CI tiers + canary all-green path)

redle82:
## Summary

Two related fixes, both pointed at the same root cause: a skill reports
"done"/"healthy" without a human ever actually looking at the outcome,
which is worse than no report at all — it looks finished when it isn't, so
more work stacks on top of something never really closed.

### `/land-and-deploy` (4 fixes)

- **Step 2a — CI tiers.** Resolve required checks from branch protection
  (`gh api .../branches/{base}/protection`) instead of treating every check
  `gh pr checks` reports as equally blocking. Repos that run a fast
  required tier plus a slower advisory/extended tier no longer wait on the
  advisory tier to gate the merge. Falls back to the old all-checks
  behavior if the API call fails.
- **Step 2b — budget-exhaustion detection.** A required check can come back
  FAILING because the code broke, or because CI ran out of budget/minutes
  and the job never really executed. Detects the signature (0 steps, run
  completes in seconds instead of its normal duration) and offers to run
  an equivalent local CI script instead of reporting a red X that means
  nothing as a real code failure.
- **Step 3.5c-bis — project-required PR report.** If a project documents a
  mandatory PR report/checklist format (in CLAUDE.md or CONTRIBUTING.md),
  verify it's present and complete before recommending merge.
- **Step 7a — human smoke gate (the core fix).** After the automated
  canary passes, open the live URL for the user and wait for an explicit
  reply before marking the deploy VERIFIED. Step 9's verdict now
  distinguishes `DEPLOYED AND VERIFIED` (human confirmed) from `DEPLOYED
  (AUTOMATED CHECKS ONLY — AWAITING HUMAN SMOKE TEST)`.

### `/canary` (1 fix, same pattern)

Canary's alert path already routes through AskUserQuestion — but the
quiet, all-green path (nothing broke across the whole monitoring window,
the most common outcome) never puts the live page in front of anyone.
Phase 6 now opens the URL and asks for explicit pass/fail after
monitoring completes; Phase 7's baseline-update offer is gated on that
confirmation so an unconfirmed all-clear can't quietly become the new
reference point other runs compare against.

### Audit of the rest

Before scoping the second fix to canary, I grepped all ~53 skill templates
for the pattern (automated verdict of HEALTHY/VERIFIED/PASSED with no
human-confirmation gate on the all-green path) and read the real
candidates: `qa`, `design-review`, `ios-design-review`, `ios-qa`,
`benchmark`. None share this gap — they already report per-item evidence
(verified/best-effort/reverted/deferred with before/after screenshots per
issue, or objective metrics like FCP/LCP/bundle-size) instead of a single
global "it works in production" claim a human never checked. Applying the
same patch there would be forcing a fix onto a shape that doesn't have the
problem.

All four `/land-and-deploy` fixes and the `/canary` fix fall back cleanly
to prior behavior when the relevant signal isn't available (no branch
protection, no documented PR checklist, non-interactive/spawned sessions
where nobody can reply).

## Test plan

- [x] `bun run gen:skill-docs` regenerates both `SKILL.md` files from the
      edited `.tmpl`s cleanly
- [x] `bun run gen:skill-docs --dry-run` reports `FRESH` on every file,
      exit 0 (the exact check `.github/workflows/skill-docs.yml` runs)
- [x] `bun test` (tier 1, full suite): 1148 pass / 6 fail — all 6
      confirmed pre-existing by stashing each patch and re-running against
      clean `main` (gbrain sandbox-PATH issues + a pre-existing
      `package.json`/`VERSION` mismatch on `main`, unrelated to this diff)
- [x] Targeted runs after each commit: `land-and-deploy-postfail.test.ts`,
      `skill-parser.test.ts`, `skill-validation.test.ts`,
      `gen-skill-docs.test.ts`, `skill-size-budget.test.ts` — same 1
      pre-existing failure, zero new failures from either patch
- [ ] Tier 2/3 evals (E2E, LLM-judge) not run — no `ANTHROPIC_API_KEY`
      configured in this environment; flagging for maintainer review since
      `CONTRIBUTING.md` recommends them for skill-behavior changes

### DIFF
--- a/canary/SKILL.md
+++ b/canary/SKILL.md
@@ -1023,9 +1023,29 @@ Per-Page Results:
 
 Alerts Fired:  [N] (X critical, Y high, Z medium)
 Screenshots:   .gstack/canary-reports/screenshots/
+Human smoke:   [CONFIRMED (user said pass) / AWAITING (screen opened, no reply yet)]
 
 VERDICT: [DEPLOY IS HEALTHY / DEPLOY HAS ISSUES — details above]
 ```
+
+**A HEALTHY status from Phases 1-5 is the automated read, not the final word.**
+Every alert path already routes through AskUserQuestion (Phase 5) — but the
+quiet, all-green path never puts anything in front of the user, which means
+the most common outcome (nothing broke) is also the one nobody ever actually
+looks at. Before finalizing the report:
+
+```bash
+open "<url>" 2>/dev/null || xdg-open "<url>" 2>/dev/null || echo "Couldn't auto-open — here's the URL: <url>"
+```
+
+Tell the user: "Automated monitoring found nothing wrong across {N} checks over
+{duration} — but that's a script's opinion, not a look. I've opened the live
+page. Take a glance, then tell me pass or fail so I can close this out as
+actually verified." If the user confirms, mark `Human smoke: CONFIRMED` in the
+report. If this is a non-interactive/spawned session, or the user doesn't
+reply before you need to finish the report, write `Human smoke: AWAITING` and
+say so plainly — do not silently treat "no alerts fired" as "a human looked
+and it's fine."
 
 Save report to `.gstack/canary-reports/{date}-canary.md` and `.gstack/canary-reports/{date}-canary.json`.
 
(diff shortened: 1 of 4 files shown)


redle82:
Bumping this — it's been open ~24h with no CI run yet (fork PRs need a maintainer approval to trigger Actions, which I don't have). Happy to address any feedback once it runs. In the meantime, applied the same patch locally in `~/.claude/skills/gstack` to confirm the human-gate behavior works as expected on `/land-and-deploy` and `/canary`.

PR #2581, opened 2026-08-15
fix: document iOS QA bridge compatibility fallback

itstimwhite:
## Summary
- add a compatibility preflight before installing the iOS QA DebugBridge
- stop without app modification when the project uses unsupported observation models or non-SwiftPM target wiring
- preserve installed production/TestFlight apps and prefer an existing real-device XCUITest harness with an isolated QA bundle
- distinguish fixture state, provider UI, and actual provider success in QA evidence

## Proven gap
A physical-device pass on an Xcode-project iOS app using `ObservableObject`/`@StateObject` found that the current generator produces accessors only for file-scope `@Observable`, while the documented dependency wiring assumes an app-level `Package.swift`. Following the prior bootstrap path would either produce no useful accessors or require invented project wiring.

## Validation
- generated skill docs and host freshness checks run locally
- skill health check run (iOS QA outputs fresh; unrelated pre-existing missing Claude-generated-file warning remains)
- `bun test --timeout 20000 ios-qa/scripts/gen-accessors.test.ts` — 57 passed, 0 failed

### DIFF
--- a/ios-qa/SKILL.md
+++ b/ios-qa/SKILL.md
@@ -876,7 +876,20 @@ fi
 
 ## Phase 1: Read source, plan codegen
 
-1. Walk the app source (passed as `--source <dir>`) and identify all `@Observable`
+1. Before changing the app or replacing an installed build, verify that the
+   bridge is compatible with the project:
+   - The generator currently supports file-scope `@Observable` classes only;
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2585, opened 2026-08-15
ios-qa: guard DebugBridgeTouch.m on DEBUG, not just TARGET_OS_IOS

Bastea:
## The problem

`DebugBridgeTouch.m` and its header both promise the code never ships:

> `Uses these private UIKit selectors (DEBUG-only; never shipped to App Store)`
> `DEBUG-only — never link in Release.`

Nothing enforced it. The only guard was `#if TARGET_OS_IOS`, so a **Release** build for iOS compiles the whole implementation in — private API included.

Measured on a real app, `nm -j` against an iOS Release binary:

| | count |
|---|---|
| `DebugBridge` symbols | **15** |
| `IOHIDEventCreateDigitizer` | 2 |
| `AXSSetAutomationEnabled` | 1 symbol, 2 strings |
| `IOKit.framework` | 4 strings |

Including `+[DebugBridgeTouch sendTapAtPoint:inWindow:]` and `_OBJC_CLASS_$_DebugBridgeTouch`.

That is a Guideline 2.5.1 private-API exposure in a shippable binary, and it fails `Package.swift`'s own stated CI invariant:

```
nm -j build/Release/<binary> | grep -q DebugBridge && exit 1
```

## Why the documented guard doesn't cover it

`Package.swift` describes the protection as `.when(configuration: .debug)` on the consuming target's dependency. That works for **SwiftPM consumers**.

It cannot be expressed by an app integrating DebugBridge as a local package inside an **`.xcodeproj`**. Xcode's Filters column under *Frameworks, Libraries, and Embedded Content* offers platform conditions only — iOS, macOS, visionOS — never build configuration. So for xcodeproj consumers the documented guard silently does nothing, which is exactly the case measured above.

The Swift targets were already safe: all four `.swift` files are `#if DEBUG` guarded and `Package.swift` defines `DEBUG` for them via `swiftSettings`. Only the Objective-C target — the one that actually links private API — was unguarded.

## The fix

1. `DebugBridgeTouch.m.template` branches on `#if !defined(DEBUG)` first and emits **nothing** in Release, falling through to the existing iOS / non-iOS branches only in Debug.
2. `Package.swift.template` declares `DEBUG` explicitly for the ObjC target:
   ```swift
   cSettings: [.define("DEBUG", .when(configuration: .debug))]
   ```
   The two Swift targets already did this. Relying on SwiftPM's implicit `DEBUG` for C-family targets isn't worth betting a private-API exposure on.

Putting the guard in the source means it holds regardless of how the package is integrated, rather than depending on the consumer getting it right.

## Verification

Compiled the generated file for iOS both ways:

```
xcrun -sdk iphoneos clang -c DebugBridgeTouch.m -arch arm64 -I include -fobjc-arc [-DDEBUG=1]
```

| config | DebugBridge symbols | private-API symbols | object size |
|---|---|---|---|
| Release (no `-DDEBUG`) | **0** | **0** | 448 B |
| Debug (`-DDEBUG=1`) | 7 | 6 | 13,104 B |

The harness is unchanged in Debug. Release emits an empty translation unit.

## Note for maintainers

The comments in both templates now record the measured failure, so the guard isn't removed later as redundant. Happy to adjust wording or split the two changes if you'd prefer them separately.

### DIFF
--- a/ios-qa/templates/DebugBridgeTouch.m.template
+++ b/ios-qa/templates/DebugBridgeTouch.m.template
@@ -20,7 +20,29 @@
 #import "DebugBridgeTouch.h"
 #import <TargetConditionals.h>
 
-#if TARGET_OS_IOS
+#if !defined(DEBUG)
+
+// RELEASE BUILD: this file deliberately emits NOTHING — no class, no symbols, no
+// private-API references. The header still declares the interface, which is inert
+// on its own; every consumer of DebugBridgeTouch is itself `#if DEBUG` guarded, so
+// nothing references it here and nothing fails to link.
+//
+// This guard was added because the comments at the top of this file and in the
+// header both PROMISED "DEBUG-only; never shipped to App Store" and "never link in
+// Release" — and nothing enforced it. Only `#if TARGET_OS_IOS` was tested, so a
+// Release build for iOS compiled the whole implementation in. Measured on a real
+// app (Seize Day, 2026-08-15), `nm -j` on a Release binary returned 15 DebugBridge
+// symbols including +[DebugBridgeTouch sendTapAtPoint:inWindow:], plus
+// IOHIDEventCreateDigitizer, AXSSetAutomationEnabled and IOKit.framework strings.
+// That is a Guideline 2.5.1 private-API exposure in a shippable binary.
+//
+// Package.swift's stated guard — `.when(configuration: .debug)` on the consuming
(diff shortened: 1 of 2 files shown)


garrytan:
Thank you — this landed in v1.67.0.0 (credited in the CHANGELOG), plus a release-guard tripwire added in v1.67.1.0. Closing since the fix is on main. 🙏

PR #2586, opened 2026-08-15
fix(test): hermetic-wiring skill-seeding tripwire can never pass on default ~/.claude installs

sneakygriff:
## Problem

The tripwire `skill seeding stays under runRoot and reads the live repo tree, never operator ~/.claude` asserts every seeded symlink target must NOT start with `~/.claude` — and then that it MUST resolve into the live repo tree. On the default global-git install the repo itself lives at `~/.claude/skills/gstack`, so every **correct** symlink carries the banned prefix: the two assertions can never both hold.

## Evidence

Pristine `c118e240` (v1.64.1.0), verified 2026-08-15 on macOS arm64:
- worktree under `~/.claude/skills/` → the seeding test **fails** (`pair-agent: symlink escapes to .../pair-agent/SKILL.md`)
- worktree anywhere else → **passes**

CI never sees this because dev checkouts don't live under `~/.claude` — but every user following the README install path who runs the suite does.

## Fix

Targets that `realpath` into the resolved repo ROOT are exempted **before** the operator-`~/.claude` ban (both sides realpath'd, so a symlinked $HOME can't dodge the tripwire). Genuine escapes — a target under `~/.claude` but outside the repo — still fail with the specific escape message.

## Verification

`bun test test/hermetic-wiring.test.ts`: 5/5 pass both at `~/.claude/skills/gstack` and at a non-`~/.claude` worktree of this branch.

### DIFF
--- a/test/hermetic-wiring.test.ts
+++ b/test/hermetic-wiring.test.ts
@@ -125,10 +125,21 @@ describe('hermetic wiring tripwire', () => {
     expect(configDir.startsWith(runRoot + path.sep)).toBe(true);
     expect(configDir.startsWith(operatorClaude)).toBe(false);
     const skillsDir = path.join(configDir, 'skills');
+    const repoRootReal = fs.realpathSync(ROOT) + path.sep;
     for (const entry of fs.readdirSync(skillsDir)) {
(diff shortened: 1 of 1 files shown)


garrytan:

PR #2587, opened 2026-08-15
fix(brain-cache): honest 'missing' instead of fabricated-empty digests on gbrain failure

sneakygriff:
## Problem

`fetchRecentDecisions` and `fetchSalience` in `bin/gstack-brain-cache` return hardcoded empty digests (`_No prior skill runs recorded._` / `_No salient pages in last 14d._`) when `!result?.pages` — i.e. when the gbrain call **failed**, not when it genuinely returned zero rows. `refreshEntity()` then caches that digest and stamps `last_refresh`, so a gbrain-unreachable failure becomes a 'successful' empty answer that survives every TTL cycle, indistinguishable from real emptiness.

## Fix

- Both fetchers return `null` on `!result?.pages`, letting `cmdGet`'s existing missing/stale-fallback machinery report the true state — the same polarity `fetchGoals`/`fetchSimplePage` already have.
- Genuine `pages.length === 0` still renders the honest empty digest.
- New `Array.isArray(result.pages)` guard: a malformed payload (`{pages: {}}`) classifies as failure instead of throwing inside `refreshEntity` (which has no try/catch) and crashing the CLI `get`.

## Verification

`bun test test/brain-cache-roundtrip.test.ts test/brain-cache-stale-but-usable.test.ts`: 33/33 pass; `bun bin/gstack-brain-cache --help` smoke-tested.

### DIFF
--- a/bin/gstack-brain-cache
+++ b/bin/gstack-brain-cache
@@ -521,6 +521,21 @@ function fetchRecentDecisions(projectSlug: string | null): string | null {
     '--json',
   ]);
   if (!result?.pages) {
+    // F10 bug fix: this branch used to return the hardcoded
+    // "_No prior skill runs recorded._" string here, which is indistinguishable
+    // from a genuine zero-rows result. That silently converted a gbrain-
+    // unreachable FAILURE into a "successful" cached digest — refreshEntity()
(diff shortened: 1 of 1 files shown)


garrytan:

PR #2588, opened 2026-08-15
fix(security): guard brain-sync arithmetic against injected .brain-last-pull content; sanitize _GBRAIN_HOST

sneakygriff:
## Problem

The generated brain-sync preamble bash (every host, every skill) feeds the raw file content of `~/.gstack/.brain-last-pull` straight into `$(( ))` arithmetic. Bash arithmetic evaluates array subscripts and command substitution, so a writable state file is an execution vector — e.g. content like `x[$(cmd)]` runs `cmd` at preamble time. `_GBRAIN_HOST` is likewise interpolated unsanitized into the trusted status line.

## Fix

- `case "$_BRAIN_LAST" in ''|*[!0-9]*) _BRAIN_LAST=0 ;; esac` before the arithmetic — non-numeric content degrades to 0 instead of evaluating.
- `_GBRAIN_HOST` normalized via `head -1 | tr -cd 'A-Za-z0-9._-'`.
- Existing `quoteSafePath()` behavior preserved; all host SKILL.md outputs regenerated (49 files) and the three ship golden fixtures re-blessed in the same commit.

## Verification

`bun run gen:skill-docs --host all --dry-run`: 0 STALE. `bun test test/host-config.test.ts test/gen-skill-docs.test.ts test/brain-sync.test.ts`: 503/503 pass.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -401,6 +401,7 @@ if [ -d "$_GSTACK_HOME/.git" ] && [ "$_BRAIN_SYNC_MODE" != "off" ]; then
   _BRAIN_DO_PULL=1
   if [ -f "$_BRAIN_LAST_PULL_FILE" ]; then
     _BRAIN_LAST=$(cat "$_BRAIN_LAST_PULL_FILE" 2>/dev/null || echo 0)
+    case "$_BRAIN_LAST" in ''|*[!0-9]*) _BRAIN_LAST=0 ;; esac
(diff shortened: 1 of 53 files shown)


garrytan:
Thank you — this landed in v1.67.0.0, re-derived under the generated-file screening rule and credited in the CHANGELOG. Closing since the fix is on main. 🙏

Issue #1778, opened 2026-05-28
bug: generated SKILL.md descriptions with colons are invalid YAML


kesslerio:
## What happened

Codex/OpenAI skill loading skips several gstack skills because generated `SKILL.md` frontmatter contains inline plain YAML scalars with unescaped colons.

Example generated frontmatter:

```yaml
---
name: setup-gbrain
preamble-tier: 2
version: 1.0.0
description: Set up gbrain for this coding agent: install the CLI, initialize a local PGLite or Supabase brain, register MCP, capture per-remote trust policy. (gstack)
triggers:
  - setup gbrain
---
```

A strict YAML parser treats the second colon (`agent: install`) as a mapping separator and fails with:

```text
invalid YAML: mapping values are not allowed in this context
```

## Affected examples

Observed in exported gstack skills:

- `setup-gbrain/SKILL.md`
- `guard/SKILL.md`
- `design-html/SKILL.md`
- `design-shotgun/SKILL.md`
- `design-review/SKILL.md`
- `ship/SKILL.md`
- `design-consultation/SKILL.md`
- `plan-tune/SKILL.md`

## Expected

Generated `SKILL.md` frontmatter should parse as YAML in strict loaders.

## Suggested fix

Quote one-line generated `description:` values or emit block scalars consistently. The generator should also parse every generated frontmatter block in tests, not just check that `name:` and `description:` strings exist.

Locally, this shape fixed the issue:

- Quote inline `description:` values with `JSON.stringify(...)`
- In `gen-skill-docs.ts`, normalize inline descriptions before writing generated content
- In `gen-skill-docs.test.ts`, parse generated frontmatter with `Bun.YAML.parse(...)`

## Verification

Local validation after patching exported skills:

```text
checked 545 SKILL.md files, bad=0
```

Also no remaining unquoted inline descriptions with colons were found by:

```bash
rg -n '^description: [^|"'"'].*:' /path/to/exported/gstack --glob 'SKILL.md'
```

time-attack:
@16francej Please close this issue as resolved. Generated skill descriptions with YAML-sensitive characters are now quoted safely, and strict parser tests protect the behavior. This shipped in #1808.

PR #2589, opened 2026-08-15
fix(gen-skill-docs): quote YAML inline scalars containing '...' (strict parsers reject bare ellipsis)

sneakygriff:
## Problem

`toYamlInlineScalar` doesn't quote scalars containing `...` — a YAML document-end marker that strict parsers (including `Bun.YAML`) reject mid-scalar. The catalog-trim path appends `...` to any >200-char description lead, so the first skill whose description gets truncated generates a SKILL.md with unparseable frontmatter.

## Fix

One condition added to `needsQuote` (`/\.\.\./ `), plus a `toYamlInlineScalar` unit-test describe in `test/catalog-trim.test.ts` covering ellipsis quoting, interior-colon quoting (#1778), plain-scalar passthrough, leading-indicator, and trailing-whitespace cases.

## Verification

`bun test test/catalog-trim.test.ts test/gen-skill-docs.test.ts`: 426/426 pass. `gen:skill-docs --host all --dry-run`: 0 STALE (no current description truncates, so generated output is byte-identical — the fix is prophylactic until the first long description lands).

### DIFF
--- a/scripts/gen-skill-docs.ts
+++ b/scripts/gen-skill-docs.ts
@@ -413,6 +413,7 @@ export function toYamlInlineScalar(s: string): string {
     s !== s.trim() ||                       // leading/trailing whitespace
     /:(\s|$)/.test(s) ||                    // "foo: bar" / trailing colon → mapping ambiguity
     /\s#/.test(s) ||                        // " #" → inline comment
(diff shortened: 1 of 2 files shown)


garrytan:

garrytan:

PR #2590, opened 2026-08-15
fix(test): repair three browse test files stranded by the v1.63–v1.64.1 refactors

sneakygriff:
## Problem

Three browse test files fail deterministically on pristine `c118e240` on any machine that actually runs them (24 failures total):

1. **compare-board.test.ts** (16 fails): the v1.64 signature-bridge wrapper for `handleReadCommand` drops its 4th argument, so the now-unconditional JS-origin assertion TypeErrors on `bm.hasCookieImports()`. The sibling write bridge two lines down passes it correctly.
2. **extension-sender-auth.test.ts** (2 fails): the behavioral "gate answers" loops iterate `sidebar-command`, which `background.js` no longer routes (dropped from `ALLOWED_TYPES` with the chat rip) — the listener goes silent on unknown types before the sender gate ever runs.
3. **security-sidepanel-dom.test.ts** (6 fails): five tests drive the shield/banner off the ripped chat path; `sidepanel.js` now deliberately leaves the shield "hidden by default, not driven". The first test waits 15s for a `data-status` that never arrives, and bun's timeout handler SIGTERMs the **shared** `beforeAll` browser (exit 143), cascading "browser has been closed" into every later test.

CI never sees #3 because the file self-skips wherever Playwright chromium isn't installed.

## Fix

- compare-board: forward `b` through the read bridge (one argument).
- sender-auth: behavioral loops iterate routed types; `sidebar-command` keeps a nothing-usable-leaks check and stays in the sender-auth policy as defense-in-depth.
- sidepanel: replaced the five stranded tests with pins of the current contract (shield exists, stays hidden, stays undriven) plus a tombstone documenting what was removed, why it can never pass, and how to resurrect the old cases if the shield is ever re-driven.

## Verification

`bun test browse/test/security-sidepanel-dom.test.ts browse/test/extension-sender-auth.test.ts browse/test/compare-board.test.ts --timeout 30000`: 38/38 pass (Playwright chromium-backed, macOS arm64).

### DIFF
--- a/browse/test/compare-board.test.ts
+++ b/browse/test/compare-board.test.ts
@@ -16,7 +16,10 @@ import { handleReadCommand as _handleReadCommand } from '../src/read-commands';
 import { handleWriteCommand as _handleWriteCommand } from '../src/write-commands';
 
 const handleReadCommand = (cmd: string, args: string[], b: BrowserManager) =>
-  _handleReadCommand(cmd, args, b.getActiveSession());
+  // Regression: the bridge must forward `b` — read-commands' now-unconditional
+  // JS-origin assertion calls bm.hasCookieImports(); omitting it TypeErrors
+  // every read (all 16 tests here failed on pristine v1.64.1.0).
+  _handleReadCommand(cmd, args, b.getActiveSession(), b);
 const handleWriteCommand = (cmd: string, args: string[], b: BrowserManager) =>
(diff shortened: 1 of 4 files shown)


garrytan:
Closing as superseded: all three target files were repaired or deleted on main before this PR (v1.67.0.0 close-only reclassification; survivors run green at HEAD). Thank you for the report/patch — the class is covered on main.

PR #2591, opened 2026-08-15
fix(redact): parcel IDs are not phone numbers

Two-Six-Alpha-1115:
## The problem

A county tax-map parcel ID (APN) reads as a national-format phone number to `pii.phone.e164` — the same collision class as the digit-only UUID that `insideUuid` already guards.

```
parcel_id: "12-3456789.000"   →  MEDIUM pii.phone.e164
norm:      "123456789000"     →  MEDIUM pii.phone.e164
```

The regex is deliberately loose to catch international formats, and an APN satisfies it in both its punctuated and normalized forms.

This isn't a rare edge. Land, title and property-tax repos carry APNs by the hundred — one title branch tripped 2 MEDIUM findings on nothing but synthetic fixtures, and the same shape recurs in every fixture, mart and smoke in the domain. A guardrail that cries wolf on the domain's primary identifier is one people learn to wave through, which is how a real HIGH finding eventually gets ignored. That's the reason to fix it rather than live with the noise.

## The guard

`looksLikeParcelId(span, match)`, wired into the existing `validate` alongside `insideUuid` and `looksLikeCompactTimestamp`, following the same context-window approach `insideUuid` uses. Two deliberately separate tiers:

1. **Dotted form, exempt on shape.** No phone convention puts a dot before a trailing 3–4 digit group after a 4–8 digit middle, so this cannot swallow a real number. Hyphen-only variants (`22-0001-000`) are **not** shape-exempted — those genuinely are phone-shaped.

2. **Digits-only form, exempt only on evidence.** A bare digit run is phone-shaped in isolation, so it must be the exact digit-normalization of a punctuated APN within a 400-char window. Fixtures and marts always carry the pair; a real phone number has no such twin nearby. This reads the document's own evidence instead of guessing from digits.

## Verification

**A/B against the unmodified engine** over inputs spanning every rule family — AWS key, PEM, GitHub PAT, email, public IP, credit card, SSN, compact timestamp, UUID, and nine phone formats. Exactly **one** behavior changed: the APN pair. Everything else byte-identical.

**New test pins both directions** (`test/redact-parcel-id-false-positive.test.ts`, 15 cases):
- 9 real phone shapes stay flagged — US dashed/parens/dotted, E.164 US/UK/DE/PT, bare 11-digit
- 3 APN shapes clean
- 3 narrowness controls, including the load-bearing one: **a bare digit run with no punctuated APN nearby is still flagged**

**Proven red under mutation.** Stubbing the guard to `return true` — the dangerous blanket-exemption failure, which would quietly gut the pattern — fails 12 of 15. Stubbing to `return false` fails 3. Restored: 15/15 with the source byte-identical.

**Existing suites green:** `redact-pattern-lint` 49/49, `gstack-config-redact-keys` 6/6, `browse/proxy-redact` 8/8.

HIGH detection is untouched — PEM, GitHub PAT and Slack token all still block.

### DIFF
--- a/lib/redact-patterns.ts
+++ b/lib/redact-patterns.ts
@@ -168,6 +168,58 @@ function looksLikeCompactTimestamp(span: string): boolean {
   );
 }
 
+/** Context window for pairing a normalized parcel ID with its punctuated form. */
+const PARCEL_CONTEXT_CHARS = 400;
+
+/**
+ * County tax-map parcel ID (APN). Ohio's dominant form is NN-NNNNNNN.NNN, and
+ * some counties use a hyphen before the 3-4 digit suffix instead of a dot.
+ */
+const PARCEL_PUNCT_RE = /\b\d{2}-\d{4,8}[.\-]\d{3,4}\b/g;
+
+/**
+ * A parcel ID reads as a national-format phone number to pii.phone.e164 — the
+ * same collision class as the digit-only UUID that `insideUuid` already guards.
+ * Land/title repos carry these by the hundred, so the false positives are not
+ * incidental: they arrive on every branch that touches title, and a guardrail
+ * that cries wolf on the domain's primary identifier trains people to wave the
+ * warning through, which is how a real HIGH finding eventually gets ignored.
+ *
+ * Deliberately narrow, in two tiers:
+ *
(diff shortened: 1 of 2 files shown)


garrytan:

PR #2596, opened 2026-08-16
fix: Windows support - Node.js CLI fallback for Smart App Control

### LINKED ISSUE #2595: Windows: Smart App Control blocks browse.exe; the CLI has no Node lane (the server does)
> **Corrected 2026-08-16.** Two claims below were wrong: Smart App Control does **not** block `bun.exe` (it is signed), and the "Residual gap" section was therefore also wrong. The real cause is narrower: the *compiled artifact* is unsigned, not the toolchain. See the correction comment for detail. PR #2596 has been reworked accordingly.

## Summary

On Windows 11 with Smart App Control in enforcement mode, `browse/dist/browse.exe` cannot launch. Every `$B` command in `browse/SKILL.md` fails immediately, so `/browse` and every skill that depends on it (`/qa`, `/design-review`, `/benchmark`, `/canary`) is dead on that machine.

The server half already has a Node lane for Windows. The CLI half does not. Adding one fixes this without disabling any security feature.

## Environment

| | |
|---|---|
| OS | Windows 11 Pro 10.0.26200 (build 26200) |
| gstack | 1.57.10.0 (`a5833c4`) |
| browse dist | `a5833c413f98b13f105beac96262e8098b628461` |
| Node | v24.19.0 |
| Smart App Control | enforcement (`VerifiedAndReputablePolicyState = 1`) |

## Reproduction

1. Windows 11 with Smart App Control on (default for many clean installs since 22H2).
2. Install gstack, build browse.
3. Run any browse command.

```
PS> & "$env:USERPROFILE\.claude\skills\gstack\browse\dist\browse.exe" status
Program 'browse.exe' failed to run: An Application Control policy has blocked this file
```

Under Git Bash the same block surfaces as a confusing `Permission denied`, which sends you looking at file permissions instead of code integrity:

```
$ ~/.claude/skills/gstack/browse/dist/browse status
bash: .../browse/dist/browse.exe: Permission denied
```

The real cause is in `Microsoft-Windows-CodeIntegrity/Operational`:

```
Event 3077: Code Integrity determined that a process
(\Device\HarddiskVolume3\Windows\System32\WindowsPowerShell\v1.0\powershell.exe)
attempted to load \Device\HarddiskVolume3\Users\<user>\.claude\skills\gstack\browse\dist\browse.exe
that did not meet the Enterprise signing level requirements or violated code
integrity policy (Policy ID:{0283ac0f-fff1-49ae-ada1-8a933130cad6}).

Event 3118: Smart App Control Block Details
```

`browse.exe` is a Bun-compiled binary with no Authenticode signature. Smart App Control rejects unsigned executables.

## Why the obvious workarounds do not apply

- **Smart App Control has no per-file allowlist.** There is no UI or registry path to exempt one binary.
- **Turning it off is a one-way door.** Once disabled, Smart App Control cannot be re-enabled without a clean Windows reinstall. Asking users to permanently drop a system-wide protection to run one CLI is not a reasonable ask.
- **Self-signing does not help.** Smart App Control does not trust self-signed certificates.

## Root cause

`browse` is two processes, and only one of them already avoids Bun on Windows.

The **server** has a Node lane, added because Bun cannot drive Playwright's Chromium on Windows (oven-sh/bun#4253, #9911):

- `browse/scripts/build-node-server.sh` produces `dist/server-node.mjs`
- `src/cli.ts` has an `IS_WINDOWS` branch that spawns it with `node`, and hard-fails if the bundle is missing:

```ts
// src/cli.ts
const NODE_SERVER_SCRIPT = IS_WINDOWS ? resolveNodeServerScript() : null;
if (IS_WINDOWS && !NODE_SERVER_SCRIPT) {
  throw new Error('server-node.mjs not found. Run `bun run build` ...');
}
```

The **CLI** has no equivalent. It only ships as the Bun-compiled `dist/browse.exe`. So on Windows the architecture is already "signed Node runs the server", but an unsigned binary is still required to start it.

`node.exe` is signed and Smart App Control allows it. Building the CLI as a Node bundle removes the last unsigned executable from the path.

## Proposed fix

Add a Node lane for the CLI, mirroring the one that already exists for the server.

Good news on feasibility: `src/cli.ts` imports only Node built-ins at the top (`fs`, `path`, `child_process`). Across its whole import graph (`error-handling`, `file-permissions`, `config`, `proxy-config`, `proxy-redact`, `terminal-agent-control`, `socks-bridge`) the only Bun APIs used are:

| API | Already in `bun-polyfill.cjs`? |
|---|---|
| `Bun.spawn` | yes |
| `Bun.spawnSync` | yes |
| `Bun.sleep` | yes |
| `Bun.stdin` | **no** (used only by `chain` reading stdin) |

So the existing polyfill covers everything except a small `Bun.stdin.text()` shim. `import.meta.dir` needs the same rewrite `build-node-server.sh` already does.

### Build script

I wrote this in Node rather than bash on purpose. In PowerShell, `bash` resolves to WSL's `bash.exe`, which fails outright on machines with no WSL distro installed, and that is exactly the population hitting this bug. It also uses esbuild rather than `bun build` so the affected machine can run it at all. If you prefer `bun build` for consistency with `build-node-server.sh`, that works for maintainer and CI builds, but it will not help an affected user build locally (see "Residual gap" below).

<details>
<summary><code>browse/scripts/build-node-cli.mjs</code></summary>

```js
/**
 * Build a Node.js-compatible CLI bundle for Windows.
 *
 * The shipped dist/browse.exe is a Bun-compiled binary. It is unsigned, so
 * Windows Smart App Control (enforcement mode) refuses to load it. SAC has no
 * per-file allowlist, and disabling it is irreversible without a Windows
 * reinstall.
 *
 * The server half already runs under Node. node.exe is signed, so SAC allows
 * it. This builds the missing other half: the CLI itself, as a Node bundle, so
 * nothing unsigned ever needs to launch.
 */

import { spawnSync } from 'node:child_process';
import { fileURLToPath } from 'node:url';
import * as path from 'node:path';
import * as fs from 'node:fs';

const SCRIPT_DIR = path.dirname(fileURLToPath(import.meta.url));
const BROWSE_DIR = path.resolve(SCRIPT_DIR, '..');
const SRC_DIR = path.join(BROWSE_DIR, 'src');
const DIST_DIR = path.join(BROWSE_DIR, 'dist');
const OUT = path.join(DIST_DIR, 'cli-node.mjs');

console.log('Building Node-compatible CLI bundle...');

// npx is a .cmd shim on Windows, which Node will only launch via a shell. Node
// deprecates (DEP0190) passing an args array alongside shell:true, so build one
// quoted command string instead.
const npx = process.platform === 'win32' ? 'npx.cmd' : 'npx';
const quote = (s) => `"${s}"`;
const command = [
  npx, '--yes', 'esbuild',
  quote(path.join(SRC_DIR, 'cli.ts')),
  '--bundle', '--platform=node', '--format=esm',
  quote(`--outfile=${OUT}`),
  '--external:playwright',
  '--external:playwright-core',
  '--external:diff',
  '--external:bun:sqlite',
  '--external:@ngrok/ngrok',
].join(' ');

const build = spawnSync(command, { cwd: BROWSE_DIR, stdio: 'inherit', shell: true });
if (build.status !== 0) {
  console.error('esbuild failed.');
  process.exit(build.status ?? 1);
}

// import.meta.dir is a Bun-ism. Point it at browse/src so the dev-mode branches
// in resolveServerScript()/resolveNodeServerScript() resolve correctly.
let bundle = fs.readFileSync(OUT, 'utf8').replaceAll('import.meta.dir', '__browseNodeSrcDir');

const header = `// ── Windows Node.js compatibility (auto-generated by build-node-cli.mjs) ──
import { fileURLToPath as _ftp } from "node:url";
import { dirname as _dn } from "node:path";
import { createRequire as _cr } from "node:module";
const __browseNodeSrcDir = _dn(_dn(_ftp(import.meta.url))) + "/src";
{
  const _r = _cr(import.meta.url);
  _r("./bun-polyfill.cjs");
  if (globalThis.Bun && !globalThis.Bun.stdin) {
    globalThis.Bun.stdin = {
      text() {
          let data = "";
          process.stdin.setEncoding("utf8");
          process.stdin.on("data", (chunk) => { data += chunk; });
          process.stdin.on("end", () => resolve(data));
          process.stdin.on("error", reject);
        });
      },
    };
  }
}
// ── end compatibility ──
`;

fs.writeFileSync(OUT, header + bundle);
fs.copyFileSync(path.join(SRC_DIR, 'bun-polyfill.cjs'), path.join(DIST_DIR, 'bun-polyfill.cjs'));

// Launcher wrappers. dist/browse keeps the SKILL.md `$B` resolution working
// verbatim, so no SKILL.md change is needed.
const bashWrapper = `#!/usr/bin/env bash
set -e
DIR="$(cd "$(dirname "\${BASH_SOURCE[0]}")" && pwd)"
CLI="$DIR/cli-node.mjs"
if [ ! -f "$CLI" ]; then
  echo "cli-node.mjs missing. Run: node \\"$DIR/../scripts/build-node-cli.mjs\\"" >&2
  exit 1
fi
# node.exe cannot read MSYS-style /c/... paths.
if command -v cygpath >/dev/null 2>&1; then
  CLI="$(cygpath -w "$CLI")"
fi
exec node "$CLI" "$@"
`;

const cmdWrapper = `@echo off
node "%~dp0cli-node.mjs" %*
`;

fs.writeFileSync(path.join(DIST_DIR, 'browse'), bashWrapper);
fs.writeFileSync(path.join(DIST_DIR, 'browse.cmd'), cmdWrapper);
try { fs.chmodSync(path.join(DIST_DIR, 'browse'), 0o755); } catch {}

console.log(`Node CLI bundle ready: ${OUT}`);
console.log('Wrappers restored: dist/browse, dist/browse.cmd');
```

</details>

### Why a `dist/browse` wrapper

The SETUP block in `browse/SKILL.md` resolves `$B` to `browse/dist/browse` and gates on `[ -x "$B" ]`. Writing the bash wrapper to that exact path means the documented workflow keeps working with no SKILL.md change and no template regeneration. `dist/browse.cmd` covers PowerShell and cmd users.

`browse/dist/` is already gitignored, so only `scripts/build-node-cli.mjs` is a new tracked file.

## What I verified

Run on the affected hardware, both shells:

| Check | Result |
|---|---|
| `browse/SKILL.md` SETUP block, verbatim | `READY: .../dist/browse` |
| `$B goto` + `text`, static page | full text returned |
| `$B goto` + `wait --networkidle` + `text`, JS-rendered SPA | full text returned |
| `$B snapshot -i` | `@e1 [link] "Learn more"` |
| `$B screenshot` | 10,403 byte PNG written |
| `$B console --errors` | CSP errors captured correctly |
| `$B status` (cold start) | server started, `Status: healthy` |
| exit codes on `goto` / `text` | 0 |
| PowerShell via `dist/browse.cmd` | working |
| Git Bash via `dist/browse` | working |

`browse.exe` is left untouched. Nothing unsigned launches.

**I could not run `bun test`.** bun is not installed on this machine, and installing it would very likely hit the same Smart App Control wall since `bun.exe` is also unsigned. So this is verified against real usage but not against your tier 1 suite. Worth a maintainer running `bun test` before merging.

## Residual gap (separate issue, not fixed here)

This restores browse on a machine that already has a built `dist/`. It does not make a **fresh** install work on a Smart App Control machine, because:

- `setup` hard-fails with `Error: bun is required but not installed`
- `bun run build` is the only documented build path
- `bun.exe` is unsigned, so Smart App Control blocks it too
- `dist/` is gitignored, so there is no prebuilt artifact to fall back on

A complete story needs either a bun-free build path or signed release artifacts. That is a much larger change and I did not want to bundle it into this one. Happy to open a separate issue if useful.

## Suggested follow-ups

1. Wire `build-node-cli.mjs` into `scripts/build.sh` on the Windows branch so `bun run build` produces both bundles.
2. Improve the error when the CLI cannot launch. Right now Git Bash reports `Permission denied`, which points at file permissions rather than code integrity. Checking `Microsoft-Windows-CodeIntegrity/Operational` for a matching 3077 and printing "Smart App Control blocked this binary" would save the next person a lot of time.
3. Add a note to the Windows section of the docs. This will hit more people over time, since Smart App Control is on by default for clean Windows 11 installs and only evaluates itself out of enforcement in some cases.

knetics9000:
Fixes #2595

> **Updated.** The first version of this PR used Node + esbuild and claimed Smart App Control blocks `bun.exe`. That was wrong: bun is signed and runs fine. The PR now uses bash + `bun build`, matching `build-node-server.sh`, and adds no new dependency. Details in the comment below.

## Problem

`bun build --compile` emits an executable with no Authenticode signature, and Windows Smart App Control in enforcement mode refuses to load unsigned binaries:

```
PS> browse.exe status
Program 'browse.exe' failed to run: An Application Control policy has blocked this file
```

Confirmed in `Microsoft-Windows-CodeIntegrity/Operational` (event 3077, plus Smart App Control block 3118). Every `$B` command dies, taking `/browse`, `/qa`, `/design-review`, `/benchmark` and `/canary` with it.

Under Git Bash the same block surfaces as `Permission denied`, which sends you looking at file permissions instead of code integrity.

Smart App Control has no per-file allowlist, and cannot be re-enabled once disabled without a clean Windows reinstall. Self-signing is not trusted. So there is no user-side workaround.

## The precise cause

Measured on the affected machine with `Get-AuthenticodeSignature`:

| Binary | Signature | Smart App Control |
|---|---|---|
| `bun.exe` | Valid, CN=Codeblog CORP | allowed |
| `node.exe` | Valid, CN=OpenJS Foundation | allowed |
| `browse.exe` | **NotSigned** | **blocked** |

bun and node both run fine. The toolchain is not the problem. The problem is the *artifact*: every locally compiled `browse.exe` is unsigned, and because it is unique to the machine that built it, it can never accumulate the file reputation Smart App Control accepts as an alternative to a signature.

That is why this cannot be fixed by rebuilding, and why it will hit every user who has Smart App Control on, which is the default for many clean Windows 11 installs.

## Why this is a small fix

The server half already exploits the fact that `node.exe` is signed. #255 added `build-node-server.sh` and the `IS_WINDOWS` branch in `src/cli.ts` that spawns `node server-node.mjs`, because Bun cannot drive Playwright's Chromium there.

So Windows was already **"signed Node runs the server"** while an unsigned binary was still required to start it. This PR adds the missing other half.

`src/cli.ts` imports only Node built-ins at the top. Across its whole import graph the only Bun APIs used are:

| API | Already in `bun-polyfill.cjs`? |
|---|---|
| `Bun.spawn` | yes |
| `Bun.spawnSync` | yes |
| `Bun.sleep` | yes |
| `Bun.stdin` | no (reached only by `chain` reading stdin) |

Three of four are already covered, and `import.meta.dir` needs the same rewrite `build-node-server.sh` already does.

## Changes

Two files, 130 insertions, no deletions.

- **`browse/scripts/build-node-cli.sh`** (new) mirrors `build-node-server.sh` structure exactly: same `bun build --target=node` invocation, same externals, same `perl -pi -e` rewrite of `import.meta.dir`, same polyfill header injection. Adds a `Bun.stdin` shim.
- On Windows it also writes `dist/browse` and `dist/browse.cmd` launchers. `dist/browse` keeps the SETUP block in `browse/SKILL.md` working unchanged, since it resolves `$B` to that path and gates on `-x`. **No SKILL.md or template change needed.**
- **`scripts/build.sh`** calls it on the line after the server bundle.

`browse/dist/` is gitignored, so `build-node-cli.sh` is the only new tracked file. No new dependencies.

### One thing to check in review

**The wrapper writes are Windows-gated** (`case "$(uname -s)"`). On macOS and Linux `build.sh:27` compiles the real binary to `browse/dist/browse`, so writing a wrapper there would clobber it. On Windows that step emits `browse.exe` instead, leaving the extensionless path free. I reasoned this from `build.sh` rather than observing it, so it is worth a second pair of eyes.

## Verification

Windows 11 Pro 26200, Smart App Control enforcing, Node v24.19.0, bun 1.3.14.

**`bun test` on this branch is identical to `main`:**

| | pass | skip | fail | errors |
|---|---|---|---|---|
| `main` | 5241 | 450 | 1435 | 12 |
| this branch | 5241 | 450 | 1435 | 12 |

Same 7126 tests across 507 files. This change is test-neutral.

Separately worth knowing: **1435 tests fail on a fresh clone on Windows before any change**, largely from binaries that are not built yet. Happy to open a separate issue with detail if that is useful.

Functional checks, both shells:

| Check | Result |
|---|---|
| `browse/SKILL.md` SETUP block, verbatim | `READY: .../dist/browse` |
| `$B goto` + `text`, static page | full text returned |
| `$B goto` + `wait --networkidle` + `text`, JS-rendered SPA | full text returned |
| `$B snapshot -i` | `@e1 [link] "Learn more"` |
| `$B screenshot` | 10,403 byte PNG written |
| `$B console --errors` | CSP errors captured |
| `$B status` from cold | server started, `Status: healthy` |
| exit codes on `goto` / `text` | 0 |
| PowerShell via `dist/browse.cmd` | working |
| Git Bash via `dist/browse` | working |

`browse.exe` is untouched and still builds.

## What I could not verify

**Non-Windows builds were not executed.** The macOS/Linux path is reasoned from `build.sh`, not observed. The wrapper stage is gated behind a `uname -s` check, so the intended effect there is one extra bundle in `dist/` and nothing else, but a maintainer running `bun run build` on macOS would confirm that cheaply.

### DIFF
--- a/browse/scripts/build-node-cli.sh
+++ b/browse/scripts/build-node-cli.sh
@@ -1,0 +1,129 @@
+#!/usr/bin/env bash
+# Build a Node.js-compatible CLI bundle for Windows.
+#
+# `bun build --compile` emits an executable with no Authenticode signature, and
+# Windows Smart App Control in enforcement mode refuses to load unsigned
+# binaries. Every locally compiled browse.exe is unique to the machine that
+# built it, so it can never accumulate the reputation SAC accepts either. The
+# result is that browse.exe cannot launch at all: every $B command dies, taking
+# /browse, /qa, /design-review, /benchmark and /canary with it. SAC has no
+# per-file allowlist and cannot be re-enabled once disabled without a clean
+# Windows reinstall.
+#
+# bun.exe and node.exe are both signed (Codeblog CORP and OpenJS Foundation),
+# so they run fine. The server half already exploits that: build-node-server.sh
+# plus the IS_WINDOWS branch in src/cli.ts spawns `node server-node.mjs`. This
+# script builds the missing other half, so no unsigned binary has to launch.
+
+set -e
+
+GSTACK_DIR="$(cd "$(dirname "$0")/../.." && pwd)"
+SRC_DIR="$GSTACK_DIR/browse/src"
+DIST_DIR="$GSTACK_DIR/browse/dist"
+
+echo "Building Node-compatible CLI bundle..."
+
+# Step 1: Transpile cli.ts to a single .mjs bundle (externalize runtime deps).
+# Same externals as build-node-server.sh.
+bun build "$SRC_DIR/cli.ts" \
+  --target=node \
+  --outfile "$DIST_DIR/cli-node.mjs" \
+  --external playwright \
+  --external playwright-core \
+  --external diff \
+  --external "bun:sqlite" \
+  --external "@ngrok/ngrok"
+
+# Step 2: Post-process
+# Replace import.meta.dir with a resolvable reference
+perl -pi -e 's/import\.meta\.dir/__browseNodeSrcDir/g' "$DIST_DIR/cli-node.mjs"
+
+# Step 3: Create the final file with polyfill header injected after the first line.
+# bun-polyfill.cjs already covers Bun.spawn, Bun.spawnSync and Bun.sleep, which is
+# everything the CLI's import graph uses except Bun.stdin (reached only by `chain`
+# when it reads from stdin), so shim that one here.
+{
+  head -1 "$DIST_DIR/cli-node.mjs"
+  echo '// ── Windows Node.js compatibility (auto-generated) ──'
+  echo 'import { fileURLToPath as _ftp } from "node:url";'
+  echo 'import { dirname as _dn } from "node:path";'
+  echo 'import { createRequire as _cr } from "node:module";'
+  echo 'const __browseNodeSrcDir = _dn(_dn(_ftp(import.meta.url))) + "/src";'
+  echo '{'
+  echo '  const _r = _cr(import.meta.url);'
+  echo '  _r("./bun-polyfill.cjs");'
+  echo '  if (globalThis.Bun && !globalThis.Bun.stdin) {'
+  echo '    globalThis.Bun.stdin = {'
+  echo '      text() {'
+  echo '        return new Promise((resolve, reject) => {'
+  echo '          let data = "";'
+  echo '          process.stdin.setEncoding("utf8");'
+  echo '          process.stdin.on("data", (chunk) => { data += chunk; });'
+  echo '          process.stdin.on("end", () => resolve(data));'
+  echo '          process.stdin.on("error", reject);'
+  echo '        });'
+  echo '      },'
+  echo '    };'
+  echo '  }'
+  echo '}'
+  echo '// ── end compatibility ──'
+  tail -n +2 "$DIST_DIR/cli-node.mjs"
+} > "$DIST_DIR/cli-node.tmp.mjs"
+
+mv "$DIST_DIR/cli-node.tmp.mjs" "$DIST_DIR/cli-node.mjs"
+
+# Step 4: Copy polyfill to dist/
+cp "$SRC_DIR/bun-polyfill.cjs" "$DIST_DIR/bun-polyfill.cjs"
+
+echo "Node CLI bundle ready: $DIST_DIR/cli-node.mjs"
+
+# Step 5: Launcher wrappers, Windows only.
+#
+# On macOS and Linux, build.sh compiles the real binary to browse/dist/browse,
+# so writing a wrapper there would clobber it. On Windows that same step emits
+# browse.exe instead, leaving the extensionless path free — which is what keeps
+# the SETUP block in browse/SKILL.md working unchanged, since it resolves $B to
+# browse/dist/browse and gates on -x.
+case "$(uname -s)" in
+  MINGW*|MSYS*|CYGWIN*|Windows_NT)
+    cat > "$DIST_DIR/browse" <<'WRAPPER'
+#!/usr/bin/env bash
+# Node launcher for gstack browse (Windows / Smart App Control).
+#
+# dist/browse.exe is unsigned, so Smart App Control blocks it. node.exe is
+# signed, so running the CLI as a Node bundle works.
+#
+# Rebuild with: bash browse/scripts/build-node-cli.sh
+set -e
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+CLI="$DIR/cli-node.mjs"
+
+if [ ! -f "$CLI" ]; then
+  echo "cli-node.mjs missing. Run: bash browse/scripts/build-node-cli.sh" >&2
+  exit 1
+fi
+
+# node.exe is a native Windows binary and cannot read MSYS-style /c/... paths.
+if command -v cygpath >/dev/null 2>&1; then
+  CLI="$(cygpath -w "$CLI")"
+fi
+
+exec node "$CLI" "$@"
+WRAPPER
+
+    cat > "$DIST_DIR/browse.cmd" <<'WRAPPER'
+@echo off
+REM Node launcher for gstack browse (Windows / Smart App Control).
+REM dist\browse.exe is unsigned and blocked by Smart App Control; node.exe is signed.
+REM Rebuild with: bash browse/scripts/build-node-cli.sh
+node "%~dp0cli-node.mjs" %*
+WRAPPER
+
+    chmod +x "$DIST_DIR/browse"
+    echo "Launcher wrappers written: $DIST_DIR/browse, $DIST_DIR/browse.cmd"
+    ;;
+  *)
+    echo "Skipping launcher wrappers (non-Windows: dist/browse is the compiled binary)."
+    ;;
+esac
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -30,6 +30,7 @@ esac
 "$BUN_CMD" build --compile make-pdf/src/cli.ts --outfile make-pdf/dist/pdf
 "$BUN_CMD" build --compile bin/gstack-global-discover.ts --outfile bin/gstack-global-discover
 bash browse/scripts/build-node-server.sh
+bash browse/scripts/build-node-cli.sh
 bash scripts/write-version-files.sh browse/dist/.version design/dist/.version make-pdf/dist/.version
 chmod +x browse/dist/browse browse/dist/find-browse design/dist/design make-pdf/dist/pdf bin/gstack-global-discover
 rm -f .*.bun-build


knetics9000:
﻿## Correction to the original version of this PR

I got two things wrong in the first version and want to flag them explicitly rather than quietly force-push over them.

**1. I claimed Smart App Control blocks `bun.exe`. It does not.** I asserted that without testing it. bun installed and ran fine on the affected machine:

| Binary | Signature | SAC |
|---|---|---|
| `bun.exe` | Valid, CN=Codeblog CORP | allowed |
| `node.exe` | Valid, CN=OpenJS Foundation | allowed |
| `browse.exe` | NotSigned | blocked |

The real cause is narrower and, I think, a better argument for this PR: it is not the toolchain, it is the *artifact*. `bun build --compile` produces an unsigned executable that is unique per machine, so it can never be signed and can never earn file reputation. Rebuilding cannot fix it.

**2. That invalidated my justification for using Node + esbuild.** The original version argued a bash + bun build script "would not help the users hitting this bug." Since bun runs fine, that was wrong, and it was the only reason not to match your existing convention. **The PR now uses bash + `bun build`, mirroring `build-node-server.sh` line for line, and drops the esbuild dependency I had added to `package.json`.** Two files changed instead of three.

**3. I can now run the test suite, which I originally said I could not.** `bun test` on this branch is identical to `main` on Windows: 5241 pass / 450 skip / 1435 fail / 12 errors on both.

**4. The "residual gap" section is gone.** I had claimed a fresh install could not work on a Smart App Control machine because `setup` requires bun. Since bun runs, `setup` and `bun run build` both work. `browse.exe` is the only thing that cannot launch, and this PR is a complete fix for it rather than a partial one.

Issue #2595 has been corrected too.

Sorry for the churn. The current version is what I would have opened if I had checked bun before asserting.

PR #2606, opened 2026-08-17
fix(redact-prepush): scan the pushed range, not the whole repo, on a new ref

fordpiggy1:
## The bug

Pushing a **new branch** can fail with `HIGH engine.input_too_large`, reported under the heading "credential(s) in the pushed diff". The only documented way past it is `GSTACK_REDACT_PREPUSH=skip`.

On a new ref there is no remote SHA to diff against, so the hook approximated the branch point as `merge-base(local, origin/HEAD)`. In a repo whose default branch has drifted from the branch it actually ships from, that base sits hundreds of commits below the real branch point.

Measured in a repo that ships from a long-lived release branch while `origin/main` sits 200 commits behind it:

```
commits in the approximated range = 201        (real range: 1)
added-line bytes handed to scan() = 3,652,518
engine cap                        = 1,048,576   -> 3.48x over
```

The push itself was **2,141 bytes**. Input size tracked the repo, not the diff.

## Why this is worse than a false alarm

A new ref that **genuinely contained a credential** produced the same size error, so the credential was never scanned at all.

The chain is: a capacity bug manufactures a bogus "credential in the pushed diff" finding, the only escape is the documented skip flag, and the resulting habit of skipping then lets a real secret through. The operator believes a guard ran when none did. A guard that fails closed and hands out a bypass in the same breath ends up weaker than one that fails open.

## The fix

Three changes. None of them reduces detection.

**1. Correct base — `pushedRangeBase()`.** Replaces the merge-base guess with `git rev-list --boundary <local> --not --remotes=<remote>`, which asks git directly which commits this push adds to *this* remote. Falls back to octopus merge-base, then the old merge-base, then the empty tree; every fallback scans **more**, never less. The remote name comes from `process.argv[2]`, which git already passes and the managed wrapper already forwards.

This is a sizing fix rather than a detection reduction: the bytes it stops scanning are bytes the remote **already holds and this push does not send**, which is exactly what the existing-ref path (`remote..local`) has always excluded. Everything the push introduces is still scanned byte for byte.

**2. Chunked scanning.** The engine's 1 MiB fail-closed cap is untouched, so any other caller still fails closed. The hook slices added lines into 512 KiB chunks with 16 KiB overlap and scans **every** chunk, deduping findings by global line and column. A single added line larger than a chunk (minified bundle, base64 blob) is split into overlapping windows rather than skipped. No threshold was raised, no file is skipped, no severity downgraded. A 3.4 MiB scan takes 0.44 s.

**3. Honest messaging.** Size is no longer printed under "rotate the credential". If a region genuinely cannot be scanned, a separate `UNSCANNED` block names the byte count and line offset, says plainly that nothing was found because nothing was read, and **still exits 1**.

## Verification

All cases run by invoking the hook with synthetic stdin. No push was made and no remote branch was created or deleted during testing.

| case | before | after |
|---|---|---|
| new ref, all-zeros remote SHA | **exit 1** `engine.input_too_large` | **exit 0**, clean |
| existing ref, small diff | exit 0 | exit 0 |
| new ref **containing** fabricated credentials | **exit 1** `engine.input_too_large`, secret never scanned | **exit 1**, names `db.url_with_password` and `aws.access_key` |

The third row is the one that matters: it is the proof the fix did not gut the guard.

Supporting cases, same method: a 3.4 MiB clean push (was blocked, now passes); a credential 3.2 MiB deep, past the old cap (was invisible, now named); one 1.6 MiB single added line (found via windowing).

Test secrets are fabricated and point at `example.invalid`. No real credential was read or written.

**Suite: 185 pass, 0 fail across the 8 redact test files** (up from 176). 5 of the 9 new tests fail against the pre-fix binary, so they genuinely pin the behaviour.

## Known limits, documented in the source

- The 64 MiB `maxBuffer` on the diff call is a second, much rarer ceiling that still ends in block-plus-bypass. Post-fix it needs a >64 MiB single push. Fixing it properly means streaming the diff through a temp file; left alone to keep this change narrow.
- The new-ref base trusts `refs/remotes/<remote>/*`. After a force-push or branch deletion since the last fetch, it may believe the remote holds a commit it no longer does. Not a regression — the old merge-base path read the same stale refs more coarsely.
- Pushing commits the remote already holds under a **new branch name** is now deliberately not re-scanned. That matches this hook's documented scope as a pushed-diff scanner rather than a history scanner, but it is a conscious narrowing rather than an accident.

### DIFF
--- a/bin/gstack-redact-prepush
+++ b/bin/gstack-redact-prepush
@@ -10,16 +10,19 @@
  *
  * Git pre-push interface: refs are read from STDIN, one per line:
  *   <local ref> <local sha> <remote ref> <remote sha>
+ * plus argv: <remote name> <remote url>.
  * We scan the ADDED lines of <remote sha>..<local sha> per ref (what's being
  * pushed). Special cases:
- *   - remote sha all-zeroes  → new branch: diff against merge-base with the
- *     remote's default branch (fallback: scan all commits unique to local ref).
+ *   - remote sha all-zeroes  → new branch: git does not tell us a base, so we
+ *     derive one from what THIS remote already has (see pushedRangeBase).
  *   - local sha all-zeroes   → branch delete: nothing to scan, skip.
  *   - force-push             → remote..local still gives the net new content.
  *
  * Behavior:
  *   - HIGH finding in added lines → print + exit 1 (block), for public AND private.
  *   - MEDIUM → warn (non-blocking). LOW/WARN → silent.
+ *   - Any byte of the pushed diff we could NOT scan → print exactly what was
+ *     missed + exit 1. Never reported as a "credential".
  *   - GSTACK_REDACT_PREPUSH=skip → log + exit 0 (escape valve).
  *
  * Installed/uninstalled via `gstack-redact install-prepush-hook` (see the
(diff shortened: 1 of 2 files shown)


garrytan:
Closing as superseded: main's redact-prepush already scans the pushed range via unknownRemoteTipBase() with chunking (bin/gstack-redact-prepush:111-150, :294); the 16KiB overlap this PR adds was evaluated and declined (a line boundary cannot bisect a detectable secret — comment at :311). Thank you for the report/patch — the class is covered on main.

garrytan:
Closing as superseded: main's redact-prepush already scans the pushed range via unknownRemoteTipBase() with chunking (bin/gstack-redact-prepush:111-150, :294); the 16KiB overlap this PR adds was evaluated and declined (a line boundary cannot bisect a detectable secret — comment at :311). Thank you for the report/patch — the class is covered on main.

Issue #2490, opened 2026-08-08
Split plan-eng-review-arch-finding: one-way architecture forks vs two-way code-derived defect findings (the only safely-tunable class is currently unreachable)


alexvibecommit:
## Summary

`plan-eng-review-arch-finding` is declared `door_type: 'one-way'` and absorbs two behaviorally different question types. One genuinely deserves the one-way gate. The other is the single highest-value candidate for `never-ask` in the whole registry, and the current classification makes it permanently untunable.

Proposal: split it into `plan-eng-review-arch-fork` (one-way) and `plan-eng-review-code-defect` (two-way).

## Evidence

Hand-audited from one `/plan-eng-review` session on a real infrastructure plan. Every question that would have fired as `plan-eng-review-arch-finding` sorted cleanly into two columns:

**Genuinely one-way — architecture forks, data-model changes, surface-area removal**

| Question | Outcome |
|---|---|
| Retire 3 MCP tools + drop `traces` | **overrode the recommendation** |
| New `ingest_sessions` persistent object | took recommendation |
| Drop traces, ship prompt-text search instead | **overrode the recommendation** |

**Code-derived defect findings — "the code says this is broken, here is the fix"**

| Question | Outcome |
|---|---|
| N+1 in `loadCandidateSequence` → cache + batch | took recommendation |
| `min_machines_running` 0→1, swap decompress lib | took recommendation |
| Server-side tail buffer for dropped records | took recommendation |
| `readable_repository_ids` gate on new tables | took recommendation |
| (+4 more of the same shape) | took recommendation |

**Right column: 8 for 8.** Left column: contains both of the session's overrides.

The reason the split is clean is not agreement rate, it is information asymmetry. In the right column the reviewer has read code the human has not, and the finding carries its own answer — there is no judgment left for the human to add. In the left column the human holds context the reviewer does not: product direction, what a surface is worth keeping, what is cheap to walk back.

## Why this blocks the feature rather than merely limiting it

Filtering the registry for what a `/plan-eng-review` user could actually tune today:

- `plan-eng-review-arch-finding` — one-way, so `never-ask` is refused. Also see the separate issue where such a write silently succeeds and never fires.
- `plan-eng-review-scope-reduce` — two-way, but this is the override class. Tuning it is the wrong move.
- `plan-eng-review-outside-voice` — two-way, and it was overridden. Wrong move.
[...]

Issue #2489, opened 2026-08-08
Question Tuning preamble points at a relative `scripts/question-registry.ts` that does not resolve from a project cwd, so agents silently fabricate question_ids (21/21 unregistered in one session)


alexvibecommit:
## Summary

The Question Tuning preamble tells the agent to look up `question_id` at a **relative path that does not resolve from a user's project directory**, and offers "make one up" as the alternative in the same sentence. The result is that agents silently fabricate `question_id`s instead of using the registry, and every fabricated id is a singleton that no per-question preference can ever attach to.

`scripts/resolvers/question-tuning.ts:26`:

```
Before each AskUserQuestion, choose `question_id` from `scripts/question-registry.ts` or
`{skill}-{slug}`, then run `${bin}/gstack-question-preference --check "<id>"`.
```

Note the asymmetry inside that one sentence: the **bin** path is interpolated absolute (`${bin}` → `~/.claude/skills/gstack/bin`), the **registry** path is hardcoded relative. Agents run with cwd set to the user's project, not the gstack install:

```console
$ pwd
/Users/<me>/Documents/<my-project>
$ ls scripts/question-registry.ts
ls: scripts/question-registry.ts: No such file or directory
```

The file exists only at `~/.claude/skills/gstack/scripts/question-registry.ts`. The lookup fails, the documented `or {skill}-{slug}` fallback succeeds, and the agent proceeds with an invented id and no error anywhere.

This line is emitted into every generated `SKILL.md` — I count it in ~40 of them (`plan-tune`, `autoplan`, `cso`, `plan-eng-review`, `design-*`, `ios-qa`, `learn`, `pair-agent`, …), so it is not skill-specific.

## Observed impact

One `/plan-eng-review` session on a real plan emitted **21 question_ids, 0 of them registered**:

```
plan-eng-review-arch-1-delta-lineage        plan-eng-review-quality-1-hook-durability
plan-eng-review-arch-2-parity-gate          plan-eng-review-perf-1-cumulative-rebuild
plan-eng-review-ov-3-subagent-transcripts   plan-eng-review-tests-2-client-tier
plan-eng-review-scope-challenge             plan-eng-review-todos-batch            …
```

[...]

PR #2607, opened 2026-08-17
fix(plan-tune): reject never-ask on one-way ids at --write

### LINKED ISSUE #2488: gstack-question-preference --write accepts never-ask on one-way ids: preference stores, --stats reports it configured, --check refuses it forever
## Summary

`bin/gstack-question-preference --write` accepts `never-ask` on a `door_type: 'one-way'` question_id, exits 0, and reports success. `--check` then refuses that same preference forever. The result is a preference that is **stored, counted by `--stats`, surfaced by `/plan-tune` as configured, and can never fire**.

The asymmetry is inside one file:

- `--check` (bin/gstack-question-preference:61) calls `isOneWayDoor()` **first** and returns `ASK_NORMALLY` before the `switch (pref)` is reached.
- `--write` (bin/gstack-question-preference:139) validates only that the string is in `ALLOWED_PREFS = ['always-ask', 'never-ask', 'ask-only-for-one-way']`. It never looks at `door_type`.

## Reproduction

Verified on `1.60.1.0`, macOS arm64, starting from an empty `question-preferences.json`:

```console
$ cat ~/.gstack/projects/<slug>/question-preferences.json
{}

$ gstack-question-preference --write '{"question_id":"plan-eng-review-arch-finding","preference":"never-ask","source":"plan-tune"}'
OK: plan-eng-review-arch-finding → never-ask (source: plan-tune)
$ echo $?
0

$ gstack-question-preference --stats
TOTAL: 1
ALWAYS_ASK: 0
NEVER_ASK: 1          # ← reads as configured and working
ASK_ONLY_ONE_WAY: 0

$ gstack-question-preference --check "plan-eng-review-arch-finding"
ASK_NORMALLY
NOTE: one-way door overrides your never-ask preference for safety.

$ cat ~/.gstack/projects/<slug>/question-preferences.json
{ "plan-eng-review-arch-finding": "never-ask" }
```

`plan-eng-review-arch-finding` is declared `door_type: 'one-way'` in `scripts/question-registry.ts`. 12 of the 53 registered ids are one-way, so this is reachable on ~23% of the registry.

## Why this is worse than the failure it protects against

The one-way gate exists so a `never-ask` cannot silently auto-approve something destructive. That is the right gate and it works. But the failure it produces on the write side is the more dangerous shape of the two:

- A preference that **wrongly auto-decides** is loud. The user sees a decision they did not make, within one session, and goes looking.
- A preference that is **silently inert** is self-congratulatory. `--stats` says `NEVER_ASK: 1`. `/plan-tune` renders it as set. Nobody audits a setting that reports it is working.

The `NOTE:` on `--check` is real, but it prints into the agent's check output at question time, not into the surface a human inspects. `--stats` is the human-facing surface and it carries no warning.

## Suggested fix

Any one of these closes it; (1) plus (3) is the honest combination:

1. **Reject at write.** `--write` resolves `door_type` via `getQuestion(id)` / `isOneWayDoor()` and exits non-zero for `never-ask` and `ask-only-for-one-way` on a one-way id, with a message naming the door type. Cheapest, and it fails at the moment the user can still do something about it.
2. **Accept but annotate.** Store it and have `--stats` / `/plan-tune` render it as `never-ask (INERT — one-way door)`. Preserves the write for a future registry reclassification, which matters if an id later flips to two-way.
3. **Make `--stats` truthful regardless.** Have `--stats` run each stored preference through the same one-way check `--check` uses and report the inert count separately. This is the piece that fixes the inspection surface, and it is needed even if (1) lands, for preferences written before the fix.

## Related observation on `ask-only-for-one-way`

Reading the same block: on a one-way id both `never-ask` and `ask-only-for-one-way` are refused above the `switch`; on a two-way id both fall through to `AUTO_DECIDE` (:81-92). So `ask-only-for-one-way` has no behavior distinct from `never-ask` at any door type — the one-way protection it names is already unconditional.

That is arguably fine as a no-op alias, but the name reads as the safer middle setting, which is exactly why a user reaches for it. If the redundancy is intentional, documenting it as an alias would help; if not, it is a separate bug. Happy to split it out if you would prefer it tracked on its own.

szsunyuan:
## Why (in your own words)

I set `never-ask` on `plan-eng-review-arch-finding` and `--stats` said it worked. The next `--check` still asked, with a note that one-way doors ignore that pref. So the file lied: it stored a preference that can never fire.

`--write` is what I want to stop. If the id is one-way, refuse `never-ask` and `ask-only-for-one-way` (exit 1, name the door type, don't touch the file). `--stats` should say `INERT_ONE_WAY` for leftovers already on disk, not `NEVER_ASK: 1`. `always-ask` on a one-way id is fine — that matches what `--check` already does.

**Fixes #2488.** Related: #2489 (closed — registry path made this easier to hit), #2490 (product split, out of scope), #2390 / #2429 (same `--write` function, unknown-source exit 2, already on tip).

### Why this shape

Issue author asked for reject-at-write plus honest `--stats` for prefs written before the fix. Storing the value and labeling it INERT is the weaker pair — not done.

CONTRIBUTING's wave rule: if two PRs fix the same thing, keep the smaller one. This is 2 files, on current `main`. Fine to absorb into a wave.

### What it does

- `--write never-ask` / `ask-only-for-one-way` on a one-way id exits 1, names `door_type: one-way`, does not write, does not append an event.
- `--write always-ask` on a one-way id still succeeds.
- Origin gate still runs first: a poisoning source on a one-way id exits 2, not 1.
- `--stats` counts leftover one-way suppressing prefs as `INERT_ONE_WAY`. Working `NEVER_ASK` is only prefs that `--check` would actually `AUTO_DECIDE`.
- `--check` tests that still need a planted one-way pref write the file directly, because `--write` now refuses them.

### What it deliberately does not do

- Split `plan-eng-review-arch-finding` (#2490)
- Change `ask-only-for-one-way` alias semantics
- Reject `--write` on `*-split-*` ids (that carve-out stays `--check`-only)
- Keyword-net `--write` (no summary on the write path)
- Hooks, registry, skill templates, `VERSION` / `CHANGELOG` / README
- A migration (schema unchanged; `--stats` is honesty for old files)
- `/ship` (wave stamps version)

AI was used for assistance.

## Live evidence

Same isolated `GSTACK_HOME`, same payload, `cwd` = repo:

```
{"question_id":"plan-eng-review-arch-finding","preference":"never-ask","source":"plan-tune"}
```

**Before** (`upstream/main` binary):

```
$ bin/gstack-question-preference --write '{"question_id":"plan-eng-review-arch-finding","preference":"never-ask","source":"plan-tune"}'
OK: plan-eng-review-arch-finding → never-ask (source: plan-tune)

$ bin/gstack-question-preference --read
{
  "plan-eng-review-arch-finding": "never-ask"
}

$ bin/gstack-question-preference --stats
TOTAL: 1
ALWAYS_ASK: 0
NEVER_ASK: 1
ASK_ONLY_ONE_WAY: 0

$ bin/gstack-question-preference --check plan-eng-review-arch-finding
ASK_NORMALLY
NOTE: one-way door overrides your never-ask preference for safety.
```

**After** (this branch):

```
$ bin/gstack-question-preference --write '{"question_id":"plan-eng-review-arch-finding","preference":"never-ask","source":"plan-tune"}'
gstack-question-preference: cannot set never-ask on one-way question "plan-eng-review-arch-finding" (door_type: one-way)

$ bin/gstack-question-preference --read
{}

$ bin/gstack-question-preference --stats
TOTAL: 0
ALWAYS_ASK: 0
NEVER_ASK: 0
ASK_ONLY_ONE_WAY: 0
INERT_ONE_WAY: 0
```

Planted leftover file (a pref written before this reject existed):

```
$ bin/gstack-question-preference --stats
TOTAL: 1
ALWAYS_ASK: 0
NEVER_ASK: 0
ASK_ONLY_ONE_WAY: 0
INERT_ONE_WAY: 1

$ bin/gstack-question-preference --check plan-eng-review-arch-finding
ASK_NORMALLY
NOTE: one-way door overrides your never-ask preference for safety.
```

## Scope

- **Changed:** `bin/gstack-question-preference`, `test/gstack-question-preference.test.ts`
- **Verified live by:** the transcripts above, plus `bun test test/gstack-question-preference.test.ts` (52 pass)
- **Did NOT test:** Tier 2/3 evals. Full `bun run test` browse/playwright/pdf shards failed here (missing Chromium, `make-pdf` timeouts) — none of those files are in this diff.

## Liveness proof (required)

image

## Checklist

- [x] This is not a generated-file-only diff (I edited the source, not a generated SKILL.md)
- [x] Linked issue or reproduction: #2488

## Test plan

- [x] `bun test test/gstack-question-preference.test.ts` → 52 pass
- [x] Live `--write` on `plan-eng-review-arch-finding` + `never-ask` → exit 1, file stays `{}`
- [x] Live `--write always-ask` on the same id still succeeds
- [x] Live `--stats` on a planted one-way `never-ask` → `INERT_ONE_WAY: 1`, `NEVER_ASK: 0`
- [x] Poisoning source on a one-way id still exits 2

Fork PRs do not get eval secrets; the free file above is the gate.

### DIFF
--- a/bin/gstack-question-preference
+++ b/bin/gstack-question-preference
@@ -8,10 +8,10 @@
 #   --check <id> [--summary-stdin] → emit ASK_NORMALLY | AUTO_DECIDE | ASK_ONLY_ONE_WAY
 #                                    (--summary-stdin pipes the question text so the
 #                                    keyword net can catch ad-hoc destructive ids, #2024)
-#   --write '{...}'                → set a preference (user-origin gate enforced)
+#   --write '{...}'                → set a preference (user-origin gate + one-way write reject)
 #   --read                         → dump preferences JSON
 #   --clear [<id>]                 → clear one or all preferences
-#   --stats                        → short summary
+#   --stats                        → short summary (inert one-way prefs counted separately)
 #
 # User-origin gate
 # ----------------
@@ -21,6 +21,12 @@
 #   - "inline-tool-output"— tune: prefix seen in tool output / file content (REJECTED)
 #   - "inline-file"       — tune: prefix seen in a file the agent read (REJECTED)
 # This is the profile-poisoning defense from docs/designs/PLAN_TUNING_V0.md.
+#
+# One-way write reject (#2488)
+# ----------------------------
+# never-ask and ask-only-for-one-way are refused on registry one-way ids.
+# --check already ignores those prefs; storing them made --stats lie.
+# always-ask on a one-way id is fine (it agrees with the safety override).
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
@@ -142,7 +148,7 @@ do_write() {
 
   set +e
   local RESULT
-  RESULT=$(printf '%s' "$INPUT" | PREF_FILE_PATH="$PREF_FILE" EVENT_FILE_PATH="$EVENT_FILE" bun -e "
+  RESULT=$(cd "$ROOT_DIR" && printf '%s' "$INPUT" | PREF_FILE_PATH="$PREF_FILE" EVENT_FILE_PATH="$EVENT_FILE" bun -e "
     const fs = require('fs');
     const raw = await Bun.stdin.text();
     let j;
@@ -176,6 +182,16 @@ do_write() {
     if (!ALLOWED_SOURCES.includes(j.source)) {
       process.stderr.write('gstack-question-preference: rejected — source \"' + j.source + '\" is not user-originated (profile poisoning defense)\n');
       process.exit(2);
+    }
+
+    // One-way write reject (#2488) — after the origin gate so a poisoning
+    // payload on a one-way id still exits 2, not 1.
+    if (j.preference === 'never-ask' || j.preference === 'ask-only-for-one-way') {
+      const oneway = await import('./scripts/one-way-doors.ts');
+      if (oneway.isOneWayDoor({ question_id: j.question_id })) {
+        process.stderr.write('gstack-question-preference: cannot set ' + j.preference + ' on one-way question \"' + j.question_id + '\" (door_type: one-way)\n');
+        process.exit(1);
+      }
     }
 
     // Optional free_text — sanitize (no injection patterns, no newlines, <=300 chars)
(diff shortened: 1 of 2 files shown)


garrytan:
Absorbed: the fix that landed in v1.68.0.0 (28d59ad5) matches this PR's gate placement and error string (one-way write reject after the origin gate, exit 1; --stats INERT_ONE_WAY reclassification). Credited in the #2488 close — thank you!

PR #2608, opened 2026-08-17
feat(codex): raise reasoning-effort defaults and unify wrapper timeouts at 570s

jerry-sam:
## What

Raises the codex skill's per-mode reasoning-effort defaults to the API's new top tier and unifies all five `_gstack_codex_timeout_wrapper` budgets at 570s, with every dependent number (Bash gates, log labels, stall messages, prose) updated so the skill stays internally consistent.

| | Before | After |
|---|---|---|
| Review (2A) effort | `high` | `max` |
| Challenge (2B) effort | `high` | `max` |
| Consult (2C) effort | `medium` | `high` |
| Review wrapper budget | 330s | 570s |
| Challenge/Consult wrapper budget | 600s | 570s |
| Review Bash gate guidance | `timeout: 360000` | `timeout: 600000` |
| Challenge/Consult Bash gate | `timeout: 660000` | unchanged (already > 570s) |
| Escalation flag | `--xhigh` → forces `xhigh` | `--max` → forces `max` (`--xhigh` kept as legacy alias) |

Edited `codex/SKILL.md.tmpl`; `codex/SKILL.md` is the `bun run gen:skill-docs` regeneration of it (only these two files change).

## Why

**`max` is a real tier now.** Verified against codex-cli 0.144.3: the CLI passes the value through unvalidated, and the API's own invalid-enum error enumerates the supported set —

```
[ReasoningEffortParam] [reasoning.effort] [invalid_enum_value] Invalid value: 'bogus_probe'.
Supported values are: 'none', 'minimal', 'low', 'medium', 'high', 'xhigh', and 'max'.
```

— and a `model_reasoning_effort="max"` run completes normally. The skill's whole premise is codex as the maximally-rigorous second opinion ("200 IQ autistic developer"); its value scales directly with reasoning depth, and the bounded-diff modes (Review/Challenge) are exactly where the extra depth pays for its latency. Consult stays one notch down (`high`) because it's interactive and large-context.

**Timeouts follow effort.** Deeper reasoning runs longer; the Review lanes' 330s budget is the first thing to break under `max`. All five sites unify at 570s — comfortably under the Challenge/Consult 660s Bash gate, with the Review gate guidance raised to 600000ms to preserve the skill's own gate-above-wrapper invariant (a 570s wrapper under the old 360s gate would invert it, and a stall would surface as a silent harness kill instead of the diagnosable exit-124 path the hang-detection blocks exist for).

**Flag coherence.** With `max` as the Review/Challenge default, a user typing `--xhigh` would be *downgrading* those modes — surprising for an escalation flag. Both flags now force `max`, keeping the "crank it" semantic; `--xhigh` remains accepted so habits don't break.

## Consistency sweep + tripwire

Everything that encoded the old numbers moved together: the gate-ordering comment at the Review site, all five `codex_timeout` telemetry labels (`"330"`/`"600"` → `"570"`), the six user-facing stall messages (5.5/10 minutes → 9.5 minutes), **both** per-mode defaults tables (the override block *and* the Model/effort section — including its ~23x-tokens / 50-minute-hang rationale, which is kept and now explains why Consult stays a notch down), the three per-site override sentences, the Error Handling gate table, and the "Bash gate above the wrapper" rule. `test/codex-hardening.test.ts`'s only `330` is a historical comment about a fixed bug, deliberately left as history.

Because the budget and defaults are stated across ~20 surfaces with no single source, this drift class is now pinned: **`test/codex-effort-timeout-consistency.test.ts`** (new, runs in the free tier) asserts the five bash sites, five telemetry labels, six stall messages, both defaults tables, the override sentences, and the Bash-gate prose all agree — and that the escalation flag never forces a tier below a mode's default. Negative-tested: it fails 4 checks against a deliberately half-updated state and passes at HEAD. (An adversarial codex review of an earlier draft of this branch found exactly this drift — bash sites updated, second table stale — which is what motivated the test.)

## Trade-off, stated plainly

This raises per-invocation cost and latency for every gstack user on the affected modes — that's the point of the PR, but it's a defaults change, not a pure fix. If you'd rather not move the defaults for everyone, I'm happy to rework this as a config knob (e.g. `gstack-config get codex_effort`, defaults unchanged) with the timeout/gate consistency fixes kept as-is — say the word and I'll amend.

## Testing

- `bun run test` (Tier 1, full free suite): 6,916/6,918 on this branch. The two failures — `browse/test/stop-dead-daemon.test.ts` and `ios-qa/scripts/gen-accessors.test.ts` — are unrelated to this diff and shard-load flakes on my machine: both pass in isolation at clean `origin/main` *and* on this branch (60/60 both runs).
- New tripwire + `test/codex-hardening.test.ts` + `test/gen-skill-docs.test.ts`: 474/474.
- Runtime probes of codex-cli 0.144.3 confirming enum validity of `max` (above).
- Regeneration verified: `bun run gen:skill-docs` after the tmpl edit produces exactly the committed `codex/SKILL.md`; `git status` clean apart from the committed files.

### DIFF
--- a/codex/SKILL.md
+++ b/codex/SKILL.md
@@ -982,13 +982,13 @@ Parse the user's input to determine which mode to run:
    - Otherwise, ask: "What would you like to ask Codex?"
 4. `/codex <anything else>` — **Consult mode** (Step 2C), where the remaining text is the prompt
 
-**Reasoning effort override:** If the user's input contains `--xhigh` anywhere,
-note it and remove it from the prompt text before passing to Codex. When `--xhigh`
-is present, use `model_reasoning_effort="xhigh"` for all modes regardless of the
-per-mode default below. Otherwise, use the per-mode defaults:
-- Review (2A): `high` — bounded diff input, needs thoroughness
-- Challenge (2B): `high` — adversarial but bounded by diff
-- Consult (2C): `medium` — large context, interactive, needs speed
+**Reasoning effort override:** If the user's input contains `--max` (or its legacy
+alias `--xhigh`) anywhere, note it and remove it from the prompt text before passing
+to Codex. When present, use `model_reasoning_effort="max"` for all modes regardless
+of the per-mode default below. Otherwise, use the per-mode defaults:
+- Review (2A): `max` — bounded diff input, needs thoroughness
+- Challenge (2B): `max` — adversarial but bounded by diff
+- Consult (2C): `high` — large context, interactive; `high` balances depth and latency
 
 ---
 
@@ -1051,15 +1051,15 @@ contradicting this skill's read-only contract (#2496, #2524):
 ```bash
 _REPO_ROOT=$(git rev-parse --show-toplevel) || { echo "ERROR: not in a git repo" >&2; exit 1; }
 cd "$_REPO_ROOT"
-# The 330s wrapper sits BELOW the 360s Bash gate so the wrapper fires FIRST
+# The 570s wrapper sits BELOW the 600s Bash gate so the wrapper fires FIRST
 # and a stall surfaces as a diagnosable exit 124 with an explicit message,
 # never as a silent harness kill that downstream reads as "no findings".
-_gstack_codex_timeout_wrapper 330 codex review --base <base> -c 'sandbox_mode="read-only"' -c 'model_reasoning_effort="high"' -c 'web_search="cached"' < /dev/null 2>"$TMPERR"
+_gstack_codex_timeout_wrapper 570 codex review --base <base> -c 'sandbox_mode="read-only"' -c 'model_reasoning_effort="max"' -c 'web_search="cached"' < /dev/null 2>"$TMPERR"
 _CODEX_EXIT=$?
 if [ "$_CODEX_EXIT" = "124" ]; then
-  _gstack_codex_log_event "codex_timeout" "330"
+  _gstack_codex_log_event "codex_timeout" "570"
   _gstack_codex_log_hang "review" "$(wc -c < "$TMPERR" 2>/dev/null || echo 0)"
-  echo "Codex stalled past 5.5 minutes. Common causes: model API stall, long prompt, network issue. Try re-running. If persistent, split the prompt or check ~/.codex/logs/."
+  echo "Codex stalled past 9.5 minutes. Common causes: model API stall, long prompt, network issue. Try re-running. If persistent, split the prompt or check ~/.codex/logs/."
 elif [ "$_CODEX_EXIT" != "0" ]; then
   # Surface non-zero exits (parse errors, arg-shape breaks, etc.) so the
   # calling agent doesn't read "no output" as a silent model/API stall and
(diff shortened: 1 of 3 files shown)


LimpingWolf:
Datapoint in support of this PR: the 330s review budget is no longer a
prospective problem at `effort="max"` — it is **already breaking today at
`effort="high"`**, after the Codex CLI 0.151.0 upgrade swapped the default
model from `gpt-5.5` to `gpt-5.6-sol`.

**Environment.** Codex CLI 0.151.0, model `gpt-5.6-sol` (default),
`model_reasoning_effort="high"`, `sandbox_mode="read-only"`, gstack v1.77.0.0.
The CLI upgrade landed on my machine 2026-08-31; every session before it ran
0.131.0 with `gpt-5.5` as default.

**A/B, same commit, everything else held constant.** The subject is a small
diff (2 files, ~40 lines), run as `codex review --commit`:

| CLI / default model | runs | wall-clock |
|---|---|---|
| 0.131.0 / `gpt-5.5` | 3 | 175s, 205s, 219s |
| 0.151.0 / `gpt-5.6-sol` | 5 | 324s, 324s, 327s, 380s, 473s |

Two of those five were killed at exit 124 by the 330s wrapper (324s, 327s);
the rest only finished because I had already raised the budget locally. A
larger branch-scope review on the same model/effort took **776s** (4.77M input
tokens).

So the 330s budget wasn't merely tight — it was calibrated against a model
that no longer runs by default. Same prompt, same diff, ~1.8x the wall-clock.

**Two caveats worth folding into the PR, both arguing it is still right to
merge:**

1. **570s does not cover the 776s case.** Unifying the ceiling removes the
   demonstrated small-review regression; it does not make large reviews a
   completion guarantee. That's fine — a timeout is a policy limit — but the
   PR description reads as though 570s settles the question, and it doesn't.
   Branch-scope reviews on big diffs will still hit the wall, just less often.

2. **570 / 600000 is the correct pair, and the margin is thin.** 600000ms is
   the Claude Code Bash tool's maximum permitted timeout, so the outer gate
   cannot go above it. I initially patched my local copy to 600s wrapper /
   660000ms gate; that's invalid — the 660000 clamps down to 600000 and the
   wrapper stops firing first, which is the whole point of the two-layer
   arrangement (the wrapper's explicit exit-124 message vs. a silent harness
   kill). 570/600000 preserves a 30s margin and is what I now run.

**This PR likely needs a rebase.** It edits `codex/SKILL.md` and
`codex/SKILL.md.tmpl`, but v1.71.0.0 (#2691) carved those into
`codex/sections/*`. The review invocation now lives at
`codex/sections/review-mode.md` (line 48 for `codex review`, line 89 for the
`codex exec` fallback — both still 330s, and the `timeout: 360000` gate is at
line 109). The `.md` files are generated from `.md.tmpl`, so both need the
change.

Since raising the budget locally, reviews that were dying at 324s/327s now
complete: 473s and 321s (exit 0) on 2026-08-31, 181s on 2026-09-01. The
570/600000 pair specifically has only been in place a few hours, so treat that
as directional rather than as a soak test.

PR #2610, opened 2026-08-17
fix(redact): a typo'd subcommand no longer exits 0 having done nothing

kinoko-studio:
## The bug

`main()` in `bin/gstack-redact` recognises exactly two subcommands and lets everything else fall through to the stdin scan. On empty stdin that prints `(no findings)` and exits 0:

```console
$ gstack-redact install-prepush-hooks    # plural typo
gstack-redact scan — repo UNKNOWN
  (no findings)
  HIGH=0 MEDIUM=0 LOW=0 WARN=0
$ echo $?
0
$ ls .git/hooks/pre-push
ls: .git/hooks/pre-push: No such file or directory
```

No hook was installed, and the operator has every reason to believe the credential guard is armed. A guard that silently no-ops must never exit 0 — this is the failure mode the guard itself exists to prevent, in the tool that installs it.

## Two smaller faults in the same dispatch

Both of them lead people into the first one.

**No `--help` handler.** `gstack-redact --help` fell through to the scanner, so piping a credential to it scanned the secret and exited 3 rather than printing help.

**No TTY guard.** With nothing piped in and no `--from-file`, `readInput()` blocks on `readSync(fd 0)` waiting for an EOF an interactive terminal never sends. It prints nothing at all, so it reads as a hang rather than as "this is a filter, feed it something". That is how I arrived here: I ran `gstack-redact scan`, watched it sit for three minutes, and assumed the scanner was wedged.

## What changed

- `--help` / `-h` / `help` → usage, exit 0
- unrecognised positional → names the offender on stderr, exit 1
- stdin is a TTY and no `--from-file` → usage instead of blocking
- `scan` stays accepted, because the human output header reads `gstack-redact scan — repo …` and that is what people type. Rejecting it would break that muscle memory for no gain.

Usage errors exit **1**, deliberately not 2 or 3. Those mean MEDIUM and HIGH findings and callers gate dispatch on them (per the exit-code table at the top of the file), so a usage error exiting 2 would be read as "medium findings — prompt the user". There is a test pinning that.

`install-prepush-hook` and `uninstall-prepush-hook` are unchanged, as is the piped filter path.

## Tests

Four tests written failing first, in `test/gstack-redact-cli.test.ts`:

| | before | after |
|---|---|---|
| typo'd subcommand | exit 0, "no findings" | exit 1, `unknown subcommand` |
| unknown positional | clean scan report | no scan, exit 1 |
| `--help` with a credential piped in | scanned it, exit 3 | usage, exit 0 |
| `scan` alias, `--json`, flags | pass | pass (unchanged) |

The TTY guard is not unit-tested — spawned stdin is a Buffer, never a TTY. Verified manually under a pty:

```console
$ script -q /dev/null bun bin/gstack-redact
gstack-redact — scan text for secrets/PII/legal content.
...
```

Suites: `test/gstack-redact-cli.test.ts` 15 pass / 0 fail; the nine redact-related files 133 pass / 0 fail; full `bun run test` **7,722 pass / 0 fail** across 7/7 shards.

## Note on the diff's own scan

`git show <branch> | gstack-redact --repo-visibility public` reports 2 HIGH. Both are `AKIA1234567890ABCDEF`, the fixture this test file already ships at lines 31, 40 and 62 — the added test follows the existing convention rather than introducing a new secret.

### DIFF
--- a/bin/gstack-redact
+++ b/bin/gstack-redact
@@ -199,13 +199,67 @@ function humanTable(findings: Finding[]): string {
   return rows.join("\n");
 }
 
+/**
+ * Usage. Exits 0 when asked for (--help), 1 when the invocation was wrong.
+ *
+ * Deliberately NOT 2 or 3: those mean MEDIUM and HIGH findings, and callers
+ * gate dispatch on them (see the exit-code table at the top). A usage error
+ * that exited 2 would be read as "medium findings — prompt the user".
+ */
+function printUsage(code: number): never {
+  const out = code === 0 ? process.stdout : process.stderr;
+  out.write(
+    "gstack-redact — scan text for secrets/PII/legal content.\n" +
+      "\n" +
+      "Reads the text to scan from STDIN, or from --from-file PATH. It is a\n" +
+      "filter: with nothing piped in it has nothing to scan.\n" +
+      "\n" +
+      "  git diff | gstack-redact --repo-visibility private\n" +
+      "  gstack-redact --from-file notes.md --json\n" +
+      "\n" +
+      "Subcommands:\n" +
+      "  install-prepush-hook     install the managed git pre-push credential guard\n" +
+      "  uninstall-prepush-hook   remove it\n" +
+      "\n" +
+      "Flags: --json --repo-visibility V --from-file PATH --allowlist PATH\n" +
+      "       --self-email EMAIL --repo-public-emails PATH --auto-redact IDS\n" +
+      "       --max-bytes N\n" +
+      "\n" +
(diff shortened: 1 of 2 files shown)


garrytan:
Landed in v1.68.0.0 (28d59ad5) with this PR's reasoning preserved in the code comments: the help/-h/help alias, unknown-subcommand → usage + exit 1, TTY-with-no-input → usage + exit 1, and the exit-code table pinning usage-error = 1 (the 'a guard that no-ops must never exit 0' framing is verbatim). Closing as absorbed — thank you!

PR #2611, opened 2026-08-17
Fix: gstack-config get returns "" with exit 0 for keys that have no default

benjaminberes-bp:
## Why (in your own words)

Four config keys that skill preambles read every run — `question_tuning`,
`team_mode`, `transcript_ingest_mode`, `repo_mode` — are missing from
`lookup_default`'s table, so they hit the `*) echo "" ;;` catch-all and come
back as an empty string with **exit 0**.

That exit code is the whole bug. Every preamble reads config with the same
pattern:

```bash
_QUESTION_TUNING=$(gstack-config get question_tuning 2>/dev/null || echo "false")
```

`|| echo` only fires on a **non-zero** exit. A catch-all that prints `""` and
exits 0 means the fallback never runs and the variable is left empty — so the
written default is unreachable, and the skill branches on `""` instead of on
`false`/`off`/`unknown`. Nothing errors; the setting simply reads as blank
forever. I hit this on `question_tuning`, where a documented default of `false`
never reached the skill that gates on it.

Two changes, and they are deliberately separate:

1. Add the four missing keys to the table, using the values the callers already
   assume in their own `|| echo "<default>"` fallback — so behaviour is
   unchanged for anyone whose fallback happened to work.
2. Make the **unknown-key** path `return 1` instead of printing `""` with exit
   0, so the caller's fallback can fire at all.

Change 2 is scoped strictly to the unknown-key catch-all. The keys whose default
is *intentionally* empty (`cross_project_learnings`, `salience_allowlist`,
`user_slug_at_*`, `redact_repo_visibility`) keep exit 0, because `""` is their
real answer and their callers depend on it — `cross_project_learnings` uses
empty-vs-set to trigger the first-time prompt, and flipping it would re-prompt
every existing user.

## Live evidence

Reproduced against `origin/main` at v1.67.0.0 (`ae8914a`), unmodified, then
against this branch. `gstack-config-before` below is literally
`git show origin/main:bin/gstack-config`.

```
########## BEFORE (origin/main @ v1.67.0.0) ##########
  get question_tuning        -> '' (rc=0)   | pattern '|| echo FALLBACK' -> ''
  get team_mode              -> '' (rc=0)   | pattern '|| echo FALLBACK' -> ''
  get transcript_ingest_mode -> '' (rc=0)   | pattern '|| echo FALLBACK' -> ''
  get repo_mode              -> '' (rc=0)   | pattern '|| echo FALLBACK' -> ''

########## AFTER (this branch) ##########
  get question_tuning        -> 'false'   (rc=0)
  get team_mode              -> 'false'   (rc=0)
  get transcript_ingest_mode -> 'off'     (rc=0)
  get repo_mode              -> 'unknown' (rc=0)
```

Note the BEFORE line that matters most: the `|| echo FALLBACK` column is empty.
The caller's own documented fallback does not fire, which is why this is
invisible rather than noisy.

Non-regression, run with an isolated `GSTACK_HOME` so the user's own config
cannot mask the defaults:

```
########## intentionally-empty defaults -- must stay rc=0 ##########
  get cross_project_learnings -> '' (rc=0)
  get salience_allowlist      -> '' (rc=0)
  get redact_repo_visibility  -> '' (rc=0)

########## unknown key -- must be rc=1 so the fallback arms ##########
  get totally_unknown_key                      -> '' (rc=1)
  get totally_unknown_key || echo "FALLBACK"   -> 'FALLBACK'
```

Test suite added with the fix:

```
$ bun test test/gstack-config-defaults.test.ts
bun test v1.3.14 (0d9b296a)
 4 pass
 0 fail
 11 expect() calls
Ran 4 tests across 1 file. [2.63s]
```

## Scope

- **Changed:** `bin/gstack-config` (+23/-2) — four keys added to
  `lookup_default`, unknown-key catch-all returns 1, and `get` propagates that
  non-zero exit. New `test/gstack-config-defaults.test.ts` (+137).
- **Verified live by:** running the unmodified `origin/main` copy of
  `bin/gstack-config` side by side with the patched one (output above), on
  Windows 11 / git-bash, gstack at v1.67.0.0; plus the new test file, and the
  four-key behaviour through the real `|| echo` pattern the preambles use.
- **Did NOT test:** macOS and Linux — the change is pure POSIX shell with no
  platform-conditional code, but I only have a Windows/git-bash box. I also did
  not audit every caller of `gstack-config get` across all skills for one that
  might *rely* on an unknown key exiting 0; I checked the preamble pattern,
  which is the dominant caller, and the four keys above.

## Liveness proof (required)

✅ Attached in a follow-up comment on this PR: `GSTACK PR 2611` typed live into my
PowerShell prompt. Unedited, cropped only. Say the word if you want a wider frame
showing the typed command and the window chrome, and I will re-shoot it.

To be explicit about how this PR was produced, since the section exists to
establish exactly that: I am a human (Benjamin Bérès), this is my account, and I
directed the work. The investigation, the diff, and the evidence above were done
with Claude Code in my terminal — which is also how the bug surfaced, on my own
Windows box. Close this if the workflow does not meet your bar; I would rather say
how it was made than have you find out.

## Checklist

### DIFF
--- a/bin/gstack-config
+++ b/bin/gstack-config
@@ -161,7 +161,23 @@ lookup_default() {
     brain_trust_policy*) echo "unset" ;;
     salience_allowlist) echo "" ;;
     user_slug_at_*) echo "" ;;
-    *) echo "" ;;
+    # Read by skill preambles but missing from this table, so they fell through
+    # to the catch-all and came back "" with exit 0. Values below are the ones
+    # the callers already assume in their own `|| echo "<default>"` fallback.
+    question_tuning) echo "false" ;;
+    team_mode) echo "false" ;;
+    transcript_ingest_mode) echo "off" ;;
+    repo_mode) echo "unknown" ;;
+    # Unknown key: exit non-zero instead of printing "". The fallback pattern
+    # the preambles use,
+    #   VAR=$(gstack-config get <key> 2>/dev/null || echo "<default>")
+    # only fires on a non-zero exit, so a catch-all echoing "" with exit 0 left
+    # VAR empty and the written default unreachable.
+    # Deliberately *only* the unknown-key path: the keys above whose default is
+    # intentionally empty (cross_project_learnings, salience_allowlist,
+    # user_slug_at_*, redact_repo_visibility) keep exit 0, because "" is their
+    # real answer and their callers rely on it.
+    *) return 1 ;;
   esac
 }
 
@@ -297,7 +313,12 @@ case "${1:-}" in
     fi
     VALUE=$(read_config_value "$KEY" || true)
     if [ -z "$VALUE" ]; then
-      VALUE=$(lookup_default "$KEY")
+      # lookup_default exits non-zero for a key it does not know. Propagate
+      # that, so the caller's `|| echo "<default>"` can fire. A known key whose
+      # default is empty still exits 0 and prints "".
+      if ! VALUE=$(lookup_default "$KEY"); then
+        exit 1
+      fi
     fi
     printf '%s' "$VALUE"
     ;;
--- a/test/gstack-config-defaults.test.ts
+++ b/test/gstack-config-defaults.test.ts
@@ -1,0 +1,137 @@
+/**
+ * gstack-config default-table completeness (gate, free).
+ *
+ * Skill preambles read configuration with
+ *
+ *     VAR=$(gstack-config get <key> 2>/dev/null || echo "<default>")
+ *
+ * and that fallback only fires on a NON-ZERO exit. `get` used to answer a key
+ * it did not know with "" and exit 0, so VAR came back empty and the default
+ * written right there in the preamble was unreachable. The skill then branched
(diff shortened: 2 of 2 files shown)


benjaminberes-bp:
image

benjaminberes-bp:
Full-frame version of the liveness proof: the typed command, the error trace, and the prompt. The first one was cropped too tight.

image

garrytan:
Landed in v1.68.0.0 (28d59ad5): the question_tuning / team_mode / transcript_ingest_mode defaults plus unknown-key non-zero propagation. One deliberate divergence from this diff: repo_mode keeps the empty-string default — the empty answer is load-bearing (gstack-repo-mode treats any non-empty answer as a user override and a synthesized 'unknown' would turn its own classifier into dead code; rationale now lives as a comment at bin/gstack-config:173-176). The diagnosis here drove the fix — thank you!

PR #2622, opened 2026-08-18
fix(setup): quote the install path in the team-mode SessionStart hook command

### LINKED ISSUE #2619: Team mode's auto-update silently never runs if gstack is installed under a path with a space — no error, ever
**Impact:** Team mode's entire value proposition is that everyone on a team stays on the same gstack version automatically. If gstack is installed under any path with a space in it, that auto-update hook silently never registers correctly — and nothing tells you. No error during `setup`, no error at session start, no error anywhere. You believe team mode is on and syncing; it isn't. Over time this produces exactly the failure mode team mode exists to prevent: team members silently drifting onto different gstack versions, each seeing different behavior, with no signal pointing at the cause.

This isn't a rare edge case — spaces in install paths are common in exactly the environments most likely to run gstack in team mode: Windows usernames (`C:\Users\Jane Doe\...`), OneDrive/iCloud-synced folders, and any org-mandated folder naming convention with a space in it.

**Root cause:** `setup` builds the hook command as an unescaped string and writes it straight into `~/.claude/settings.json` (`setup:1859-1863`, `bin/gstack-settings-hook:81-83`). The space breaks the command when Claude Code runs it, and the failure is swallowed.

**To reproduce:** install gstack under a path containing a space, run `./setup --team`, start a new session, and confirm the hook never actually runs.

**Fix:** escape the path before writing it into the hook command. `test/setup-windows-fallback.test.ts:59` currently asserts the unescaped form, which is why this was never caught.

harjothkhara:
## Why (in your own words)

Team mode promises that everyone on a team stays on the same gstack version automatically. But if gstack is installed under a path with a space (Windows `C:\Users\Jane Doe\...`, OneDrive/iCloud folders), `setup --team` writes the SessionStart hook command unquoted into `~/.claude/settings.json`. The shell splits it at the space, the hook exits 127 every session, and nothing ever reports it — `setup` discards the registration's stderr and the hook runner's failure is invisible. The team believes auto-update is on; it silently never runs. This change quotes the path where the command is built, and makes re-running `./setup --team` heal an already-broken registration in place instead of dedupe-skipping it.

Fixes #2619.

## Live evidence

Before (main, `c86e647`) — build `HOOK_CMD` exactly as `setup:1862` does under an install at `.../space dir/gstack`, register via the real `gstack-settings-hook add`, execute the stored string the way the hook runner does:

```
$ bash -c "$STORED_CMD"
bash: /private/tmp/.../scratchpad/space: No such file or directory
exit=127
```

After (this branch) — same install path, same flow:

```
$ bash -c "$STORED_CMD"
exit=0
$ cat hook.log
HOOK_RAN
```

Self-heal — seed a settings.json holding the old unquoted entry, re-run `gstack-settings-hook add` with the corrected command:

```
entries: 1
healed command: "/private/tmp/.../space dir/gstack/bin/gstack-session-update"
```

Tests: `bun test test/team-mode.test.ts test/setup-windows-fallback.test.ts` → 36 pass, 0 fail (3 new regression tests).

## Scope

- **Changed:** `setup` (quote `$SOURCE_GSTACK_DIR` in both `HOOK_CMD` branches), `bin/gstack-settings-hook` (legacy `add` refreshes a matching entry's command in place), plus the test that pinned the unquoted form (`setup-windows-fallback.test.ts`) and 3 new regression tests.
- **Verified live by:** the before/after repro above on macOS — real `gstack-settings-hook`, real stored settings.json, executed via `bash -c` exactly as the hook runner does; plus the self-heal repro.
- **Did NOT test:** a real Windows Git-Bash session (the Windows branch gets the same quoting; the D7 test pins the `bash "…"` form), or a full `./setup --team` run against a live Claude Code session.

## Liveness proof (required)

`GSTACK PR` typed live at my terminal prompt:

GSTACK PR typed live at a terminal prompt

## Checklist

- [x] Linked issue or reproduction: #2619

P.S. — you should hire me. 115+ contributions to open source: https://github.com/harjothkhara

### DIFF
--- a/bin/gstack-settings-hook
+++ b/bin/gstack-settings-hook
@@ -74,10 +74,16 @@ case "$ACTION" in
       try { settings = JSON.parse(fs.readFileSync(settingsPath, "utf8")); } catch {}
       if (!settings.hooks) settings.hooks = {};
       if (!settings.hooks.SessionStart) settings.hooks.SessionStart = [];
-      const exists = settings.hooks.SessionStart.some(entry =>
+      const existing = settings.hooks.SessionStart.find(entry =>
         entry.hooks && entry.hooks.some(h => h.command && h.command.includes("gstack-session-update"))
       );
-      if (!exists) {
+      if (existing) {
+        // Refresh a stale command in place (e.g. an unquoted path written by an
+        // older setup) so re-running setup heals broken registrations.
+        for (const h of existing.hooks) {
+          if (h.command && h.command.includes("gstack-session-update")) h.command = hookCmd;
+        }
+      } else {
         settings.hooks.SessionStart.push({
           hooks: [{ type: "command", command: hookCmd }]
         });
--- a/setup
+++ b/setup
@@ -1857,9 +1857,9 @@ SETTINGS_HOOK="$SOURCE_GSTACK_DIR/bin/gstack-settings-hook"
 # launched directly by the OS — the file-association dialog appears instead.
 # Prefix with 'bash' so Claude Code's hook runner invokes Git Bash explicitly.
 if [ "$IS_WINDOWS" -eq 1 ]; then
-  HOOK_CMD="bash $SOURCE_GSTACK_DIR/bin/gstack-session-update"
+  HOOK_CMD="bash \"$SOURCE_GSTACK_DIR/bin/gstack-session-update\""
 else
-  HOOK_CMD="$SOURCE_GSTACK_DIR/bin/gstack-session-update"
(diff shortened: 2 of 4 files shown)


garrytan:
The root cause was fixed differently in v1.68.1.0 (9da66929): quoting was centralized in gstack-settings-hook's add-event (gsQuoteCmd) rather than at the setup call site — setup now deliberately passes raw paths and the healer round-trips them idempotently, which also covers $, backticks, and embedded quotes. This diff would reintroduce the caller-side quoting the code now warns against, and it targets the legacy add verb while setup calls add-event. Closing with thanks — your diagnosis is credited in the #2619 close.

PR #2630, opened 2026-08-18
fix(evidence): subtract Bun-auto-loaded dotenv keys from the child environment

### LINKED ISSUE #2624: gstack-evidence reports false test results in any repo with a .env.local: Bun auto-loads dotenv and Bun.spawn inherits it
## Summary

`gstack-evidence run` reports **false test failures** in any repo that has a
`.env.local`. It is a Bun script, Bun auto-loads dotenv files from the cwd into
`process.env`, and `Bun.spawn` inherits that polluted environment into the
child test runner. The tool whose entire job is trustworthy verification
silently returns the wrong verdict.

## Mechanism

`bin/gstack-evidence:1` is `#!/usr/bin/env bun`. Bun's runtime loads `.env`,
`.env.local` and `.env.<NODE_ENV>` from the cwd into `process.env` before user
code runs. Then at `bin/gstack-evidence:192`:

```ts
proc = Bun.spawn(spawnArgv, { stdin: "inherit", stdout: "pipe", stderr: "pipe" });
```

No `env` key is passed, so the child inherits `process.env` — already carrying
the repo's secrets and config. For the common single-token form the argv is
`["bash", "-c", cmd]` (`:170`), and it flows straight through to `npm` and the
test runner.

There is no env sanitisation anywhere in the file; the only `process.env` read
is `GSTACK_HOME` at `:95`.

## Observed

On a Next.js repo whose `.env.local` defines `PHOTO_RESERVATION_SECRET` and
`NEXT_PUBLIC_BASE_URL`, two unrelated suites went red under the wrapper and
green everywhere else:

```
gstack-evidence run --label vitest -- 'npm run test:once'   ->  2 files RED
npm run test:once                                           ->  green
CI=true npm run test:once                                   ->  green
```

Both suites assert behaviour that only holds when those vars are **unset** (a
capability degrades to `null`; a URL falls back to `http://localhost`). The
injected production values flipped both. Nothing in the diff was wrong.

Two-line confirmation that the runtime, not the repo, is responsible:

```
$ bun  -e 'console.log(process.env.PHOTO_RESERVATION_SECRET, process.env.NEXT_PUBLIC_BASE_URL)'
<secret> https://app.example.com
$ node -e 'console.log(process.env.PHOTO_RESERVATION_SECRET, process.env.NEXT_PUBLIC_BASE_URL)'
undefined undefined
```

## Why this one matters more than the usual env-leak bug

- It is **silent**. The run completes and records a verdict; it just records the
  wrong one. Both directions are possible — a false red as above, or a false
  green where an injected value masks a real failure.
- It affects **every repo with a `.env.local`**, which is most JS repos.
- It lands specifically on the evidence ledger, which exists to be believed.

`gstack-evidence` is the only script under `bin/` that runs user-supplied repo
commands, so it is the whole blast radius — the other `Bun.spawn` call sites in
`bin/` are fixed-argv tool invocations.

## Suggested fix

Subtract the keys Bun parsed from `.env*` in the cwd before spawning, and pass
an explicit `env` to `Bun.spawn`. That keeps the developer's real environment
(which the child legitimately needs) while removing exactly what the runtime
injected behind the user's back.

Worth noting `test/helpers/hermetic-env.ts` already implements a full allowlist
(`buildHermeticEnv`, `hermeticChildEnv`), but it is drop-everything-else and
lives under `test/`. That is too aggressive here: `gstack-evidence run` executes
arbitrary developer commands that may need `PATH` extensions, credentials, or
tool config. The narrow subtraction is the right shape.

A `--clean-env` opt-out flag would be a reasonable escape hatch for anyone who
genuinely wants dotenv values passed through.

harjothkhara:
## Why (in your own words)

`gstack-evidence` exists to be believed — it's the mechanical arm of /ship's "no completion claims without fresh verification evidence." But in any repo with a `.env.local` (most JS repos), it records the wrong verdict: the script runs under Bun, Bun auto-loads dotenv files from the cwd into `process.env` before user code runs, and the `Bun.spawn` call passed no `env`, so the child test runner inherited values the developer never exported. Suites that assert behavior when those vars are *unset* go falsely red; injected values can equally mask a real failure as a false green. This change subtracts exactly the keys those dotenv files define and passes an explicit `env` — the developer's real environment (PATH, credentials, tool config) still flows through untouched. The helper never throws, per the file's transparency invariant.

Fixes #2624.

## Live evidence

Repro repo containing `.env.local` with `INJECTED_SECRET=oops`; the wrapped command asserts the var is unset.

Before (main, `c86e6472`):

```
$ gstack-evidence run --label repro -- 'test -z "$INJECTED_SECRET" && echo CLEAN || { echo "POLLUTED: $INJECTED_SECRET"; exit 1; }'
POLLUTED: oops
gstack-evidence: recorded label=repro exit=1 log=.../logs/2026-08-18T18-31-53-777Z-repro-....log
wrapper-exit=1
```

After (this branch, same command, same repo):

```
CLEAN
gstack-evidence: recorded label=repro exit=0 log=.../logs/2026-08-18T18-39-13-589Z-repro-....log
wrapper-exit=0
```

Red-green on the new test: with the fix stashed (main's binary), `dotenv keys Bun auto-loaded from the cwd do not reach the child` fails; with the fix, `bun test test/evidence.test.ts` → 26 pass, 0 fail.

## Scope

- **Changed:** `bin/gstack-evidence` — new `childEnv()` helper (parses only the *keys* from `.env`, `.env.<NODE_ENV>`, `.env.local`, `.env.<NODE_ENV>.local` in the cwd, subtracts them from a copy of `process.env`; wrapped in try/catch so it can never turn a run red) and `env: childEnv()` on the `Bun.spawn` call. Plus 2 regression tests.
- **Verified live by:** the before/after repro above on macOS, driving the real `gstack-evidence run` entry path; full `test/evidence.test.ts` suite (26 pass).
- **Did NOT test:** Windows; repos using non-standard dotenv filenames (out of scope — the subtraction targets exactly the set Bun auto-loads). I deliberately did not add a `--clean-env`/opt-out flag to keep the diff minimal — easy follow-up if you want the escape hatch.

## Liveness proof (required)

`GSTACK PR` typed live at my terminal prompt:

GSTACK PR typed live at a terminal prompt

## Checklist

- [x] Linked issue or reproduction: #2624

P.S. — you should hire me. 115+ contributions to open source: https://github.com/harjothkhara

### DIFF
--- a/bin/gstack-evidence
+++ b/bin/gstack-evidence
@@ -34,7 +34,7 @@
  * in the ledger; it cannot prove that an expected lane ever ran.
  */
 
-import { mkdirSync, openSync, writeSync, closeSync, readdirSync, statSync, unlinkSync, chmodSync } from "fs";
+import { mkdirSync, openSync, writeSync, closeSync, readdirSync, readFileSync, statSync, unlinkSync, chmodSync } from "fs";
 import { join, dirname } from "path";
 import { spawnSync } from "child_process";
 import { appendJsonl, readJsonl } from "../lib/jsonl-store";
@@ -67,6 +67,33 @@ function sha256(text: string): string {
   const h = new Bun.CryptoHasher("sha256");
   h.update(text);
   return h.digest("hex");
+}
+
+// Bun auto-loads .env/.env.local/.env.<NODE_ENV> from the cwd into process.env
+// before user code runs, and an env-less Bun.spawn hands that pollution to the
+// child test runner — flipping verdicts in any repo whose suites depend on
+// those vars being unset (both directions: false red AND false green). Subtract
+// exactly the keys those files define; the developer's real environment (PATH,
+// credentials, tool config) passes through untouched. Never throws: per the
+// transparency invariant, bookkeeping must not turn a run red.
+function childEnv(): Record<string, string | undefined> {
+  const env = { ...process.env };
+  try {
+    const mode = process.env.NODE_ENV || "development";
+    for (const f of [".env", `.env.${mode}`, ".env.local", `.env.${mode}.local`]) {
+      let text: string;
+      try {
+        text = readFileSync(f, "utf-8");
+      } catch {
+        continue;
+      }
+      for (const line of text.split("\n")) {
+        const m = /^\s*(?:export\s+)?([A-Za-z_][A-Za-z0-9_]*)\s*=/.exec(line);
+        if (m) delete env[m[1]];
+      }
+    }
+  } catch {}
+  return env;
 }
 
 function git(args: string[]): string | undefined {
@@ -189,7 +216,7 @@ async function cmdRun(argv: string[]): Promise<number> {
   let exitCode: number;
   let proc: ReturnType<typeof Bun.spawn> | undefined;
   try {
-    proc = Bun.spawn(spawnArgv, { stdin: "inherit", stdout: "pipe", stderr: "pipe" });
+    proc = Bun.spawn(spawnArgv, { stdin: "inherit", stdout: "pipe", stderr: "pipe", env: childEnv() });
   } catch (e: any) {
     // Spawn failure (ENOENT on argv-direct form): record exit 127, propagate 127.
     exitCode = 127;
--- a/test/evidence.test.ts
+++ b/test/evidence.test.ts
@@ -76,6 +76,28 @@ describe('gstack-evidence run', () => {
     expect(fs.readFileSync(rec.log_path, 'utf-8')).toContain('ok');
   });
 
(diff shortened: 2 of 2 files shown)


garrytan:
Closing in favor of #2652, which shipped in the v1.69.0.0 wave (#2666): its value-equality guard preserves shell-exported overrides that a key-name match would delete, and its tests assert the scrub warning fired (the .env.local-under-NODE_ENV=test vacuity trap). Your diagnosis and before/after repro were excellent and are what put this bug in the wave — thank you!

PR #1696, opened 2026-05-25
v1.58.0.0 feat: /pr-prep — pre-PR upstream duplicate audit + /ship Step 1.5 gate

BenjaminDSmithy:
## Summary

New `/pr-prep` skill: a **pre-PR upstream duplicate audit**. It walks `git log
base..HEAD`, derives search keywords per commit from the subject + changed file
paths, queries the upstream repo's open issues and PRs via `gh`, and scores each
commit against open work (EXACT_DUP / OVERLAP / SIBLING / CLEAN). EXACT_DUP
refuses to proceed. `/ship` runs it as a **Step 1.5 gate** before any merge, test,
or version bump.

**Why it helps you, the maintainer:** it kills duplicate community PRs before
they're filed — the triage you do most. (Dogfood result below.)

## What it does

- **Step 1.4** — reads the upstream `CONTRIBUTING.md`: pre-push gates, test
  layout, branch/commit conventions, banned patterns. Used inline when writing
  the PR body.
- **Step 4** — title-token + file-overlap Jaccard scoring, state-weighted. The
  bucketing is implemented in `bin/gstack-pr-prep-score` (pure, unit-tested) and
  referenced by the skill as the canonical scorer.
- **Step 4.4** — optional `codex` second opinion on CLEAN commits (different model
  family for independent signal; soft-skips if `codex` is absent).
- **Step 4.5** — per-commit pre-push gate + trigger-path + tests-added annotation.
- **Step 4.6** — commit-message style conformance against the *upstream* repo's
  convention (sampled from its `CONTRIBUTING.md` + `git log`), not the
  contributor's personal house style.
- **/ship Step 1.5** — invokes pr-prep with `GSTACK_FROM_SHIP=1`; EXACT_DUP aborts
  ship with resolution paths; skips cleanly on solo repos and `--skip-pr-prep`.

## Dogfooded on itself

Ran `/pr-prep` on this branch against `garrytan/gstack@main`: **CLEAN — 13/13
commits, no EXACT_DUP, no OVERLAP**, in ~30-60s of `gh` queries. No existing
pr-prep / duplicate-audit skill upstream; the open `ship` PRs all touch unrelated
parts of the flow. Full audit is in a comment on this PR.

## Tests + release

- `test/pr-prep-score.test.ts` — 13 deterministic cases pinning the scorer's
  bucket thresholds (gate-tier, free).
- Registered in `AGENTS.md`, `docs/skills.md`, `scripts/proactive-suggestions.json`,
[...]

PR #2640, opened 2026-08-20
fix(scripts): stop heredoc bodies deadlocking under Homebrew bash

BenjaminDSmithy:
## The bug

`./setup --help` can hang forever on macOS. No output, no error, no way to tell it apart from a slow install. Ten more scripts carry the same latent hang.

bash 5.2+ delivers a heredoc body of 64KiB or less through a pipe. The forked child writes the whole body before `exec`, and nothing reads the other end until the command starts. Under macOS pipe-KVA pressure the kernel hands a fresh pipe a 512-byte buffer instead of the usual 16 to 64KiB, so any body of 512 bytes or more blocks `write()` permanently. The runtime capacity check bash would need to notice this (`F_GETPIPE_SZ`) is Linux-only, so it never fires.

`#!/usr/bin/env bash` resolves to Homebrew bash on a Mac with brew on PATH, which is most of them. Apple's `/bin/bash` 3.2 predates the pipe path and is unaffected, so the bug is invisible to anyone testing with the system shell. It is pressure-dependent, which is why it reads as "worked on my machine": the same script runs fine all day and then wedges.

Reproduction on any bash 5.2+:

```bash
probe() { [ -p /dev/stdin ] && echo PIPE || echo TEMPFILE; }
probe <<EOF
$(printf 'x%.0s' $(seq 1 1000))
EOF
```

On Homebrew bash 5.3.15 that prints `PIPE`. Prefix it with `BASH_COMPAT=50` and it prints `TEMPFILE`.

## The fix

`BASH_COMPAT=50` in each affected script, which restores the pre-5.2 tempfile path.

Deliberately **not** a `#!/bin/bash` shebang swap:

* That pins the script to whatever bash lives at `/bin`, which is 3.2 on macOS and absent on some Linux distributions.
* It is bypassed entirely by `bash script.sh` call sites. The variable survives those.

Other properties:

* Not exported, so child processes keep their own compat level.
* Placed below any `--help` `sed` range that reads `$0`, so usage output is unchanged. Verified on all eleven.
* Every guarded script is bash-3.2-clean (no associative arrays, case conversion, or `mapfile`), so compat level 50 costs them nothing.

## Scope

Eleven scripts, one guard each, no behaviour change:

`setup`, `bin/gstack-brain-sync`, `bin/gstack-artifacts-init`, `bin/gstack-jsonl-merge`, `bin/gstack-team-init`, `bin/gstack-distill-free-text`, `bin/gstack-brain-restore`, `bin/gstack-settings-hook`, `gstack-upgrade/migrations/v1.27.0.0.sh`, `gstack-upgrade/migrations/v1.37.0.0.sh`, `scripts/build-app.sh`

## Test

`test/heredoc-pipe-deadlock.test.ts` does two things:

1. Scans every tracked shell script for a heredoc body in the 512B to 64KiB window and fails without the guard, so a new one cannot land unguarded.
2. Proves the mechanism at runtime on bash 5.2+ by asserting the body moves from `PIPE` to `TEMPFILE`. Skipped on older bash, where the pipe path does not exist.

`bun run test:free` is green except three failures that reproduce identically on a pristine `main` checkout here: the `path-validation` symlink case, `touchfiles` `detectBaseBranch`, and the `codex-hardening` watchdog (flaky roughly 2 runs in 3 on stock main).

## Related

Split out of #1696 so each piece can be reviewed on its own merits. This one is a pure bug fix and stands alone.

### DIFF
--- a/bin/gstack-artifacts-init
+++ b/bin/gstack-artifacts-init
@@ -34,6 +34,16 @@
 #   GSTACK_HOME — override ~/.gstack
 #   USER — fallback for repo naming if $USER is unset
 
+# Heredoc delivery guard. bash 5.2+ writes a heredoc body <=64KiB through a
+# pipe in the forked child before exec, with no reader on the other end. On
+# macOS under pipe-KVA pressure a fresh pipe gets a 512-byte buffer, so any
+# body >=512B blocks write() forever and the script hangs at startup with no
+# output. Compat level 50 restores the tempfile path. These scripts are
+# bash-3.2-clean, so the compat level costs them nothing. Not exported: the
+# guard is per-script, and it survives `bash script.sh` call sites that
+# bypass the shebang.
+BASH_COMPAT=50
+
 set -euo pipefail
 
 GSTACK_HOME="${GSTACK_HOME:-$HOME/.gstack}"
--- a/bin/gstack-brain-restore
+++ b/bin/gstack-brain-restore
@@ -25,6 +25,16 @@
 # Env:
 #   GSTACK_HOME — override ~/.gstack
 
+# Heredoc delivery guard. bash 5.2+ writes a heredoc body <=64KiB through a
+# pipe in the forked child before exec, with no reader on the other end. On
+# macOS under pipe-KVA pressure a fresh pipe gets a 512-byte buffer, so any
+# body >=512B blocks write() forever and the script hangs at startup with no
+# output. Compat level 50 restores the tempfile path. These scripts are
+# bash-3.2-clean, so the compat level costs them nothing. Not exported: the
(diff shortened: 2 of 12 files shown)


garrytan:
Absorbed into the v1.69.0.0 wave (#2666) with your authorship preserved and CHANGELOG credit. Independent review verified the mechanism (bug-bash thread + an unrelated project hitting the same class) and audited all 11 guarded scripts for pre-bash-5 syntax. Wave adaptation: the pipe-probe test skips on minimal-/dev sandboxes; the bash-4.3/4.4 BASH_COMPAT warning caveat is documented in the commit body. Thank you!

Issue #2635, opened 2026-08-19
install-prepush-hook: fs.mkdirSync(recursive) throws EEXIST on Windows (bun) when .git/hooks exists


danielkorr:
## Summary

`gstack-redact install-prepush-hook` throws `EEXIST` on Windows (bun runtime) when the target `.git/hooks` directory already exists, and the hook is never written.

## Environment

- gstack: **1.67.2.0**
- bun: **1.3.12** (Windows x64)
- OS: **Windows 11** (`MINGW64_NT-10.0-26200`), running under Git Bash
- Global git install (`~/.claude/skills/gstack`)

## What happened

Enabled the hook globally (`gstack-config set redact_prepush_hook true`) and installed it into several existing repos. On one repo the install crashed:

```
function installPrepushHook(): void {
  const dir = hooksPath();
  fs.mkdirSync(dir, { recursive: true });   // <-- throws here
          ^
EEXIST: file already exists, mkdir '.git\hooks'
    path: ".git\hooks",
 syscall: "mkdir",
   errno: -17,
    code: "EEXIST"
      at installPrepushHook (…/bin/gstack-redact:58:6)
```

The hook file was not created, so that repo was left unprotected until I copied the managed hook in by hand.

## Root cause

`bin/gstack-redact` → `installPrepushHook()`:

```ts
const dir = hooksPath();
fs.mkdirSync(dir, { recursive: true });
```

Per Node semantics, `mkdirSync(dir, { recursive: true })` is a no-op when `dir` already exists. Under **bun on Windows** it instead throws `EEXIST`. Since almost every real repo already has a `.git/hooks` directory, the install is fragile on Windows. (In my run it failed on 1 of 4 repos — the other three, which also already had `.git/hooks`, happened to succeed, so the behavior looks nondeterministic under bun/Windows.)

## Suggested fix

Make the mkdir defensive so an existing directory is never fatal:

```ts
// Option A — ignore EEXIST
try {
  fs.mkdirSync(dir, { recursive: true });
} catch (e: any) {
  if (e?.code !== "EEXIST") throw e;
}

// Option B — guard on existence
if (!fs.existsSync(dir)) fs.mkdirSync(dir, { recursive: true });
```

The same pattern likely applies anywhere else the codebase relies on `recursive: true` being a no-op on Windows.

## Workaround (for other Windows users hitting this)

Copy the managed hook from a working repo:

```bash
[...]

Issue #2048, opened 2026-06-18
browse: EEXIST on Windows when .gstack exists — mkdirSecure rethrows EEXIST from recursive mkdir (Bun)


JeyRat:
## Summary
On **Windows (Bun-compiled `browse`)**, every `browse` command aborts with:

```
[browse] EEXIST: file already exists, mkdir 'C:\...\<repo>\.gstack'
```

once a `<repo>/.gstack/` directory already exists and contains a **stale daemon record** (a `browse.json` / `terminal-*` set whose recorded daemon PID is dead). The daemon-relaunch path then can't recreate the state dir and the whole CLI is dead until the user manually clears `.gstack`.

## Environment
- gstack **v1.58.1.0**
- Windows 11, prebuilt `browse/dist/browse.exe` (Bun-compiled)
- Invoked from inside a git repo that has a pre-existing `<repo>/.gstack/`

## Root cause
`mkdirSecure()` in `browse/src/file-permissions.ts`:

```ts
export function mkdirSecure(dirPath: string): void {
  fs.mkdirSync(dirPath, { recursive: true, mode: 0o700 });
  restrictDirectoryPermissions(dirPath);
}
```

On **Bun + Windows**, a *recursive* `mkdirSync` throws `EEXIST` when the directory already exists (Node returns silently — recursive mkdir is supposed to be idempotent). `ensureStateDir()` in `browse/src/config.ts` only special-cases `EACCES`/`ENOTDIR` and **re-throws everything else**, so the `EEXIST` becomes fatal. The function's own docstring even says *"Safe to call on an existing directory"*, so `EEXIST` should be a no-op.

## Reproduce
1. On Windows, run any `browse` command from a repo dir (creates `<repo>/.gstack/` + a live daemon record).
2. Let the daemon die uncleanly (machine sleep / crash / kill) so the record goes stale and no live daemon remains.
3. Run any `browse` command again from that dir → `EEXIST: mkdir '...\.gstack'`, repeats on every invocation.

Running from a fresh dir (no `.gstack` yet) works, which is the tell.

## Fix
Treat `EEXIST` as success in `mkdirSecure` (matches its idempotent contract; covers all call sites — `ensureStateDir`, auth dir, skill staging):

```ts
export function mkdirSecure(dirPath: string): void {
  try {
    fs.mkdirSync(dirPath, { recursive: true, mode: 0o700 });
  } catch (err: any) {
[...]

PR #2641, opened 2026-08-20
fix(redact): tolerate EEXIST from recursive mkdir in install-prepush-hook on bun/Windows

Lockyer228:
**Linked issue**: #2635

## Why (in your own words)

On Windows, bun's `fs.mkdirSync(dir, { recursive: true })` throws `EEXIST` when the directory already exists - Node treats the same call as a no-op. Since virtually every real repo already has a `.git/hooks` directory, `gstack-redact install-prepush-hook` crashes on Windows before writing the hook, leaving the repo without the credential-scan guard while the operator believes it is armed (#2635). This change makes that one mkdir defensive: `EEXIST` is tolerated only when the path is confirmed to be an existing directory; a regular file occupying the path, a stat failure, or any other errno still throws, so real collisions still fail loudly.

## Live evidence

`bun --preload` fixture (committed as `test/helpers/emulate-bun-windows-eexist.ts`) emulates bun-on-Windows `mkdirSync` semantics - recursive mkdir on an existing directory throws EEXIST - so the exact crash path runs on any platform. Transparent emulation; no other fs behavior is altered.

Before (baseline `main` @ 9da66929, in a repo with pre-existing `.git/hooks` + a non-managed `pre-push`):

```
$ bun --preload test/helpers/emulate-bun-windows-eexist.ts bin/gstack-redact install-prepush-hook
EEXIST: file already exists, mkdir '.git/hooks'
 code: "EEXIST"
      at installPrepushHook (bin/gstack-redact:58:6)
exit=1            # hook never written; repo left unguarded
```

After (this fix, same repo setup):

```
$ bun --preload test/helpers/emulate-bun-windows-eexist.ts bin/gstack-redact install-prepush-hook
gstack-redact: preserved existing hook as pre-push.local (chained).
gstack-redact: installed pre-push hook at .git/hooks/pre-push
exit=0            # managed hook written; original preserved as pre-push.local
```

Committed regression test (`test/fs-utils.test.ts`) reproduces the above at test level and fails on baseline / passes on the fix; unit tests additionally pin that a regular file occupying the mkdir path still throws `EEXIST`.

## Scope

- **Changed:** `bin/gstack-redact` (`installPrepushHook` uses the new helper), `lib/fs-utils.ts` (new `mkdirpSync`), `test/fs-utils.test.ts` + `test/helpers/emulate-bun-windows-eexist.ts` (new tests).
- **Verified live by:** the before/after CLI run above; `bun test test/fs-utils.test.ts` (4/4); `bun test test/redact-prepush-hook.test.ts test/gstack-redact-cli.test.ts` (41/41, install/uninstall/chaining unchanged); full free suite `bun run test` - no new failures (5 pre-existing `browse/test/*` failures reproduce on clean `main`, browser-launch environmental).
- **Did NOT test:** a real Windows machine (none available). The Windows crash path is exercised through the transparent `--preload` emulation above; the issue reporter's stack trace matches it frame-for-frame.

## Checklist

- [x] Not a generated-file-only diff(pure source + tests)
- [x] No ETHOS.md edits, no voice / founder perspective / YC changes
- [x] No new public command / external service / host adapter(N/A)
- [x] Linked issue: #2635

## Notes

- Deliberately scoped to the one call site from #2635. The same bare `mkdirSync(recursive)` pattern exists elsewhere (e.g. #2048's `browse`/`mkdirSecure`, and open PR #2329 adds a `mkdirpSync` under `scripts/` for the doc generators) - if you'd rather unify, `lib/fs-utils.ts` is a natural home and I'm happy to extend it in a follow-up.

### DIFF
--- a/bin/gstack-redact
+++ b/bin/gstack-redact
@@ -37,6 +37,7 @@ import {
   type ScanOptions,
   type Finding,
 } from "../lib/redact-engine";
+import { mkdirpSync } from "../lib/fs-utils";
 
 const MAX_STDIN_BYTES = 16 * 1024 * 1024; // hard ceiling before the engine cap
 
@@ -55,7 +56,9 @@ function hooksPath(): string {
 
 function installPrepushHook(): void {
   const dir = hooksPath();
-  fs.mkdirSync(dir, { recursive: true });
+  // mkdirpSync, not bare mkdirSync: bun on Windows throws EEXIST from a
+  // recursive mkdir when .git/hooks already exists (#2635).
+  mkdirpSync(dir);
   const hookPath = path.join(dir, "pre-push");
   const prepushBin = path.join(import.meta.dir, "gstack-redact-prepush");
 
--- a/lib/fs-utils.ts
+++ b/lib/fs-utils.ts
@@ -1,0 +1,27 @@
+import { mkdirSync, statSync } from "fs";
+
+/**
+ * mkdir -p that tolerates the target directory already existing.
+ *
+ * Node's mkdirSync(dir, { recursive: true }) is a no-op when dir already
+ * exists, but bun on Windows throws EEXIST in the same situation (#2635),
+ * which crashed `gstack-redact install-prepush-hook` on any repo whose
+ * .git/hooks already existed. Swallow EEXIST only when statSync confirms the
+ * path is an existing directory; anything else - a regular file occupying the
+ * path, a stat failure, a different errno - rethrows the original error, so a
+ * real collision still fails loudly.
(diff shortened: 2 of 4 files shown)


garrytan:
Absorbed into the v1.69.0.0 wave (#2666) with your authorship preserved and CHANGELOG credit. The wave also swept your mkdirpSync to the other Windows-reachable sites (decision-log, evidence, prepush skip-log) with a wiring-level regression test on your preload fixture. Fixes #2635 ships with the wave — thank you!

PR #2647, opened 2026-08-20
fix: use Claude CLI auth status for benchmark availability

rocke2020:
## Why

`ClaudeAdapter.available()` guesses where credentials are stored, so a working OAuth login can be reported as unavailable. This asks Claude Code for its own auth status, while keeping the API-key and legacy fallbacks for older CLIs.

This differs from #1894: it is cross-platform, and an explicit `loggedIn: false` wins over stale credential files or Keychain entries.

## Live evidence

```text
claude auth status --json (private fields omitted) -> {"exit_status":0,"loggedIn":true}
ClaudeAdapter.available() with ANTHROPIC_API_KEY unset -> {"ok":true}
```

The regression test failed before the change on both the OAuth and stale-credentials cases. Final focused run: 37 passed, 0 failed. `bun run build` also passes.

## Scope

- Changed: Claude benchmark adapter auth detection and its hermetic cross-platform tests.
- Did not test: logging the real CLI out; both nonzero-exit true/false JSON paths are covered with the fake CLI.
- Full free suite is not green locally; the provider-adjacent Sourcebot failure reproduces unchanged on clean `upstream/main`, and the remaining failures are outside this adapter.

## Liveness proof
image

## Checklist

- [x] Liveness screenshot attached: `GSTACK PR` typed live into a real surface
- [x] This is not a generated-file-only diff
- [x] No ETHOS.md or voice/founder/YC changes
- [x] New public command/external service/host adapter issue: N/A
- [x] Related reproduction: #1890; prior partial PR: #1894


rocke2020:
This PR fixes a bug in gstack’s cross-model benchmark, which runs the same prompt through Claude, Codex, and Gemini. Acutallly, i use benchmark-models for claude code cli and want trigger it in my codex. This PR fixes incorrect Claude availability detection in benchmark-models.

  Before:

  - gstack guessed authentication from ANTHROPIC_API_KEY, a credentials file, or macOS Keychain.
  - Claude CLI could be genuinely logged in via OAuth and work normally, yet gstack reported Claude as “NOT READY” and
    excluded it from benchmarks.

  - Conversely, a stale credentials file could incorrectly make Claude appear available.

  After:

  - gstack asks Claude CLI directly using claude auth status --json.
  - loggedIn: true/false is authoritative.
  - API-key and legacy checks remain as fallbacks for older CLI versions.
  - It handles nonzero exit codes and informational output around the JSON.

  It only fixes the benchmark provider adapter—not the actual model routing. See PR #2647
  (https://github.com/garrytan/gstack/pull/2647).

PR #2652, opened 2026-08-20
fix(evidence): stop bun's dotenv autoload from reaching the spawned command

namtrok:
`bin/gstack-evidence` has a `#!/usr/bin/env bun` shebang, and **bun auto-loads `.env`, `.env.<NODE_ENV>` and `.env.local` from the cwd into `process.env`**. The wrapper spawned the command with no `env` override, so everything run through `gstack-evidence run` inherited those variables — and a repo `.env.local` routinely holds production credentials.

Reproduce in three lines:

```bash
mkdir t && cd t
printf 'SECRET_PROBE=leaked\n' > .env.local
printf 'Bun.spawnSync(["bash","-c","echo child SECRET_PROBE=$SECRET_PROBE"],{stdout:"inherit"})\n' > c.ts
bun c.ts        # -> child SECRET_PROBE=leaked
```

## Two problems, and the second is worse than the leak

1. **Secrets reach a child that would not otherwise have them.** `npm test` run by hand in the same shell sees none of them; the same command through the wrapper sees all of them.
2. **The command under test behaves differently, so the ledger certifies a run that is not the run CI performs.** Observed in a Next.js repo on 2026-08-20: four tests failed **4/4** through the wrapper and passed **5/5** without it, because app code branched on env vars only the wrapper supplied. It reads exactly like a flake, and nearly an hour went into chasing one — the measuring instrument was the bug. `gstack-evidence` exists to record trustworthy evidence, so silently altering the environment defeats its purpose.

## The fix, and why the detection is exact rather than heuristic

The child env is now `process.env` minus the keys bun injected. Verified on **bun 1.3.11**: a dotenv file does **not** override a variable the shell already exported — the shell's value wins.

```
$ SECRET_PROBE=from_shell bun p.ts
SECRET_PROBE=from_shell          # .env.local said "leaked"
```

So a key whose live value equals the dotenv file's value was injected by bun, and dropping it restores the environment the user's own shell would have given the command. A key whose live value **differs** is genuinely the caller's and survives untouched.

`BUN_DOTENV_FILES()` mirrors bun's precedence, including that `.env.local` is skipped when `NODE_ENV` is `test` (measured: `NODE_ENV=test` loads `.env` but not `.env.local`). Scrubbing a key bun never loaded would strip a variable the caller legitimately provided.

- **Escape hatch:** `GSTACK_EVIDENCE_KEEP_DOTENV=1` restores the old pass-through behaviour.
- When keys are scrubbed the wrapper warns with the **key names only**, so the diagnostic cannot become the leak it prevents.
- A key that fails to parse (multi-line values) is left in the child env — the safe direction.

## Tests

Six cases in `test/evidence.test.ts`, **mutation-verified**: removing `env: spawnEnv` reddens exactly the two leak tests, restoring it gives 30/30. The three control tests correctly stay green under that mutation, since they guard against over-reach rather than testing the scrub.

⚠️ Worth knowing for anyone extending these: `bun test` sets `NODE_ENV=test`, so **the first version of these tests passed vacuously** against a `.env.local` bun had never loaded. Every leak test now asserts the scrub warning fired, so a fixture bun ignores fails instead of passing silently. One test pins the `NODE_ENV=test` skip behaviour deliberately.

Verified so far: `bun test test/evidence.test.ts` → **30 pass / 0 fail**, plus the mutation run above. The full suite is running as I open this; I'll confirm in a comment rather than claim it here.

### DIFF
--- a/bin/gstack-evidence
+++ b/bin/gstack-evidence
@@ -34,7 +34,7 @@
  * in the ledger; it cannot prove that an expected lane ever ran.
  */
 
-import { mkdirSync, openSync, writeSync, closeSync, readdirSync, statSync, unlinkSync, chmodSync } from "fs";
+import { mkdirSync, openSync, writeSync, closeSync, readdirSync, statSync, unlinkSync, chmodSync, readFileSync } from "fs";
 import { join, dirname } from "path";
 import { spawnSync } from "child_process";
 import { appendJsonl, readJsonl } from "../lib/jsonl-store";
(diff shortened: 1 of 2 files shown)


garrytan:
Absorbed into the v1.69.0.0 wave (#2666) with your authorship preserved and CHANGELOG credit. Wave additions on top of your diff: a doc comment on the ${VAR}-expansion limitation and a regression pin for the unreadable-.env fail-open path. Your mutation-verified tests and the value-equality guard were exactly right — thank you!

PR #2668, opened 2026-08-22
v1.71.0.0 feat: --reconcile backfills office-hours tenure from timeline.jsonl (#2657)

### LINKED ISSUE #2657: developer-profile undercounts office-hours tenure: Phase-4.5-only logging + no timeline.jsonl reconciliation (4 logged vs 14 real runs)
`/office-hours` greets a user who has run gstack daily since April with "this is your third session." The relationship-tier feature understates the relationship for exactly the long-tenure users it exists to reward. Found on v1.67.2.0, still present in v1.68.2.0.

## Real-user numbers (my machine)

| Source | Count |
|---|---|
| `~/.gstack/developer-profile.json` `sessions[]` | 4 (3 startup + 1 resources) |
| `office-hours` events in `~/.gstack/projects/*/timeline.jsonl` | 14 runs across 5 project slugs |
| `office-hours` lines in `~/.gstack/analytics/skill-usage.jsonl` | 20 |
| Total skill runs logged since 2026-04-30 | 269 |

`gstack-builder-profile` reports `SESSION_COUNT: 3`, `TIER: welcome_back`. The real history says this user should be `inner_circle`.

## Why

`sessions[]` has exactly one writer: the `--log-session` call in office-hours **Phase 4.5**, which runs only after the diagnostic conversation, premise agreement, and approach selection. Two structural losses follow:

1. **Early-exit sessions never count.** Any office-hours run that ends before Phase 4.5 (user interrupt, escape hatch, context death, host crash) is invisible to the tier system, even though the timeline logged its `started` event.
2. **No reconciliation with history that predates the feature.** `do_migrate()` in `bin/gstack-developer-profile` migrates the legacy `builder-profile.jsonl` (the #1677 writer/reader split), but nothing ever backfills from `timeline.jsonl`, where the complete per-run history (`started`/`completed` events per skill, per project) already exists. Sessions from before the `--log-session` era, and sessions lost to the #1677 window, are permanently uncounted.

#1677 closed its stranded-entries question with "loss is aesthetic." With 4 months of usage the loss compounds into the greeting being factually wrong by ~3x, and `welcome_back`/`regular`/`inner_circle` copy ("This is session N", accumulated-signal reflections, the session-5 journey summary) asserts the wrong history as fact.

Related: #1677 (writer/reader split, the original stranding), #2067 (resources rows double-counted, closed).

## Proposed fix (two independent parts)

**(a) Log a stub at Phase 1, upgrade it at Phase 4.5.** Office-hours appends a minimal session entry (`{date, mode: "started", project_slug}`) right after mode selection, and Phase 4.5 upgrades that entry in place with signals/design_doc/assignment. Early exits then still count toward tenure; signal aggregates stay truthful because stubs carry `signal_count: 0`.

**(b) Add `--reconcile` (or extend `--migrate`) to backfill count-only entries from `timeline.jsonl`.** For each `{"skill":"office-hours","event":"completed"}` timeline event with no matching `sessions[]` entry within the same hour, append `{date, mode, project_slug, signal_count: 0, signals: [], backfilled: true}`. Never fabricate signals; `backfilled: true` lets the reader exclude these from signal reflections while counting them for tenure. Idempotent, and it also heals the #1677 stranded window retroactively.

Happy to send a PR for either or both; (b) is self-contained in `bin/gstack-developer-profile`, (a) touches the office-hours SKILL template. Flagging first in case you would rather define tenure as "completed design-doc sessions only" on purpose; if so, the greeting copy should stop claiming "this is session N".

briacSck:
Closes #2657 (part b — the `--reconcile` backfill you accepted; part a, the Phase-1 stub logging, is untouched and stays yours to spec).

## Summary

**`--reconcile` (new subcommand, `bin/gstack-developer-profile`)** — walks `~/.gstack/projects/*/timeline.jsonl` for office-hours `completed` events and appends count-only rows (`{date, mode, project_slug, signal_count: 0, signals: [], backfilled: true}`) for runs with no `sessions[]` entry in the same hour. Idempotent; never fabricates signals. On the reporting machine from the issue: 3 logged sessions → 9 (`welcome_back` → `inner_circle`), second run backfills 0, every reflection field byte-identical.

**Reader (`--read`)** — backfilled rows count for `SESSION_COUNT`/`TIER`, and are excluded from everything the greeting asserts as fact: `LAST_PROJECT`, `LAST_ASSIGNMENT`, `DESIGN_*`, `CROSS_PROJECT`, and the builder nudge. Signals were already safe (`signals_accumulated` is an aggregate the backfill never touches).

**Greeting guard (`office-hours/sections/design-and-handoff.md` + `.tmpl`, kept in sync)** — a backfilled-only profile now has tier > introduction with empty `LAST_*`; the welcome_back/regular templates would have interpolated a blank assignment. The guard acknowledges tenure without asserting content.

### Judgment calls on your spec sketch, flagged for review

- **"Same hour" = ±60-minute proximity, not clock-hour buckets** — a Phase 4.5 write at 15:50 matches a `completed` event at 16:05 across the :00 boundary.
- **Long sessions:** Phase 4.5 logs mid-run; `completed` fires at skill end, sometimes hours later. When the event carries a sane `duration_s` (< 8h), the match window extends back across the run's span so a 2.5-hour session doesn't double-count. Garbage durations (real timelines carry year-length values) fall back to the plain window.
- **A resumed run's two completions collapse into one session**; back-to-back runs on *different* projects each count (run coverage is slug-scoped, and the scoping is persisted on the row so late-arriving synced timelines behave the same as a single pass).
- **Phase 6 `resources` rows never cover the hour** — the failed-Phase-4.5/successful-resources case is exactly the partial-run gap this command closes. Symmetrically, a backfilled row never takes `mode: 'resources'` (it would neither count nor cover its own run, breaking idempotency).
- **A no-op reconcile leaves the file untouched** — no mtime churn, no resort, no brain-enqueue.
- **Accepted + pinned by test:** legacy date-only entries can only match at midnight-UTC granularity, so a same-day run double-counts; the data can't support tighter matching. (Production `--log-session` has always written full ISO timestamps, so real exposure should be rare.)
- **No invocation site, by design** — this ships part (b) as a manual command, matching your comment. Wiring it into the office-hours preamble (`--reconcile 2>/dev/null || true` before the profile read — cheap and no-op-safe) can ride with part (a).

### Known residuals (not addressed here, flagged honestly)

- The unlocked read-modify-write race on `developer-profile.json` is shared with `--log-session`/`--derive`; `--reconcile` widens the window slightly. Matching the file's existing concurrency model felt right for a first PR — happy to add a lock shared by all writers if you want it.
- `do_migrate`/`do_log_session`/`do_reconcile` now carry three copies of the mktemp/trap/mv/enqueue ritual; extracting a shared writer helper is a mechanical follow-up I left out to keep this diff reviewable.
- Root-resolution note documented in the header: this binary honors `GSTACK_STATE_ROOT` before `GSTACK_HOME` (D16) while `gstack-timeline-log` writes under `GSTACK_HOME` only — timeline discovery assumes they point at the same root.

## Test Coverage

24 new tests in `test/gstack-developer-profile.test.ts` (42 → 66 in the file, all green). The mid-review coverage audit scored 18/26 paths (69%); every gap it named was then closed: malformed/no-ts/unparseable-ts lines, missing-timeline dirs, a directory named `timeline.jsonl`, the 60/61-minute boundary, mode passthrough and the `'unknown'` fallback, legacy date-only matching (both the covered and the accepted-double-count case), chronological-resort assertions, `NUDGE_ELIGIBLE` isolation, and the backfilled-only `--read` contract. Also hardened: an ambient `GSTACK_STATE_ROOT` no longer points the whole test file at the developer's real state dir.

## Pre-Landing Review

Ran the full gstack review army (4 specialists + always-on Claude adversarial + red team, `/ship` pipeline). Everything actionable was fixed in this diff before the PR:

- **Adversarial:** long-session double-count (→ duration-aware window), cross-project collapse (→ slug-scoped coverage), unconditional resort with epoch-0 dates (→ write gated on `added > 0`, NaN dates sort to tail), resources rows blocking backfill (→ tenure-row coverage), tier-with-empty-`LAST_*` greeting break (→ template guard), one unreadable timeline aborting the run (→ per-file tolerance).
- **Red team (on the fixed diff):** the `mode:'resources'` idempotency interaction above, cross-invocation slug parity, the duration cap tightened 24h → 8h.
- **Testing/maintainability specialists:** all findings addressed or noted in residuals above.

## Design Review

No frontend files changed — design review skipped.

## Eval Results

Not run locally: the dev machine is Windows-on-ARM (the bash-spawning suites run in WSL; `test:evals` needs a full `claude -p` rig). The prompt-touching surface is one guard paragraph in `design-and-handoff.md(.tmpl)` — flagged here so the eval gate on your side gets a look. The tmpl/rendered pair was verified drift-free by regenerating through `gen-skill-docs` and byte-comparing.

## Plan Completion

No plan file — this PR implements the spec sketch from your comment on #2657.

## TODOS

No TODO items completed in this PR (the #2657 residual you filed for v1.69.0.0 isn't on pushed main yet).

## Documentation

Post-ship doc sync (`/document-release`, commit `bc98815d`), driven by a cross-model doc review of the shipped diff:

- **office-hours/sections/design-and-handoff.md + .md.tmpl** — moved the backfilled-history greeting guard above the Step 2 tier dispatch: it said "applies to every tier below" but sat inside the welcome_back section, and Step 2 instructs "Follow ONE tier path... Do not mix tiers", so regular/inner_circle readers never saw it and would render an empty `[LAST_ASSIGNMENT]` or "You've iterated 0 designs" (the CHANGELOG's own hero case). The guard now also covers DESIGN_TITLES/DESIGN_COUNT and ACCUMULATED_SIGNALS. Blocks byte-identical across .md/.md.tmpl; `gen:skill-docs` round-trip clean.
- **office-hours/SKILL.md + .md.tmpl** — the Phase 4.5 `sessions[]` writer doc now names the second writer: `--reconcile` backfilled rows (`backfilled: true`) count toward SESSION_COUNT/TIER but are excluded from every reflection (LAST_*, DESIGN_*, CROSS_PROJECT, nudge).
- **CHANGELOG.md** (1.70.0.0 entry, two precision fixes, entry otherwise untouched) — entry shape now includes `signals: []` (the code writes it); dedupe table row now says "backfilled rows' coverage is slug-scoped" (logged-session coverage is global, per the implementation comment).
- **Verified, no change needed** — "24 new regression tests" matches the diff exactly; VERSION `1.70.0.0` = CHANGELOG heading = package.json `1.70.0`; `--reconcile` is documented in the bin header and surfaced via `--help`; README/CLAUDE.md/AGENTS.md don't document this surface.

### Documentation Debt
- ⚠️ **No automatic `--reconcile` caller** — only `--help` text and the dispatch table reference it; the #2657 reporter still sees `welcome_back` until they run `gstack-developer-profile --reconcile` by hand. Decide: wire it into the read path (like the builder-profile shim's auto `--migrate`) or keep manual-first and let it ride with part (a). (Deliberately not decided by this PR.)
- ⚠️ `gstack-developer-profile` has no reference entry in README.md (pre-existing, not a regression).

## Test plan

- [x] Full free suite (`bun run scripts/test-free-shards.ts`) — `[test:free] PASS` on the exact final tree (native-bun WSL rig; the suite is POSIX-only by repo convention)
- [x] `test/gstack-developer-profile.test.ts` — 66 pass / 0 fail (42 pre-existing + 24 new)
- [x] Functional smoke on fixtures: duration matching, cross-project, resources exclusion, dir-named timeline, idempotency
- [x] Real-data dry-run on the issue's reporting machine: 6 backfilled from 9 runs, `welcome_back` → `inner_circle`, reflections unchanged, re-run backfills 0

Note on the version: 1.69.0.0 is claimed by #2666, so the queue advanced this to v1.70.0.0.


briacSck:
Rebased onto v1.69.0.0. The conflicts were `VERSION` / `package.json` / `CHANGELOG.md` only — #2666 touched none of the files this PR touches, so the code you accepted is unchanged: `git diff` over `bin/gstack-developer-profile`, `test/gstack-developer-profile.test.ts` and the four office-hours files against the pre-rebase tip is empty. Retitled to v1.71.0.0 since #2691 claims 1.70.0.0; the CHANGELOG gap closes when that lands.

One thing that needs your click before this can report anything: all ten workflow runs on this PR have sat at `action_required` since 2026-08-22 — only PR Title Sync executes, on `pull_request_target`. CI has never run here, so the BLOCKED is the missing approval, not a failing check. Version Gate in particular is the one I'd want to see run.

Local evidence meanwhile: 66/66 in `test/gstack-developer-profile.test.ts`, 158 assertions. Free suite run differentially against `upstream/main` on the same box — no new failures.

Still yours to call, unchanged from the PR description: auto-wiring `--reconcile` into office-hours, part (a) Phase-1 stub logging, and the shared-atomic-writer refactor if you want it.

PR #2670, opened 2026-08-22
fix(brain-sync): fail loud when artifacts source has zero pages

jzeisweiss:
## Summary

- stop reporting a successful artifact git command as gbrain indexing
- verify the maintained artifact source page count with a read-only, five-second bounded probe
- fail loud with `index_failed` when that source reports zero pages; otherwise report the git result and explicitly mark index freshness as unverified
- terminate the full probe process tree on timeout, including Git Bash/MSYS native cleanup

This deliberately does not run `gbrain sync` at every skill boundary: that operation is expensive and owns a single-writer database. Missing, unavailable, invalid, or nonzero probe results remain truthful `indexing not verified` outcomes rather than false indexing claims.

## Verification

Red first on upstream `main`: the zero-page regression returned status 0 before this change. The final test proves a real artifact commit reaches the remote, then the command fails because the maintained source still has zero pages.

- `bun test test/brain-sync.test.ts test/gstack-gbrain-sync.test.ts test/brain-sync-windows-paths.test.ts test/gbrain-spawn-windows-shell.test.ts` — 122 pass, 0 fail
- `bash -n bin/gstack-brain-sync`
- `shellcheck --severity=error bin/gstack-brain-sync`
- `git diff --check`

The broader free suite completed every shard except one unrelated timeout-wrapper timing assertion under six-way load (`rc=143` vs `rc=124`); that assertion passes standalone on both this commit and pristine upstream `main`.

### DIFF
--- a/bin/gstack-brain-sync
+++ b/bin/gstack-brain-sync
@@ -71,6 +71,213 @@ remote_auth_hint() {
   esac
 }
 
+# Derive the artifacts source id without creating, advancing, or re-registering
+# anything. The explicit override is the same one used by
+# gstack-gbrain-source-wireup; otherwise the artifacts remote basename is the
+# canonical source id.
+artifacts_source_id() {
+  if [ -n "${GSTACK_BRAIN_SOURCE_ID:-}" ]; then
+    printf '%s\n' "$GSTACK_BRAIN_SOURCE_ID"
+    return 0
+  fi
+
+  local remote_url=""
+  remote_url=$(git -C "$GSTACK_HOME" remote get-url origin 2>/dev/null) || true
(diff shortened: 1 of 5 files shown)


jzeisweiss:
Rebased onto current `main` (`ad840054`, v1.69.0.0). Same commit, new parent.

Local verification after rebase: `bun test test/brain-sync.test.ts test/gstack-gbrain-sync.test.ts` → **97 pass, 0 fail**.

jzeisweiss:
Privacy scrub: removed a private-repository reference from the PR description that was not resolvable for external readers. No other description changes.

PR #2705, opened 2026-08-27
fix: shared preamble telemetry lost across separate tool-call boundaries

exGeni:
_Implementation: Claude Code (Sonnet 5). Independent technical review: Codex — findings from three separate review passes (initial plan, initial diff, and this revision's diff) are folded into this PR (see Live evidence). This PR was revised after a design-comparison pass explicitly asked "is the current implementation the simplest way to guarantee the required properties, or just the first one that worked" — the revision below replaced the original mtime-based recovery with something smaller and safer; details in the "Revision" section._

## Why (in your own words)

Every skill's preamble sets _TEL_START/_SESSION_ID/_TEL once, and the
"Telemetry (run last)" block reads them again at the end. On a host that runs
each fenced bash block in a SKILL.md as its own subprocess (observed: Claude
Code — only cwd persists between tool calls, not shell state), the
completion block sees none of them. Worst case: _TEL (telemetry on/off) is
lost the same way, and the completion gate is `[ "$_TEL" != "off" ]` — an
empty _TEL reads as "not off", so a user who explicitly set telemetry: off
still gets local analytics writes and a remote gstack-telemetry-log call.

## Live evidence

Before (the actual arithmetic bug, this machine):

```
$ env -u _TEL_START bash -c '_TEL_END=$(date +%s); _TEL_DUR=$(( _TEL_END - _TEL_START )); echo "TEL_START=[$_TEL_START] TEL_DUR=$_TEL_DUR"'
TEL_START=[] TEL_DUR=1787837147
```

(an unset _TEL_START is 0 in Bash arithmetic, so duration becomes the raw
current epoch, not near-zero as the symptom first suggested)

Full end-to-end live run on this machine (real telemetry: off setting),
preamble and completion run as genuinely separate processes sharing one
parent PID:

```
$ (preamble) echo "TELEMETRY=$_TEL PPID=$PPID EXPECT_SESSION_ID=$_SESSION_ID"
TELEMETRY=off PPID=3427913 EXPECT_SESSION_ID=3427913-1787849499

$ (completion, separate process, ~4s later, no inherited vars) echo "RECOVERED _TEL=$_TEL _TEL_DUR=$_TEL_DUR _SESSION_ID=$_SESSION_ID"
RECOVERED _TEL=off _TEL_DUR=17 _SESSION_ID=3427913-1787849499

analytics before=26 after=26 (must match, TELEMETRY=off)
```

(session id matches exactly across the two separate-process events; 17s
duration matches the real wall-clock gap; zero new analytics lines despite
telemetry: off being invisible to the completion process)

Regression test (test/preamble-cross-process-telemetry.test.ts) run against
this exact diff:

```
bun test v1.3.14 (0d9b296a)
 4 pass
 0 fail
 30 expect() calls
Ran 4 tests across 1 file.
```

## Revision: mtime-based recovery replaced with marker-content recovery

The first version of this PR recovered _TEL_START/_SESSION_ID by reading
`~/.gstack/sessions/"$PPID"`'s file MTIME, which needs an OS branch (GNU
`date -r` vs BSD `stat -f %m`) — exactly the kind of platform-specific
parsing this fix exists to remove. A design-comparison pass (explicitly
framed as "don't take the current implementation as given — compare
alternatives on: amount of new logic, blast radius, edge cases, host-path
dependency, per-skill byte cost, fit with upstream's architecture") found
that since the preamble already controls both the write and read side of
the marker file, there's no need to infer anything from filesystem
timestamps: write the real values as text instead of an empty `touch`, read
them back with `read -r`. This dropped the OS branch entirely from both
files and reduced the per-skill byte cost by roughly a quarter (measured
below).

One candidate considered and rejected: pushing the recovery into a `bin/`
script (matching the existing `gstack-repo-mode`/`gstack-slug`
`eval "$(...)"` pattern). Verified this doesn't work: `hosts/define-host.ts`
defaults `usesEnvVars: true` for every host except Claude, so
`${ctx.paths.binDir}` resolves to `$GSTACK_BIN` — a variable set only inside
the preamble's own `runtimeRoot` block, which doesn't survive to a separate
completion-process for 9 of 10 hosts. (This exact pre-existing bug is
already reproducible today in this repo's own `test/fixtures/golden/codex-
ship-SKILL.md`, lines 837/843-844 — the completion block's existing
timeline-log/telemetry-log calls already silently no-op on env-var hosts
for the same reason. Filed as a separate, pre-existing, out-of-scope
observation, not part of this fix.)

A second `/codex review` pass on the marker-content revision itself caught
a real fail-open risk: if `_TEL` were ALSO recovered from the marker (as
originally drafted), a completion process whose $PPID happens to collide
with an unrelated or stale session's marker (confirmed possible on this
machine — `~/.gstack/sessions/1` exists, meaning some Bash-tool subprocess
was at some point reparented to init before reading $PPID) could inherit
that foreign session's telemetry opt-IN, leaking analytics for a user who
currently has telemetry: off. Fixed by keeping `_TEL` recovery on its own,
separate path: re-reading the live `~/.gstack/config.yaml` directly (never
the marker), so a marker collision can only ever produce a wrong-but-
harmless duration/session-id, never a privacy leak. Added a behavioral test
that plants a foreign marker with an old-format telemetry-looking third
field and asserts it's ignored.

Measured impact (real generator run, not estimated):
- `ship/SKILL.md`: 91,954 B vs main's 91,267 B (delta +687) vs this PR's
  original version's 92,204 B (delta +937) — about a quarter smaller, while
  closing the fail-open gap above that the original version didn't have.
- All 9 previously-bumped `maxSkeletonBytes` budgets re-measured and
  reduced (still bumped, but by roughly half as much; exact values in each
  comment in `test/helpers/carve-guards.ts`).

A third `/codex review` pass on this final diff raised one advisory (P2,
non-blocking) finding: the `$PPID` marker is shared across concurrent or
rapidly-sequential skill invocations in one host process, so a second
invocation's preamble can overwrite a first invocation's marker before the
first's completion reads it, misattributing duration/session-id. This is
the same trade-off already disclosed in this PR's first commit (reusing the
existing shared marker rather than introducing a new per-invocation file) —
independently reconfirmed here, not a new finding, and still judged
acceptable: it degrades to wrong-but-harmless local analytics, never a
crash, never a privacy issue (that path is now hard-separated per the fix
above).

## Scope

- **Changed:** scripts/resolvers/preamble/generate-preamble-bash.ts,
  scripts/resolvers/preamble/generate-completion-status.ts (writes real
  values into the existing `~/.gstack/sessions/"$PPID"` marker instead of an
  empty touch; recovers duration/session-id via `read -r` from that marker;
  recovers `_TEL` separately and only from the live config file, never the
  marker) + regenerated all 50 preamble-tier SKILL.md files + 3 refreshed
  `ship` golden fixtures + adjusted `maxSkeletonBytes` for the 9 skills
  whose skeleton was already close to budget (same convention as the prior
  v1.68 #2402 bumps, each with its measured value in the comment)
- **Verified live by:** the commands above, run directly on this machine,
  across three review/revision cycles
- **Did NOT test:** the marker-sharing race between concurrent skills in
  one host process (documented, accepted, see "Revision" above); Windows
  (this PR's changed files aren't in the curated Windows CI subset, and the
  new regression test uses `spawn('bash', ...)`, which is excluded from
  that subset by the runner's own POSIX-only curation)

## Liveness proof (required)

изображение

## Checklist

- [x] Liveness screenshot attached: `GSTACK PR` typed live into a real surface
- [x] This is not a generated-file-only diff (edited the two resolver .ts
      sources, regenerated)
- [x] No ETHOS.md edits, no voice/founder-perspective/YC changes
- [x] N/A — not a new public command/service/host adapter
- [x] Linked issue or reproduction: reproduction above (no existing issue found)

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -546,7 +546,9 @@ if [ -z "$_SESSION_ID" ]; then
   _SESSION_ID="$PPID-$_TEL_START"
 fi
 if [ -z "$_TEL" ]; then
-  _TEL=$(~/.claude/skills/gstack/bin/gstack-config get telemetry 2>/dev/null || echo "off")
+  _TEL_CFG="${GSTACK_STATE_ROOT:-${GSTACK_HOME:-${GSTACK_STATE_DIR:-$HOME/.gstack}}}/config.yaml"
+  _TEL=$(grep -E '^telemetry:' "$_TEL_CFG" 2>/dev/null | tail -1 | sed -E 's/^telemetry:[[:space:]]*//; s/[[:space:]]+$//')
+  [ -z "$_TEL" ] && _TEL="off"
 fi
 _TEL_DUR=$(( _TEL_END - _TEL_START ))
 rm -f ~/.gstack/analytics/.pending-"$_SESSION_ID" 2>/dev/null || true
--- a/SKILL.md
+++ b/SKILL.md
@@ -69,8 +69,13 @@ _LAKE_SEEN=$([ -f ~/.gstack/.completeness-intro-seen ] && echo "yes" || echo "no
 echo "LAKE_INTRO: $_LAKE_SEEN"
 _TEL=$(~/.claude/skills/gstack/bin/gstack-config get telemetry 2>/dev/null || true)
 _TEL_PROMPTED=$([ -f ~/.gstack/.telemetry-prompted ] && echo "yes" || echo "no")
-_TEL_START=$(date +%s)
-_SESSION_ID="$$-$(date +%s)"
+_SESS_MARKER=~/.gstack/sessions/"$PPID"
+if [ "$(uname -s 2>/dev/null)" = "Darwin" ]; then
+  _TEL_START=$(stat -f %m "$_SESS_MARKER" 2>/dev/null || date +%s)
+else
+  _TEL_START=$(date -r "$_SESS_MARKER" +%s 2>/dev/null || stat -c %Y "$_SESS_MARKER" 2>/dev/null || date +%s)
+fi
+_SESSION_ID="$PPID-$_TEL_START"
 echo "TELEMETRY: ${_TEL:-off}"
 echo "TEL_PROMPTED: $_TEL_PROMPTED"
 _EXPLAIN_LEVEL=$(~/.claude/skills/gstack/bin/gstack-config get explain_level 2>/dev/null || echo "default")
@@ -529,6 +534,20 @@ Run this bash:
 
 ```bash
 _TEL_END=$(date +%s)
+_SESS_MARKER=~/.gstack/sessions/"$PPID"
+if [ -z "$_TEL_START" ]; then
+  if [ "$(uname -s 2>/dev/null)" = "Darwin" ]; then
+    _TEL_START=$(stat -f %m "$_SESS_MARKER" 2>/dev/null || echo "$_TEL_END")
+  else
+    _TEL_START=$(date -r "$_SESS_MARKER" +%s 2>/dev/null || stat -c %Y "$_SESS_MARKER" 2>/dev/null || echo "$_TEL_END")
+  fi
+fi
+if [ -z "$_SESSION_ID" ]; then
+  _SESSION_ID="$PPID-$_TEL_START"
+fi
+if [ -z "$_TEL" ]; then
+  _TEL=$(~/.claude/skills/gstack/bin/gstack-config get telemetry 2>/dev/null || echo "off")
+fi
 _TEL_DUR=$(( _TEL_END - _TEL_START ))
 rm -f ~/.gstack/analytics/.pending-"$_SESSION_ID" 2>/dev/null || true
 # Session timeline: record skill completion (local-only, never sent anywhere)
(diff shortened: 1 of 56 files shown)


exGeni:
Closing as superseded by v1.71.0.0 (#2691). That release replaced the entire inline preamble bash this PR touched with the compiled `bin/gstack-skill-start` / `bin/gstack-skill-end` pair, which independently solves both problems this PR targeted:

- Session-id/duration surviving separate Bash-tool-call subprocesses: `gstack-skill-start` now echoes `SESSION_ID`/`TEL_START` as STATUS lines that the calling agent passes explicitly to `gstack-skill-end --session-id ... --tel-start ...` as CLI args, rather than relying on shell state or a file marker.
- The `_TEL`-off fail-open violation: both `gstack-skill-start` (~line 107) and `gstack-skill-end` (~line 34) re-read telemetry state fresh via `gstack-config get telemetry` (a live grep of `~/.gstack/config.yaml`) on every invocation — `gstack-skill-end` even fails closed (`|| echo off`) if that read errors. Neither script recovers telemetry consent from a per-session marker file, so the $PPID-collision leak class this PR's /codex review found doesn't apply to the new architecture.

Thanks for the fast turnaround on #2417 — closing this one since the bash it patches no longer exists.

PR #2708, opened 2026-08-28
fix: make squash a merge-method default

JimmyJuan:
## Summary
- Treat squash as the recommended default instead of a mandatory merge method.
- Honor explicit user choices and documented project constraints after intersecting them with GitHub repository capabilities.
- Bind every non-interactive merge command to an explicit PR target and exactly one resolved method.
- Keep ordinary git push behavior separate from PR merge policy.

## Tests
- Full bun test suite passes.
- Merge-command regression test: 2 passed, 0 failed.
- Generated skill freshness passes for all supported hosts.
- git diff --check passes.

## Behavior examples
- Repositories that allow squash default to squash.
- A repository that only allows rebase automatically uses rebase and reports the constraint.
- When squash is unavailable and multiple alternatives remain, the workflow asks instead of guessing.

### DIFF
--- a/land-and-deploy/sections/merge-and-deploy.md
+++ b/land-and-deploy/sections/merge-and-deploy.md
@@ -5,10 +5,32 @@
 Record the start timestamp for timing data. Also record which merge path is taken
 (auto-merge vs direct) for the deploy report.
 
-Try auto-merge first (respects repo merge settings and merge queues):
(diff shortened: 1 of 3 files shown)


JimmyJuan:
Superseded by #2712, which splits the source/test change from generated-file regeneration and updates the carve guard.

PR #2751, opened 2026-08-31
redact-prepush: repo-committed scan exclusions via .gstack/redact-ignore

Yugz29:
## Problem

Repos whose test suites deliberately contain **fake credentials** — redaction test fixtures, placeholder connection strings — trip `gstack-redact-prepush` whenever those files are edited: MEDIUM noise on every push at best, a **blocking HIGH on a fixture** at worst. The hook's own header (the *FALSE HIGH FINDINGS* note) already names the consequence: it trains the `--no-verify` reflex, which defeats the guardrail far more thoroughly than an ignored fixture would. Hit in practice on a repo whose redaction test suite carries fake `postgres://user:pass@…` URLs and fake tokens by design.

## Mechanism

Opt-in, repo-scoped: a `.gstack/redact-ignore` file at the repo root, one git pathspec per line (`#` comments), appended as `:(exclude)` pathspecs to **both** added-line diff paths (the per-commit `git show --cc` and the range `git diff`), so listed paths never enter the scan.

- **Consent model is the file itself**: it is committed and reviewed like any code change — excluding a path is a visible diff, not a local bypass.
- Unreadable file → no exclusions (fail closed). No file → behavior unchanged for every existing user.
- Resolved once per run (cached), from `git rev-parse --show-toplevel`.

## Verification

Poisoned-commit test on a real repo: a fake `postgres://svc:…@db.internal` line committed under an excluded `tests_v2/` is skipped entirely (exit 0, silent); the **same line** committed under a non-excluded directory still blocks HIGH with the usual message.

### DIFF
--- a/bin/gstack-redact-prepush
+++ b/bin/gstack-redact-prepush
@@ -190,6 +190,36 @@ function unknownRemoteTipBase(localSha: string): string | null {
  * resolving a merge is still caught, while an ordinary merge contributes
  * nothing. Returns null when the notion does not apply, so callers fall back.
  */
+/**
+ * Repo-committed scan exclusions: `.gstack/redact-ignore` at the repo root,
+ * one git pathspec per line, `#` comments. Listed paths are appended as
+ * `:(exclude)` pathspecs to the added-line diffs, so they never enter the
+ * scan — meant for test fixtures that deliberately contain fake credentials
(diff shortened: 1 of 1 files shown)


Yugz29:
Related: #2899 (env.kv false positives on function calls, same hook).

PR #2771, opened 2026-09-01
browse: honor GSTACK_CHROMIUM_PATH in the headless launch path (macOS 13 fix)

oli548:
## Problem

On macOS 13, `./setup` and `/browse` fail because Playwright no longer ships a
bundled Chromium / `chrome-headless-shell` build for that OS:

```
launch: Executable doesn't exist at .../ms-playwright/chromium_headless_shell-1234/chrome-headless-shell-mac-x64/chrome-headless-shell
Looks like Playwright was just installed or updated. Please run: npx playwright install
```

`setup` hard-exits at the `ensure_playwright_browser` gate before any skills are
linked, and even after linking, `browse goto` can't start its daemon.

`GSTACK_CHROMIUM_PATH` already lets `launchHeaded()` (the persistent-context /
headed path) point at a custom Chromium/Chrome bundle, but the default
**headless** `launch()` path deliberately ignored it and always used the
Playwright cache. So there was no way to run headless browse against a stock
system Chrome on an OS Playwright no longer bundles for.

## Fix

Make the headless `launch()` path honor `GSTACK_CHROMIUM_PATH` too, mirroring
`launchHeaded()`:

- when the env var is set, pass it as `executablePath`;
- scope out the XProtect self-heal for that launch via
  `usesCustomExecutable: Boolean(headlessExecutablePath)` — a
  `GSTACK_CHROMIUM_PATH` bundle belongs to the wrapper/embedder and must never
  be quarantine-cleared or reinstalled over (same contract `launchHeaded()` and
  `probePoisonedChromiumBundle` already follow).

When `GSTACK_CHROMIUM_PATH` is unset, behavior is unchanged: `executablePath` is
omitted and `usesCustomExecutable` is `false`, so the Playwright cache is used
and the heal applies exactly as before.

## Testing

Manually verified on **macOS 13.7.8 (22H730), Google Chrome 152**:

```
export GSTACK_CHROMIUM_PATH="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
browse goto https://example.com      # -> Navigated to https://example.com (200)
browse text                          # -> "Example Domain ..."
browse status                        # -> healthy, Mode: launched, Tabs: 1
browse js "navigator.userAgent"      # -> HeadlessChrome/152 ...
```

Also confirmed a direct Playwright probe launches system Chrome headless via
`chromium.launch({ executablePath })` on this OS.

Note: the `launch()` method isn't currently under unit test upstream (it calls
`chromium.launch` directly), so this change ships without one; happy to add a
seam/test if you'd like.

### DIFF
--- a/browse/src/browser-manager.ts
+++ b/browse/src/browser-manager.ts
@@ -520,13 +520,23 @@ export class BrowserManager {
       launchArgs.push(...headlessGpuArgs(process.platform, process.env));
     }
 
+    // Support custom Chromium/Chrome binary via GSTACK_CHROMIUM_PATH env var
+    // in the headless path too (macOS 13 fix): Playwright dropped its bundled
+    // Chromium/chrome-headless-shell builds for older macOS, but a stock system
+    // Chrome (which is Chromium-based) launches fine here via executablePath.
+    // When the env var supplies the bundle, scope out the XProtect self-heal —
+    // that bundle belongs to the wrapper/embedder, exactly as launchHeaded does.
+    const headlessExecutablePath = process.env.GSTACK_CHROMIUM_PATH || undefined;
+
     // XProtect self-heal wrapper (P0 #2554): a macOS definition update can
     // start SIGKILLing the pinned Chromium at spawn. On the classified
     // signature, clear quarantine on the Playwright cache + force-reinstall
-    // once, then retry this launch once. This headless path always uses the
-    // Playwright cache (no executablePath), so the heal is never scoped out.
+    // once, then retry this launch once. With no GSTACK_CHROMIUM_PATH this path
+    // uses the Playwright cache, so the heal applies; a custom executable is
+    // scoped out via usesCustomExecutable below.
     this.browser = await launchWithXProtectHeal(() => chromium.launch({
       headless: useHeadless,
+      ...(headlessExecutablePath ? { executablePath: headlessExecutablePath } : {}),
       // #2220: the daemon owns signal policy, not Playwright. Playwright's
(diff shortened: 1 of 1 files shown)


bluesourcebeck:
Thanks for fixing this. I independently tested a matching headless-launch change
using an installed Chromium-compatible browser (Helium, Chromium 152) and found
one more launch path worth covering: `BrowserManager.handoff()`.

Against upstream commit `71f6048e8ada25180e61438abc1d98cb151fe9a7`, handoff still
probes the bundled Chromium and starts `launchPersistentContext()` without the
custom executable. Thus headless startup can work while switching to a visible
browser fails when the Playwright browser cache is intentionally absent.

The companion fix I tested locally:

- Resolve the existing `GSTACK_CHROMIUM_PATH` in handoff too.
- Pass it as `executablePath` to the persistent-context launch.
- Set `usesCustomExecutable` on the XProtect healing wrapper, as in this PR.
- Skip the bundled-browser poison probe when a custom executable is selected.

An opt-in live test used only a loopback page and an isolated temporary profile.
It checked URL, synthetic cookie and localStorage retention, a click, screenshot,
extension service worker, and browser health after handoff. An invalid custom
path also left the original headless browser usable. Live + source-wiring tests
passed (2 tests, 21 assertions); four related test files passed (50 tests).
These were local macOS checks with existing dependencies, not a full upstream
suite or cross-platform certification.

Would you prefer to include this remaining handoff path here, or keep it as a
small follow-up? Posting here first to avoid another duplicate headless fix.

PR #2774, opened 2026-09-02
Fix global binary fallback path for env-var hosts

blockchainian:
## Summary

The browse, design, and make-pdf setup blocks build the global-install fallback path as `$HOME` plus the host's binary directory with a leading `~` stripped. Hosts with `usesEnvVars` (codex today) store `$GSTACK_BROWSE`-style variables there instead of `~` paths, so nothing is stripped and the generated shell reads:

```bash
[ -z "$B" ] && B="$HOME$GSTACK_BROWSE/browse"
```

which expands to `/Users/me/Users/me/.codex/skills/gstack/browse/dist/browse`. The `-x` check fails and every affected Codex skill (browse, benchmark, canary, the four design skills, make-pdf) reports `NEEDS_SETUP` or `DESIGN_NOT_AVAILABLE` even though the binary is installed, prompting the user for a pointless rebuild on each session.

## Fix

Add `globalBinDir()` in `scripts/resolvers/types.ts`, which expands a leading `~` to `$HOME` and passes env-var paths through unchanged, and use it at all six sites in `browse.ts`, `design.ts`, and `make-pdf.ts`.

Generated Codex output after the fix:

```bash
[ -z "$B" ] && B="$GSTACK_BROWSE/browse"
```

Claude and other literal-path hosts are unchanged (`$HOME/.claude/skills/gstack/browse/dist/browse`).

## Test plan

- [x] New `test/resolver-global-bin-path.test.ts` fails on main, passes with the fix, for both env-var and literal hosts
- [x] `bun test` unit suite: no new failures (the 7 pre-existing failures in gbrain-detect, parity-suite, and user-slug-fallback are identical on clean main)
- [x] `bun run gen:skill-docs --host codex` regenerated; setup block prints `READY` in a Codex CLI session

https://claude.ai/code/session_01Ko25nKXJXcKY2haaWwXDcG

### DIFF
--- a/scripts/resolvers/browse.ts
+++ b/scripts/resolvers/browse.ts
@@ -1,4 +1,4 @@
-import type { TemplateContext } from './types';
+import { globalBinDir, type TemplateContext } from './types';
 import { COMMAND_DESCRIPTIONS } from '../../browse/src/commands';
 import { SNAPSHOT_FLAGS } from '../../browse/src/snapshot';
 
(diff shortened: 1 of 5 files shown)


blockchainian:
Closing: this was already fixed upstream in v1.64.0.0 (#2571) via `toShellPath`. My local install was behind.

PR #2784, opened 2026-09-02
fix(redact): four-part version strings are not public IPv4 addresses

cdecook23:
## The problem

A repo using a four-part `MAJOR.MINOR.PATCH.BUILD` version trips `pii.ip_public` on **every release**. `1.128.1.0` is a syntactically valid public IPv4 address, so each file the version is written to lands as a MEDIUM finding:

```
MEDIUM pii.ip_public   9:6   1.12*****     ## [1.128.1.0] - 2026-09-02
MEDIUM pii.ip_public  27:2   1.12*****     VERSION
MEDIUM pii.ip_public  53:16  1.12*****     package.json
MEDIUM pii.ip_public  60:20  1.12*****     package-lock.json  (top level)
MEDIUM pii.ip_public  72:16  1.12*****     package-lock.json  (packages."")
```

Five findings, every push, indefinitely — the count is structural, since those are exactly the files a version has to be written to. The practical cost isn't the noise: it's that a scanner reporting the same findings every time is one people stop reading, which is how a genuine finding gets waved through. (This surfaced on a repo where the standing advice had become "it's always five, ignore it" — and nobody had checked what the five were.)

Four-part schemes aren't unusual: .NET/NuGet assembly versions, Electron builds, and plenty of hand-rolled schemes all look like this.

## The fix

`pii.ip_public` now rejects a dotted quad that is *provably* a version.

The bias is deliberate and I'd rather state it than have it inferred: **a leaked address costs more than a nuisance finding**, so an exemption requires proof of version context, never mere proximity to the word "version". Two ways to earn it:

1. **This occurrence sits in a version declaration** — a `*version` key (`"version": "1.2.3.4"`, `app_version = 1.2.3.4`), a `v` prefix (`v1.2.3.4`), or a Keep-a-Changelog heading (`## [1.2.3.4] - date`).
2. **The identical string is declared a version elsewhere in the same scanned text.** This covers the bare `1.128.1.0` line of a `VERSION` file without teaching the validator anything about diff or file formats — if those exact characters are proven a version in one place, a bare repetition of them is the same version.

Consequences worth checking in review:

- A bare `8.8.8.8` stays flagged — nothing proves it a version.
- A real address **cannot** be laundered by sharing a diff with an unrelated version string, because rule 2 matches on the exact value.
- `version: 1.2.3.4 via 8.8.8.8` still flags the address — the key exempts its own value, not the line.

## Shape

No engine or `RedactPattern` change. It lives entirely in the existing `validate` hook, which already receives the `RegExpExecArray` (and therefore `.input`/`.index`). `nearRegex` is positive-only, so it couldn't express this.

## Tests

Four new cases in `test/redact-engine.test.ts`: the declaration forms, the bare-line-plus-proof case, and two laundering attempts.

- All **164 redact tests pass** across the 8 redact test files, including `redact-pattern-lint.test.ts` (the linear-time regex lint).
- Verified against the real-world diff that motivated this: **5 MEDIUM → 0**.
- Full `test/` suite: 6 failures, identical before and after this change (`gstack-gbrain-detect`, `resolve-user-slug` — environment-dependent, none touching redaction). Compared against a stashed baseline rather than assumed.

### DIFF
--- a/lib/redact-patterns.ts
+++ b/lib/redact-patterns.ts
@@ -159,6 +159,56 @@ export function isPublicIPv4(ip: string): boolean {
   return true;
 }
 
+/**
+ * True when a dotted quad is a four-part VERSION rather than an IP address.
+ *
+ * Four-segment version schemes (`MAJOR.MINOR.PATCH.BUILD` — .NET/NuGet
+ * assembly versions, Electron builds, and hand-rolled schemes) are
+ * syntactically indistinguishable from IPv4: `1.128.1.0` is a valid public
+ * address. A repo on such a scheme therefore trips `pii.ip_public` on EVERY
+ * release, once per file the version is written to — the changelog heading,
+ * a VERSION file, package.json, and both package-lock.json fields. A scanner
+ * that reports the same findings on every push is one nobody reads.
+ *
+ * The bias here is deliberate: a leaked IP costs more than a nuisance
+ * finding, so an exemption requires PROOF of version context, never mere
+ * proximity to the word "version". Two ways to earn it:
+ *
+ *   1. This occurrence sits in a version declaration — a `*version` key, a
+ *      `v` prefix, or a `[…]` changelog heading.
+ *   2. The IDENTICAL string is declared as a version elsewhere in the same
+ *      scanned text. That covers the bare `1.128.1.0` line of a VERSION file
+ *      without teaching this validator anything about diff or file formats:
+ *      if the same characters are proven a version in one place, a bare
+ *      repetition of them is the same version.
+ *
+ * So `8.8.8.8` on a line by itself stays flagged — nothing proves it a
+ * version — while a real address never becomes exempt by sharing a diff with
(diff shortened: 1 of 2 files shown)


cdecook23:
Rebased onto current `main` — was conflicting after 57 commits, now `MERGEABLE`.

The conflict was purely additive in `test/redact-engine.test.ts`: the digit-only-UUID suppression tests landed in the same block. Both sides kept — that file now runs **100 tests**, and the full redact suite is **230 passing across 9 files**, including the linear-time pattern lint.

Re-verified after the rebase: the motivating release diff still goes 5 MEDIUM → 0, and a bare `8.8.8.8` still flags.

Worth noting the two changes are neighbours in intent. The UUID-fixture suppression comment says it plainly — *"14 of 21 MEDIUM findings on one ordinary branch were exactly this, all from test files — the volume that makes people stop reading MEDIUM output at all."* The four-part-version case is the same failure in a different disguise: it fires on every release, in exactly the files a release has to touch.

PR #2857, opened 2026-09-12
fix(browse): Aside readiness probe fails under zsh and discards its own error output

TheAppMakerPro:
## What's broken

The `{{ASIDE_SETUP}}` readiness probe reports `ASIDE_NOT_RUNNING` for a healthy, signed-in Aside. Two independent defects, both in the probe's bash block.

**1. The timeout prefix is not word-split under zsh.**

```bash
...
```

bash splits unquoted `$_T` into `gtimeout` + `30`. zsh does not word-split unquoted parameters, so it tries to exec a command literally named `gtimeout 30` and fails with "command not found". Any harness running skill blocks under zsh on macOS hits this every time.

**2. The probe discards the evidence it tells you to quote.**

`2>&1 | grep -q` throws the output away, so every failure collapses into `ASIDE_NOT_RUNNING` regardless of cause. Step 2 then instructs the agent to "quote the probe output verbatim", which is impossible, because it was already discarded.

The two compound. The zsh failure is invisible, so a running Aside looks like a closed app and the skill falls through to the `$B` fallback engine: a lower-fidelity driver with no real sessions and no extensions. The degradation is silent and easy to miss.

## What changed

- Bind the deadline through a function applied to `"$@"` instead of an unquoted variable prefix. Splits correctly in bash and zsh. The chain is unchanged: gtimeout, then timeout, then perl alarm.
- Capture the probe output, match against it, and print it under `PROBE_OUTPUT:` on failure, capped at 5 lines.
- Step 2 now says how to read that output. Aside's own "isn't running on this machine" means the app is closed and should be opened. A permission, sandbox, or helper/signing error means Aside is healthy and the shell could not reach it, so opening the app will not help.

`{{ASIDE_RESEARCH}}` lifts the probe from `{{ASIDE_SETUP}}` at render time, so one edit corrects both copies.

## Verification

Same machine, same shell (`/bin/zsh`), Aside 1.26.906.1630 running and signed in:

- before: `ASIDE_NOT_RUNNING`
- after: `READY: aside 1.26.906.1630`

In a shell that genuinely cannot reach Aside, the new probe prints the real cause under `PROBE_OUTPUT:` instead of a misleading one.

## Note on the test change

`test/aside-driver.test.ts` asserted the literal strings `_T="gtimeout 30"`, `_T="timeout 30"` and `_T="perl -e ..."`. Those pinned the exact syntax the bug lived in, so a correct probe failed them while the broken one passed. I repinned them to the behaviour the surrounding comment says they were meant to protect (the ordered deadline chain), added a regression guard that fails if any unquoted variable is used as a command prefix before `aside repl`, and pinned that a failed probe keeps its output.

Commits are split per the bisect convention: source fix, test repin, then generated output (`bun run gen:skill-docs`; 20 SKILL.md plus the 3 ship goldens, no hand edits).

## CI expectations

This is a fork PR, so the eval and E2E lanes will fail with empty-env auth rather than on the change itself. Per CONTRIBUTING that is expected for fork PRs; free-tests, check-freshness, workflow-lint and windows-tests still give real signal.

Local results:

- `test/aside-driver.test.ts` + `test/host-config.test.ts`: 105 pass, 0 fail.
- `test/gstack-design-detect.test.ts`: 3 failing, pre-existing. Verified by stashing this branch's changes and re-running at a clean HEAD: identical 62 pass / 3 fail. They are realpath (`/private/var` vs `/var`) and mirror-download assertions, untouched by this change.
- `test/parity-suite.test.ts`: unverified. The free runner killed its shard at the 460s wall-clock deadline and reported it TIMED-OUT rather than failed, still in flight at the kill. Flagging it rather than claiming it passes.

https://claude.ai/code/session_018HtyLHJaPqMef3aoQUwzxx

### DIFF
--- a/benchmark/SKILL.md
+++ b/benchmark/SKILL.md
@@ -161,19 +161,23 @@ Skills that run plan reviews (`/plan-*-review`, `/codex review`) include the EXI
 gstack drives the Aside AI browser first. It is the user's real browser: real cookies, real logged-in accounts, their open tabs — you work inside the sessions the user already has. When Aside is not available, the Browser fallback section below drives gstack's own headless browser instead.
 
 ```bash
-_T=""; command -v gtimeout >/dev/null 2>&1 && _T="gtimeout 30"; [ -z "$_T" ] && command -v timeout >/dev/null 2>&1 && _T="timeout 30"
-[ -z "$_T" ] && command -v perl >/dev/null 2>&1 && _T="perl -e alarm(shift);exec(@ARGV) 30"
+_gs_timeout() { if command -v gtimeout >/dev/null 2>&1; then gtimeout 30 "$@"; elif command -v timeout >/dev/null 2>&1; then timeout 30 "$@"; elif command -v perl >/dev/null 2>&1; then perl -e 'alarm(shift); exec(@ARGV)' 30 "$@"; else "$@"; fi; }
 if [ "${GSTACK_SKIP_ASIDE:-}" = "1" ] || ! command -v aside >/dev/null 2>&1; then
   echo "NEEDS_ASIDE"
-elif $_T aside repl 'console.log("ASIDE_READY " + pwd)' 2>&1 | grep -q '^ASIDE_READY'; then
-  echo "READY: aside $(aside --version 2>/dev/null)"
 else
-  echo "ASIDE_NOT_RUNNING"
+  _OUT=$(_gs_timeout aside repl 'console.log("ASIDE_READY " + pwd)' 2>&1)
+  if echo "$_OUT" | grep -q '^ASIDE_READY'; then
+    echo "READY: aside $(aside --version 2>/dev/null)"
+  else
+    echo "ASIDE_NOT_RUNNING"
+    echo "PROBE_OUTPUT:"
+    echo "$_OUT" | head -5
+  fi
 fi
 ```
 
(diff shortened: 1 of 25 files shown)


TheAppMakerPro:
Withdrawing this. The probe bug it describes is real and reproducible, but this patch is not the right fix: it regresses the parity gate.

Verified locally, test/parity-suite.test.ts vs the v1.64.1.0 baseline:
- on main: 2 pass, 0 fail
- with this branch: 4 skills over their ceilings (plan-eng-review +110B, office-hours +219B, design-consultation +613B and ratio 1.082, qa ratio 1.086)

The cause is mine, not the probe change: I expanded step 2 of the BROWSER SETUP prose from one line to a paragraph, and that cost lands in every skill rendering {{ASIDE_SETUP}}.

Leaving the underlying report here in case it is useful to whoever picks it up: in scripts/resolvers/aside.ts the probe holds its timeout prefix in an unquoted variable (_T="gtimeout 30", invoked as `$_T aside repl`). bash word-splits it; zsh does not, so under zsh it execs a command literally named "gtimeout 30" and a healthy signed-in Aside is reported as ASIDE_NOT_RUNNING. The failure is invisible because the probe pipes its output into `grep -q`, which also makes step 2's instruction to quote the probe output verbatim impossible to follow.

Sorry for the noise.

PR #2867, opened 2026-09-14
v1.84.2.0 Fix OSV Scanner failures: bump sharp and adm-zip, expire override pins

### LINKED ISSUE #2866: OSV Scanner failing on main: sharp and adm-zip overrides have fallen behind published fixes
The weekly OSV Scanner run on `main` has been failing since 2026-09-14 ([run 34819144240](https://github.com/garrytan/gstack/actions/runs/34819144240)). The 2026-09-07 run was green, so this is fresh.

Two advisories, both in packages that reach the tree only through `package.json` `overrides`:

| Package | Pinned | Advisory | Severity | Fixed in |
|---|---|---|---|---|
| `sharp` | `0.35.0` (exact) | GHSA-rgj7-g3m4-5g8c, bundled libheif RCE | 8.9 HIGH | 0.35.4 |
| `adm-zip` | `^0.6.0` | GHSA-vwc7-r8mq-g2x9 / CVE-2026-76845, symlink-follow on extract lets a crafted archive overwrite files outside the target directory | 6.8 MEDIUM | 0.6.1 |

One detail worth flagging, because the scan output actively misleads here: the `adm-zip` advisory records a `last_affected` event rather than a `fixed` event, so the scanner table prints `FIXED VERSION = --` and the summary reads as though the finding is unfixable. It is not. 0.6.1 is published and sits outside the affected range; `POST https://api.osv.dev/v1/query` against `adm-zip@0.6.1` returns zero vulns, as it does for `sharp@0.35.4`.

### Real exposure is low, but that is not the interesting part

The only `sharp` consumer is `browse/src/screenshot-size-guard.ts`, which downscales full-page PNG screenshots and never decodes the AVIF or HEIF paths the libheif advisory covers. Nothing in the tree imports `adm-zip` at all; it arrives via `onnxruntime-node` under `@huggingface/transformers`. Both fixes are a one-line version change, so exposure was never the deciding question.

### The recurring problem

An exact pin is the one dependency shape that cannot drift forward on its own. Dependabot will not move `sharp: "0.35.0"`, and `dependency-review.yml` only inspects newly *added* dependencies at PR time. So an aging pin stays invisible until a weekly cron turns red days later, and a weekly cron that only goes red is a signal nobody is subscribed to.

`.osv-scanner.toml` already solved the mirror-image problem: every `[[IgnoredVulns]]` entry carries an `ignoreUntil`, which forces re-justification. Nothing applied that discipline to the `overrides` block.

### Proposed fix

PR incoming. Three parts:

1. Bump `sharp` to 0.35.4 and `adm-zip` to 0.6.1.
2. Add `.override-pins.json` (a reason and a `reviewBy` date per override) plus `test/override-freshness.test.ts` in the free suite, so a stale pin fails on the PR rather than in a cron.
3. Have the OSV Scanner workflow open an issue on a failed scheduled run, guarded on `repository.has_issues` so forks with issues disabled skip the job.

smsmatt:
Closes #2866.

The weekly OSV Scanner run on `main` has been failing since 2026-09-14. Two advisories, both in packages that reach the tree only through `package.json` `overrides`, both already fixed upstream.

## What changed

**`sharp` 0.35.0 to 0.35.4** clears GHSA-rgj7-g3m4-5g8c (bundled libheif RCE, CVSS 8.9). **`adm-zip` 0.6.0 to 0.6.1** clears GHSA-vwc7-r8mq-g2x9 / CVE-2026-76845 (symlink-follow on extract). `POST https://api.osv.dev/v1/query` returns zero vulns for both new versions.

The `adm-zip` advisory records `last_affected` rather than a `fixed` event, so the scanner table prints `FIXED VERSION = --` and reads as though the finding is unfixable. It is not; 0.6.1 is published and outside the affected range.

**`.override-pins.json` + `test/override-freshness.test.ts`** give the `overrides` block the discipline `.osv-scanner.toml` already applies to suppressions. An exact pin is the one dependency shape that cannot drift forward on its own: Dependabot will not move `sharp: "0.35.0"`, and `dependency-review.yml` only inspects newly *added* dependencies at PR time. So each override now carries a reason and a `reviewBy` date, and the free suite fails when a pin has no rationale, when a rationale outlives its override, when a `reviewBy` has passed, or when an exact pin no longer matches what the lockfile resolved. That moves the signal from an unwatched weekly cron to a gate a reviewer is already reading.

**`osv-scanner.yml`** opens an issue on a failed scheduled run, or comments on the open one, guarded on `repository.has_issues` so forks with issues disabled skip the job rather than erroring.

## Exposure

Low, and deliberately not the deciding factor. The only `sharp` consumer is `browse/src/screenshot-size-guard.ts`, which downscales full-page PNG screenshots and never decodes the AVIF or HEIF paths the libheif advisory covers. Nothing imports `adm-zip` directly; it arrives via `onnxruntime-node` under `@huggingface/transformers`. Both fixes are a one-line version change, so there was no reason to weigh exposure against cost.

## Verification

`bun install` relocks cleanly, and the lockfile diff is scoped to `sharp`, `adm-zip`, and the `@img/sharp-*` platform packages. `test/deps-smoke.test.ts` (sharp import, `metadata()`, resize round-trip), `test/override-freshness.test.ts`, and `test/osv-config-wiring.test.ts` pass: 11 tests, 32 assertions.

On the full free suite, honest reporting: it is not green in my container, and it is not green on the base commit either. Base `71f6048e` fails 52 tests, this branch fails 46, and diffing the two failure sets leaves exactly one branch-only entry, a PTY WebSocket test in `browse/test/terminal-agent-integration.test.ts` that passes 3 of 3 in isolation on the branch and 2 of 2 on base. It is flake under shard load, not a regression. The shared baseline failures are environmental (no `file` binary for the compiled-binary check, `~/.gstack` home state, TMPDIR-sensitive path-traversal tests), so CI should be the authority on suite health here rather than my numbers.

VERSION is bumped to 1.84.2.0 with a CHANGELOG entry. Commits are bisected: the dependency bump, the freshness gate, the workflow change, and the release notes are each independently revertable.


smsmatt (commented):
/trunk merge

PR #2870, opened 2026-09-15
fix(aside): the readiness probe can never detect Aside under zsh, and hides why it fails

### LINKED ISSUE #2842: Aside probe always reports ASIDE_NOT_RUNNING under zsh: unquoted $_T timeout wrapper is never word-split

AntonioVitalic:
Fixes #2842, #2941. Supersedes #2824 — same branch, same author, same fix; that PR was auto-closed when a force-push from a shallow clone left the branch without a common ancestor with `main`, and GitHub refuses to reopen a PR whose branch was force-pushed after the close. The history is repaired here: this head is a normal child of `main`.

## Why (in your own words)

On any machine whose shell is zsh — the macOS default, and what agent harnesses run their shell tool as — the BROWSER SETUP probe answers `ASIDE_NOT_RUNNING` while Aside is running fine. The status line is the small part. The damage is that every browsing skill then silently drops to the bundled headless browser, which has none of the user's cookies or logged-in sessions, so a page behind a login comes back signed-out and the skill blames the site instead of the probe. #2810 made Aside the browser gstack drives first; under zsh that contract was unreachable, and Aside only ships for macOS, where zsh is the shell.

The probe holds its deadline in a variable and expands it unquoted, so it only becomes two words in a shell that word-splits unquoted expansions. zsh does not — it looks for one command named `gtimeout 30`. The variable is the vehicle for all three deadline arms, so under zsh every arm fails, including the perl arm that is the only one present on a stock Mac.

This change makes the deadline a function, `_gs_d`, which takes the command as `"$@"`, already split, and behaves identically in sh, bash and zsh.

It also stops the probe from hiding *why* it failed. "Not running" covers states with different fixes — no window open for the profile, a `NODE_OPTIONS` preload that kills the Node-built CLI — and a bare verdict sends every one of them to "open the Aside app". The failure line now carries the CLI's reason, `ASIDE_NOT_RUNNING: <reason>`, the same shape `gstack-render` already reports; the BROWSER SETUP prose, which already asked the agent to "quote the probe output verbatim" — output the probe used to throw away — now quotes that reason before asking the user to open the app.

## Live evidence

Environment: macOS 15 (Darwin 25.6.0), zsh 5.9, bash 3.2.57, aside 1.26.906.1630 — installed, signed in, answering `aside repl` in ~5 ms throughout. **Measured against `main` at v1.88.1.0 (`b9706f36`)**, with this branch rebased onto it. Each row narrows PATH so a different deadline arm is selected; the probe is extracted verbatim from the rendered `browse/SKILL.md` of each tree, not retyped.

The mechanism, isolated:

```
$ _T="gtimeout 30"; $_T echo hi
zsh:1: command not found: gtimeout 30
$ ${=_T} echo hi          # zsh word-split forced, to confirm the cause
hi
```

Before — `main` at v1.88.1.0, still broken on the current release. Aside was up the whole time:

```
deadline arm            zsh                  bash    sh
gtimeout (coreutils)    ASIDE_NOT_RUNNING    READY   READY
timeout only            ASIDE_NOT_RUNNING    READY   READY
perl only (stock Mac)   ASIDE_NOT_RUNNING    READY   READY
none of the three       READY                READY   READY
```

After — this branch:

```
deadline arm            zsh      bash     sh
gtimeout (coreutils)    READY    READY    READY
timeout only            READY    READY    READY
perl only (stock Mac)   READY    READY    READY
none of the three       READY    READY    READY

$ GSTACK_SKIP_ASIDE=1 zsh probe.sh    → NEEDS_ASIDE
$ PATH without aside,  zsh probe.sh   → NEEDS_ASIDE
```

`READY` is the full line `READY: aside 1.26.906.1630` in every cell; trimmed here for width. The `none of the three` row is why the 4th arm exists: today's empty `_T` still runs the call, so a function without `else "$@"` would drop it.

The failure reason, with the real CLI and the preload from #2941 (a `--require` whose file is gone). The aside binary is a Node single-executable and honors `NODE_OPTIONS`, so it dies before it runs; the useful line sits below Node's loader frame, which is why the probe prints the first line that starts with a capital letter rather than the first line:

```
$ NODE_OPTIONS=--require=/nonexistent/restore-node-options.cjs zsh probe.sh   # main
ASIDE_NOT_RUNNING
$ NODE_OPTIONS=--require=/nonexistent/restore-node-options.cjs zsh probe.sh   # this branch
ASIDE_NOT_RUNNING: Error: Cannot find module '/nonexistent/restore-node-options.cjs'
```

Same result under bash for both trees.

**Why not `eval`.** `eval` splits in all three shells and is the obvious one-liner — this PR carried it until I measured the perl arm. `eval` re-parses the string, so the parens and the `;` in `perl -e alarm(shift);exec(@ARGV) 30` stop being data and become syntax:

```
deadline arm            eval, zsh            eval, bash
gtimeout                READY                READY
timeout only            READY                READY
perl only (stock Mac)   ASIDE_NOT_RUNNING    ASIDE_NOT_RUNNING
```

A stock Mac has no Homebrew coreutils and no GNU `timeout`, but does have `/usr/bin/perl` — so `eval` trades the zsh bug for a regression on the default macOS install, and takes bash down with it. Recorded here so the next person doesn't reach for it. The same reasoning rules out `${=_T}` (zsh-only): anything that keeps the deadline a string has to re-split or re-parse it.

**Those tables are now a test.** The old pins asserted the probe's *text*, which is how a shape that had never run under zsh satisfied every one of them. `test/aside-driver.test.ts` now renders the probe and executes it: 3 shells × 4 deadline arms, with a stub `aside` and stub `gtimeout` / `timeout` on a narrowed PATH so the arm under test is the only one reachable, plus both `NEEDS_ASIDE` exits, plus two failing CLIs in every shell — one that prints its own sentence, one that crashes like Node with the useful line below the frame. Put `main`'s resolver back under this branch's test and the first zsh cell fails while sh and bash stay green in the same run — the bug, in one assertion:

```
$ git show origin/main:scripts/resolvers/aside.ts > scripts/resolvers/aside.ts
$ bun test test/aside-driver.test.ts
Expected: "zsh/gt: READY: aside 9.9.9"
Received: "zsh/gt: ASIDE_NOT_RUNNING"
(fail) the rendered probe answers READY in sh, bash and zsh, on every deadline arm
```

(The text pins go red in that run too — same swap.) Printing the first line instead of the capital-letter one fails the Node case with `node:internal/modules/cjs/loader:1573`. On this branch: 29 pass, 0 fail, 1.7 s idle for the file with its 20 probe runs; the test carries a 30 s ceiling for a loaded CI box, not a wait.

## What it costs in the skeleton budgets

The probe renders into 19 browsing skeletons, several with single-digit headroom, so each half is sized on purpose.

- **The deadline costs zero bytes.** The function is written to the byte count of the two lines it replaces: two lines, the same three arms in the same order, the perl arm in `main`'s exact spelling (`alarm(shift);exec(@ARGV)`), and no `2>&1` on `command -v`, which writes nothing to stderr when a command is missing (checked in zsh, bash, sh, dash and ksh). The rationale lives in `scripts/resolvers/aside.ts` as a TS comment, not in the emitted bash.
- **The reason costs 53 bytes per copy of the probe** — capturing the output and printing one line of it — and 44 in the skeletons whose BROWSER SETUP prose line, reworded, gives 9 back. That moves two guards in `test/helpers/carve-guards.ts` by exactly the measured amount, with the measurement in the comment as the other entries carry it: `plan-devex-review`'s skeleton cap 68,500 → 68,550 (measured 68,543) and `plan-ceo-review`'s union ratio 1.08 → 1.081 (measured 1.0802). Nothing else moves.

| vs cap | main | this PR |
|---|---:|---:|
| `plan-devex-review` skeleton | 68,490 / 68,500 | **68,543 / 68,550** |
| `plan-ceo-review` skeleton | 80,043 / 80,100 | **80,096 / 80,100** |
| `plan-ceo-review` union ratio | 1.0799 / 1.08 | **1.0802 / 1.081** |
| `qa` union ratio | 1.0793 / 1.08 | **1.0798 / 1.08** |

```
$ bun test test/aside-driver.test.ts test/parity-suite.test.ts test/skill-size-budget.test.ts test/host-config.test.ts
[skill-size-budget] corpus OK: 3114885 → 2428662 bytes (×0.780)
[skill-size-budget] catalog OK: ~4524 tokens (target ≤7000, 53 tracked skills)
[parity] 21/21 skills passed parity vs v1.64.1.0
 113 pass
 0 fail
 562 expect() calls
```

## Scope

- **Changed:** `scripts/resolvers/aside.ts` — the deadline becomes `_gs_d()`, written to the byte count of the lines it replaces, same gtimeout → timeout → perl alarm order, plus a 4th arm that runs the call unbounded when none of the three exists (what the empty `_T` already did); the failure line carries the CLI's reason, and the BROWSER SETUP prose quotes it first. `test/helpers/carve-guards.ts` — the two guards above, by the measured amount. `test/aside-driver.test.ts` — the assertion pinned the broken invocation verbatim, so it now pins the function, asserts the 4th arm, and asserts neither `$_T aside repl` nor an `eval … aside repl` form can come back; and a second test **executes** the rendered probe across 3 shells × 4 deadline arms and two failing CLIs with stubbed binaries, so a text pin is no longer the only thing standing between this bug and a release. The 19 generated SKILL.md files and the 3 `test/fixtures/golden/` fixtures are regenerated, not hand-edited (`bun run gen:skill-docs`; goldens rendered the way `test/host-config.test.ts` renders them).
- **Rebased onto `main` at v1.88.1.0** (`b9706f36`), no conflicts; every number above is from that tree.
- **Verified live by:** running the probe extracted from each tree's rendered `browse/SKILL.md` under zsh, bash and sh, once per deadline arm (PATH narrowed with symlink shims to select each), before and after, with Aside live — the tables above. Plus both `NEEDS_ASIDE` paths, the broken-preload case against the real CLI, and the same matrix as an executing test in the suite. **Paired full run, same machine, back to back:** `bun run test` exits 1 on this branch. On `main` at v1.88.1.0 it exits 124 twice in a row, with nothing else running: shard 3 wedges on `test/pty-output-wake.test.ts`. Because the killed shard never finishes, `main`'s failure list comes up short, so the files that failed only here were re-run alone on both trees. All 9 give identical results on each, among them `aside-render` at 68 pass / 0 fail; its one timeout in the full run also shows up on `main`, in a different case. `aside-driver`, `parity-suite`, `host-config`, `skill-size-budget`, `third-party-actions` and the spawn-timeout tripwire have no failures in either run, and `bun run build` finishes clean in 5 s.
- **Did NOT test:** Linux or Windows hosts for the probe itself (Aside is macOS-only; there the probe short-circuits on `command -v aside` and the fallback browser takes over — that path is unchanged). No `claude -p` E2E run. The no-window state is covered by the stubbed test, not reproduced live: that means closing the windows of a signed-in Aside profile. On this machine `main` itself fails `bun run test` broadly — the `cso` suites need Docker and the PTY/WebSocket suites are flaky — so that run separates a regression from the noise only as the paired delta above, not as a green tick.

## Liveness proof (required for external contributors)

Authenticated login: `gh api user --jq .login` → AntonioVitalic.

`GSTACK PR` typed live at my own zsh prompt — the `command not found` underneath is the shell answering the keystrokes, not an overlay:

GSTACK PR typed live in Terminal

## Checklist

- [x] Liveness screenshot attached (GSTACK PR typed live into a real surface) or PR author is @garrytan (owner exemption)
- [x] New public command / external service / host adapter has an accepted issue linked (or N/A) — N/A
- [x] Linked issue or reproduction: #2842, #2941

Introduced in #2810.


AntonioVitalic:
Superseded by #2824 — same branch, same commit. Closing this one to keep the original thread, which has the earlier history on this bug.

PR #2899, opened 2026-09-18
fix(redact): env.kv no longer fires on a function call assigned to a credential-shaped name

Yugz29:
## Problem

The `env.kv` value capture is `[^\s'"]{8,}` — any run of non-space, non-quote characters — so it matches **code** as well as literals. With a credential-shaped name in front (bare `session`, `token`, `password` all pass `isCredentialShapedEnvName` as bare suffixes), an ordinary source line is a MEDIUM finding whenever the expression's mixed case clears the 3.0-bit entropy gate.

Two pushes in a row were refused on a project where `session` is a domain word (a local pre-push stage that treats MEDIUM as blocking on a public repo), on these exact lines:

```python
# intelligence/tests/test_core_client_retry.py — a fake requests.Session for a retry test
    session = _FlakySession(requests.ReadTimeout("Read timed out"))
    session = _FlakySession()

# intelligence/pulse_intelligence/cli.py — a lookup, not a secret
    session = find_session(client, args.session_id, now=now, state=state, day=args.date)
```

```
env.kv  MEDIUM  Secret-named assignment (env/YAML/JSON) with high-entropy value · _Fla********…
env.kv  MEDIUM  Secret-named assignment (env/YAML/JSON) with high-entropy value · find********…
```

Neither line contains a secret. The first was worked around by renaming the variable; the second is production code where `session` is the right name.

## Fix

`validate` sees only the span, never the file, so the calibration is on the span's shape — as a **negative** test on exactly one code shape: a call, i.e. an identifier or dotted path immediately followed by `(`.

```ts
const ENV_KV_CALL_SHAPED = /^[A-Za-z_$][A-Za-z0-9_$]*(?:\.[A-Za-z_$][A-Za-z0-9_$]*)*\(/;
```

Nothing else is excluded, deliberately:
- a character allow-list was tried first and silenced real passwords carrying `!#$&*` (`Tr0ub4dor&3!xyz`);
- excluding dotted attribute paths (`settings.API_KEY`) silenced passphrases (`correct.horse.battery.staple`) and fixed no false positive actually met.

## Measurements

Four variants scanned side by side (upstream `v1.79.0.0`, the allow-list attempt, call + dotted-path, **call only**):

| corpus | upstream | allow-list | call + paths | **call only** |
|---|---|---|---|---|
| 27 lines: the blocking lines above, other calls, and literals (hex, base64 `==`, JWT, `sk-…`, `%`-encoded DSN, five `!#$&*` passwords, a dotted passphrase) | 6 false positives | 8 (5 real passwords silenced) | 3 (passphrases silenced) | **0** |
| the 22 `env.kv` cases already in `test/redact-engine.test.ts` | 0 | 0 | 0 | **0** |

## Tests

- `env.kv ignores function calls assigned to credential-shaped names`: the four calls above plus a FastAPI `Depends(...)` stay silent.
- `env.kv still fires on literal secrets whatever punctuation they carry`: hex, base64 with padding, JWT, `sk-…`, percent-encoded DSN, five punctuation-heavy passwords, `correct.horse.battery.staple` (bare and quoted).
- `bun test` over the 16 redact/telemetry files: 325 pass, 0 fail. Tier 1 static checks otherwise untouched (no SKILL.md change).


Yugz29:
Related: #2751. Both address gstack-redact-prepush false positives. #2751 lets a repo exclude deliberate test fixtures; this one stops env.kv from firing on function calls. They touch different files and can land in either order.

Yugz29:
Friendly bump: this still reproduces on `v1.79.0.0` (current `main`), and it keeps costing a manual review step downstream.

This week alone, `gstack-redact-prepush` reported four more `env.kv` MEDIUM findings on test lines of this shape, across two pushes:

```python
session = SessionView(raw=session_view("ffffffffffffffff"), day=REFERENCE.date())
```

No secret in any of them: `session` is a domain word and the value is a constructor call. The captured span starts with `SessionView(`, so the call-only rule in this PR silences exactly these lines, while the literal cases in its tests still fire.

On our side we filter this shape in a local pre-push stage, but `gstack-redact-prepush` runs after it without that filter, so every push shows a MEDIUM count we have to reconcile by hand. The only lever on the gstack stage is `GSTACK_REDACT_PREPUSH=skip`, which turns the whole scan off. That is not something we want to do, since HIGH findings must keep blocking.

The PR is still mergeable: `lib/redact-patterns.ts` has not changed on `main` since `v1.78.0.0`. Happy to rebase, narrow it further, or have it folded into the next community wave, whichever is easiest for you.

PR #2933, opened 2026-09-22
land-and-deploy: poll CI checks every 60s instead of 10s

manav-jain:
## What

Adds `--interval 60` to the `gh pr checks --watch` call in Step 3 of `/land-and-deploy` (both `SKILL.md` and the `SKILL.md.tmpl` it is generated from).

## Why

`gh pr checks --watch` defaults to a **10 second** refresh interval:

```
-i, --interval int   Refresh interval in seconds in watch mode (default 10)
```

The skill budgets a 15-minute CI wait, so a single watch costs ~90 polls, and each poll is several GitHub API requests. That lands on the 5,000/hr **per-user** primary quota, which is shared across every credential the user has — so it competes with the rest of the toolchain rather than having its own budget.

I hit this on a machine running several concurrent agent sessions against one `gh` token: the quota drained, and unrelated tooling started failing with `API rate limit exceeded` while the watch loop kept polling.

CI runs here take minutes, not seconds, so a 60s interval costs nothing in practice and cuts the poll count 6x. `--fail-fast` still exits on the first check failure, so a red build is reported as promptly as the interval allows.

## Testing

Flag verified against `gh version 2.x` (`gh pr checks --help`). Docs-only change — no code paths affected.


manav-jain:
Closing — not needed on our side after all. Thanks anyway.

PR #2936, opened 2026-09-22
browse type: refuse a selector as text, add --into <sel>

DizzyWesterwald:
## What

`browse type` joined every argument into the typed text, so `browse type "#q" hello` typed the selector into whatever element had focus and still reported success. A tool that types the wrong thing and says ok is the worst kind of wrong.

## Change

- `type` refuses a first argument that both looks like a CSS selector and matches an element on the current page, and says what to do instead.
- New `type --into <sel> <text>`: focuses that element first, then sends the keystrokes.
- Six tests in `browse/test/commands.test.ts`; the red path was proven by reverting the fix and watching them fail.
- `gstack/llms.txt` updated for the new form. The SKILL.md table that documented `type <text>` no longer exists on main, so no doc change there.

Rebased onto v1.87.5.0.


DizzyWesterwald:
Test note: `bun test test/commands.test.ts` on this branch gives 248 pass / 2 fail, and the same two ("Path traversal prevention" for eval and cookie-import) fail identically on a clean checkout of v1.87.5.0 in this environment (the file-not-found path wins before the traversal check when the test runs outside the repo's safe directory), so they are unrelated to this change. The nine `type` tests pass.

